PR #17, opened 2024-08-04
Serde Serialize and split client packets into their own files

lukas0008:
## Serde Serialize
Now ClientPacket is split into two traits:
1. Packet - Which only has a const VarInt as the PACKET_ID
2. ClientPacket - Has the write function

If a struct implements Packet and serde::Serialize, it will have an automatic implementation of ClientPacket. This means that to make a new packet, as long as it does not infer anything from the context, you can simply do:
```rust
#[derive(serde::Serialize)]
struct CMyPacket {
  ...
}
impl Packet for CMyPacket {
  const PACKET_ID: VarInt = 0x...;
}
```
And this will act as if you had implemented ClientPacket manually. If some behaviour requires you to add custom serialization, simply remove `#[derive(serde::Serialize)]` and implement ClientPacket manually like before.

Note that you can still do `T::PACKET_ID` on a type that implements ClientPacket, since ClientPacket requires Packet to be implemented.

## Splitting client packets into their own files

While this may be unnecessary in some places, in client/play there was a large mod.rs file which was extremely hard to navigate.

Why add `c_` to the beginning of every filename? This is because client/play/player_action.rs shares an object between packets, so indicating it with its name allows for easier navigation.

## Todo
- [x] Still need to implement other serialization types
- [x] Switch existing packets to use automatic serialization
- [ ] Need to do the same thing for the server packets (use a deserializer and split them up to their individual files) (In a Seperate PR)


Snowiiii:
Hey @lukas0008. Thanks for the PR. Mostly looks good so far. But i think we should use an macro to generate the packet id. At the end we could also automatically just generate 
```rust
impl Packet for CMyPacket {
      const PACKET_ID: VarInt = ...;
}
```
i think the macro should look something like `#[packet(id = 123)]`. Feel free for suggestions. And let me know if you need help with that.

I also think that splitting everything into separate files improves readability :+1: .

I would like to merge/review when everything is done. To better keep track of what is already finished i implemented Checkboxes into your todo

Snowiiii:
todo: add bounded string

PR #21, opened 2024-08-07
Add Deserializer for server bound packets

lukas0008:
Identical implementation as #17 but for deserializing server bound packets, for details look there.


Snowiiii:
Looks good, Thanks  :heart:

PR #34, opened 2024-08-19
adjust file descriptor limits

Bruce0203:


Snowiiii:
Hey, Can you provide a description?

Bruce0203:
The file descriptor is used to manage TCP sockets in the operating system. The number of file descriptors is limited per process, which can lead to a maximum limit of TCP sockets.

Snowiiii:
Should'nt we let the user decide whats the File descriptors limit, I can imagine that common server distros change this value or unlimit it. I wonder how Netty/Vanilla Minecraft handles it

lukas0008:
Correct me if I'm wrong, but isn't the soft limit of file handles on linux usually unlimited? Only the hard limit which cannot be changed by unprivileged users is usually set (although also usually to an unnecessarily high value like >=2048). But to change those, pumpkin would need to run as root, which is also against security practices. The soft limit being lower also would probably mean that the user themselves set it that way for a reason.

Snowiiii:
Netty also uses the global OS value. I think we should use it aswell. There may will be useres who want to have an connection limit. I also don't really want to deal with libc. I will close this for now. Feel free to reopen or make an issue. To give good points to set the limits ourselfs

Issue #15, opened 2024-08-04
Tracking: Add Commands


Snowiiii:
Our goal is to implement all Vanilla Commands. Im sure this will take a while :D

- [x] Add Serverbound Chat Command (https://wiki.vg/Protocol#Chat_Command)

## Command Senders
- [x] Players can send Commands
- [x] Console can send Commands 
- [x]  RCON can send Commands

### Custom
- [x] /pumpkin | gives information about current pumpkin version
### Vanilla
https://minecraft.wiki/w/Commands#Available_in_Java_Edition
- [x] /[advancement](https://minecraft.wiki/w/Commands/advancement)
- [x] /[attribute](https://minecraft.wiki/w/Commands/attribute)
- [x] /[ban](https://minecraft.wiki/w/Commands/ban)
- [x] /[ban-ip](https://minecraft.wiki/w/Commands/ban-ip)
- [x] /[banlist](https://minecraft.wiki/w/Commands/banlist)
- [x] /[bossbar](https://minecraft.wiki/w/Commands/bossbar)
- [x] /[clear](https://minecraft.wiki/w/Commands/clear)
- [x] /[clone](https://minecraft.wiki/w/Commands/clone)
- [x] /[damage](https://minecraft.wiki/w/Commands/damage)
- [x] /[data](https://minecraft.wiki/w/Commands/data)
- [x] /[datapack](https://minecraft.wiki/w/Commands/datapack)
- [x] /[debug](https://minecraft.wiki/w/Commands/debug)
- [x] /[defaultgamemode](https://minecraft.wiki/w/Commands/defaultgamemode)
- [x] /[deop](https://minecraft.wiki/w/Commands/deop)
- [x] /[dialog](https://minecraft.wiki/w/Commands/dialog)
- [x] /[difficulty](https://minecraft.wiki/w/Commands/difficulty)
- [x] /[effect](https://minecraft.wiki/w/Commands/effect)
- [x] /[enchant](https://minecraft.wiki/w/Commands/enchant)
- [x] /[execute](https://minecraft.wiki/w/Commands/execute)
- [x] /[experience](https://minecraft.wiki/w/Commands/experience)
- [x] /[fetchprofile](https://minecraft.wiki/w/Commands/fetchprofile)
- [x] /[fill](https://minecraft.wiki/w/Commands/fill)
- [x] /[fillbiome](https://minecraft.wiki/w/Commands/fillbiome)
[...]

PR #39, opened 2024-08-20
Command Parser

user622628252416:
Regarding #15, I implemented a tree-style command parser, as well as the /help command for reference.

There is a central CommandDispatcher instance, that is contained inside of a OnceLock (pumpkin::commands::DISPATCHER), from where all commands are dispatched. There is currently no access to the level/world the command is executed in, so what a command can actually do is quite limited. This is more of a temporary solution.


Snowiiii:
Hey, Mostly looks good :+1: . Could you may fix all Clippy warns ?. I also just pushed a commit which fixes the one warn in packet_encoder

pumpkin/src/commands/arg_player.rs
```diff
@@ -0,0 +1,43 @@
...
+    None
+}
+
+/// todo: implement (so far only own name + @s is implemented)
+pub fn parse_arg_player<'a>(src: &'a mut CommandSender, arg_name: &str, consumed_args: &ConsumedArgs) -> Result<&'a mut Client, InvalidTreeError> {
+    let s = consumed_args.get(arg_name)
+        .ok_or(InvalidConsumptionError(None))?;
+
+    if let Player(client) = src {
+        if s == "@s" {
```
Bryntet, line 32: This should probably be converted to a match statement

user622628252416: Done 👍

user622628252416:
I fixed all the warns that I could fix. I did have to suppress two dead_code warns tho because no command needing literals is implemented as of now.

Snowiiii:
> I fixed all the warns that I could fix. I did have to suppress two dead_code warns tho because no command needing literals is implemented as of now.

Great. I just don't really like that in help command, Commands are send multiple times if they have multiple arguments. I find that very confusing. CI is btw unhappy

Snowiiii:
Okay. Everything other looks good. Thank you :+1:

PR #42, opened 2024-08-20
feat: initial docker support

Erb3:
Adds support for docker

- [x] Compile in docker
- [x] Load favicon
- [x] Document usage
- [x] Fix log spam
- [x] Make SIGTERM work
- [x] Alpine

I've decided that GitHub actions can be done later. Since we aren't making releases yet, it's hard to publish docker images.


Snowiiii:
We want to give people the option to choose their own icons. So compiling with Pumpkin's icon data is not an option

Erb3:
Well in the case we can't have a default icon. I am not suggesting to hardcode it, but we can't both have a default loaded dynamically, and be able to compile to a binary

Snowiiii:
So im not very fimiliar with docker, Why it isn't possible or so easy to just use the icon file directly?

Erb3:
It's not a docker limitation, we totally could include the icon, but:

1. You can't just give the binary to someone, and say "run this". It will panic due to the missing icon. This is unrelated to Docker.
2. It uses `CARGO_MANIFEST_DIR` to find the icon, which is injected by cargo.

Again, neither of these are limitations of Docker. I could set the `CARGO_MANIFEST_DIR` myself, but it feels jank. I also think that being able to run Pumpkin standalone would be a benefit. It would be like you running `java -jar paper.jar`, but it crashes because you didn't include a default icon.

Snowiiii:
Your right, I made the icon optional so you can start the server with only the executable
https://github.com/Snowiiii/Pumpkin/commit/b37039126ac1e5a4c1c0e104f78923e3b3c5f4f9

Snowiiii:
> It now loads successfully, however now the console is being spammed by the following log:
> 
> ```
> 2024-08-21T15:56:27.941Z INFO  [pumpkin::commands] Text { text: "Empty Command" }
> ```

Does it spamm?. Or only says when sending something in the terminal?

Erb3:
It spams

> > It now loads successfully, however now the console is being spammed by the following log:
> > ```
> > 2024-08-21T15:56:27.941Z INFO  [pumpkin::commands] Text { text: "Empty Command" }
> > ```
> 
> Does it spamm?. Or only says when sending something in the terminal?

Spam, many each ms

Snowiiii:
> > > It now loads successfully, however now the console is being spammed by the following log:
> > > ```
> > > 2024-08-21T15:56:27.941Z INFO  [pumpkin::commands] Text { text: "Empty Command" }
> > > ```
> > 
> > 
> > Does it spamm?. Or only says when sending something in the terminal?
> 
> Spam, many each ms

Can you may apply this patch and test ?

[test.patch](https://github.com/user-attachments/files/16695118/test.patch)

Snowiiii:
> > > > It now loads successfully, however now the console is being spammed by the following log:
> > > > ```
> > > > 2024-08-21T15:56:27.941Z INFO  [pumpkin::commands] Text { text: "Empty Command" }
> > > > ```
> > > 
> > > 
> > > Does it spamm?. Or only says when sending something in the terminal?
> > 
> > 
> > Spam, many each ms
> 
> Can you may apply this patch and test ?

[test.patch](https://github.com/user-attachments/files/16695134/test.patch)

Erb3:
Your .patch file didn't work, perhaps CRLF/LF? Well I manually did the change and it worked :)

Snowiiii:
Great, So whats needs to be fixed next?

Erb3:
> Great, So whats needs to be fixed next?

The biggest problem now is termination not working, for some reason. This leads to the docker container being slow to stop (10s timeout)

Erb3:
It suspect this is caused by running pumpkin with PID 1, which can't terminate like normal. You have to install a special CTRL+C handler, similar to what is used for graceful shutdown. See https://robertying.com/post/sigterm-docker/

Snowiiii:
@Erb3 Can you may test with latest commit ?

https://github.com/Snowiiii/Pumpkin/commit/fa1ca0782e3f6418706e66ca3137da4b5f90a0f9

Erb3:
> @Erb3 Can you may test with latest commit ?
> 
> [fa1ca07](https://github.com/Snowiiii/Pumpkin/commit/fa1ca0782e3f6418706e66ca3137da4b5f90a0f9)

Works like a charm!

Snowiiii:
Okay, Whats next ?

Erb3:
I've decided to post-pone GitHub actions to publish, since we aren't making releases. I'll try to get it working on alpine, since it has a smaller footprint. If that doesn't work, we're ready to merge :)

Snowiiii:
Looks good. I just wonder if it makes sense to may use `RUSTFLAGS="-C target-cpu=native"` ?

Snowiiii:
Thank you @Erb3

PR #45, opened 2024-08-21
Fix basic chat message packet

kralverde:
Fixes basic chat message packets.


Snowiiii:
Hey, Im sure what was the problem?, Was it that is use 255 bytes instead of 256 ?

kralverde:
it looks like the 256 and some optional types do not have a boolean in-front of them; they are conditional on other things

Snowiiii:
So, It generally works now?

kralverde:
For normal chat messages, yes

Snowiiii:
Thanks so much. I really see the effort you put into this. Good job 👍

PR #47, opened 2024-08-21
Fix compilation of individual packages

Luk-ESC:
Previously `pumpkin-text`, `pumkin-protocol` and `pumkin-world` would not compile on their own.
This happened because we didn't specify the required features, but since a dependency (or `pumkin`) enabled them, the error went unnoticed.
Check out https://doc.rust-lang.org/cargo/reference/features.html#feature-unification for more information.

This PR enables the `sync` feature for `tokio` which is required by `pumkin-world` and moves `uuid` to the workspace so that every crate uses the same feature set (specifically, `pumkin-text` and `pumkin-protocol` also require it's `serde` feature).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,5 +12,5 @@ lto = true
 codegen-units = 1
 
 [workspace.dependencies]
-tokio = { version = "1.39.2", features = ["net", "macros", "rt-multi-thread", "fs", "io-util"] }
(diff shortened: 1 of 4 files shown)


Snowiiii:
Hey, Your right and i have nothing to add. Thanks for the PR @Luk-ESC 👍

PR #51, opened 2024-08-21
Add permissions check for SSetCreativeSlot

Bryntet:
make sure the players gamemode is creative when client sends SSetCreativeSlot

### DIFF
--- a/pumpkin/src/client/player_packet.rs
+++ b/pumpkin/src/client/player_packet.rs
@@ -402,7 +402,7 @@ impl Client {
     pub fn handle_set_creative_slot(&mut self, _server: &mut Server, packet: SSetCreativeSlot) {
         let gamemode = self.player.as_ref().unwrap().gamemode;
         if gamemode != GameMode::Creative {
-            self.kick("CHEATER")
+            self.kick("Invalid action, you can only do that if you are in creative")
         }
         let inventory = &mut self.player.as_mut().unwrap().inventory;
 
--- a/pumpkin/src/client/player_packet.rs
+++ b/pumpkin/src/client/player_packet.rs
@@ -400,11 +400,12 @@ impl Client {
     }
 
     pub fn handle_set_creative_slot(&mut self, _server: &mut Server, packet: SSetCreativeSlot) {
-        let gamemode = self.player.as_ref().unwrap().gamemode;
-        if gamemode != GameMode::Creative {
-            self.kick("Invalid action, you can only do that if you are in creative")
+        let player = self.player.as_mut().unwrap();
+        if player.gamemode != GameMode::Creative {
+            self.kick("Invalid action, you can only do that if you are in creative");
+            return;
         }
-        let inventory = &mut self.player.as_mut().unwrap().inventory;
+        let inventory = &mut player.inventory;
 
         inventory.set_slot(packet.slot as usize, packet.clicked_item.to_item(), false);
     }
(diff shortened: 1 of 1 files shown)


Snowiiii:
You have a good commit message, Why you don't use that for the PR description. There are many people using "self explanatory commit" or something like this in PR and i really hate it. So would be cool if you could use good commit messages and good PR descriptions in the future 👍

pumpkin/src/client/player_packet.rs
```diff
@@ -400,6 +400,10 @@ impl Client {
     }
 
     pub fn handle_set_creative_slot(&mut self, _server: &mut Server, packet: SSetCreativeSlot) {
+        let gamemode = self.player.as_ref().unwrap().gamemode;
+        if gamemode != GameMode::Creative {
+            self.kick("CHEATER")
+        }
         let inventory = &mut self.player.as_mut().unwrap().inventory;
```
Snowiiii, line 407: Why we should `&mut self.player.as_mut()` twice when we also could just do it once ?

Bryntet:
updated my PR message ;)

pumpkin/src/client/player_packet.rs
```diff
@@ -400,6 +400,10 @@ impl Client {
     }
 
     pub fn handle_set_creative_slot(&mut self, _server: &mut Server, packet: SSetCreativeSlot) {
+        let gamemode = self.player.as_ref().unwrap().gamemode;
+        if gamemode != GameMode::Creative {
+            self.kick("CHEATER")
```
Snowiiii, line 405: Im also not sure about the kick message, I generally like to use a "more professional", I want to check for vanilla impossible actions at multiple places. I prefer to use "Invalid  ... action". In this case we could use. "Invalid Gamemode". A modder/cheat developer will properbly understand what he did wrong

pumpkin/src/client/player_packet.rs
Bryntet: because the borrow checker yells at me if i try to have a mut ref to player (that gets used later, so it can't be dropped) when doing self.kick (which requires &mut)

pumpkin/src/client/player_packet.rs
Bryntet: fixed :)

pumpkin/src/client/player_packet.rs
Snowiiii: > because the borrow checker yells at me if i try to have a mut ref to player (that gets used later, so it can't be dropped) when doing self.kick (which requires &mut)

Use
```rust
pub fn handle_set_creative_slot(&mut self, _server: &mut Server, packet: SSetCreativeSlot) {
        let player = self.player.as_mut().unwrap();

        if player.gamemode != GameMode::Creative {
            self.kick("Invalid action, you can only do that if you are in creative");
            return;
        }

        player
            .inventory
            .set_slot(packet.slot as usize, packet.clicked_item.to_item(), false);
    }
 ```

Snowiiii (changes requested):
```suggestion
  let player = self.player.as_mut().unwrap();

        if player.gamemode != GameMode::Creative {
            self.kick("Invalid action, you can only do that if you are in creative");
            return;
        }

        player
            .inventory
            .set_slot(packet.slot as usize, packet.clicked_item.to_item(), false);
 ```

Bryntet:
that should be fixed now

Snowiiii:
Looks good, Thanks

PR #53, opened 2024-08-21
Chat packet updates + Serialization additions

kralverde:
Adds ability to filter chat messages, kicks players for sending too large messages, and adds support for enums in the serializer.


pumpkin-protocol/src/bytebuf/serializer.rs
```diff
@@ -121,7 +129,8 @@ impl<'a> ser::Serializer for &'a mut Serializer {
     where
         T: ?Sized + Serialize,
     {
-        unimplemented!()
+        self.output.put_var_int(&_variant_index.into());
+        _value.serialize(self)
```
Snowiiii, line 133: Can you may use value instead of _value ?,

pumpkin-protocol/src/bytebuf/serializer.rs
```diff
@@ -241,11 +254,11 @@ impl<'a> ser::SerializeTuple for &'a mut Serializer {
     where
         T: ?Sized + Serialize,
     {
-        todo!()
+        _value.serialize(&mut **self)
```
Snowiiii, line 257: Can you may use value instead of _value ?,

pumpkin-protocol/src/bytebuf/serializer.rs
kralverde: This is from the Serializer trait, I'm not sure I can change it

kralverde: Can you rename trait method parameters in rust? I've never tried

Snowiiii: Its so that clippy does not complain about dead code. You can just change the argument name

Snowiiii: > Can you rename trait method parameters in rust? I've never tried

yeah, Im pretty sure you can rename every argument in rust aslong it has the same type

Snowiiii:
Thank you @kralverde

PR #55, opened 2024-08-22
Update CONTRIBUTING.md

Bryntet:
The message was a bit to vague before

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ There are several ways you can contribute to Pumpkin:
   If you'd like to contribute code changes, fork the Pumpkin repository on GitHub.
   Install Rust at [rust-lang.org](https://www.rust-lang.org/).
   Make your changes on your local fork and create a pull request to the main repository.
-  Ensure your code adheres to our project structure and style guidlines.
+  Ensure your code adheres to our project structure and style guidelines.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -22,7 +22,7 @@ There are several ways you can contribute to Pumpkin:
   If you'd like to contribute code changes, fork the Pumpkin repository on GitHub.
   Install Rust at [rust-lang.org](https://www.rust-lang.org/).
   Make your changes on your local fork and create a pull request to the main repository.
-  Ensure your code adheres to the project's coding style guidelines (if any).
+  Ensure your code adheres to our project structure and style guidlines.
```
Luk-ESC, line 25: Typo: `guidlines` -> `guidelines`

Bryntet:
fixed :)

StripedMonkey:
You say "The message was a bit too vague before", this does not address that. Right now "Our project structure and style guidelines" isn't actually defined in this document (and only structure is defined elsewhere, in `STRUCTURE.md`). If someone were to read this, it's completely vapid because there is nothing meaningful being said here besides "Follow my rules, if you don't know what those rules are, too bad."

A better fix, that would make this less vague, is to link to `STRUCTURE.md`, and explicitly say, for example, that rustfmt must be run as a first pass at "our" style guidelines.

PR #56, opened 2024-08-22
Fix `set_slot` armor indexing (Off-by-one error)

Luk-ESC:
The correct value was already used in the second match arm.
`slot - 4` goes up to value `4` (`8` - `4`), `self.armor` has 4 elements (max index is `3`)


Snowiiii:
Looks good. Thanks @Luk-ESC

PR #58, opened 2024-08-23
Print packet id in hex instead of decimal

neeleshpoli:
Whenever a packet is unimplemented, the packet id is printed out in decimal. This makes it difficult to find the respective packet to implement (when needed to) on the protocol documentation so instead of printing in decimal, it is printed in hex now. It should be easier to find packets on the protocol documentation now.

Also I understand really small changes are annoying in repositories, but I feel like this change can make development easier down the road.


Snowiiii:
Hey, Yeah makes sense. Thank you @neeleshpoli

PR #59, opened 2024-08-23
Add packets CCloseContainer, CSetContainerProperty and SCloseContainer

Bryntet:
I have added the packets CCloseContainer and CSetContainerProperty and made abstractions for ContainerProperty.


Bryntet:
I think the abstraction I made is okay, but I'd love to get feedback on it

Bryntet:
Just wrote the packet for SContainerClose too

pumpkin-inventory/src/lib.rs
```diff
@@ -51,3 +52,37 @@ impl WindowType {
...
+            15 => Ok(WindowType::Grindstone),
+            16 => Ok(WindowType::Hopper),
+            17 => Ok(WindowType::Lectern),
+            18 => Ok(WindowType::Loom),
+            19 => Ok(WindowType::Merchant),
+            20 => Ok(WindowType::ShulkerBox),
+            21 => Ok(WindowType::SmithingTable),
+            22 => Ok(WindowType::Smoker),
+            23 => Ok(WindowType::CartographyTable),
+            24 => Ok(WindowType::Stonecutter),
```
StripedMonkey, line 84: You're already using `num_derive::ToPrimitive;`, why do this at all? This is just `WindowType::from_u8()`

pumpkin-inventory/src/window_property.rs
```diff
@@ -0,0 +1,103 @@
+pub trait WindowPropertyTrait: Sized {
+    fn to_id(self) -> i16 {
+        0
```
kralverde, line 3: I think this should be `self as i16`. Then remove all the other implementations that are this.

kralverde (changes requested):

pumpkin-inventory/src/window_property.rs
Bryntet: I tried doing that, but we'd have to constrain the trait further than just Sized in that case. Do you know what supertrait is for `enum as u8`?

kralverde: I'm not around a computer right now sorry, but I'm pretty sure you just put `#[repr(u8)]` on the Enum, then hit it with the Enum::from_u8

StripedMonkey: Why do you need sized at all? In fact, why have an explicit type at all for windowproperties? Just implement to_packet on the WindowPropertyTrait.

To get the conversion you just need to restrict the bounds to an Into<u8> or something similar.

Bryntet: It is Sized so that we can take self instead of &self in the trait

Bryntet:
ah, trying to solve the merge conflicts broke things

pumpkin/src/client/container.rs
```diff
@@ -82,4 +85,20 @@ impl Player {
             &item.into(),
         ))
     }
+
+    /// The official Minecraft client is weird, and will always just close *any* window that is opened when this gets sent
```
Snowiiii, line 89: Why that is weird ?

Bryntet: I should probably change the wording from weird, but it's mainly just a note so that people won't try and depend on behaviour that only a specific window will be closed, when which window_id you send doesn't actually matter to the client

Bryntet: What I think I meant to comment on as weird is that the packet still requires an ID, but it never gets used for anything.

PR #60, opened 2024-08-24
Add packet SClickContainer and abstractions

Bryntet:
Only the packet and abstraction has been implemented so far, the logic will be implemented shortly.


Bryntet:
Tried to get double-clicking to work, but since everything else works, and I'm kind of tired of Inventory work, I'll just leave my PR as this. 

Also, do note that due to how Pumpkin currently handles packets in quick succession (it doesn't, it sometimes drops them), dragging items only partially works.

Bryntet:
In this PR, I have also added the command `/echest` which is very helpful when testing out the inventory system. 

It is a global container, that all players can look at at the same time.

pumpkin/src/server/mod.rs
```diff
@@ -47,6 +48,8 @@ pub struct Server {
     /// Cache the registry so we don't have to parse it every time a player joins
     pub cached_registry: Vec<Registry>,
 
+    pub open_containers: HashMap<u64, OpenContainer>,
+    pub drag_handler: DragHandler,
```
Snowiiii, line 52: *Luckly*. I refactored the Server struct. Now we split most of the things game related into the world. We should put this into World aswell

pumpkin-inventory/Cargo.toml
```diff
@@ -4,10 +4,9 @@ version.workspace = true
...
 
 [dependencies]
-# For items
-pumpkin-world = { path = "../pumpkin-world"}
-
 num-traits = "0.2"
 num-derive = "0.4"
-
-
+thiserror = "1.0.63"
```
Snowiiii, line 9: i really should make thiserror a workspace dependency. For now you should just use `thiserror = "1.0"`

pumpkin-inventory/Cargo.toml
```diff
@@ -4,10 +4,9 @@ version.workspace = true
...
+itertools = "0.13.0"
+# For items
+pumpkin-world = { path = "../pumpkin-world"}
```
Snowiiii, line 12: I usually prefer local crate imports above all other deps

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+                slot: Slot::Normal(slot.try_into().unwrap()),
+            },
+            4 => Self::new_drop_item(button),
+            5 => Self::new_drag_item(button, slot),
+            6 => Self {
+                click_type: ClickType::DoubleClick,
+                slot: Slot::Normal(slot.try_into().unwrap()),
+            },
+            // TODO: Error handling
+            _ => unreachable!(),
```
Snowiiii, line 26: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+                } else {
+                    Slot::OutsideInventory
+                }
+            }
+        };
+        let button = match button {
+            0 => MouseClick::Left,
+            1 => MouseClick::Right,
+            // TODO: Error here
+            _ => unreachable!(),
```
Snowiiii, line 47: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+            slot: Slot::Normal(slot.try_into().unwrap()),
+        }
+    }
+
+    fn new_drop_item(button: i8) -> Self {
+        let drop_type = match button {
+            0 => DropType::SingleItem,
+            1 => DropType::FullStack,
+            // TODO: Error handling
+            _ => unreachable!(),
```
Snowiiii, line 82: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+            click_type: ClickType::ShiftClick,
+        }
+    }
+
+    fn new_key_click(button: i8, slot: i16) -> Self {
+        let key = match button {
+            0..9 => KeyClick::Slot(button.try_into().unwrap()),
+            40 => KeyClick::Offhand,
+            // TODO: Error handling here
+            _ => unreachable!(),
```
Snowiiii, line 68: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+        Self {
+            click_type: ClickType::MouseClick(button),
+            slot,
+        }
+    }
+
+    fn new_shift_click(slot: i16) -> Self {
+        Self {
+            // TODO: Error handle this
+            slot: Slot::Normal(slot.try_into().unwrap()),
```
Snowiiii, line 58: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+            click_type: ClickType::ShiftClick,
+        }
+    }
+
+    fn new_key_click(button: i8, slot: i16) -> Self {
+        let key = match button {
+            0..9 => KeyClick::Slot(button.try_into().unwrap()),
```
Snowiiii, line 65: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+            40 => KeyClick::Offhand,
+            // TODO: Error handling here
+            _ => unreachable!(),
+        };
+
+        Self {
+            click_type: ClickType::KeyClick(key),
+            slot: Slot::Normal(slot.try_into().unwrap()),
```
Snowiiii, line 73: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    fn new_drag_item(button: i8, slot: i16) -> Self {
+        let state = match button {
+            0 => MouseDragState::Start(MouseDragType::Left),
+            4 => MouseDragState::Start(MouseDragType::Right),
+            8 => MouseDragState::Start(MouseDragType::Middle),
+            1 | 5 | 9 => MouseDragState::AddSlot(slot.try_into().unwrap()),
+            2 | 6 | 10 => MouseDragState::End,
+            // TODO: Error handling
+            _ => unreachable!(),
```
Snowiiii, line 98: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+            slot: Slot::OutsideInventory,
+        }
+    }
+
+    fn new_drag_item(button: i8, slot: i16) -> Self {
+        let state = match button {
+            0 => MouseDragState::Start(MouseDragType::Left),
+            4 => MouseDragState::Start(MouseDragType::Right),
+            8 => MouseDragState::Start(MouseDragType::Middle),
+            1 | 5 | 9 => MouseDragState::AddSlot(slot.try_into().unwrap()),
```
Snowiiii, line 95: Never trust the client. We should handle invalid packets

pumpkin-inventory/src/container_click.rs
```diff
@@ -0,0 +1,155 @@
...
+            3 => Self {
+                click_type: ClickType::CreativePickItem,
+                slot: Slot::Normal(slot.try_into().unwrap()),
+            },
+            4 => Self::new_drop_item(button),
+            5 => Self::new_drag_item(button, slot),
+            6 => Self {
+                click_type: ClickType::DoubleClick,
+                slot: Slot::Normal(slot.try_into().unwrap()),
+            },
```
Snowiiii, line 24: I wonder if we may should replace 0..6 with an enum. Would be may more cool for plugin api in the future

pumpkin-inventory/src/lib.rs
```diff
@@ -52,3 +62,185 @@ impl WindowType {
...
+
+    fn print_all_contents(&self) {
+        self.all_slots_ref()
+            .into_iter()
+            .enumerate()
+            .filter_map(|(slot, item)| item.map(|item| (slot, item)))
+            .for_each(|(slot, item)| {
+                dbg!(slot, item);
+            });
+    }
```
Snowiiii, line 103: i think we don't need this :thinking:

pumpkin-inventory/src/lib.rs
```diff
@@ -52,3 +62,185 @@ impl WindowType {
...
+
+    fn internal_pumpkin_id(&self) -> u64 {
+        0
+    }
```
Snowiiii, line 107: I still don't got for what that used for

pumpkin-inventory/src/lib.rs
```diff
@@ -52,3 +62,185 @@ impl WindowType {
...
+            *item_slot = None;
+        }
+        MouseClick::Right => {
+            let half = item.item_count / 2;
+            item.item_count -= half;
+            new_item.item_count = half;
+        }
+    }
+    *carried_item = Some(new_item);
+}
```
Snowiiii, line 131: This looks like an not rusty approche. I you can just get not mut input values and then return the the value. Something like
```rust
pub fn handle_item_take(
    carried_item: Option<ItemStack>,
    item_slot: Option<ItemStack>,
    mouse_click: MouseClick,
) -> (Option<ItemStack>, Option<ItemStack>) {
    let Some(item) = item_slot else {
        return (None, None);
    };
   let item_slot = item_slot;
    match mouse_click {
        MouseClick::Left => {
            *item_slot = None;
        }
        MouseClick::Right => {
            let half = item.item_count / 2;
            item.item_count -= half;
            new_item.item_count = half;
        }
    }
    (new_item, item_slot);
}
```

pumpkin-inventory/src/lib.rs
```diff
@@ -52,3 +62,185 @@ impl WindowType {
...
+                let mut new = *carried;
+                new.item_count = 1;
+                *current_slot = Some(new);
+            }
+        },
+        // Take stack into carried
+        (Some(_current), None) => handle_item_take(carried_slot, current_slot, mouse_click),
+        (None, None) => (),
+    }
+}
```
Snowiiii, line 165: I think replacing this as return value would be also better here as well

pumpkin-inventory/src/lib.rs
```diff
@@ -52,3 +62,185 @@ impl WindowType {
...
+        carried_item.item_count -= carried_change;
+        if carried_item.item_count == 0 {
+            *carried_slot = None;
+        }
+    } else {
+        let left_over = slot.item_count + carried_change - 64;
+        slot.item_count = 64;
+        carried_item.item_count = left_over;
+    }
+}
```
Snowiiii, line 193: I think replacing this as return value would be also better here as well

pumpkin-inventory/src/open_container.rs
```diff
@@ -0,0 +1,69 @@
...
+        &WindowType::Generic9x3
+    }
+
+    fn all_slots(&mut self) -> Vec<&mut Option<ItemStack>> {
+        self.0.iter_mut().collect()
+    }
+
+    fn all_slots_ref(&self) -> Vec<Option<&ItemStack>> {
+        self.0.iter().map(|slot| slot.as_ref()).collect()
+    }
```
Snowiiii, line 68: Does this really need to be an Vec?

pumpkin-inventory/src/player.rs
```diff
@@ -84,23 +91,78 @@ impl PlayerInventory {
...
+        self.slots()
+    }
+
+    fn all_combinable_slots(&self) -> Vec<Option<&ItemStack>> {
+        self.items.iter().map(|item| item.as_ref()).collect()
+    }
+
+    fn all_combinable_slots_mut(&mut self) -> Vec<&mut Option<ItemStack>> {
+        self.items.iter_mut().collect()
+    }
```
Snowiiii, line 167: Does this really need to be an Vec?

pumpkin-protocol/src/server/play/s_click_container.rs
```diff
@@ -0,0 +1,86 @@
...
+                    length_of_array,
+                    array_of_changed_slots,
+                    carried_item,
+                })
+            }
+        }
+
+        deserializer.deserialize_seq(Visitor)
+    }
+}
```
Snowiiii, line 86: Why you did this?. Do you need to Deserialize a custom way?. We use an custom ServerPacket impl theń

pumpkin/src/client/container.rs
```diff
@@ -27,78 +39,405 @@ impl Player {
...
+                let all_slots = container.all_slots();
+                if let Some(item_in_pressed_slot) = all_slots[slot].to_owned() {
+                    let slots = all_slots.into_iter().enumerate();
+                    // Hotbar
+                    let find_condition = |(slot_number, slot): (usize, &mut Option<ItemStack>)| {
+                        // TODO: Check for max item count here
+                        match slot {
+                            Some(item) => {
+                                if item.item_id == item_in_pressed_slot.item_id
+                                    && item.item_count != 64
```
Snowiiii, line 230: Stack sized does varie. We parse it from items.json

pumpkin/src/client/container.rs
```diff
@@ -27,78 +39,405 @@ impl Player {
...
+    ) -> Result<(), InventoryError> {
+        let changing_slot = match key_click {
+            KeyClick::Slot(slot) => slot,
+            KeyClick::Offhand => 45,
+        };
+        let mut changing_item_slot = self.inventory.get_slot(changing_slot as usize)?.to_owned();
+        let mut container = OptionallyCombinedContainer::new(&mut self.inventory, opened_container);
+
+        container.handle_item_change(&mut changing_item_slot, slot, MouseClick::Left)?;
+        *self.inventory.get_slot(changing_slot as usize)? = changing_item_slot;
```
Snowiiii, line 272: You could also return this

pumpkin-inventory/src/container_click.rs
Bryntet: Here, the error handling for negative slot value, is done in the function above, I will change this to .expect, though

pumpkin-inventory/src/container_click.rs
Bryntet: This whole file is about converting the packets to enums 👍

pumpkin-inventory/src/lib.rs
Bryntet: True, it is left over from me debugging.

pumpkin-inventory/src/lib.rs
Bryntet: Placeholder for now. My thought process goes that instead of storing Containers on the chunks in the `World` struct, the blocks only store a unique ID, which can be used to access the inventory. So we don't need to block `World` and chunks when interacting with inventories.

pumpkin-inventory/src/lib.rs
Bryntet: I don't agree. This function ensures correct behaviour for anyone using it. Otherwise correct behaviour can not be enforced as easily.

pumpkin-inventory/src/player.rs
Bryntet: It gets combined with a separate `Vec<&mut Option<ItemStack>>` later, so we'd have to convert it into a Vec anyways

pumpkin/src/client/container.rs
Bryntet: Yes, I thought i put a TODO in the places where needed 👍

pumpkin-protocol/src/server/play/s_click_container.rs
Bryntet: See line 53 to 65 for the reason why it has to be manually deserialized.

pumpkin/Cargo.toml
```diff
@@ -62,3 +62,4 @@ mio = { version = "1.0.2", features = ["os-poll", "net"]}
 uuid.workspace = true
 tokio.workspace = true
 rayon.workspace = true
+itertools = "0.13.0"
```
StripedMonkey, line 65: Workspace dependency for itertools?

pumpkin-inventory/src/open_container.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn new() -> Self {
+        Self([None; 27])
+    }
+}
+impl Container for Chest {
+    fn window_type(&self) -> &'static WindowType {
+        &WindowType::Generic9x3
+    }
+
+    fn all_slots(&mut self) -> Vec<&mut Option<ItemStack>> {
```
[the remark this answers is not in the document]
Bryntet: My thinking is that maybe a plugin will want to have their own custom rules, and hook in to item presses and such, then they'd just have to define their own struct that implements `Container` 

and the thought is that other containers that require special things be placed/taken from special places, will handle it here.

I get that `Chest` as it is now is only a wrapper, but since you can't impl a trait for a specific enum variant, it's much easier if you abstract it like this in the long run IMO

pumpkin/Cargo.toml
Bryntet: Good catch, I just let my editor add the dependency and I didn't notice it landed there instead of the crate 👍

pumpkin/src/server/mod.rs
Bryntet: IMO, containers should not have to be locked on World. World should save a unique ID for each container, and try to lock the container that way.

pumpkin-inventory/src/lib.rs
lukas0008: I feel like it is a good approach, both `&mut Option<ItemStack>` are inputs and outputs, and it is necessary for them to switch anyway. I don't think this is not a rusty approach, since the situation warrants this kind of api, I mean in rust you already need to mark something mut, so I also don't think the usage of this function would be annoying.

I do think however that there shoudld be a clear code doc to explain this behaviour.

PR #61, opened 2024-08-25
Implement Use Item (packet id 0x39) packet

neeleshpoli:
The server can now read use item packets that are sent by the client. The server doesn't handle the packet currently and prints an error to console.


Snowiiii:
Could you may implement an function which is linked to the packet?. Basiclly like all other packets have. You can then leave the function empty

neeleshpoli:
Sure!

PR #64, opened 2024-08-28
Added remaining entity type ids

earomc:
Not that crazy. I wrote a quick lil program to generate them for me from the registry.json. So that I could copy paste them over. Maybe you're interested in that. We should probably use a proper registry for that stuff in the future tho.


Snowiiii:
Looks good. But sure there are that few entities ?. I tought mc had like 1000 of them. I also would like to have the generator somewhere in case of an new update

earomc:
No, no, 129 sounds about right in terms of entities, not counting block entities or entity variants (e.g. cats, wolfs, llamas etc.). And there are 129 entries in the entity_type entries list as of now.

About the generator: It's a separate Rust project with about 20 lines of code. Didn't want to include it in the commit tho. If you want I can send it as a gist or whatever.

Snowiiii:
> No, no, 129 sounds about right in terms of entities, not counting block entities or entity variants (e.g. cats, wolfs, llamas etc.). And there are 129 entries in the entity_type entries list as of now.
> 
> About the generator: It's a separate Rust project with about 20 lines of code. Didn't want to include it in the commit tho. If you want I can send it as a gist or whatever.

If would like to have this tool included in pumpkin somewhere (if you want)

Snowiiii:
Hey, Until CI is failing nothing will be merged. This is also the case for your other open PR

PR #65, opened 2024-08-28
create rust-toolchain.toml

StripedMonkey:
This restricts people to using the stable toolchain.

If someone, (like myself) has the default toolchain be `nightly`, this will prevent the project from being built on a different toolchain accidentally.

You can see an example of this here:
```
~/Projects/Pumpkin> rustup show 
Default host: x86_64-unknown-linux-gnu
rustup home:  /home/monkey/.rustup

installed toolchains
--------------------

stable-x86_64-unknown-linux-gnu
nightly-x86_64-unknown-linux-gnu (default)
1.65.0-x86_64-unknown-linux-gnu
[snip]

installed targets for active toolchain
--------------------------------------

x86_64-apple-darwin
x86_64-pc-windows-gnu
x86_64-pc-windows-msvc
x86_64-unknown-linux-gnu

active toolchain
----------------

stable-x86_64-unknown-linux-gnu (overridden by '/home/monkey/Projects/Pumpkin/rust-toolchain.toml')
rustc 1.80.1 (3f5fd8dd4 2024-08-06)
```

Notice that the `(default)` by `nightly-...`

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,7 @@
+[toolchain]
+# Anything in the latest stable version of rust is fine to use.
+channel = "stable"
+targets = [
+    "x86_64-unknown-linux-gnu",
+    "x86_64-pc-windows-msvc",
+]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -2,6 +2,8 @@
 # Anything in the latest stable version of rust is fine to use.
 channel = "stable"
 targets = [
-    "x86_64-unknown-linux-gnu",
+    "x86_64-apple-darwin",
+    "x86_64-pc-windows-gnu",
     "x86_64-pc-windows-msvc",
+    "x86_64-unknown-linux-gnu",
 ]


Snowiiii:
We definitely want to support more target than Windows and Linux. At best we want to support all target's rust has. Is it possible to just remove that target category ?

StripedMonkey:
The toolchains specified in this file are generally considered the minimum subset supported, it's not a maximal restriction. 

I also don't think you understand what you're asking either, are you saying you want to support all T1 Targets? All T2 Targets? If you actually mean "all targets" then that's just not going to ever work unless we drop Tokio, Mio, and a bunch of other libraries we're currently using and slap `#[no_std]` onto everything.

That doesn't even begin to touch packaging requirements for "all platforms". We might be able to build a valid executable for Android, for example, but we're not signing and slapping the NDK on top of this so it's meaningless to claim "support" if you can't build a runnable executable for it.

If you don't test on X platform, you don't support X platform. Removing the default targets won't magically make this project support all targets.

Unless you're actually putting in the effort to build (and test!) for these platforms they're just "not supported". Regardless of whether it's listed somewhere.

Eventually one of the projects I'm getting around to is adding in `cross` support and CI to do cross compilation for different targets, but until that happens this is the subset we know that this project is being tested on, so that's what this project "supports".

Snowiiii:
You have 2 target's there. Look what mio does support
```
targets = [
  "aarch64-apple-ios",
  "aarch64-linux-android",
  "wasm32-wasi",
  "x86_64-apple-darwin",
  "x86_64-pc-windows-gnu",
  "x86_64-pc-windows-msvc",
  "x86_64-unknown-dragonfly",
  "x86_64-unknown-freebsd",
  "x86_64-unknown-illumos",
  "x86_64-unknown-linux-gnu",
  "x86_64-unknown-netbsd",
  "x86_64-unknown-openbsd",
  "x86_64-unknown-hermit",
]
```

StripedMonkey:
Yes, and that is not all the targets rust supports. That's not even listing the M2 macs on that list. My point about needing to drop Mio to support "all the targets rust supports" still stands. If you actually want "all targets", you need to consider what that actually means.

Snowiiii:
> Yes, and that is not all the targets rust supports. That's not even listing the M2 macs on that list. My point about needing to drop Mio to support "all the targets rust supports" still stands. If you actually want "all targets", you need to consider what that actually means.

Okay i probably expressed my self wrong. We want to support as many targets as we can and as our deps allow

StripedMonkey:
And I think it's admirable to push for that, this is not preventing other toolchains or platforms from being used.

The target list is strictly additive. You can read the documentation [here](https://rust-lang.github.io/rustup/overrides.html#:~:text=The%20host%20platform%20is%20automatically%20included%3B%20the%20targets%20listed%20here%20are%20additive)

Snowiiii:
Okay, But i would still add `x86_64-apple-darwin` and `x86_64-pc-windows-gnu` both are also tested on Pumpkin

PR #66, opened 2024-08-28
implement xoroshiro128 RNG

kralverde:
A rust implementation of one of Minecraft's RNGs: XOROshiro128, including tests cases against the results of equivalent java code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -773,6 +773,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "md5"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "490cc448043f947bae3cbee9c203358d62dbee0db12107a74be5c30ccfd09771"
+
+[[package]]
 name = "memchr"
 version = "2.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


Snowiiii:
Hey i actually never heard of XOROshiro128. Can you may explain a bit?. It looks like its an random generator. I wonder where Mojang uses this implementation

Snowiiii:
I also found that rust `rand` crate has xoshiro128 aswell
https://github.com/rust-random/rand/blob/79f1b0ffdbc8605c99ce8348d40ee073a2a7bd0f/src/rngs/xoshiro128plusplus.rs#L17

kralverde:
I never heard of it either, my code is basically 1-1 of the Minecraft Java. I saw the rust crate too, but it gives different values for the same seed as the Minecraft one

kralverde:
Yeah, I just used this as a wrapper around the rust implementation and the tests failed. It may have been user error, but this works and I think is good enough (TM)

kralverde:
We can always come back to it and change the internals, I just wanna get a stepping stone for chunk generation out of the way

kralverde:
I can confirm that the rust implementation uses different magic numbers than the java version. I refactored a bit, but I am confident that this is what we should use for the XorOshiro128 implementation

Snowiiii:
Thank you @kralverde

PR #69, opened 2024-08-31
Removed unused #[allow()]'s

DaniD3v:
title


DaniD3v:
This only fails on the stable from githubs CI. Works on my nightly.
I think they patched the broken lint.

Snowiiii:
1.81 is out. still failing

DaniD3v:
I'm using the nightly toolchain so it might've been fixed in 1.82.
I'm just gonna remove the one case

I'll also replace the existing allows with expect.

DaniD3v:
expect was newly added in `1.81` and warns if an `#[allow]` hint is not necessary anymore.

Snowiiii:
Looks good. Thank you @DaniD3v

PR #70, opened 2024-08-31
Remove lazy static in favor of LazyLock

DaniD3v:
## Remove lazy static in favor of `LazyLock`

- Increases MSRV
- Removes a dependency (-compiletime)

there's no change in program behavior in this PR.


Snowiiii:
thank you @DaniD3v

PR #71, opened 2024-09-01
implement legacy_random and java string hash

kralverde:
Adds the legacy minecraft RNG and the java string hash


Snowiiii:
Thank you @kralverde

PR #72, opened 2024-09-01
Created Identifier to help keep track of plugin resources. 

neeleshpoli:
Created Identifier to help keep track of plugin resources. 
### Use cases: 
- Can be used in the event handler api to make it easy to keep track of events
- Make it easier to keep track of commands.
- \+ many other uses as well

Serde was added so that structs can be serialized and deserialized when transporting between plugin and server. Currently message pack is used, but protocol buffers should be used in the future due to smaller size and efficiency. The only reason why protocol buffers are not used right now is because it requires some work to setup.


pumpkin-plugin/Cargo.toml
```diff
@@ -5,4 +5,5 @@ edition.workspace = true
 
 [dependencies]
 extism = "1.5.0"
-log.workspace = true
\ No newline at end of file
+log.workspace = true
+serde ={ version = "1.0.209", features = ["derive"] }
```
Snowiiii, line 9: Hey can you may just use serde version `1.0`. I really should move serde into workspace

neeleshpoli: Sure!

neeleshpoli: I can move it into the workspace if you're fine with it

neeleshpoli:
I have moved serde into the workspace as well

Snowiiii:
Okay looks good then, Thank you @neeleshpoli

PR #73, opened 2024-09-02
docs: structure documentation

Erb3:
More can be customized later, but here is an initial structure for the documentation.

image
image


lukas0008:
Looks good :+1:

PR #74, opened 2024-09-03
Networking Refactor Pt. 1

StripedMonkey:
This is the first in a series of refactors to improve the networking stack for Pumpkin.

This first PR takes the first step towards extracting TCP connections from the `server` and client code. Nothing significant here, as I'd like to start taking small steps that make sense to you guys.


pumpkin/src/server/connection.rs
Snowiiii, line 1: I don't think ConnectionHandler is a good name. It does not handle anything

StripedMonkey: The intent is for this type to manage all client connections. I agree it isn't an apt name at the moment. If you would like me to change the name for the moment, I'm open to that. However, the next revisions of this will include putting Tokio connection handles (maybe the next step is to just move the Mio Tokens and connections, depends how much I want to rip the band aid off)

Snowiiii: I think it would be may better to just remove the "ConnectionHandler" for now. I mean its just a wrapper struct. Feel free to reimplememt it in the PR you mentioned

StripedMonkey: I can, the problem is that these PRs touch more files than I like already, and massive refactors are hard to understand and review. I'd much prefer to chunk it out with stubbed (but still functional) code than dropping a massive PR bomb that nobody is going to properly review.

Snowiiii: Have larger changes in parts is better. In this PR the biggest change prbly is the change from Arc<Token> to just Token, Which is fine. ConmectionHandler is not big. Its fine to have it in a seperate PR

StripedMonkey: Migrating to a dedicated Connection handler for all connections is a massive refactor over the mess that is a World right now. That's why I only introduced the type with a small change here lol.

Snowiiii: Why World is a mess?. I think it makes sense that you can add and remove players in a world

StripedMonkey: `World` should not interact with low-level tcp primitives and packet manipulation. The OSI model exists for a reason. Use it.

Snowiiii:
Thanks

PR #75, opened 2024-09-03
Migrate completely to tokio

neeleshpoli:
- [x] Migrate to tokio networking
- [x] Remove mio from dependancies
- [ ] Read packets from the client
- [x] Write packets to the client
- [ ] Read RCON packets from the client
- [x] Write RCON packets to the client
- [x] Make functions that send packets async (see changed files for more details)
- [ ] Make command Apis take async closures

Migrate everything to tokio. Currently mio is used for networking. This PR aims to remove mio and use tokio in its place. This PR also makes a ton of changes to the code base as many functions are converted to async functions. Everything has been migrated to tokio networking and all the need functions are made async but currently packets are not read from the client and rcon client. Reading is not implemented yet because I am not sure on how to properly implement it. I'm open to suggestions and feedback!

<sub>I probably should have opened an issue first<sub>

### DIFF
--- a/pumpkin-core/src/random/gaussian.rs
+++ b/pumpkin-core/src/random/gaussian.rs
@@ -1,18 +1,14 @@
-use super::Random;
+use super::RandomImpl;
 
-pub trait GaussianGenerator: Random {
-    fn has_next_gaussian(&self) -> bool;
+pub trait GaussianGenerator: RandomImpl {
+    fn stored_next_gaussian(&self) -> Option<f64>;
 
-    fn set_has_next_gaussian(&mut self, value: bool);
-
-    fn stored_next_gaussian(&self) -> f64;
-
-    fn set_stored_next_gaussian(&mut self, value: f64);
+    fn set_stored_next_gaussian(&mut self, value: Option<f64>);
 
     fn calculate_gaussian(&mut self) -> f64 {
-        if self.has_next_gaussian() {
-            self.set_has_next_gaussian(false);
-            self.stored_next_gaussian()
+        if let Some(gaussian) = self.stored_next_gaussian() {
+            self.set_stored_next_gaussian(None);
+            gaussian
         } else {
             loop {
                 let d = 2f64 * self.next_f64() - 1f64;
@@ -21,8 +17,7 @@ pub trait GaussianGenerator: Random {
 
                 if f < 1f64 && f != 0f64 {
                     let g = (-2f64 * f.ln() / f).sqrt();
-                    self.set_stored_next_gaussian(e * g);
-                    self.set_has_next_gaussian(true);
+                    self.set_stored_next_gaussian(Some(e * g));
                     return d * g;
                 }
             }
--- a/pumpkin-core/src/random/legacy_rand.rs
+++ b/pumpkin-core/src/random/legacy_rand.rs
@@ -1,11 +1,10 @@
 use super::{
-    gaussian::GaussianGenerator, hash_block_pos, java_string_hash, Random, RandomSplitter,
+    gaussian::GaussianGenerator, hash_block_pos, java_string_hash, RandomDeriverImpl, RandomImpl,
 };
 
-struct LegacyRand {
+pub struct LegacyRand {
     seed: u64,
-    internal_next_gaussian: f64,
-    internal_has_next_gaussian: bool,
+    internal_next_gaussian: Option<f64>,
 }
 
 impl LegacyRand {
@@ -18,29 +17,20 @@ impl LegacyRand {
 }
 
 impl GaussianGenerator for LegacyRand {
-    fn has_next_gaussian(&self) -> bool {
-        self.internal_has_next_gaussian
-    }
-
-    fn stored_next_gaussian(&self) -> f64 {
+    fn stored_next_gaussian(&self) -> Option<f64> {
         self.internal_next_gaussian
     }
 
-    fn set_has_next_gaussian(&mut self, value: bool) {
-        self.internal_has_next_gaussian = value;
-    }
-
-    fn set_stored_next_gaussian(&mut self, value: f64) {
+    fn set_stored_next_gaussian(&mut self, value: Option<f64>) {
         self.internal_next_gaussian = value;
     }
 }
 
-impl Random for LegacyRand {
+impl RandomImpl for LegacyRand {
     fn from_seed(seed: u64) -> Self {
         LegacyRand {
             seed: (seed ^ 0x5DEECE66D) & 0xFFFFFFFFFFFF,
-            internal_has_next_gaussian: false,
-            internal_next_gaussian: 0f64,
+            internal_next_gaussian: None,
         }
     }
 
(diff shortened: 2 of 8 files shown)


neeleshpoli:
I'll take a look in a few hours! Thanks!

pumpkin/src/main.rs
```diff
@@ -111,105 +96,61 @@ fn main() -> io::Result<()> {
             });
         }
         loop {
```
neeleshpoli, line 98: Although this code should work, I have no idea why it doesn't. The packets are being read but aren't being handled for some reason.

Snowiiii: This code is bit older. I moved to a new design where i only actullay read the packets and process them in the main. See https://github.com/Snowiiii/Pumpkin/blob/master/pumpkin/src/main.rs#L157. When you only process in the poll client function no player packets will be handled

neeleshpoli: I see, I'll take a look. Thank you!

neeleshpoli: Also yep I'm aware that no player packets are handled, I'm just starting by trying to get status packets working.

Snowiiii: The code should work then

neeleshpoli: Thats really weird, it's not working for some reason. I'll continue trying to find the problem.

kralverde: theres not a deadlock in there is there?

neeleshpoli: Nope, there shouldn't be a deadlock. After further testing, I realized that the client is never sending a status request packet. Not sure what that means as the client should be sending it immediately after the handshake.

kralverde: Are packets being dropped by us between the creation of the client and the polling of the client?

neeleshpoli: Its a possibility that packets are being lost on the polling of the client, I'll do some more testing.

neeleshpoli: Alright I have tested and determined that the packet is not being handled on the server. Looking at Wireshark the client is definitely sending the packet to the server.  The server always handles the handshake (switches states). The server correctly deserializes the send packet. But for some reason when handling the packet, the server never responds to the status packet.

Snowiiii: Had this myself aswell

neeleshpoli: Alright, I think the next step is to use a debugger to see what's going on.

Snowiiii: @neeleshpoli @kralverde 
The recent PR i made solves the locking issue. This was also the reason me and lukas stopped the port to tokio. I already did much debugging when porting to tokio. The code i gave you should work mostly and it did work. I had a Status response and could Join the World.

neeleshpoli: Alright, thank you. I'll finish porting to tokio

neeleshpoli:
Soo apparently parking lot's mutex isn't send + sync, so I've kind of hit a roadblock with porting to tokio.

StripedMonkey:
> Either we would need to go back to tokio mutex (which is send + sync) or we could use channels as @StripedMonkey has suggested instead of a bunch of locking mechanisms (Rwlock, Mutex etc.).

Its difficult to make progress on Rote / Mechanical ports of code with an entirely different pattern and mechanism to them. That is something I've personally felt with complex code, and why I've taken to improving the existing structure of code before submitting major changes like this.

In my opinion channels are part of the solution, but so might other kinds of sync primitives in other places. 

However, at this current stage I believe that the effort should be placed in eliminating synchronization primitives until we have the context and are able to correctly identify places to inject concurrency that it logically fits in.

Tl;dr The Tokio transition is hard because it's so entangled with all the existing code. Baby steps are easier to understand the implications of. 

I'll be giving things a look tomorrow if I can, but I sympathize with your problem.

Asurar0:
> Soo apparently parking lot's mutex isn't send + sync, so I've kind of hit a roadblock with porting to tokio (ofc I only noticed this after a few hours of porting). This in turn would make the client struct not thread safe (not send + sync), which complicates things.
> 
> Either we would need to go back to tokio mutex (which is send + sync) or we could use channels as @StripedMonkey has suggested instead of a bunch of locking mechanisms (Rwlock, Mutex etc.).

Are you sure Rust analyzer is not hallucinating ? `parking_lot::Mutex` and `parking_lot::RwLock` are [Send](https://docs.rs/parking_lot/latest/parking_lot/type.Mutex.html#impl-Sync-for-Mutex%3CR,+T%3E) + [Sync](https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html#impl-Send-for-RwLock%3CR,+T%3E)

`parking_lot` is completely compatible with tokio, it's even recommended for use in their tokio documentation:
> The primary use case for the async mutex is to provide shared mutable access to IO resources such as a database connection. If the value behind the mutex is just data, it’s usually appropriate to use a blocking mutex such as the one in the standard library or [parking_lot](https://docs.rs/parking_lot).
https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use

Asurar0:
> > Either we would need to go back to tokio mutex (which is send + sync) or we could use channels as @StripedMonkey has suggested instead of a bunch of locking mechanisms (Rwlock, Mutex etc.).
> 
> Its difficult to make progress on Rote / Mechanical ports of code with an entirely different pattern and mechanism to them. That is something I've personally felt with complex code, and why I've taken to improving the existing structure of code before submitting major changes like this.
> 
> In my opinion channels are part of the solution, but so might other kinds of sync primitives in other places.
> 
> However, at this current stage I believe that the effort should be placed in eliminating synchronization primitives until we have the context and are able to correctly identify places to inject concurrency that it logically fits in.
> 
> Tl;dr The Tokio transition is hard because it's so entangled with all the existing code. Baby steps are easier to understand the implications of.
> 
> I'll be giving things a look tomorrow if I can, but I sympathize with your problem.

For what it is worth. I'm waiting for this branch to finish to start migrating some components into Actors. Cleaning the architecture incrementally

neeleshpoli:
> Are you sure Rust analyzer is not hallucinating ? `parking_lot::Mutex` and `parking_lot::RwLock` are [Send](https://docs.rs/parking_lot/latest/parking_lot/type.Mutex.html#impl-Sync-for-Mutex%3CR,+T%3E) + [Sync](https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html#impl-Send-for-RwLock%3CR,+T%3E)
I misread the error by rust analyzer. Its complaining that the mutexgaurd isn't send + sync. In this case, I was awaiting on packet decoding (which I rewrote and simplified, not pushed yet), which requires the struct to be send + sync. What I'll probably end up doing is making the packetdecoder/encoder thread safe so that it doesn't need to be wrapped in a mutex.

> I've personally felt with complex code, and why I've taken to improving the existing structure of code before submitting major changes like this.
I've simplified the networking code a lot so it should make it more manageable 👍

PR #77, opened 2024-09-04
Add S->C keep alive packets

kralverde:
Sends keep alive packets to the client every 15 seconds like the notchian server does


Snowiiii:
Hey. I wanted to implement this also myself. I really don't like the approche you made here with. a new thread. I would rather first implement a tick system and then tick every player connection. You can look at vanilla code for reference. they also kick the player if they don't responde with keepalive for 15sec. For the tick system i would consider a simple Trait which has an tick() function

kralverde:
That's fair, I just got tired of getting timed out when testing things.

Snowiiii:
Yeah. I can understand. but still want the tick system first before merging this. Feel free to open a new PR for the tick system and later refactoring this one to use the new system. Ofc you don't have to

Bryntet:
IMO, Keep alive should *not* be tied to ticks. 

Say the server is having really low TPS for some reason, you still want to tell each client to stick around.

I like the idea of having it on a separate thread.

PR #81, opened 2024-09-07
Command system rewrite

Snowiiii:


pumpkin/src/commands/cmd_help.rs
lukas0008, line 1: Don't keep commented code, i mean this is what git is for, if you want an old version, use the history

lukas0008 (commented):
Overall very nice

Bryntet:
I understood it as if this would allow async within commands, but it doesn't seem like the case? Maybe I'm missing something.

pumpkin/src/commands/cmd_help.rs
Snowiiii: Yeah. I just knew that i would reimplement it soon and im too lazy to search trough commits

lukas0008: > Don't keep commented code, i mean this is what git is for, if you want an old version, use the history

Resolved by 870059a31762289e774199cb61880c6fce1b77cb

Snowiiii:
> I understood it as if this would allow async within commands, but it doesn't seem like the case? Maybe I'm missing something.

I said async is now way easier to implement. Be now can actually implement it without big problems if there where no async clouseres thing. Using the old approche async would be impossible

Bryntet:
Anyways, other than me still wanting async in commands, this looks good to me.

Snowiiii:
> Anyways, other than me still wanting async in commands, this looks good to me.

No worries. It will come

PR #84, opened 2024-09-08
Cleanup Server struct

lukas0008:
This cleanup's goal is to make all `Server` functions only require &self. This means that `Server` can now be a `Arc<Server>` instead of `Arc<Mutex<Server>>`, which solves a lot of deadlock problems.


Snowiiii:
Looks good to me. Server was mutable in Client mostly because of the old golden days where Players and Clients where not separate and i needed to have mutable Server in Clients(Players)

PR #87, opened 2024-09-09
Add a general error trait for simplified use of Pumpkin errors

Bryntet:
This PR aims to begin providing a general framework for errors, so that they can provide their own logging severity, and additional attributes such as if the client should get kicked or not.


Bryntet:
Something I'd like to see in the future, is `PumpkinError` somehow helping with backtracing. I am unsure of how to do that though

lukas0008:
I am not sure what this pr is about, is this for errors that the plugins might send back?

Snowiiii:
Hey @Bryntet. Im not sure if its really necessary to have this. I see no problem of the current implementation and just check kick the player instead of returning an Err

Bryntet:
It's mainly because, right now, in player packet handler, any error that isn't `DeserializerError` gets unwrapped (i.e. my `InventoryError`) this PR aims to make a standardised trait that we implement for all errors, so we can always return `Box<dyn PumpkinError>`

Snowiiii:
> It's mainly because, right now, in player packet handler, any error that isn't `DeserializerError` gets unwrapped (i.e. my `InventoryError`) this PR aims to make a standardised trait that we implement for all errors, so we can always return `Box<dyn PumpkinError>`

We should check if the client input is correct or not. When not you can kick the player yourself, I don't see any issues with the current way its implemented

Bryntet:
> We should check if the client input is correct or not. When not you can kick the player yourself, I don't see any issues with the current way its implemented

How do you suppose that we do this when we need to return a dynamic error (`Box<dyn std::error::Error>`) and the error might not be so bad that you need kick?

Snowiiii:
@Bryntet Fix CI

Bryntet:
Okay, I'm at a social event rn but I will fix CI when I get home.

pumpkin/src/main.rs
```diff
@@ -33,6 +33,7 @@ use std::time::Duration;
 pub mod client;
 pub mod commands;
 pub mod entity;
+mod error;
```
Snowiiii, line 36: I think this should be public

Bryntet: Done.

PR #88, opened 2024-09-09
Solve Pumpkin's locking problem

Snowiiii:
Finally. A Pumpkin without locks Problems.

This also allows the Tokio port to work.


pumpkin/src/commands/cmd_echest.rs
```diff
@@ -27,7 +27,8 @@ pub(crate) fn init_command_tree<'a>() -> CommandTree<'a> {
                     }
                 }
             }
-            player.open_container(server, "minecraft:generic_9x3");
+            // TODO
+            // player.open_container(server, "minecraft:generic_9x3");
```
Bryntet, line 31: Don't break my code pls

PR #89, opened 2024-09-10
Networking Refactor Pt. 2

StripedMonkey:
This PR is the second step towards a slightly improved networking API.

The majority of the changes present here are simply extracting discrete components into their own types. This also improves and pushes packet creation into types which actually have the information to generate the required information.

I would describe the current method of building packets as "outside in", when it should be "inside out" for easier refactoring. The high level code in the `Server` struct shouldn't be managing low level information like packet creation, as it makes everything significantly more unwieldy.

Names are bikeshedded since that seemed to be a point of contention last time.


Snowiiii:
What does this improve?. Wrapping things into their own structs is not a real improvement to me

StripedMonkey:
We agree that the current network design is difficult to maintain, and contains a lot of deadlocking code yes? Do we agree that there are principles of software design that might be valuable to take inspiration from to write maintainable code? 

Surely we can apply [SOLID](https://en.wikipedia.org/wiki/SOLID) design principles to code to make it better to use. The [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) or [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) principles suggest that a single struct or object should have a single concern or task in maintainable code. Refactoring or rewriting any one of these new structs requires less effort than it took to do so in the first place. It touches fewer pieces of code, and has a well encapsulated data and a simple and clear interface which improves our ability to refactor and add new features later.

That's not to say that we should follow SOLID principles blindly, after all Rust isn't an Object Oriented Programming language. Things like [Responsibility Driven Design](https://en.wikipedia.org/wiki/Responsibility-driven_design) exist more for Data Driven languages too. Personally I'm more of a fan of RDD than SOLID, or even DDD but we all have our favorites. The point is that these guiding principles inform our software design and lend themselves to certain types of programming problems and refactoring. None of them are perfect in every situation, but any of them is a hell of a lot better than now.

A prime example of where this refactor improves the situation, is that now it's trivial for actual caching to be done on the server status reliably. Since its contents are no longer `pub` and directly accessed, a simple `update_status()` can be written to update the cached status, instead of having to worry about any of the places the status may have been directly updated without interacting with the previous API interface.

I encourage you to learn more about actual software development practices and why they're useful and important. The current `Server` struct very much resembles a [God Object](https://en.wikipedia.org/wiki/God_object) which is problematic for a reason.

pumpkin-config/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition.workspace = true
 
 [dependencies]
 pumpkin-core = { path = "../pumpkin-core" }
+pumpkin-protocol = { path = "../pumpkin-protocol" }
```
Snowiiii, line 8: Pumpkin config should not have the Protocol implemented

pumpkin-config/src/resource_pack.rs
```diff
@@ -25,6 +27,21 @@ impl ResourcePackConfig {
...
+            Some(TextComponent::text(&self.prompt_message))
+        };
+        CConfigAddResourcePack::new(
+            uuid::UUID::new(self.resource_pack_url.as_bytes()),
+            &self.resource_pack_url,
+            &self.resource_pack_sha1,
+            self.force,
+            prompt_message,
+        )
+    }
```
Snowiiii, line 44: Pumpkin Config should not have to use any Packets at all

pumpkin/src/server/bikeshed_server_listing.rs
```diff
@@ -0,0 +1,93 @@
...
+use image::GenericImageView as _;
+use pumpkin_config::{BasicConfiguration, BASIC_CONFIG};
+use pumpkin_protocol::{
+    client::{config::CPluginMessage, status::CStatusResponse},
+    Players, Sample, StatusResponse, VarInt, Version, CURRENT_MC_PROTOCOL,
+};
+
+use super::CURRENT_MC_VERSION;
+
+pub struct BikeShedServerListing {
```
Snowiiii, line 13: `cached_server_brand` has nothing to do with Server Listing

pumpkin-protocol/src/uuid.rs
```diff
@@ -11,3 +11,9 @@ impl Serialize for UUID {
         serializer.serialize_bytes(self.0.as_bytes())
     }
 }
+
+impl UUID {
+    pub fn new(data: &[u8]) -> Self {
+        Self(uuid::Uuid::new_v3(&uuid::Uuid::NAMESPACE_DNS, data))
+    }
```
Snowiiii, line 18: This code should be the equivalent to Java's `UUID::nameUUIDFromBytes`. I think its wrong to put this as an new

pumpkin-config/src/resource_pack.rs
StripedMonkey: The alternative is a new type wrapper around the config in the `pumpkin` crate. While I agree there should be strict dependency rules around crates, (after all I was the one who proposed making such rules), this is the location which provides exactly enough information to produce this packet. Refactoring either the packet itself or the resource pack data would require changes from both, so coupling the two together means you only have to change two locations instead of three.

If you actually want me to create a newtype for this I can.

pumpkin/src/server/bikeshed_server_listing.rs
StripedMonkey: I'm afraid I don't understand what you mean. The wiki clearly shows that the server status is a part of the [Server List Ping Protocol](https://wiki.vg/Server_List_Ping). If you disagree with this, why? What do you believe it should be called instead?

Additionally, where is branding given to players outside of the server listing?

pumpkin-protocol/src/uuid.rs
StripedMonkey: Hmm. I agree somewhat. This wrapper is overly generic for a "UUID", but this is it's only usage right now. Do you believe there's a better suited name for it?

pumpkin-config/src/resource_pack.rs
lukas0008: > The alternative is a new type wrapper around the config in the `pumpkin` crate. While I agree there should be strict dependency rules around crates, (after all I was the one who proposed making such rules), this is the location which provides exactly enough information to produce this packet. Refactoring either the packet itself or the resource pack data would require changes from both, so coupling the two together means you only have to change two locations instead of three.
> 
> If you actually want me to create a newtype for this I can.

you can drop the `pumpkin-protocol` dep, then add `pumpkin-config` in protocol, then you can impl `AsRef<CConfigAddResourcePack>` or something similar, maybe there is a better way to do this though.

pumpkin-protocol/src/uuid.rs
lukas0008: Imo using `new` is not bad in this situation, but you should at least add a doc comment explicitly saying that it is using `uuid::Uuid::NAMESPACE_DNS`, since that might be misleading otherwise.

pumpkin-config/src/resource_pack.rs
StripedMonkey: I would tend to believe that's an even worse structuring to have the low level protocol API dependent on the high level server configuration.

In terms of layers of abstraction, it should be

Protocol -> Config -> pumpkin

Having 

Config -> Protocol -> pumpkin

Puts config much further down the dependency tree.

Why do you believe it's more appropriate to have the config class as a protocol dependency?

Ostensibly, the protocol will change little, if any, and only additively. Using it as a foundational crate for everything else makes significantly more sense than the other way around.

pumpkin/src/server/bikeshed_server_listing.rs
Snowiiii: The brand has nothing to do with the Status/Ping. The brand is send in the Config state. Every common Code editor should also let you see all used references in code

pumpkin-protocol/src/uuid.rs
StripedMonkey: I don't believe a Namespaced uuidv3 UUID is the sole UUID type in pumpkin. It should change regardless of the mechanism. Considering that this is only a newtype wrapper around an existing UUID type you should either completely encapsulate the underlying UUID type or not encapsulate it at all. Otherwise there's not much gained by a newtype wrapper here.

pumpkin/src/server/bikeshed_server_listing.rs
StripedMonkey: That's fine then. For exactly this reason I bike shedded the name.

pumpkin-config/src/resource_pack.rs
lukas0008: > Protocol -> Config -> pumpkin
> 
> Having
> 
> Config -> Protocol -> pumpkin
> 
> Puts config much further down the dependency tree.

I would argue that having config down the chain is better, protocol will most likely depend on config anyway, config will also contain lower-level configuration from features.toml, which alex wants to be able to configure **everything** with.

> Why do you believe it's more appropriate to have the config class as a protocol dependency?

Its because protocol depends on a lot of things, which would mean that if config depends on protocol, those things would not be able to depend on config.

> Ostensibly, the protocol will change little, if any, and only additively. Using it as a foundational crate for everything else makes significantly more sense than the other way around.

Thats true in theory, but for packets where serialization is more complicated, when we change the internal data structure, we will also need to update the serialization.

---

Imo config should only ever depend on core, otherwise it will be annoying to use it, and pumpkin-protocol should only be a dep in the main crate, this is not because of some rule, but due to how everything is currently structured.

If you think theres a better solution then im open to suggestions!

PS: please dont just suggest switching all things around (from pumpkin-protocol depending on everything to everything depending on pumpkin-protocol), that is not a better solution, just a different one, and it also makes it impossible to compose packets using types from other crates in the packets

pumpkin-protocol/src/uuid.rs
lukas0008: Yeah alr, but just use doc comment if this stays as it is

pumpkin-config/src/resource_pack.rs
StripedMonkey: > Its because protocol depends on a lot of things, which would mean that if config depends on protocol, those things would not be able to depend on config.

This is just not true. The protocol is defined by Mojang and the Minecraft API. There is nothing to configure. These are fixed types that are not configurable. You can build a packet or message based on what you may care about, but the inputs and serialization are not configured.

Put another way, there is only one way to write a packet to a Minecraft client, and there is only one way a Minecraft server will receive that packet from the client. A packet's shape, or layout will never, and can never change based on server configuration. The *only* way you can configure a server is to decide what packets you will send or not.

Serialization doesn't change based on configuration. You just change *what* you serialize based on configuration. Something which belongs outside of the `protocol` crate, because it's not a part of the protocol itself.

Everything *but* protocol types are configurable here.

It might be good for you to understand what "ostensibly" means.

lukas0008: > This is just not true. The protocol is defined by Mojang and the Minecraft API. There is nothing to configure. These are fixed types that are not configurable. You can build a packet or message based on what you may care about, but the inputs and serialization are not configured.
> 
> Put another way, there is only one way to write a packet to a Minecraft client, and there is only one way a Minecraft server will receive that packet from the client. A packet's shape, or layout will never, and can never change based on server configuration. The only way you can configure a server is to decide what packets you will send or not.

but pumpkin-protocol does include serialization, it was agreed upon this way, so that serialization isnt placed around in random places around the codebase.
also, some packets cant be serialized from their serde layout, so using only types that are described in wiki.vg is impossible, idk if youve ever read the wiki.vg, but its filled with "which varies by context", "depends on context", "must be known from the context"

> Serialization doesn't change based on configuration. You just change *what* you serialize based on configuration. Something which belongs outside of the protocol crate, because it's not a part of the protocol itself.

alr, but pumpkin-protocol depends on other crates that will depend on config so you cant do it anyway

> It might be good for you to understand what "ostensibly" means.

Communication is a skill in and of itself, using English words that are archaic or uncommon isn't adroit in most situations, so using a *fancy* word just to use them is actually very counterproductive. 
also fyi i know what it means, if you had actually read what i wrote comprehensively, youd know that i disagreed with what you have said

StripedMonkey: > but pumpkin-protocol does include serialization, it was agreed upon this way, so that serialization isnt placed around in random places around the codebase.
also, some packets cant be serialized from their serde layout, so using only types that are described in wiki.vg is impossible, idk if youve ever read the wiki.vg, but its filled with "which varies by context", "depends on context", "must be known from the context"

Nothing said here is contradicting anything I said. I agree that pumpkin-protocol should handle serialization of packet types. That's what a protocol is. It's absolutely true that some packets are context dependent, and cannot be serialized without knowing, for example, connection state. But nothing said here argues otherwise.

PR #91, opened 2024-09-11
Improved synchronization primitives

Asurar0:
> [!WARNING]
> This pull request requires review as it has a high likelihood of introducing a bug and should be thoroughly examined before approval.

This PR includes the following changes:

- Added `parking_lot` and `crossbeam` dependencies.
- Replaced all `std::sync::Mutex` by `parking_lot::Mutex` and refactored implementation accordingly.
- Replaced all `Mutex<T> where T: Copy` by `crossbeam::AtomicCell<T>` and refactored implementation accordingly.

Significant portions of the code involve setting Mutex on frequently updated fields and certain types that implement Copy. This optimization is expected to reduce locking overhead and improve overall performance.

## Method

- Types that do not implement `Copy` are now wrapper into `parking_lot::Mutex` and the overall `.unwrap()` or `expect()` around poisoning has been removed.
- Types that implements `Copy` trait has been wrapped into `AtomicCell` and the following conversions have been made: `*self.field.lock().unwrap() = value` -> `self.field.store(value)`, `*self.field.lock.unwrap()` -> `self.field.lock()` (There are exceptions to this rule as some methods required to update the data in middle of their routine)

## Bug

I've done my best to understand the code, but the number of changes makes it likely I've introduced an issue. Please review and test thoroughly. However, I am highly confident that the changes will result in performance improvements and are worth the time spent on reviewing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -535,6 +535,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1137cd7e7fc0fb5d3c5a8678be38ec56e819125d8d7907411fe24ccb943faca8"
+dependencies = [
+ "crossbeam-channel",
+ "crossbeam-deque",
+ "crossbeam-epoch",
+ "crossbeam-queue",
+ "crossbeam-utils",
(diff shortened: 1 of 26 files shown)


Snowiiii:
Looks good. Thanks @Asurar0

PR #95, opened 2024-09-15
World persistence and .MCA region files

Asurar0:
Related to: https://github.com/Snowiiii/Pumpkin/issues/92

This pull request is intended to introduce a ChunkStorage actor, which will be responsible for retrieving and storing chunks to and from disk. The implementation of this feature is a crucial step towards achieving world persistence, a key requirement for Pumpkin to function as a vanilla server in the future.

Current state:
- [ ] Implement serialization of ChunkData (temporary at the moment)
- [x] Implement skeleton ChunkStorage trait 
- [x] Implement LMDB
- [ ] Implement .MCA storage
    - [ ] Implement Anvil format parsing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -148,6 +148,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


Snowiiii:
> Could there perhaps be Linear region file format support planned?

Hey, What do you mean by Linear region file?. We definitly want to import and export Vanilla world files

Bryntet:
That is really cool, I think we most definitely should support it.

Snowiiii:
@Asurar0 any progress ?

Snowiiii:
> Could there perhaps be Linear region file format support planned?

Now it's supported :D

PR #96, opened 2024-09-15
Migrate to tokio

neeleshpoli:
- [ ] Migrate to tokio networking
- [x] Remove mio from dependancies
- [ ] Read packets from the client
- [x] Write packets to the client
- [ ] Read RCON packets from the client
- [ ] Write RCON packets to the client
- [ ] Make functions that send packets async (see changed files for more details)
- [ ] Make command Apis take async closures

Migrate everything to tokio. Currently mio is used for networking. This PR aims to remove mio and use tokio in its place. This PR also makes a ton of changes to the code base as many functions are converted to async functions. Everything has been migrated to tokio networking and all the need functions are made async but currently packets are not read from the client and rcon client. Reading is not implemented yet because I am not sure on how to properly implement it. I'm open to suggestions and feedback!
I made a new PR as I accidentally closed the orginal one (#75) because I reverted my orginal changes to add new ones.
<sub>I probably should have opened an issue first<sub>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1567,7 +1567,6 @@ checksum = "80e04d1dcff3aae0704555fe5fee3bcfaf3d1fdf8a7e521d5b9d2b42acb52cec"
 dependencies = [
  "hermit-abi",
  "libc",
- "log",
  "wasi",
  "windows-sys 0.52.0",
 ]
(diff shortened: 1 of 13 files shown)


neeleshpoli:
Unfortunately, I will be busy for a little while, so if anyone else wants to complete the PR, they can go ahead. Feel free to ask any questions and I can answer.

PR #98, opened 2024-09-17
implement density functions for terrain generation

kralverde:
WIP

This PR has the end goal of generating plains biomes


kralverde:
I bit off more than I can chew. Here is another stepping stone towards chunk generation: the density functions. I will keep on working towards terrain generation.

pumpkin-core/src/math/mod.rs
```diff
@@ -20,8 +21,20 @@ pub fn squared_magnitude(a: f64, b: f64, c: f64) -> f64 {
...
     coord >> 4
 }
+
+pub fn squared_hypot(a: f64, b: f64) -> f64 {
+    a * a + b * b
+}
+
+pub fn hypot(a: f64, b: f64) -> f64 {
+    squared_hypot(a, b).sqrt()
+}
```
Snowiiii, line 40: Rust already provides hypot functions, Why use own instead ?

pumpkin-world/src/world_gen/noise/density/end.rs
```diff
@@ -0,0 +1,78 @@
...
+                    let r = 100f32 - (h * h + q * q).sqrt() * g;
+                    let s = r.clamp(-100f32, 80f32);
+
+                    f = f.max(s);
+                }
+            }
+        }
+
+        f
+    }
```
Snowiiii, line 55: Is is vanilla ?

Snowiiii:
I see you almost have `function1`, `function2` `function3`... everywhere. Did you got this code from somewhere or why var names are so weird?

kralverde:
Straight from Java lol

pumpkin-world/src/world_gen/noise/density/end.rs
kralverde: Yes

pumpkin-core/src/math/mod.rs
kralverde: I'll use rust's

Snowiiii:
I don't like if someone just copies Java code and then even say that he don't know what the code does. So you said your goal is to have the Plains biome generation working. Do you need everything here for this, Or is something missing ?

Snowiiii:
I would like to have good function names at least

kralverde:
Functions have been renamed

Snowiiii:
noice, can you may fix the conflicts ? 
then it should be ready to merge

kralverde:
rebased, should be good

PR #100, opened 2024-09-18
Move to own NBT crate

Snowiiii:
**What Advantages this has:**
- If a new game update releases and breaks/changes NBT System, We don't depend on other crates to update
- We may can fix this https://github.com/Snowiiii/Pumpkin/blob/cb60b02da45615c44147b09fa2ad31bda1b7a247/pumpkin-core/src/text/mod.rs#L142
- The NBT format does not support booleans, You have to write 1 or 0 u8 for a boolean. fastnbt just crashes when using booleans without any info. This took me 2 Days to debug when i first created the registry. I definitely want to support booleans, This will make Coding easier
- We may can make our NBT crate faster than fastnbt

(Not fully serious)
- Learn new things
- Have Fun


StripedMonkey:
I advocate keeping this a module in `pumpkin-protocol`. In principle this is just a data format we should be (de)serializing to and from. We don't want to tie any structs to a nbt-specific data type if at all possible.

Snowiiii:
Disclaimer: `pumpkin-world` still uses fastnbt **for now**. I have problems implementing Long Arrays, But Serialization using our own crate in for example `pumpkin-registry` works great and also allows us to use booleans

PR #101, opened 2024-09-18
Add `block_id!` macro

lukas0008:
Added a macro which allows getting `BlockId` from blocks.json at compile-time.

Basic usage:
```rust
let grass_block = block_id!("minecraft:grass_block");
```

with block properties:
```rust
let button = block_id!(
    "minecraft:acacia_button",
    "facing" = "north",
    "powered" = "true",
    "face" = "floor"
);
```


Snowiiii:
Hey, We need this for Chunks right?. Or for what usecases is this?

Snowiiii:
Can't we refactor BlockId::new also use Compile time ?

lukas0008:
but... how?

StripedMonkey:
In principle it's possible to build the data as a bindgen-esque crate. I would advocate for not embedding block types/properties directly into the built binary though. Pre-generating them and caching the results somewhere feels like a better approach.

lukas0008:
I do store the parsed blocks.json statically, in the same build it only has to be parsed once, so the compile length hit should be negligible, since each macro is literally just a hashmap lookup with a bit of logic to parse the input.

PR #103, opened 2024-09-20
Create Item Entities and make them interactable 

Bryntet:
This PR aims to add the following:

- [x] ItemEntities
- [x] Player Dropping of items
- [x] Entity collision checking
- [x] Block breaking drops item
- [x] Player able to pick up item


Bryntet:
To clarify, collisions *kinda* work. They work fully with the ground, but they cancel out velocity, and don't act exactly as they should.

Dropped Items also do not merge into one yet.

Snowiiii:
Hey @Bryntet any updates on this?

Bryntet:
Currently stuck on getting collisions to work exactly like vanilla. Chunks saving to ram changes is also blocking the Block-breaking drops items.

Bryntet:
Block dropping now partially works.

pumpkin-protocol/src/client/play/c_entity_metadata.rs
```diff
@@ -5,13 +5,13 @@ use crate::VarInt;
...
 #[packet(0x58)]
-pub struct CSetEntityMetadata<T> {
-    entity_id: VarInt,
-    metadata: Metadata<T>,
-    end: u8,
+pub struct CSetEntityMetadata<T: Serialize> {
+    pub entity_id: VarInt,
+    pub metadata: Metadata<T>,
+    pub end: u8,
 }
```
Snowiiii, line 12: I see that you like to make the fields public, But its more of an "interal convention" in pumpkin to use the new function

pumpkin/src/entity/item.rs
```diff
@@ -0,0 +1,304 @@
...
+
+        let empty = Vector3 {
+            x: 0.,
+            y: 0.,
+            z: 0.,
+        };
+
+        let entity = Arc::new(Entity {
+            entity_id: server.new_entity_id(),
+            uuid: Uuid::new_v4(),
```
Snowiiii, line 53: There should be a check if the UUID is not matching any existing player UUID on the server, I also wonder what to do if a new player join which has the same UUID as an entity

pumpkin/src/entity/item.rs
```diff
@@ -0,0 +1,304 @@
...
+    let yaw_cos = f64::cos(f64::from(player.yaw.load()) * (PI / 180.0));
+    let random_angle = random_float() * (2.0 * PI);
+    let random_offset = 0.02 * random_float();
+
+    Vector3 {
+        x: (-yaw_sin * pitch_cos).mul_add(0.3, f64::cos(random_angle) * random_offset),
+        y: (-pitch_sin).mul_add(0.3, (random_float() - random_float()).mul_add(0.1, 0.1)),
+        z: (yaw_cos * pitch_cos).mul_add(0.3, f64::sin(random_angle) * random_offset),
+    }
+}
```
Snowiiii, line 304: Is this from vanilla ?

pumpkin/src/entity/mod.rs
```diff
@@ -47,6 +55,9 @@ pub struct Entity {
     /// Indicates whether the entity is on the ground (may not always be accurate).
     pub on_ground: AtomicBool,
 
+    /// Indicates whether the entity is inside of ground, and needs to be pushed out.
+    pub in_ground: AtomicBool,
```
Snowiiii, line 59: I assume this generally checks if a Entity is in blocks not only ground, right?, Then a better name would be maybe `inside_block`

pumpkin/src/entity/to_packet.rs
```diff
@@ -0,0 +1,28 @@
...
+        velocity.y,
+        velocity.z,
+    )
+}
+
+impl From<&Entity> for CSpawnEntity {
+    fn from(entity: &Entity) -> Self {
+        spawn_entity_with_data(entity, None)
+    }
+}
```
Snowiiii, line 28: I really don't like this, First you have a whole new file for only wrapping around some values, Also i don't think its that hard to write this every time, If you think this is really necessary i would much more prefer this in the entity file

Snowiiii:
Closing due to inactivity. Feel free to re-open if the work is still needed.

PR #104, opened 2024-09-22
Add keep alive packets

Bryntet:
This PR adds keep alive packets so that clients don't disconnect after 20s of inactivity.


Bryntet:
Adding keep alive packets, required me to wrap `Client` in an Arc, that is the main thing that this PR touches outside of sending the packet.

PR #105, opened 2024-09-26
Network Refactor Pt. 3: Introduce a packet Codec and Connection type

StripedMonkey:
This introduces a series of Codecs for encoding and decoding packets, as well as a Connection<T> type which frames streams with encryption and protocol framing.

I have withheld hooking this up to anything, in favor of some tests to show that it works due to the large burden of refactoring code to use it. Partial refactors containing code hooking this up can be found in a dump commit on my fork of Pumpkin.


Snowiiii:
So what benefits this gives us?

StripedMonkey:
This makes connections composable, and encapsulates connections as frameable streams of any type. Encryption is handled transparently.

This is a part of my continued efforts to make the networking stack better through more encapsulation and rigid API boundaries.

All the code here does is handle taking an incoming TCP (or any `impl AsyncStream`) Connection and turn them into `UncompressedPacket`s to then be serialized.

StripedMonkey:
This PR is not designed to actually add features that did not exist before, but make it easier to maintain and extend the existing code. I'm trying to get to the point where working with network connections is completely abstracted out of the core game logic like it currently is.

Bryntet:
That crate paste has the same functionality as unstable concat_idents 👍

StripedMonkey:
It saves like 10 lines of repetitive code, but honestly I don't think it's worth the extra dependency. It's nice to have but not necessary here.

pumpkin-protocol/src/server/handshake/mod.rs
```diff
@@ -1,25 +1,13 @@
...
+use crate::{ConnectionState, VarInt};
 
 #[packet(0x00)]
+#[derive(Deserialize)]
 pub struct SHandShake {
     pub protocol_version: VarInt,
     pub server_address: String, // 255
     pub server_port: u16,
     pub next_state: ConnectionState,
 }
```
Snowiiii, line 13: Hey, The only reason i used manuel Deserialization here is because the server address has a max length of 255. There where plans of implementing maybe some kinda struct called `BoundedString`, Not sure tough how

StripedMonkey: I've been playing with a solution to this problem for fixed length bitsets or similar, but not addressed it in depth.

In theory, it's possible to make a newtype wrapper like
```rust
struct BoundedLengthString<const LEN: usize>(String);
```

and implement a string visitor that restricts the size, implementing `From`/`Into` would allow you to use 
```
#[serde(try_from = BoundedLengthString<255>)]
pub server_address: String
```

but there are some considerations to make regarding strings in general. The current implementation is actually incorrect in that it assumes string length is in bytes. However, according to the wiki Strings are actually prefixed with their length in *characters*, making the implementation faulty on Non-ASCII/english characters.

lukas0008: @Snowiiii Imo @StripedMonkey 's solution is cleaner, since we dont need to cast to a different type when using the packet.

---

@StripedMonkey 
> but there are some considerations to make regarding strings in general. The current implementation is actually incorrect in that it assumes string length is in bytes. However, according to the wiki Strings are actually prefixed with their length in characters, making the implementation faulty on Non-ASCII/english characters.

^^ regarding this, i dont think thats true tho, on wiki.vg it says that a string is a `UTF-8 string prefixed with its size in bytes as a VarInt.`, the size here does not refer to amount of characters, but amount of bytes. This can be easily checked, by simply sending a chat message in pumpkin with the current implementation, with characters outside of the single byte range of UTF-8
image

StripedMonkey: Sorry, I missread, Max string length in characters is what I was actually reading. So the limits provided in the wki aren't in bytes.

Snowiiii:
Hey @StripedMonkey, I would love to merge or close this PR soon. Do you have any plans finishing the PR soon ?

StripedMonkey:
Life has been busy for the last bit, so I've not dedicated much time to this. 

The largest outstanding issue here is that I depend on things implementing `Serialize` and `Deserialize` and not everything actually implements those traits, i.e chunk serialization is not actually using serde, and it's just a lot of work to get everything correct and reimplemented using serde.

It should be relatively straightforward to do, but it's still time consuming.

If you would like to take my code and work on it, maintainers should have commit access to the pr. Otherwise feel free to copy what you like, and discard the rest.

I have no timeline at this point. Not this weekend for sure.

PR #106, opened 2024-09-28
Create Block, BlockCategory enums. Create block state struct.

kralverde:
This PR implements enums for Block and BlockCategories via macro which will be necessary for terrain gen. Additionally adds a BlockState struct to include these enums.


Snowiiii:
There where discussions in the Discord about an Large Blocks Enum, Like @lukas0008 once did. The Problem was compile time. I generally would really like a big enum with all blocks the game has, This would be also a more error safe approche

kralverde:
I think the issue was create an enum for every single possible state. This just has blocks and categories. Try it out, I don't think compile time is that bad.

kralverde:
Unless anyone finds issue with it it's ready

PR #107, opened 2024-09-29
Add kill command

Alerty2:
This change adds the /kill command to the server. However, it only makes the sound, shows the particles and the player can't move anymore instead of properly dying. This happens because the command calls the kill function on entity/mod.rs which only does that.


Snowiiii:
Hey, Thanks for the PR :D. I currently did not implemented respawning, thats the reason the Player is proberbly stuck. Btw you forgot to format thats the reason CI is mad.

Alerty2:
Perfect, I've just formatted the code, thank you for the advice!

pumpkin/src/commands/cmd_kill.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+pub(crate) fn init_command_tree<'a>() -> CommandTree<'a> {
+    CommandTree::new(NAMES, DESCRIPTION).with_child(
+        argument(ARG_TARGET, consume_arg_target).execute(&|sender, server, args| {
+            let target = parse_arg_player(sender, server, ARG_TARGET, args)?;
+            target.entity.kill();
+            target.send_system_message(
+                TextComponent::text("You have been killed.").color_named(NamedColor::Red),
+            );
```
Snowiiii, line 24: Vanilla does not send a message to the killed target. We also should not

Alerty2:
Ok ,done. I confused and written kick instead of kill but only on the name of the commit.

Snowiiii:
Looks good to me then. We should definitely add support for killing entites as well not only players in the future.

PR #109, opened 2024-10-06
allow assets for docker build

ThePaulo1:
Fix failing build due to assets not found

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -6,6 +6,7 @@
 
 # Allow the source code folders
 !/pumpkin*/
+!/assets
 
 # Dependencies
 !Cargo.lock


Snowiiii:
Hey @ThePaulo1. Looks like this has something to do with docker, You should proberbly mention that the PR. Im not that fimilar with with Docker, The only "asset" we really have is the icon, Which is also Optional. Can you maybe explain what the issue is here?

Snowiiii:
Ah i see. So when assets are included everything works fine?

ThePaulo1:
yes the docker build succeeds when assest are included

Snowiiii:
Great. Then thanks foe the PR @ThePaulo1 👍

PR #111, opened 2024-10-08
expose docker container port

ThePaulo1:
The port of the Pumpkin server needs to be exposed to be accessable from the docker container


Snowiiii:
So the only Problem im seeing here is what when we change the actually port? (which is possible in the config). This is usually done when wanting to have Proxy support

ThePaulo1:
Thats no problem at all because the "expose" in the dockerfile is just a guideline and not actually exposing any ports so others know that this port is usually used. The actual port is set with the docker run "-p" flag so if someone would like to use another port they just have to change the 25565:25565 mapping to the specified port in the config

Snowiiii:
Ah okay, I just looked up the docs. Okay seems fine then. Thanks for the PR

PR #126, opened 2024-10-16
Add Docker Support to the Dev Container

Legonois:
Added the "ghcr.io/devcontainers/features/docker-outside-of-docker:1" feature to the dev container, so that docker build and docker run can be used in development


Snowiiii:
Looks good.
Thanks @Legonois

PR #131, opened 2024-10-16
docs: update Wikipedia link from German to English version

alpaim:
This very small PR updates the Wikipedia link in the Authentication docs page from German to English.

Changes:
- Updated the link for the "Denial of Service" (DoS) article from de.wikipedia.org to en.wikipedia.org


Snowiiii:
Upsi, My bad. Good catch @alpaim. 
Thank you

PR #135, opened 2024-10-16
Optimize floating point operations

julianbraha:
This uses Rust standard library's`mul_add`, which is more accurate and performs more efficiently than the equivalent `(a * b) + c`. Also see [the standard library docs](https://doc.rust-lang.org/stable/std/primitive.f64.html#method.mul_add) for more info.

### DIFF
--- a/pumpkin-core/src/math/mod.rs
+++ b/pumpkin-core/src/math/mod.rs
@@ -18,7 +18,7 @@ pub fn wrap_degrees(var: f32) -> f32 {
 }
 
 pub fn squared_magnitude(a: f64, b: f64, c: f64) -> f64 {
-    a * a + b * b + c * c
+    c.mul_add(c, a.mul_add(a, b * b))
 }
 
 pub fn magnitude(a: f64, b: f64, c: f64) -> f64 {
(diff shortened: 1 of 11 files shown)


Snowiiii:
Great, Was thinking of making such changes my own.
Thank you @julianbraha

Snowiiii:
Seems like, you made a little upsi. A Testing is not passing

julianbraha:
```
failures:

---- random::legacy_rand::test::test_next_triangular stdout ----
thread 'random::legacy_rand::test::test_next_triangular' panicked at pumpkin-core/src/random/legacy_rand.rs:304:13:
assertion `left == right` failed
  left: 79.11264482608077
 right: 79.11264482608078

---- random::xoroshiro128::tests::test_next_gaussian stdout ----
thread 'random::xoroshiro128::tests::test_next_gaussian' panicked at pumpkin-core/src/random/xoroshiro128.rs:351:13:
assertion `left == right` failed
  left: -0.37725128286308096
 right: -0.3772512828630807
```

It's just a minor precision difference in the final digit. This is expected. See this from the standard library docs:
> yielding a more accurate result than an unfused multiply-add.

and this:
> The result of this operation is guaranteed to be the rounded infinite-precision result. It is specified by IEEE 754 as fusedMultiplyAdd and guaranteed not to change.

Bryntet:
Yes, but the randomness has the unit tests there because we want exact parity with Java. So it should be reverted there, or be made optional

PR #137, opened 2024-10-16
Fixed velocity support

Alvsch:
Fixed the support for velocity proxies

### DIFF
--- a/pumpkin-protocol/src/bytebuf/mod.rs
+++ b/pumpkin-protocol/src/bytebuf/mod.rs
@@ -370,7 +370,6 @@ impl ByteBuffer {
         if self.buffer.len() >= len {
             Ok(self.buffer.copy_to_bytes(len))
         } else {
-            dbg!("{:?} < {:?}", self.buffer.len(), len);
             Err(DeserializerError::Message(
                 "Unable to copy bytes".to_string(),
             ))
--- a/pumpkin-protocol/src/bytebuf/mod.rs
+++ b/pumpkin-protocol/src/bytebuf/mod.rs
@@ -370,6 +370,7 @@ impl ByteBuffer {
         if self.buffer.len() >= len {
             Ok(self.buffer.copy_to_bytes(len))
         } else {
+            dbg!("{:?} < {:?}", self.buffer.len(), len);
             Err(DeserializerError::Message(
                 "Unable to copy bytes".to_string(),
             ))
--- a/pumpkin/src/client/mod.rs
+++ b/pumpkin/src/client/mod.rs
@@ -1,10 +1,8 @@
 use std::{
-    io::{self, Write},
-    net::SocketAddr,
-    sync::{
+    collections::VecDeque, io::{self, Write}, net::SocketAddr, sync::{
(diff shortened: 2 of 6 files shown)


pumpkin-protocol/src/bytebuf/mod.rs
```diff
@@ -370,6 +370,7 @@ impl ByteBuffer {
         if self.buffer.len() >= len {
             Ok(self.buffer.copy_to_bytes(len))
         } else {
+            dbg!("{:?} < {:?}", self.buffer.len(), len);
```
Snowiiii, line 373: Please remove that

pumpkin/src/client/mod.rs
```diff
@@ -221,22 +220,24 @@ impl Client {
...
+            pumpkin_protocol::ConnectionState::Status => self.handle_status_packet(server, packet).unwrap(),
             // TODO: Check config if transfer is enabled
             pumpkin_protocol::ConnectionState::Login
             | pumpkin_protocol::ConnectionState::Transfer => {
-                self.handle_login_packet(server, packet).await
+                self.handle_login_packet(server, packet).await.unwrap()
             }
             pumpkin_protocol::ConnectionState::Config => {
-                self.handle_config_packet(server, packet).await
+                self.handle_config_packet(server, packet).await.unwrap()
```
Snowiiii, line 233: -_- We don't want unwraps

pumpkin/src/proxy/velocity.rs
```diff
@@ -60,11 +60,36 @@ pub fn receive_plugin_response(
             ));
             return;
         }
-        // TODO: no unwrap
-        let addr: SocketAddr = buf.get_string().unwrap().parse().unwrap();
+        // TODO: dont default to localhost
+        let addr: SocketAddr = buf.get_string().unwrap().parse()
+            .unwrap_or_else(|_| ([127, 0, 0, 1], 25565).into());
```
Snowiiii, line 65: you probably should get the current address then `BASIC_CONFIG.server_address

Alvsch: that address is the default player address, it panics when you use localhost to connect to the server

Alvsch: didn't come up with a better solution

Snowiiii: then you should be able to use `client.address`.
Seems Paper does that as well: https://github.com/PaperMC/Paper/blob/master/patches/server/0731-Add-Velocity-IP-Forwarding-Support.patch#L202

Alvsch: I was infact wrong about it being localhost, the address we get from velocity just doesn't include a port

Snowiiii: > I was in fact wrong about it being localhost, the address we get from velocity just doesn't include a port edit: should we just default the port to 0?

looks like paper just grabs the current port from the players Socket Address, I think we should do that as well
image

Snowiiii:
Looks good to me. 
Thanks you @Alvsch

PR #140, opened 2024-10-16
Optimize some floating point operations, but maintain parity with Java

julianbraha:
This uses mul_add in some cases, but maintains parity with Java and passes the unit tests.


Snowiiii:
Looks good now.
Thanks @julianbraha

PR #143, opened 2024-10-18
Fix server icons + add server icon to config

kralverde:
This PR adds to config option to set the path to the server favicon


Snowiiii:
Why did you remove the icon by default ?

kralverde:
The previous path was the root "/icon" which is already kind of broken no?

Snowiiii:
> The previous path was the root "/icon" which is already kind of broken no? We also don't provide an image so it doesn't make sense to me to point to a path by default.

Nope, It works. Did you never saw the icon ?

kralverde:
It stopped working a while ago, but the path it checks for is "/icon.png" at the root. If you recreate the server to pumpkin.kralverde.dev, you will also see it has no icon

Snowiiii:
Not sure maybe your right. So what would be a good way to get the icon from the current file. I was may thinking implementing something similar like in my game engine: https://github.com/ventengine/Vent-Engine/blob/master/crates/vent-assets/src/io/file.rs#L6

kralverde:
What's the icon file it should be?

Snowiiii:
oh please man, we don't need a separate file for this one

Snowiiii:
Looks good to me now.
Thank you @kralverde

PR #148, opened 2024-10-18
Add broadcast command

van-sprundel:
Implementation for #15.

Supports [target selectors](https://minecraft.fandom.com/wiki/Commands/say#Arguments).
Also added functions to get `all_players` and `nearest_player`.

### DIFF
--- a/pumpkin/src/commands/cmd_say.rs
+++ b/pumpkin/src/commands/cmd_say.rs
@@ -30,7 +30,7 @@ pub fn init_command_tree<'a>() -> CommandTree<'a> {
                 let message = &format!("[Console]: {content}");
                 let message = TextComponent::text(message).color_named(NamedColor::Blue);
 
-                server.broadcast_message(message.clone());
+                server.broadcast_message(&message);
                 sender.send_message(message);
             } else {
                 sender.send_message(
--- a/pumpkin/src/commands/cmd_say.rs
+++ b/pumpkin/src/commands/cmd_say.rs
@@ -31,7 +31,10 @@ pub fn init_command_tree<'a>() -> CommandTree<'a> {
                 let message = TextComponent::text(message).color_named(NamedColor::Blue);
 
                 server.broadcast_message(&message);
-                sender.send_message(message);
+
+                if !sender.is_player() {
+                    sender.send_message(message);
+                }
             } else {
                 sender.send_message(
                     TextComponent::text("Please provide a message: say [content]")
--- a/pumpkin/src/commands/cmd_say.rs
+++ b/pumpkin/src/commands/cmd_say.rs
@@ -2,6 +2,7 @@ use crate::commands::tree::CommandTree;
 use crate::commands::tree::RawArgs;
 use crate::commands::tree_builder::argument;
 use crate::commands::CommandSender;
+use crate::entity::player::Player;
 use crate::server::Server;
 use pumpkin_core::text::{color::NamedColor, TextComponent};
 
@@ -61,8 +62,7 @@ fn parse_token<'a>(token: &'a str, sender: &'a CommandSender, server: &'a Server
     let result = match token {
         "@p" => {
             if let CommandSender::Player(player) = sender {
-                server
-                    .get_nearest_player_name(player)
+                get_nearest_player_name(player)
                     .map_or_else(Vec::new, |player_name| vec![player_name])
             } else {
                 return token.to_string();
@@ -92,6 +92,17 @@ fn parse_token<'a>(token: &'a str, sender: &'a CommandSender, server: &'a Server
     };
 
     format_player_names(&result)
+}
+
+// Gets the nearest player name in the same world
+fn get_nearest_player_name(player: &Player) -> Option<String> {
+    let target = player.last_position.load();
+
+    player
+        .living_entity
+        .entity
+        .world
+        .get_nearest_player_name(&target)
 }
 
 // Helper function to format player names according to spec
--- a/pumpkin/src/commands/cmd_say.rs
+++ b/pumpkin/src/commands/cmd_say.rs
@@ -10,7 +10,7 @@ const DESCRIPTION: &str = "Sends a message to all players.";
 const ARG_CONTENT: &str = "content";
 
 pub fn consume_arg_content(_src: &CommandSender, args: &mut RawArgs) -> Option<String> {
-    args.pop().map_or(None, |v| Some(v.to_string()))
+    args.pop().map(|v| v.to_string())
 }
 
 pub fn init_command_tree<'a>() -> CommandTree<'a> {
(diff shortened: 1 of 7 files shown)


pumpkin/src/world/mod.rs
```diff
@@ -259,6 +259,22 @@ impl World {
...
+    }
+
+    /// Gets all players
+    pub fn get_players(&self) -> Vec<Arc<Player>> {
+        self.current_players
+            .lock()
+            .values()
+            .cloned()
+            .collect::<Vec<_>>()
+    }
```
Snowiiii, line 276: I don't like copying the Vec here

pumpkin/src/commands/cmd_say.rs
```diff
@@ -0,0 +1,111 @@
...
+            _ => vec![token.to_string()],
+        };
+
+        // formatted player names
+        final_message.push_str(&format_player_names(&result));
+        final_message.push(' ');
+    }
+
+    final_message.trim_end().to_string()
+}
```
Snowiiii, line 97: We already have player arguments implemented, See for example gamemode command or kick command...
You can implement is easily by providing the target as an argument `.with_child(argument(ARG_TARGET, consume_arg_player)` and then parsing it ` let target = parse_arg_player(sender, server, ARG_TARGET, args)?;`

pumpkin/src/commands/cmd_say.rs
```diff
@@ -0,0 +1,111 @@
...
+
+pub fn init_command_tree<'a>() -> CommandTree<'a> {
+    CommandTree::new(NAMES, DESCRIPTION).with_child(
+        argument(ARG_CONTENT, consume_arg_content).execute(&|sender, server, args| {
+            if let Some(content) = args.get("content") {
+                let content = parse_selectors(content, sender, server);
+                let message = &format!("[Console]: {content}");
+                let message = TextComponent::text(message).color_named(NamedColor::Blue);
+
+                server.broadcast_message(message.clone());
```
Snowiiii, line 33: It takes a reference, It should be possible avoiding cloning here

pumpkin/src/server/mod.rs
```diff
@@ -127,6 +130,35 @@ impl Server {
...
+    }
+
+    /// Get all online players
+    pub fn get_online_players(&self) -> impl Iterator<Item = Arc<Player>> + '_ {
+        self.worlds.iter().flat_map(|world| world.get_players())
+    }
+
+    /// Gets the nearest player from the world position
+    pub fn get_nearest_player(&self, target: &Vector3<f64>) -> Option<Arc<Player>> {
+        // TODO respect which world the player is in
```
Snowiiii, line 147: You can easily get the World from a player by calling `player.living_enitity.entity.world`. But i the current plan is to search for Player on all world (e.g. Gamemode, Kick Command...)

pumpkin/src/server/mod.rs
```diff
@@ -127,6 +130,35 @@ impl Server {
...
+        world
+            .current_players
+            .lock()
+            .values()
+            .min_by(|a, b| {
+                let dist_a = distance(&a.last_position.load(), target);
+                let dist_b = distance(&b.last_position.load(), target);
+                dist_a.partial_cmp(&dist_b).unwrap()
+            })
+            .cloned()
```
Snowiiii, line 159: Please stop cloning so much

pumpkin/src/world/mod.rs
van-sprundel: Got it. Changed it to clone just the name string instead of the entire player

pumpkin/src/commands/cmd_say.rs
van-sprundel: The `say` command supports multiple target specifiers, it's not a command argument, but a way the content gets formatted

pumpkin/src/server/mod.rs
van-sprundel: In minecraft, the nearest player has to be in the same world. If that's the plan though, then I can change it

pumpkin/src/commands/mod.rs
```diff
@@ -77,6 +78,7 @@ pub fn default_dispatcher<'a>() -> CommandDispatcher<'a> {
     dispatcher.register(cmd_echest::init_command_tree());
     dispatcher.register(cmd_kill::init_command_tree());
     dispatcher.register(cmd_kick::init_command_tree());
+    dispatcher.register(cmd_say::init_command_tree());
```
StripedMonkey, line 81: This isn't your problem per-se, but When I see large blocks of repetitive code like this my eyes start to glaze over and I stop being able to read.

It sets off alarm bells that the code could be simplified some. It might not be getting to that point right now, but with how large and repetitive this is getting, another abstraction would be nice.

### List Option
Something like 
```rust
fn register_commands(&mut self, commands: impl IntoIterator<C>) 
where C: CommandTree
{
    for command in commands {
        dispatcher.register(command::default());
    }
}

// in default_dispatcher
dispatcher.register_commands(vec![ cmd_pumpkin, cmd_gamemode, ... ]);
```
### Macro Option
```rust
macro_rules! init_commands {
    (($command:ident),*) => { $( dispatcher.register($command::init_command_tree()); )*}
}

// In default_dispatcher
init_commands!(cmd_pumpkin, cmd_gamemode, cmd_stop, cmd_help, cmd_echest, cmd_kill);
```

Say what you will about macros, but as long as they're explained, they make the boilerplate much more legible.

In the current code, `dispatcher.register(...::init_command_tree());` is all noise, and what you're actually reading needs to be that tiny `cmd_say`


The macro option can be implemented as-is, but the list list one requires you to do some more significant refactors. Neither has to be implemented right now, but it's getting there.

pumpkin/src/world/mod.rs
```diff
@@ -259,6 +262,34 @@ impl World {
...
+    /// Gets all players
+    pub fn get_player_names(&self) -> Vec<String> {
+        self.current_players
+            .lock()
+            .values()
+            .map(|p| p.gameprofile.name.clone())
+            .collect::<Vec<_>>()
+    }
+
+    pub fn get_nearest_player_name(&self, target: &Vector3<f64>) -> Option<String> {
```
StripedMonkey, line 281: I understand this PR is not intended to give us the functionality for everything, but this should not be `get_nearest_player_name`, but 
```rust
let Some(name) = get_nearest_player().map(Player::get_name) else {
    return Err(...)
 }
``` 
or something. 

for a single `get_nearest_...` it works, but implementing this functionality for EVERYTHING you can get from a "nearest player" would be terrifying

StripedMonkey (changes requested):
Thanks for your efforts trying to implement this! I do have some quibbles and things that would be good to see changed.

PR #152, opened 2024-10-19
unload chunks when no players are using them

kralverde:
This PR adds a counter to how many players are loading a chunks, unloading the chunk if no players are using them


Snowiiii:
I think you should send the unload chunk packet as well

Snowiiii:
Great work @kralverde (like always). Thank you :heart:

PR #153, opened 2024-10-19
Added translations

asodugf12:
Translations. Checked against minecraft translation files


pumpkin/src/server/translation.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+
+use serde_json::Value;
+
+#[derive(Debug)]
+pub enum TranslationError {
+    InvalidFile,
+    FileRead,
+    JsonParse,
+}
```
Snowiiii, line 17: Make use of the `thiserror` crate

pumpkin/src/server/translation.rs
```diff
@@ -0,0 +1,159 @@
...
+) -> Result<HashMap<String, String>, TranslationError> {
+    let file = File::open(path.into()).map_err(|_| TranslationError::InvalidFile)?;
+    let mut bufreader = BufReader::new(file);
+    let translations = fetch_translations(&mut bufreader, message)?;
+    let results = make_hashmap(translations);
+
+    Ok(results)
+}
+
+//Read a huge object line by line and tricking serde_json into thinking they are individual objects
```
Snowiiii, line 31: doc comments are written using ///, // are for normal comments

Snowiiii:
I would like to use Config values are introduced like in the image in https://github.com/Snowiiii/Pumpkin/issues/120. You can seperate PR's so its easier for you to implemented, I would first begin to just use the client-side translations. You then can move on to server-side, I would also consider a better approche then reading hashmap's everytime, Im considering caching translations depending if they are loaded by players

asodugf12:
Added all the suggestions :>

asodugf12:
I think caching could be left to a config value that would be turned on automatically if the server has used it in the last 5 times it started. What would you recommend for caching?

asodugf12:
Finally fixed cargo fmt file diff error haizzz
Is this way of setting up TranslationsConfig okay?

Snowiiii:
this is still not the solution i wanted, Im sorry but implementing the language system i want, im pretty sure, is more complex than you think. I would like to do it myself. We have to put all languages from Minecraft into the server and copy then over. Not only provide one single path

asodugf12:
I am still unsure what solution you are thinking of, but I'm okay with you finishing the rest. I'll find some more issues to work on in the meantime. Thanks for guiding me, Alex :>

asodugf12:
Actually you know what, I still want to see this through the end. So if it's possible to make a detailed diagram of what functionality is needed and goes where, I'd love to work on this further

Snowiiii:
My Plan.

1. Check if translations are on in the Config (default yes). If not use en_us as default (copy that over)
2. If Translations are on, Check if client-side translation are on in the Config (default yes). If client-side translations are on, we can just simply use `TextComponent::translatable`, If off, we will use translations for the server (copy them over).

I would prefer if you first add the ability to use client-side translations if they are on, otherwise or if translations are off just use default en_us, Im still not 100% about copying every Minecraft translation into Pumpkin. But you can start with this

asodugf12:
Some questions that I am having right now:
1. Does the `translate()` function accept a `&str` or a `TextComponent`?
2. Is the input by default en_us?
3. Do we have to parse the `TextComponent::translatable` (aka fill in the `%s` with the data in the `with` field)
4. If not, do we have to generate the `with` field of `TextComponent::translatable`?

Snowiiii:
> Some questions that I am having right now:
> 
>     1. Does the `translate()` function accept a `&str` or a `TextComponent`?
> 
>     2. Is the input by default en_us?
> 
>     3. Do we have to parse the `TextComponent::translatable` (aka fill in the `%s` with the data in the `with` field)
> 
>     4. If not, do we have to generate the `with` field of `TextComponent::translatable`?

1. It accepts a &str (e.g. "commands.gamemode.something"...), And returns a TextComponent.
2. Yes, en_us is the default
3 & 4: You will it using the with field (which we have) 
image

asodugf12:
I think the main translation logic is complete?
Working on file handling next

pumpkin/src/server/translation.rs
```diff
@@ -0,0 +1,208 @@
...
+    Translated(HashMap<String, TextComponent<'a>>),
+}
+
+pub fn translate<'a>(
+    config: &TranslationConfig,
+    message: &'a str,
+) -> Result<Translations<'a>, TranslationError> {
+    if !config.enabled {
+        return Ok(Translations::Untranslated(TextComponent::text(message)));
+    }
```
Snowiiii, line 35: So you just return a new TextComponent with the same string, But we wanted to use the default en_us when translations is off

pumpkin/src/server/translation.rs
```diff
@@ -0,0 +1,208 @@
...
+            content: TextContent::Translate {
+                translate: std::borrow::Cow::Borrowed(message),
+                with: vec![],
+            },
+
+            style: Style::default(),
+        }));
+    }
+
+    let path = "";
```
Snowiiii, line 48: Why there is an empty path ?

pumpkin/src/server/translation.rs
```diff
@@ -0,0 +1,208 @@
...
+    InvalidFile,
+    #[error("Failed to read file. Error: {0}")]
+    FileRead(std::io::Error),
+    #[error("Invalid JSON encountered. Use only objects in the translation file.")]
+    JsonParse,
+}
+
+pub enum Translations<'a> {
+    Untranslated(TextComponent<'a>),
+    Translated(HashMap<String, TextComponent<'a>>),
```
Snowiiii, line 26: Im also not sure why it has a HashMap, Both should just return a TextComponent

pumpkin/src/server/translation.rs
asodugf12: I thought that the default input string was en_us, so I set it as unchanged. 
Is what you want like `TextComponent::Translated(TextComponent::text(message)));`

pumpkin/src/server/translation.rs
asodugf12: empty path because i haven't set up file handling yet. Should've left a comment on that, sorry.

pumpkin/src/server/translation.rs
asodugf12: HashMap is in case more than 1 string contains the message. Do I just check for exact matches instead?

pumpkin/src/server/translation.rs
asodugf12: Also, there is no en_us translation file, which is obvious now that i think about it.

Snowiiii:
> I think the main translation logic is complete? Working on file handling next

Like i said you can just use the `en_us` from Vanilla. Put it into the assets folder.

asodugf12:
Do we have to keep a en_us file for every version of minecraft?

asodugf12:
I meant like, do we need every single subversion like 1.14.2 or just use 1.14 for all of them, since I need to download all the versions and that would take a while. 
Would love it if there's an easier way tho :<

Snowiiii:
> I meant like, do we need every single subversion like 1.14.2 or just use 1.14 for all of them, since I need to download all the versions and that would take a while. Would love it if there's an easier way tho :<

Pumpkin does target only the newest Minecraft version, Which is currently 1.21.3

asodugf12:
added en_us lang files up until 1.17, will finish the rest later

asodugf12:
Okay, I have discovered that for versions from 1.12.2 back, minecraft uses a different format for en_us files, which would require a different parser. I'll work on it maybe tomorrow :D

Snowiiii:
> Okay, I have discovered that for versions from 1.12.2 back, minecraft uses a different format for en_us files, which would require a different parser. I'll work on it maybe tomorrow :D

We only support the latest Minecraft version

asodugf12:
Oh, so that means this PR is complete?

Snowiiii:
> Oh, so that means this PR is complete?

Please remove all those langauge files from all other versions than **1.21.3**. **We only need one translation file for the latest Minecraft version**

asodugf12:
Alright, please list anything else that needs to be finished in this PR, because I feel like it's been lingering for too long now :D

pumpkin/src/server/translation.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[error("File cannot be opened.")]
+    InvalidFile,
+    #[error("Failed to read file. Error: {0}")]
+    FileRead(std::io::Error),
+    #[error("Invalid JSON encountered. Use only objects in the translation file.")]
+    JsonParse,
+}
+
+pub fn translate<'a>(
+    config: &TranslationConfig,
```
Snowiiii, line 27: Please remove the Config parameter. It is a Static you can use it yourself everywhere

pumpkin/src/server/translation.rs
```diff
@@ -0,0 +1,201 @@
...
+    message: &'a str,
+) -> Result<TextComponent<'a>, TranslationError> {
+    if !config.enabled {
+        let path = "assets/lang/en_us/en_us.json";
+        let translations = get_translations(path, message)?;
```
Snowiiii, line 32: I think this should be cached, You can use LazyStatic

asodugf12: I am not sure if you're talking about the crate or static with LazyLock, but I'd assume it would be the latter?

asodugf12:
Finished with the suggestions :D

asodugf12:
Hello, any comment on the fixes? @Snowiiii

Snowiiii:
You should use serde_json to read the JSON translation file. You can fully  remove the manuel reading and parsing from the file. And also rename the static variable from `PATH` to something better like `EU_US`.

asodugf12:
I am using serde_json, the current implementation just reads it line by line and turn each one into a json object, thus avoids having to read the whole file into memory, since it is one big json object.

asodugf12:
Or actually, do we cache read it once and cache it into memory? If so do you have any suggestions?

asodugf12:
Hallo, any more suggestions? I forgot to check the PR due to exams :D

Snowiiii:
> I am using serde_json, the current implementation just reads it line by line and turn each one into a json object, thus avoids having to read the whole file into memory, since it is one big json object.

I'm fine to have it stored as a HashMap, I don't think that the file will be that large

asodugf12:
I see, do I also add a cache for the hashmap?

Snowiiii:
> I see, do I also add a cache for the hashmap?

yes

Snowiiii:
Hey @asodugf12 

I'm Sorry, but I'm closing this PR due to ongoing discussions and the need to explore alternative approaches. We need a better way to doing this since we need to support multiple arguments, And in general this is something which is really not important right now, Client-Side translations (which vanilla server uses) will be enough. 
Like i said i'm Sorry, This was already open for a very long time, But maybe you find something even cooler to work on, I would recommend to may discuss your idea in our discord

asodugf12:
It's alright, I do feel like this has dragged on for too long. I am open for any features or issues to fix, and hopefully we can make goals and implementations clear beforehand so it could be smoother

PR #154, opened 2024-10-19
Optimized Anvil parsing

ReCore-sys:
I have rewritten the anvil parsing system to read as little data as possible. It no longer stores a location table in memory since that is only useful if we are going to reuse it and since we have no caching system for it, we can't.
Instead of storing the table in memory, it now calculates and reads only the bytes it needs. I have also switched out the ZLib decompression library for Zune, which does perform slightly better. 
I have also attached my memory-mapped implementation as well for consideration, though it does use some unsafe code.
Overall, it now performs about 15% faster (~440 µs, down from ~515 µs), but still nowhere near FastAnvil or the memory map implementation (~85 µs and 75 µs respectively). Benchmarks and unit tests are included

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,3 +98,5 @@ features.toml
 docs/.vitepress/dist
 docs/.vitepress/cache
 node_modules
+
+.idea/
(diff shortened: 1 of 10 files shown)


Snowiiii:
Please use the master branch now

PR #157, opened 2024-10-20
Format numbers and dates per region

neeleshpoli:
This PR formats dates and numbers to be formatted for each locale correctly. This also converts times to the user's respective time zone.

This introduces 2 new Vue components, FmtNum and FmtDateTime. These can be used as normal Vue components in the markdown files.

### For example, here is `en_US`:
image
### And here is `de-DE`
image

### Here is a US formatted date, in the CDT time zone:
image
### And here is a London formatted date
image

Everything is adjusted based on region, using the Intl JavaScript apis. ***To be clear, there is no actual translation. Only the numbers and dates are formatted as per the user region.***

Also, there are improvements to the benchmarking page so that it flows more smoothly.

### DIFF
--- a/docs/.vitepress/components/FmtDateTime.vue
+++ b/docs/.vitepress/components/FmtDateTime.vue
@@ -1,0 +1,19 @@
+<script setup>
+import { computed, getCurrentInstance } from 'vue';
+
+const props = defineProps({
+    d: {
+        type: Date,
+        required: true
+    }
+});
+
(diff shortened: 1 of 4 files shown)


Snowiiii:
Overall good job, One thing i would change are the Notes
image
It is very irritating to have like 3 Warn sings at once

Snowiiii:
Now its better.
Thank you @neeleshpoli

PR #159, opened 2024-10-20
Fix chunk loading deadlock and dangling chunk memory

kralverde:
If the chunk fetcher spawns and calls level.fetch_chunks and the chunk sender has not spawned yet, because the function is synchronous, the channel can be filled with no consumer

### DIFF
--- a/pumpkin-world/src/level.rs
+++ b/pumpkin-world/src/level.rs
@@ -81,8 +81,9 @@ impl Level {
     /// it is removed from memory. Should only be called on chunks the player was not watching
     /// before
     pub fn mark_chunks_as_newly_watched(&self, chunks: &[Vector2<i32>]) {
-        chunks.par_iter().for_each(|chunk| {
-            match self.chunk_watchers.entry(*chunk) {
+        chunks
+            .par_iter()
+            .for_each(|chunk| match self.chunk_watchers.entry(*chunk) {
                 Entry::Occupied(mut occupied) => {
                     let value = occupied.get_mut();
                     if let Some(new_value) = value.checked_add(1) {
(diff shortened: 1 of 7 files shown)


Snowiiii:
After testing, It seems you don't unload the chunks somewhere resulting in huge memory consumption

kralverde:
@Snowiiii could you take another look at the memory consumption?

Snowiiii:
> @Snowiiii could you take another look at the memory consumption?

I will bit later, Not working on Pumpkin and just bit relaxing for the first time :D.

Snowiiii:
Can confirm, memory is now consistent at 8-10MB again

kralverde:
@StripedMonkey how do these changes look?

pumpkin-world/src/level.rs
```diff
@@ -119,77 +127,96 @@ impl Level {
...
+    fn load_chunk_from_save(
+        chunk_reader: Arc<dyn ChunkReader>,
+        save_file: SaveFile,
+        chunk_pos: Vector2<i32>,
+    ) -> Result<Option<Arc<ChunkData>>, ChunkReadingError> {
+        match chunk_reader.read_chunk(&save_file, &chunk_pos) {
+            Ok(data) => Ok(Some(Arc::new(data))),
+            Err(
+                ChunkReadingError::ChunkNotExist
+                | ChunkReadingError::ParsingError(ChunkParsingError::ChunkNotGenerated),
```
StripedMonkey, line 154: Why are these separate things?

StripedMonkey:
> @StripedMonkey how do these changes look?

I don't have any objections to the PR code itself. It's looking pretty good.

kralverde:
Streams chunks instead of collecting them in a vec and fixes a bug where chunks were not unloaded

PR #160, opened 2024-10-20
made encryption config work

Alvsch:
Made the encryption option in the config actually turn on and off encryption.

Also cleaned up some of the encryption response handling.

### DIFF
--- a/pumpkin/src/client/client_packet.rs
+++ b/pumpkin/src/client/client_packet.rs
@@ -99,12 +99,12 @@ impl Client {
                 }
             }
         } else {
-            *gameprofile = Some(GameProfile {
+            let profile = GameProfile {
                 id: login_start.uuid,
                 name: login_start.name,
                 properties: vec![],
                 profile_actions: None,
-            });
+            };
 
             if BASIC_CONFIG.encryption {
                 let verify_token: [u8; 4] = rand::random();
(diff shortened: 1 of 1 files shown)


pumpkin/src/client/client_packet.rs
```diff
@@ -106,10 +106,23 @@ impl Client {
...
                 .await;
+            } else {
+                if ADVANCED_CONFIG.packet_compression.enabled {
+                    self.enable_compression().await;
+                }
+                self.finish_login(
+                    gameprofile
+                        .as_ref()
+                        .expect("The GameProfile was just set to Some, this should never fail"),
+                )
```
Snowiiii, line 123: You could actually remove the need for an optional gameprofile here. Above a new gameprofile is creates, you can store it in a new var
```rust
            let profile = GameProfile {
                id: login_start.uuid,
                name: login_start.name,
                properties: vec![],
                profile_actions: None,
            };
            *gameprofile = Some(profile);
            
             self.finish_login(
                   profile,
                )
 ```

Alvsch: borrow checker would not like this as we move profile into gameprofile

Snowiiii: `finish_login` takes in a reference, means you also can set the gameprofile below the function

Snowiiii:
Looks good then, We should keep in mind that encryption is needed when online mode is enabled. Otherwise its Optional

PR #162, opened 2024-10-20
Block breaking and placing

Bryntet:
In this PR, I aim to add support for fast asynchronous placing & breaking of blocks.

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -6,7 +6,7 @@ use crate::{
     client::Client,
     entity::{player::Player, Entity},
 };
-use num_traits::{Euclid, ToPrimitive};
+use num_traits::ToPrimitive;
 use pumpkin_config::BasicConfiguration;
 use pumpkin_core::math::position::WorldPosition;
 use pumpkin_core::math::vector2::Vector2;
@@ -20,11 +20,12 @@ use pumpkin_protocol::{
     ClientPacket, VarInt,
 };
 use pumpkin_world::block::BlockId;
+use pumpkin_world::chunk::ChunkData;
 use pumpkin_world::coordinates::ChunkRelativeBlockCoordinates;
 use pumpkin_world::level::Level;
 use scoreboard::Scoreboard;
-use tokio::sync::mpsc;
 use tokio::sync::Mutex;
+use tokio::sync::{mpsc, RwLock};
 
 pub mod scoreboard;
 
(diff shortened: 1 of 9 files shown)


pumpkin/src/world/mod.rs
```diff
@@ -349,4 +355,39 @@ impl World {
         self.broadcast_packet_all(&CRemoveEntities::new(&[entity.entity_id.into()]))
             .await
     }
+    pub async fn set_block(&self, position: WorldPosition, block_id: BlockId) {
+        let (z_chunk, z_rem) = position.0.z.div_rem_euclid(&16);
+        let (x_chunk, x_rem) = position.0.x.div_rem_euclid(&16);
+        let chunk_coordinate = Vector2 {
+            x: x_chunk,
+            z: z_chunk,
+        };
```
StripedMonkey, line 364: I would like to see stuff like this turned into helper functions on `WorldPosition`. 

It can be a `position.get_chunk_position()` (naming here might be confusing for in-chunk position vs chunk position) or similar.

pumpkin/src/world/mod.rs
```diff
@@ -349,4 +355,39 @@ impl World {
...
+
+        // Since we divide by 16 remnant can never exceed u8
+        let relative = ChunkRelativeBlockCoordinates {
+            x: (x_rem as u8).into(),
+            z: (z_rem as u8).into(),
+            y: position.0.y.into(),
+        };
```
StripedMonkey, line 371: ditto for what I said about the `chunk_coordinate`

pumpkin/src/world/mod.rs
```diff
@@ -349,4 +355,39 @@ impl World {
...
+
+        let (sender, mut receive) = mpsc::channel(1024);
+        {
+            let level = self.level.clone();
+            tokio::spawn(
+                async move { level.lock().await.fetch_chunks(&[chunk_coordinate], sender) },
+            );
```
StripedMonkey, line 378: Does it make sense to spawn and await within a `fetch_chunk` or something instead of doing it externally?

We probably want to always fetch the chunk with a separate task or something.

pumpkin/src/world/mod.rs
```diff
@@ -271,21 +275,23 @@ impl World {
...
-            let level = level.lock().await;
-            level.fetch_chunks(&chunks, sender)
-        });
+        {
+            let level = self.level.clone();
+            tokio::spawn(async move {
+                let level = level.lock().await;
+                level.fetch_chunks(&chunks, sender)
+            });
+        }
```
StripedMonkey, line 284: Should this simply be its own function?

Bryntet: That sounds like a good idea.

pumpkin/src/world/mod.rs
Bryntet: I also thought of this, but it must've slipped my mind.

PR #163, opened 2024-10-20
Handle player packet id 0x23

tuttarealstep:
Fix for:
image

Using the protocol packet: https://wiki.vg/Protocol#Player_Abilities_.28serverbound.29


.gitignore
```diff
@@ -77,9 +77,7 @@ Cargo.lock
 
 # === PROJECT SPECIFIC ===
 # mc decompiled source
-mc-source-code/DecompilerMC/
-mc-source-code/src/
-mc-source-code/target/
+mc-source-code/
```
Snowiiii, line 80: Why you modified the .gitignore ?

pumpkin/src/client/player_packet.rs
```diff
@@ -541,6 +541,13 @@ impl Player {
         }
     }
 
+    pub async fn handle_player_abilities(&self, player_abilities: SPlayerAbilities) {
+        let mut abilities = self.abilities.lock().await;
+
+        // Set the flying ability
+        abilities.flying = player_abilities.flags & 0x02 != 0 && abilities.allow_flying;
```
Snowiiii, line 548: The abilities have `allow_flying` as well, Should should set them and not check if they true in the packet

.gitignore
tuttarealstep: Sorry I've used the "decompile.sh" file and that folder for save researches, to prevent any accidental upload I've changed to include any file and subfolder

pumpkin/src/client/player_packet.rs
tuttarealstep: I've seen that this packet is sent by the client only when the player starts/stops flying and is not sending the "allow_fly" property. I've implemented the same check that vanilla does (bit mask 0x02 on the Flags field of the packed and the additional check on the player "allow_fly" flag) to set the flying property

Snowiiii: Oh, Just looked at vanilla code and your right!, nvm then

PR #164, opened 2024-10-21
Prevent panic on client poll function

tuttarealstep:
Hi!
In pumpkin\src\client\mod.rs if an error occur in the poll function the server crash:
image

With the fix (match instead of using unwrap) the server don't panic.
image


Snowiiii:
Yeah, that was a quick tmp solution when porting to tokio.
Thank you @tuttarealstep

PR #165, opened 2024-10-21
Add Shaped and Unshaped crafting support, as well as deserializing the entire recipe format

Bryntet:
This PR aims to add deserialization of the minecraft Recipe format. It also adds support for the Shaped and Unshaped crafting forms, accessible from a craftingtable (`/craft`)


Snowiiii:
Please port the recipies to 1.21.2

Snowiiii:
Hi @Bryntet. Just wanted to say that 1. I really don't like the approche of having 1000 recipie files. 2. Im currently making an own extractor (see branch), Which can produce all recipes in one json file, With everything we need

Bryntet:
Yes sounds good! I will use the extractor :)

Bryntet:
Ready for review. 

Couple of issues left:

Dragging/shift-clicking is currently disabled in all crafting input/output slots

items are kept currently in container even when closing it.

Cargo.toml
```diff
@@ -25,6 +25,7 @@ inherits = "release"
 debug = true
 
 [workspace.dependencies]
+serde_json = "1.0"
```
Snowiiii, line 28: I would prefer to keep serde_json near serde

PR #168, opened 2024-10-21
Anvil Saving

lukas0008:
This pr aims to add world saving using minecraft .mca files to pumpkin.


lukas0008:
Comment: d0b6805455165a428504625794aafb47af2a6ecf added the ability to save chunks to anvil, but converting chunks into bytes is still not done.

pumpkin-world/src/chunk/anvil.rs
```diff
@@ -84,7 +87,11 @@ impl Compression {
     }
 }
 
-impl ChunkReader for AnvilChunkReader {
+fn modulus(a: i32, b: i32) -> i32 {
+    ((a % b) + b) % b
+}
```
Snowiiii, line 92: You never can have enough of modulus functions right :wink:

pumpkin-world/src/chunk/anvil.rs
```diff
@@ -161,3 +167,136 @@ impl ChunkReader for AnvilChunkReader {
...
+        at: pumpkin_core::math::vector2::Vector2<i32>,
+    ) -> Result<(), super::ChunkWritingError> {
+        // TODO: update timestamp
+
+        let bytes = chunk.to_bytes();
+
+        let region = (
+            ((at.x as f32) / 32.0).floor() as i32,
+            ((at.z as f32) / 32.0).floor() as i32,
+        );
```
Snowiiii, line 185: We use this also for reading, would probably make sense to have it as a function

Snowiiii:
I also noticed that there is no compression yet

Bryntet:
Yooo good job @lukas0008 :)

1Nu55et1:
Hey, I'm not an expert, but maybe you could manage the files by dividing them into regions, where each file represents a 32x32 chunk area. Instead of using floating-point division for calculations, you could use bitwise operations (like shifts and masks) to make it faster and more precise.

For compression, you might want to try something like Zlib, or if you're planning to use another method, I'd be curious to know which one and why. Also, you could organize this into separate modules if you want: one for region management, another for compression, and another one for utilities. This might make the code easier to maintain and expand. 

Sorry if this explanation is a bit messy or unclear, but I tried my best to help and give some suggestions that might be useful. :)

1Nu55et1:
And I used this to try to understand it:  
https://minecraft.fandom.com/wiki/Anvil_file_format

PR #169, opened 2024-10-22
fix tick system

kralverde:
Currently, the ticking system blocks the whole thread.... this fixes that


kralverde:
also happens when we read stdin. this frees up two more cores for us to use on more intensive stuff

Snowiiii:
Looks good.
Thanks @kralverde

PR #172, opened 2024-10-23
Handle chunks in an intermediary state

kralverde:
This PR handles two edge cases that can leave chunks in memory:
1) A client disconnected before all chunks are sent
2) A client moving so fast that a chunk that was in render distance is no longer in render distance before the chunk packet is sent

~~A future implementation would somehow cancel the work being done to generate chunks if they are no longer needed.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ num-derive = "0.4"
 
 # Concurrency/Parallelism and Synchronization
 rayon = "1.10.0"
-parking_lot = "0.12.3"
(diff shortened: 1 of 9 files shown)


kralverde:
now also closes https://github.com/Snowiiii/Pumpkin/issues/170 :p

Snowiiii:
Hey @kralverde, Lets remove the id in main.
1. We don't need it
2. Its tricky to really track down a Client with it, Much easier with a Name or UUID.
3. Main becomes cleaner

Snowiiii:
Thanks @kralverde

PR #176, opened 2024-10-24
make `get_player_by_name` async, rename old fn to `get_player_by_name_blocking`

lokka30:
Currently, the existing commands system is all that requires the existing blocking version of `get_player_by_name` function. After discussing in the Discord, it seems renaming the old blocking method with a suffix of `_blocking` in addition to adding an async version under the original (unmodified) name of `get_player_by_name`.

This change was required to work on the 'Check if player already joined' project.

[Jump to Discord conversation](https://discord.com/channels/1268592337445978193/1268611318617866261/1299028500547637359)


lokka30:
as per request, will make this part of a larger commit instead.

PR #177, opened 2024-10-24
Implement `get_player_by_uuid` for server & world

lokka30:
Similar to the existing functions named `get_player_by_name`, these functions will search for a player with a matching `Uuid`.

This implementation is asynchronous and thus better suits being combined with #176, which makes `get_player_by_name` asynchronous too.


lokka30:
as per request, will make this part of a larger commit instead.

PR #178, opened 2024-10-24
Prevent duplicate profiles

lokka30:
This PR aims to resolve the to-do item [Check if player already joined][1].

This is implemented via two checks:
1. **Prevent duplicate UUIDs.**
  Stops a player from logging in to the server if they are already playing (or if their UUID is being used in some other narrow possibility).
3. **Prevent duplicate usernames.**
  Stops a player from taking the same username as an online player.
  Of course, this is primarily made for offline-mode servers. However, it *might* be possible that this serves as a protection for online-mode servers, in case someone does a name change whilst connected, someone takes their old name, and tries to connect.

## To-do

- [ ] Code reviewed with no further suggestions / requirements.
  Please note, I'm a Rust newbie, so please check my code thoroughly for any mistakes. 🙂 
- [x] Testing completed:
  - [x] Compiles, runs, allows normal players as expected.
  - [x] Prevents duplicate UUIDs (player already logged in).
  - [x] Prevents duplicate usernames (particularly for offline mode).
- [x] Ready to merge (convert PR from draft accordingly).

### UUID duplication

image

### Username duplication

image

[1]: https://github.com/users/Snowiiii/projects/12/views/3?pane=issue&itemId=84062083

### DIFF
--- a/pumpkin/src/client/authentication.rs
+++ b/pumpkin/src/client/authentication.rs
@@ -143,8 +143,6 @@ pub enum AuthError {
     FailedParse,
     #[error("Unknown Status Code")]
     UnknownStatusCode(StatusCode),
-    #[error("You are already connected to this server")]
-    DuplicateConnection,
 }
 
 #[derive(Error, Debug)]
--- a/pumpkin/src/client/authentication.rs
+++ b/pumpkin/src/client/authentication.rs
@@ -143,6 +143,8 @@ pub enum AuthError {
     FailedParse,
     #[error("Unknown Status Code")]
     UnknownStatusCode(StatusCode),
+    #[error("You are already connected to this server")]
+    DuplicateConnection,
 }
 
 #[derive(Error, Debug)]
--- a/pumpkin/src/client/client_packet.rs
+++ b/pumpkin/src/client/client_packet.rs
@@ -163,6 +163,13 @@ impl Client {
             }
         }
 
+        // don't allow a duplicate username
+        if let Some(online_player) = &server.get_player_by_name(&profile.name).await {
+            log::debug!("A player (IP '{}', attempted username '{}') tried to log in with the same username as an online player (UUID '{}', IP '{}', username '{}')", &self.address.lock().await.to_string(), &profile.name, &profile.id.to_string(), &online_player.client.address.lock().await.to_string(), &online_player.gameprofile.name);
+            self.kick("A player with this username is already connected").await;
+            return;
+        }
+
         if ADVANCED_CONFIG.packet_compression.enabled {
             self.enable_compression().await;
         }
@@ -192,6 +199,13 @@ impl Client {
             let ip = self.address.lock().await.ip();
 
             let profile = authentication::authenticate(username, &hash, &ip, auth_client).await?;
+            
+            // Don't allow duplicate UUIDs
+            if let Some(online_player) = &server.get_player_by_uuid(profile.id).await {
+                log::debug!("Player (IP '{}', username '{}') tried to log in with the same UUID ('{}') as an online player (IP '{}', username '{}')", &ip.to_string(), &profile.name, &profile.id.to_string(), &online_player.client.address.lock().await.to_string(), &online_player.gameprofile.name);
+                return Err(AuthError::DuplicateConnection);
+            }
+            
             // Check if player should join
             if let Some(actions) = &profile.profile_actions {
                 if ADVANCED_CONFIG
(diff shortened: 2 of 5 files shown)


pumpkin/src/client/authentication.rs
```diff
@@ -143,6 +143,8 @@ pub enum AuthError {
     FailedParse,
     #[error("Unknown Status Code")]
     UnknownStatusCode(StatusCode),
+    #[error("You are already connected to this server")]
+    DuplicateConnection,
```
Snowiiii, line 147: I don't like to have the error here, This has not really something to do with Authentication. Is more just validation

lokka30: Do you have a spot in mind you would prefer the error be stored? In this case I also believe the duplicate UUID check should be moved out of `authenticate` and potentially go inside at the start of the `if BASIC_CONFIG.online_mode` block, line 153 in `client_packets.rs`?

pumpkin/src/world/mod.rs
```diff
@@ -347,6 +357,16 @@ impl World {
...
     }
 
+    /// Gets a Player by UUID
+    pub async fn get_player_by_uuid(&self, id: uuid::Uuid) -> Option<Arc<Player>> {
+        for player in self.current_players.lock().await.values() {
+            if player.gameprofile.id == id {
+                return Some(player.clone());
+            }
+        }
+        None
```
Snowiiii, line 367: Its a HashMap, You can search the player by uuid in it

pumpkin/src/client/authentication.rs
Snowiiii: The second option makes more sense to me

pumpkin/src/world/mod.rs
lokka30: Thanks, should be resolved in [7b7eab5](https://github.com/Snowiiii/Pumpkin/pull/178/commits/7b7eab5dc67c47752c52a7b2b6aa1dcf60abf748) 🙂

pumpkin/src/client/authentication.rs
lokka30: Thanks, should be resolved in [7b7eab5](https://github.com/Snowiiii/Pumpkin/pull/178/commits/7b7eab5dc67c47752c52a7b2b6aa1dcf60abf748) 🙂

pumpkin/src/client/client_packet.rs
```diff
@@ -151,6 +151,14 @@ impl Client {
         };
 
         if BASIC_CONFIG.online_mode {
+            // Don't allow duplicate UUIDs
+            if let Some(online_player) = &server.get_player_by_uuid(profile.id).await {
+                log::debug!("Player (IP '{}', username '{}') tried to log in with the same UUID ('{}') as an online player (IP '{}', username '{}')", &self.address.lock().await.to_string(), &profile.name, &profile.id.to_string(), &online_player.client.address.lock().await.to_string(), &online_player.gameprofile.name);
+                self.kick("You are already connected to this server").await;
+                return;
+            }
```
Snowiiii, line 159: Check for duplicate uuid when authentication is successful.

Snowiiii: Im thinking of even check for duplicate uuid if online mode is true **or** false. Thats what Mojang is doing.

pumpkin/src/world/mod.rs
```diff
@@ -336,9 +336,19 @@ impl World {
...
+            if player.gameprofile.name == name {
+                return Some(player.clone());
+            }
+        }
+        None
+    }
+
+    /// Gets a Player by username (Blocking - Legacy use only)
+    pub fn get_player_by_name_blocking(&self, name: &str) -> Option<Arc<Player>> {
+        // TODO: Remove this blocking functions when commands are async.
```
Snowiiii, line 351: Since https://github.com/Snowiiii/Pumpkin/commit/7347bae275e65351ced7f80098e98aab48889aa4 is merged now. You can replace blocking_lock with lock in the other function

lokka30: Sorry, just to double check, did you mean just remove the blocking function altogether?

In the linked commit, it is still using the blocking function [here](https://github.com/Snowiiii/Pumpkin/blob/7347bae275e65351ced7f80098e98aab48889aa4/pumpkin/src/commands/arg_player.rs#L29).

it is also still using the blocking lock in that commit [here](https://github.com/Snowiiii/Pumpkin/blob/7347bae275e65351ced7f80098e98aab48889aa4/pumpkin/src/world/mod.rs#L367).

Please kindly clarify for me :) 

the problematic file is arg_player.rs which is the only reference for the blocking function

pumpkin/src/client/authentication.rs
lokka30: marking as resolved

pumpkin/src/client/client_packet.rs
lokka30: marking as resolved

pumpkin/src/world/mod.rs
lokka30: marking as resolved

pumpkin/src/client/client_packet.rs
lokka30: Should be resolved in [cded477](https://github.com/Snowiiii/Pumpkin/pull/178/commits/cded477f0e149a12033c7786aebf67ac7fee6bd1) 🙂

Snowiiii:
Looks good. 
Thanks @lokka30

PR #179, opened 2024-10-24
make commands async

kralverde:
This PR fixes and makes commands async


Snowiiii:
Mhh. Its fine for now.
Thanks @kralverde

PR #181, opened 2024-10-25
docker: add compose file for development

### LINKED ISSUE #174: Supply `docker-compose.yml` in project

There is currently no Docker Compose file supplied with the project. I suggest supplying an example `docker-compose.yml` which can get people running with Pumpkin easily via Docker Compose.

Ideally, we can start working on this once #125 is complete. It may be possible to do prior, using `build: .` possibly so it builds the image locally, and then switch to using the registry image once #125 is done.

lokka30:
This resolves #174.

I've chosen to make this compose file purely for development use (and thus sticks with `build: .`, rather than using an image).

Instead, for server owners, we should provide an example `docker-compose.yml` for them to use which references a public image instead.

This way, both use cases for the compose files are separated.

***

- [x] Ready for review.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,8 @@
+services:
+  pumpkin:
+    build: .
+    ports:
+      - 25565:25565
+    volumes:
+      - ./configuration.toml:/pumpkin/configuration.toml
+      - ./features.toml:/pumpkin/features.toml


Snowiiii:
Did you test this ?

lokka30:
> Did you test this ?

Certainly, just fired it up again, here's a screenshot 🙂 
image
Currently there is an issue with Pumpkin taking a while to shut down on docker, it is fixed by #183

Snowiiii:
Okay seems fine then.
Thanks @lokka30

PR #183, opened 2024-10-25
Implement cross platform signal handling with `tokio`

### LINKED ISSUE #182: Handle SIGTERM on UNIX

- Pumpkin currently handles `SIGINT` via `ctrlc::set_handler` in `main.rs`, however, it doesn't handle `SIGTERM`
- Handling `SIGTERM` is especially important for Docker, since it will allow Docker to gracefully stop the server, instead of killing it.

- The `ctrlc` crate we are are already using can simply have its `termination` feature enabled in `cargo.toml` with no other changes required.
- Whilst working on this, may as well also comment the ctrlc part of the `main` function more, and use `expect` instead of `unwrap` for a better error message if unsuccessful.
- To my understanding, these changes do not create any caveats for OS compatibility, i.e., non-UNIX.

lokka30:
This PR enables the `termination` feature in `ctrlc` so that `sigterm` and `sighup` are also handled on UNIX systems. This in turn improves Docker compatibility since containers now listen to the `SIGTERM` signal it sends to containers when trying to stop the container via `Ctrl-C` whilst attached.

Resolves #182; please check this issue for more information.

***

- [x] Fix compilation errors on Windows
- [x] Ready for merge

### DIFF
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -215,7 +215,7 @@ async fn setup_sighandler() -> io::Result<()> {
         handle_interrupt();
     }
 
-    Ok(());
+    Ok(())
 }
 
 // Unix signal handling
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -100,7 +100,9 @@ async fn main() -> io::Result<()> {
     //     .unwrap();
 
     tokio::spawn(async {
-        setup_sighandler().await.expect("Unable to setup signal handlers");
+        setup_sighandler()
+            .await
+            .expect("Unable to setup signal handlers");
     });
 
     // ensure rayon is built outside of tokio scope
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -25,7 +25,7 @@ use client::Client;
 use server::{ticker::Ticker, Server};
 use std::io::{self};
 use tokio::io::{AsyncBufReadExt, BufReader};
-#[cfg(windows)]
+#[cfg(not(unix))]
 use tokio::signal::ctrl_c;
 #[cfg(unix)]
 use tokio::signal::unix::{signal, SignalKind};
@@ -208,8 +208,8 @@ fn handle_interrupt() {
     std::process::exit(0);
 }
 
-// Windows Ctrl-C handling
-#[cfg(windows)]
+// Non-UNIX Ctrl-C handling
+#[cfg(not(unix))]
 async fn setup_sighandler() -> io::Result<()> {
     if ctrl_c().await.is_ok() {
         handle_interrupt();
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -25,7 +25,7 @@ use client::Client;
 use server::{ticker::Ticker, Server};
 use std::io::{self};
 use tokio::io::{AsyncBufReadExt, BufReader};
-// TODO: #[cfg(windows)]
+#[cfg(windows)]
 use tokio::signal::ctrl_c;
 #[cfg(unix)]
 use tokio::signal::unix::{signal, SignalKind};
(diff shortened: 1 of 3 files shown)


lokka30:
As per conversation in #182, I've switched `ctrlc` crate to just using `tokio` in a2c5e00. Please let me know if you require any further changes @Snowiiii - i've listed any caveats in the commit description. I'll work on getting this to compile in GH actions.

Snowiiii:
CI is failing on Windows.

lokka30:
Already the case afaik @Snowiiii 🙂 
- Windows import: [here](https://github.com/Snowiiii/Pumpkin/pull/183/files#diff-87075e86df7f7dcaa204153006cd8dcad793869fc365391b8b284d64f6b251c3R28-R29)
- Unix import: [here](https://github.com/Snowiiii/Pumpkin/pull/183/files#diff-87075e86df7f7dcaa204153006cd8dcad793869fc365391b8b284d64f6b251c3R30-R31)
- Windows function: [here](https://github.com/Snowiiii/Pumpkin/pull/183/files#diff-87075e86df7f7dcaa204153006cd8dcad793869fc365391b8b284d64f6b251c3R211-R219)
- Unix function: [here](https://github.com/Snowiiii/Pumpkin/pull/183/files#diff-87075e86df7f7dcaa204153006cd8dcad793869fc365391b8b284d64f6b251c3R221-R237)

I haven't been able to test the Windows function works (besides now compiling successfully) since I use macos and debian

pumpkin/src/main.rs
```diff
@@ -25,6 +25,10 @@ use client::Client;
 use server::{ticker::Ticker, Server};
 use std::io::{self};
 use tokio::io::{AsyncBufReadExt, BufReader};
+#[cfg(windows)]
+use tokio::signal::ctrl_c;
```
Snowiiii, line 29: This is not windows specific

lokka30: oh true, this should be enabled on anything but unix. i'll try fix, thanks

lokka30: Resolved in 7ab44d0

Snowiiii: Oh man Please, I now tell you the third time. ctrlc can be included in **any** Platform. Tokio has interinal checking and uses the right platform. Your current implementation only sets ctrlc for only on Platform. Suprise ctrlc is supported on more platforms. Just take a look at https://github.com/tokio-rs/tokio/blob/master/tokio/src/signal/ctrl_c.rs

lokka30: I think I understand what you're requesting now, @Snowiiii: to alter the PR to just use ctrlc on all platforms. I advise against this,

Using ctrlc on UNIX has [caveats](https://github.com/tokio-rs/tokio/blob/master/tokio/src/signal/ctrl_c.rs#L18-L34), and thus for UNIX systems I am using a custom implementation without those caveats. Hope that explains why there are platform-dependent versions in this PR.

I'd be happy to alter the PR to just use ctrlc on all platforms and accept the caveats, but I'd personally recommend against it.

Snowiiii: Ohh so from my understanding it sends, A Singal when CTRL+C is pressed ?

lokka30: - Basically, UNIX has a whole different way of signal handling than Windows.
- The windows way of doing it afaik is purely ctrl-c requests the program to stop running.
- However, with UNIX there are a bunch more signals, and they are handled in a considerably different way to Windows.
- So although the `ctrlc` module in `tokio` 'works' cross-platform, since it tries to make everything work all in one, it can cause issues for UNIX systems in mostly edge-cases.
- This is why I have separated the UNIX signal handling from all other platforms.
- Since the UNIX signal handling code in this PR already handles the signals from UNIX's variant of 'Ctrl-C', there is no need to enable the `ctrlc` module code; if anything this may cause conflicts with the existing unix-specific signal handling code.

Hope that helps clear it up, let me know if you are still unsure about anything 🙂

Snowiiii:
I tested it on my Linux machine, It works fine. I understood that ctrl+c. Just sends a signal on unix.
Then its fine.
Thanks @lokka30

PR #191, opened 2024-10-26
world docs: corrected to 'world', not 'server'

lokka30:
These functions access only the world itself's current players, and thus the old documentation was incorrect saying that it accesses the server's online players.


Snowiiii:
Thanks @lokka30

PR #192, opened 2024-10-26
Add player join and quit messages.

### LINKED ISSUE #188: Add join and quit messages

- Pumpkin currently doesn't have join and quit messages.

- Join message
  - Will position in `client_packets.rs`, at the top of `finish_login`, above where we send them the success packet.
- Quit message
  - I can't find where to put this, please could a maintainer advise
- Kick log message
  - Only an INFO log message to note that a player was kicked.

**To-do list**
- [x] Maintainers approve idea
- [x] Contributor indicates good place to handle the quit message code; edit above
- [x] Started work on a PR (linked to issue).

lokka30:
This implementation will send a 'joined the game' and 'left the game' message akin to the vanilla server when a player joins or disconnects respectively.

One caveat with this implementation is that it only broadcasts the join/quit messages to players in the same world, unlike the vanilla server, which broadcasts it to all online players (regardless of their world).

I discussed this with Alex who suggested to keep it as per-world (seemingly for now?) - [Discord conversation](https://discord.com/channels/1268592337445978193/1268611318617866261/1299668206734147597).

Regardless, this implementation works perfectly on my end. (yes, it has been tested)

Resolves #188 

Screenshot:
image

- [x] Ready for review/merge

### DIFF
--- a/pumpkin/src/server/mod.rs
+++ b/pumpkin/src/server/mod.rs
@@ -1,7 +1,13 @@
 use connection_cache::{CachedBranding, CachedStatus};
 use key_store::KeyStore;
 use pumpkin_config::BASIC_CONFIG;
-use pumpkin_core::GameMode;
+use pumpkin_core::{
+    text::{
+        color::{Color::Named, NamedColor},
+        TextComponent,
+    },
+    GameMode,
+};
 use pumpkin_entity::EntityId;
 use pumpkin_inventory::drag_handler::DragHandler;
 use pumpkin_inventory::{Container, OpenContainer};
@@ -113,6 +119,14 @@ impl Server {
                 self.server_listing.lock().await.add_player();
             }
         }
+
+        // Handle join message
+        let msg_txt = format!("{} joined the game.", player.gameprofile.name.as_str());
+        let msg_comp = TextComponent::text(msg_txt.as_str()).color(Named(NamedColor::Yellow));
+        for player in world.current_players.lock().await.values() {
+            player.send_system_message(&msg_comp).await;
+        }
+
         (player, world.clone())
     }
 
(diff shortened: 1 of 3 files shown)


lokka30:
I'd like to add some console logging in before we merge actually.

lokka30:
image
Done, ready now assuming all compiles well 😛

Snowiiii:
I would prefer to have the message in `world` not in player

lokka30:
Thanks for the review @Snowiiii, I've moved disconnect message logic from `player` to `world` in 7c06b7a.

I think this causes implications though. Wouldn't this send players in the world a 'left the game' message when a player is removed from the world, even if they're on the same server, perhaps moving to another world?
If so, this would also cause the disconnect message to be a per-world thing (non-vanilla behaviour), when the join message is a per-server thing (vanilla behaviour).

Please let me know what you think

pumpkin/src/server/mod.rs
```diff
@@ -113,6 +116,15 @@ impl Server {
...
             }
         }
+
+        // Handle join message
+        let msg_txt = format!("{} joined the game.", player.gameprofile.name.as_str());
+        let msg_comp = TextComponent::text(msg_txt.as_str()).color_named(NamedColor::Yellow);
+        for player in world.current_players.lock().await.values() {
+            player.send_system_message(&msg_comp).await;
+        }
+        log::info!("{}", msg_comp.to_pretty_console());
```
Snowiiii, line 126: Let put this then into world as well

lokka30: Resolved in b5e01db 🙂 
Do we plan on mimicking vanilla behavior (per-server join/quit rather than per-world) in future?

Snowiiii: Not sure, We could make it  a config setting, but it would be annoying. For now i would leave it per world

Snowiiii:
Looks good now.
Thanks @lokka30

PR #193, opened 2024-10-26
CChunkData optimizations

Alvsch:
Optimized CChunkData::write
Now creating rayon threads before the tokio runtime
Added profiling profile

### DIFF
--- a/pumpkin-protocol/src/client/play/c_chunk_data.rs
+++ b/pumpkin-protocol/src/client/play/c_chunk_data.rs
@@ -3,7 +3,10 @@ use arrayvec::ArrayVec;
 use itertools::Itertools;
 
 use pumpkin_macros::client_packet;
-use pumpkin_world::{chunk::{ChunkData, TOTAL_SUBCHUNK}, DIRECT_PALETTE_BITS};
+use pumpkin_world::{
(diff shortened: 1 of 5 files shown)


pumpkin-protocol/src/client/play/c_chunk_data.rs
```diff
@@ -124,22 +146,6 @@ impl<'a> ClientPacket for CChunkData<'a> {
...
-                    continue;
-                }
-                let index = i / 2;
-                let mask = if i % 2 == 1 { 0xF0 } else { 0x0F };
-                chunk_light[index] |= mask;
-            }
-
-            lighting_subchunks.push(chunk_light);
-        });
-
```
Snowiiii, line 142: Why did you remove light ?

Alvsch: i moved it up into the other loop

Snowiiii:
I wonder how you benchmarked the speed difference and what the speed difference is

Alvsch:
I used both cargo-flamegraph and samply for profiling. (Both rust specific tools)

The total per chunk loading on my pc was reduced from ~40 ms to ~20 ms
This being flying around with 32 render distance.

Snowiiii:
Sounds very cool!
Thanks @Alvsch

PR #194, opened 2024-10-26
Implement max player limit

### LINKED ISSUE #189: Add check if the server is full

- Players can join if the server is full

- Add a check in `client_packet.rs`, after online mode authentication is done and after uuid/username duplicates checking is done. So in current master, around line 181.

**To-do**
- [ ] Maintainer likes idea
- [ ] @lokka30 started work on PR

lokka30:
Resolves #189 

Ready for review

image

TBD in future: allow operators to bypass full server. need to wait for operator/permission stuff to be implemented first.

### DIFF
--- a/pumpkin/src/client/client_packet.rs
+++ b/pumpkin/src/client/client_packet.rs
@@ -164,16 +164,25 @@ impl Client {
             }
         }
 
+        // Don't allow new logons when server is full.
+        // TODO: If client is an operator or otherwise suitable elevated permissions, allow client to bypass this requirement.
+        let player_count = server.get_player_count().await;
+        if player_count >= BASIC_CONFIG.max_players {
+            log::debug!("Player (IP '{}', username '{}') tried to log in, but the server is full ({}/{} players).", &self.address.lock().await, &profile.name, player_count, BASIC_CONFIG.max_players);
+            self.kick("The server is currently full, please try again later")
+                .await;
+        }
+
         // Don't allow duplicate UUIDs
         if let Some(online_player) = &server.get_player_by_uuid(profile.id).await {
-            log::debug!("Player (IP '{}', username '{}') tried to log in with the same UUID ('{}') as an online player (IP '{}', username '{}')", &self.address.lock().await.to_string(), &profile.name, &profile.id.to_string(), &online_player.client.address.lock().await.to_string(), &online_player.gameprofile.name);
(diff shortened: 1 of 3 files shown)


pumpkin/src/client/client_packet.rs
```diff
@@ -164,16 +164,25 @@ impl Client {
...
         }
 
+        // Don't allow new logons when server is full.
+        // TODO: If client is an operator or otherwise suitable elevated permissions, allow client to bypass this requirement.
+        let player_count = server.get_player_count().await;
+        if player_count >= BASIC_CONFIG.max_players {
+            log::debug!("Player (IP '{}', username '{}') tried to log in, but the server is full ({}/{} players).", &self.address.lock().await, &profile.name, player_count, BASIC_CONFIG.max_players);
+            self.kick("The server is currently full, please try again later")
+                .await;
+        }
```
Snowiiii, line 174: Why we should check so late and even after auth ?. It would make more sense to check directly when login starts

lokka30: Thanks for the review. Would you prefer me move it to the start of `handle_login_start`?

Snowiiii: Yes

lokka30: Resolved in 9fdc526. 🙂

pumpkin/src/server/mod.rs
```diff
@@ -163,6 +164,21 @@ impl Server {
...
+            count += world
+                .current_players
+                .lock()
+                .await
+                .len()
+                .to_u32()
+                .expect("Unable to convert to u32");
+        }
+        count
+    }
```
Snowiiii, line 180: you don't need to convert it to u32, just use usize

lokka30: `BASIC_CONFIG.max_players` is `u32` which is why I used the conversion, also `usize` is platform dependent , please let me know what you think

Snowiiii: So generally we want to use usize for such things like a player count. I also want to give the option to disable a player limit, maybe by setting it to -1 ?. the config value can stay u32 because this will be more than enough if someone wants to set a limit

Snowiiii: TL;DR return usize. If you compare it with the config value, cast it

lokka30: Cheers - should be resolved in 8aba1ff. I've also added logic that setting the max players to zero will disable the limit. (since it's unsigned, we can't use `-1`)

Snowiiii:
Looks good now.
Thanks @lokka30

PR #202, opened 2024-10-27
Added Worldborder

Alvsch:
Added worldborder packets to pumpkin-protocol
Added worldborder command
Created a worldborder struct stored in World


Snowiiii:
Is this tested ?

Alvsch:
Yes, I've now double checked every command and only found a small problem with the add command which is now fixed.

Snowiiii:
Okay, Looks good then.
Thanks @Alvsch

PR #206, opened 2024-10-28
Use #[derive(Default) instead of impl Default

suprohub:


Snowiiii:
Use the PR template. Empty descriptions are not welcome

PR #207, opened 2024-10-28
use derive macro for `Default` trait on `Scoreboard` struct and chunk optimizations

suprohub:
## Description
A Detailed Description:
when I was reading the code I accidentally came across this:
```rust
impl Default for Scoreboard {
    fn default() -> Self {
        Self::new()
    }
}
```
In my opinion, this is a useless piece of code, because we can use #[derive(Default)] instead of this, for speed up
code read time.

## Testing
`cargo run` and joining to server

## Checklist

- [yes ] Code is well-formatted and adheres to project style guidelines: `cargo fmt`
- [ yes] Code does not produce any clippy warnings `cargo clippy`


Bryntet:
Change the commit name to something like "use derive macro for `Default` trait on `Scoreboard`"

Snowiiii:
1. Use x instead of yes in checkboxes
2. Don't lie to me, Your Code does produce clippy warns

Snowiiii:
I also just noticed you not only added derive Default for Scoreboard, You added a bunch of things. This should be split up into multiple PR's

suprohub:
> I also just noticed you not only added derive Default for Scoreboard, You added a bunch of things. This should be split up into multiple PR's

how i can create another pull request? i can only view current

Snowiiii:
> how i can create another pull request? i can only view current

You should use different branches

PR #208, opened 2024-10-28
use derive macro for Default trait on Scoreboard struct

suprohub:
## Description
A Detailed Description:
when I was reading the code I accidentally came across this:
```rust
impl Default for Scoreboard {
    fn default() -> Self {
        Self::new()
    }
}
```
In my opinion, this is a useless piece of code, because we can use #[derive(Default)] instead of this, for speed up
code read time.

## Testing
`cargo run` and joining to server

## Checklist


Snowiiii:
Looks good.
Thanks @suprohub

PR #215, opened 2024-10-29
Adds "Expensive Task" API and eager cancelling to client

kralverde:
## Description
A Detailed Description:
Adds an idea of "expensive tasks" to the client where the client can watch async tasks that take a lot of compute power so they can be cancelled if no longer needed.

## Testing
A description of the tests performed to verify the changes:
Built in debug mode for slow chunk spawning, then flew around.

## Checklist
N/A

### DIFF
--- a/pumpkin-world/src/level.rs
+++ b/pumpkin-world/src/level.rs
@@ -16,7 +16,7 @@ use crate::{
     world_gen::{get_world_gen, Seed, WorldGenerator},
 };
 
-pub type ConcurrentChunkResult = Vec<(Vector2<i32>, JoinHandle<()>)>;
+pub type ConcurrentChunkResult = (Vec<Vector2<i32>>, Vec<JoinHandle<()>>);
 
 /// The `Level` module provides functionality for working with chunks within or outside a Minecraft world.
 ///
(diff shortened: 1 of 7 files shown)


Bryntet:
Love this idea! At a cursory glance this implementation looks very good too! Great job :)

Snowiiii:
So actually a Client is just a "tool" for Networking stuff, Kinda like a NetworkingHandler in Java  Minecraft. It provides simple functions for sending and receiving Packets. We also use them for all States before Play. I wonder why we should have any Expensive tasks in the Client. If you want to cancel some "Expensive" Player task e.g. Chunk generation, We have already function which get called when the Player closed the connection or we close the Player connection. For Example `player::remove()` or `world::remove_player`

kralverde:
I know, I wrote the code. I can port it to player if that's what you really want.

Snowiiii:
> I can port it to player if that's what you really want.
> 
> As explained in the discord, player disconnect isn't ideal because it's async and work we don't need will likely be done before it's invoked, it needs to be synchronous.

I would like to have this in the Player, Like i said see the `Client` more as a util for Networking.

lukas0008:
Expensive tasks shouldnt be run inside of async context, it is generally unadvised to do so, this is why rayon exists for example.

PR #216, opened 2024-10-30
small CChunkData write improvement

Alvsch:
## Description
- Replaced the HashMap in CChunkData::write with a Vec
- Added profiling profile

Even though Vec has a worse time complexity for this, the HashMap overhead is worse for performance

## Testing
Performance profiling was done using cargo-flamegraph.
If this breaks something I'll be genuinely surprised.

## Checklist

### DIFF
--- a/pumpkin-protocol/src/client/play/c_chunk_data.rs
+++ b/pumpkin-protocol/src/client/play/c_chunk_data.rs
@@ -54,9 +54,7 @@ impl<'a> ClientPacket for CChunkData<'a> {
                     // Palette length
                     data_buf.put_var_int(&VarInt(palette.len() as i32));
 
-                    let mut palette_ids = Vec::with_capacity(palette.len());
                     palette.iter().for_each(|id| {
-                        palette_ids.push(*id);
                         // Palette
                         data_buf.put_var_int(&VarInt(id.get_id_mojang_repr()));
                     });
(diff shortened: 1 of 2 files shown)


pumpkin-protocol/src/client/play/c_chunk_data.rs
```diff
@@ -55,23 +53,26 @@ impl<'a> ClientPacket for CChunkData<'a> {
                     data_buf.put_u8(block_size as u8);
                     // Palette length
                     data_buf.put_var_int(&VarInt(palette.len() as i32));
-                    let mut palette_map = HashMap::new();
-                    palette.iter().enumerate().for_each(|(i, id)| {
-                        palette_map.insert(*id, i);
+
+                    let mut palette_ids = Vec::with_capacity(palette.len());
```
BenjaminBrienen, line 57: You could probably improve this by removing this Vec and using a lazy iterator instead. This vec seems to only be used for an iterator.

Alvsch: Good idea, reusing the palette Vec works the same. No need to have the palette_ids Vec

BenjaminBrienen (approved):

Snowiiii:
I tested the changes on pre-generated worlds and our own Chunk gen and can confirm that everything works.
Thanks @Alvsch

PR #217, opened 2024-10-30
Lighting propagation

EamonHeffernan:


Snowiiii:
Empty PRs are not welcome, Use the PR template

extractor/src/main/kotlin/de/snowii/extractor/extractors/Blocks.kt
```diff
@@ -52,10 +52,14 @@ class Blocks : Extractor.Extractor {
             for (state in block.stateManager.states) {
                 val stateJson = JsonObject()
                 val id = stateIdCounter
+
                 stateIdCounter++
+
                 stateJson.addProperty("id", id)
                 stateJson.addProperty("luminance", state.luminance)
+                stateJson.addProperty("opacity", state.opacity)
```
Snowiiii, line 60: I would put this below the opaque boolean

Snowiiii: Actually we can may remove the opaque and only write opacity when the block has opaque. So in rust we could make it an Option<Value>

Snowiiii:
Feel free to reopen when you provide good informations about your PR

PR #218, opened 2024-10-31
Implement Query protocol

neeleshpoli:
## Description
This implements the Query protocol, supported by the vanilla server, to allow for a different way to get information about the server.

More info at protocol [docs](https://wiki.vg/Query)

## Testing
Send test query packets using services like
- https://mcsrvstat.us/
- https://mcstatus.io/
- And many more (just anything that checks server status with the query protocol)

## Checklist

- [x] Add query options to config
- [x] Added packet data types
- [x] Correctly handle config options
- [x] Correctly parse a handshake packet
- [x] Correctly respond to handshake packet
- [x] Correctly parse basic status packet
- [x] Correctly respond to basic status packet
- [x] Correctly parse full status packet
- [x] Correctly respond full status packet
- [x] Handle errors correctly
- [x] Update docs and README (#228)


pumpkin-protocol/src/query.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug)]
+pub enum PacketType {
+    Handshake = 9,
+    Stat = 0,
+}
+
+#[derive(Debug)]
+pub struct SBasePacket {
+    pub magic: u16,
+    pub packet_type: PacketType,
```
Snowiiii, line 12: In the docs it says that the packet type is a byte. You can still use the Enum but you should add `repr(i8)`

pumpkin-protocol/src/query.rs
```diff
@@ -0,0 +1,89 @@
...
+                session_id: reader.read_i32().await.unwrap(),
+                payload: SBasePayload::Handshake, 
+            },
+            _ => todo!(),
+        }
+    }
+}
+
+pub struct CBasePacket {
+    pub packet_type: PacketType,
```
Snowiiii, line 42: here its an byte as well

pumpkin-protocol/Cargo.toml
```diff
@@ -14,7 +14,7 @@ serde.workspace = true
 thiserror.workspace = true
 itertools.workspace = true
 log.workspace = true
-
+tokio.workspace = true
```
Snowiiii, line 17: For what you need tokio here ?

neeleshpoli: For the convenience functions provided by `AsyncReadExt` and `AsyncWriteExt` in tokio. This includes functions like read_u8, read_i32, write_i32, write_u16_le etc. Although thinking about it, adding tokio for convenience seems a little excessive. I'll probably remove tokio 👍

neeleshpoli: On second thought, I'll probably keep tokio since it would overcomplicate the code. But if you feel like tokio shouldn't be in pumpkin-protocol, then I'll go ahead and get rid of it 👍

Snowiiii:
I think you don't need to save the clients into a HashMap, Your example we also don't save RCON Clients into a hashmap: https://github.com/Snowiiii/Pumpkin/blob/master/pumpkin/src/rcon/mod.rs#L33. I also very dislike to have a Struct named "QueryClients"

neeleshpoli:
Looking at RCON, since it uses TCP each client can take a TcpStream and not have to be saved in a central place. The thing is UDP is stateless, so there is nothing like a UdpStream where each client can own their connection. This means that the clients need to be stored in a central place. 

What should I rename `QueryClients` to then?

StripedMonkey (changes requested):
Overall this looks great! I appreciate how self-contained this PR is. I have a couple of requested changes, but it's mostly to try and clean things up.

I won't complain if you said no to the extra work, but I would truly appreciate it if you wrote a handful of tests for some of the (de)serialization to validate the format. Just something which can be used to verify that a packet gets read (or written) in a way that means if someone comes through "cleaning" up this code, we can know it matches what we expect from your original implementation.

Snowiiii:
Just tested and it works pretty well. I also like the new changed code much more (Thanks @StripedMonkey for helping reviewing).
Overall excellent work @neeleshpoli, Thank you

Snowiiii:
@neeleshpoli I just noticed that for some reason Query is on by default. It should be off by default

Snowiiii:
> @neeleshpoli I just noticed that for some reason Query is on by default. It should be off by default

Currently don't have time, if you want you can make a PR, I think it should be pretty easy to fix

PR #224, opened 2024-11-01
Add damage and re-spawning

Alvsch:
## Description
Adds damage, re-spawning and fall damage.
I feel like improvements can be made but don't remember what I was thinking about as it was a couple days ago.

## Testing
Damage, fall damage and re-spawning gone through basic testing and works.
In the basic testing  I also checked and it seems like the weird fall damage calculations are there.

## Checklist


Snowiiii:
https://github.com/Snowiiii/Pumpkin/pull/224/commits/26252a2f776d98036007d433de57d30c13bc80bf you should then also not set the player to flying

Alvsch:
should i keep the allow_flying or remove that too?

pumpkin/src/world/mod.rs
```diff
@@ -192,7 +192,6 @@ impl World {
         {
             let mut abilities = player.abilities.lock().await;
             abilities.allow_flying = true;
-            abilities.flying = true;
         }
         player.send_abilties_update().await;
```
Snowiiii, line 197: At this point you we can actually just not send the abilities at all

Alvsch: if we dont want survival mode player to allow flight we can remove the mutating of abilities here. But there are still other variables that need to be sent

pumpkin/src/entity/living.rs
```diff
@@ -72,6 +97,38 @@ impl LivingEntity {
         amount > 0.0
     }
 
+    pub async fn update_fall_distance(&self, dont_damage: bool) {
+        let y = self.entity.pos.load().y;
+        let last_y = self.entity.last_pos.load().y;
```
Snowiiii, line 102: You moved `last_pos` to `entity` to calculate the fall distance, Lets move `last_pos` to `living_entity` instead

Alvsch: sounds like a good idea.
I think there was some reason i put it in `entity` instead of `living_entity`. But it not needed now and neither can i remember the reason.

Snowiiii:
Looks good. Big thanks @Alvsch.
Good work like always

Issue #94, opened 2024-09-13
Add tab support for commands


Snowiiii:
Support Name Tabing when writing Commands. We want to add Support for both Console and In-Game

## Game Packets
- [x] https://wiki.vg/Protocol#Command_Suggestions_Response
- [x] https://wiki.vg/Protocol#Commands (Optional i think)

wendulem:
@Snowiiii Hey, first time contributor. Could you document this a bit more. Do you mean that when processing stdin in main.rs the command dispatcher (given the user pressed tab) should output the command names that match - or complete stdin with the uniquely matching command in that case

Snowiiii:
> @Snowiiii Hey, first time contributor. Could you document this a bit more. Do you mean that when processing stdin in main.rs the command dispatcher (given the user pressed tab) should output the command names that match - or complete stdin with the uniquely matching command in that case

Hey, Great your interested in contributing. Pumpkin is currently heavily in WIP. That's the Reason documentation is so poor. I also opend this issue recently and completely forgot about it. I will edit it a bit to give more Informations.

So the Plan is to support both Tab support in the Server Console and In-Game.

We have a Command Dispatcher struct which is stored in `pumpkin/server.rs`. The Dispatcher has all Command

marc-marcos:
If @Snowiiii isn't actively working on it I would like to give it a try!

Snowiiii:
> If @Snowiiii isn't actively working on it I would like to give it a try!

No, Problem. go ahead

user622628252416:
@marc-marcos Hi, are you still working on this? Else I'd pick this up

marc-marcos:
Go for it!

PR #227, opened 2024-11-02
Commands: Fix Bug, Refactor Arguments, Add /tp

### LINKED ISSUE #210: Worldborder Doesn't Check for Negatives

### Operating System

Arch Linux

79d2f91

### What happened?

Worldborder command can be executed using negative numbers, which can cause the player to be kicked from the game and unable to rejoin. Vanilla minecraft doesn't allow a worldborder less one.

### To Reproduce

/worldborder set -1000

### Expected behavior

log syntax error "Worldborder cannot be smaller than 1 block wide"

user622628252416:
## Description
- Fix bug in command dispatch
- refactor command argument parsers: replace one-dimensional posititon by position2d and position3d
- support relative coordinates in position3d (using ~)
- refactor and implement arg_player/arg_entity
- add arg_entities as a placeholder for when other entities than players are a thing
- remove some curly braces where they are not necessary
- add teleport command (#15)
- add bounded number argument type
- replace simple args in cmd_worldborder with appropriate number arg types
- move args to seperate module/folder
- remove second argument parsing step for better performance and more concise code
- fix #210

## Testing
ran commands and evaluated their effects a lot (especially affected commands: help/tp/worldborder)

## Checklist


Snowiiii:
I would prefer to have all arguments in a folder like the commands, makes it much cleaner

pumpkin/src/command/commands/cmd_teleport.rs
```diff
@@ -0,0 +1,276 @@
...
+const ARG_TARGETS: &str = "targets";
+
+/// rotation: yaw/pitch
+const ARG_ROTATION: &str = "rotation";
+
+/// single entity
+const ARG_FACING_ENTITY: &str = "facingEntity";
+
+/// position
+const ARG_FACING_LOCATION: &str = "facingLocation";
```
Snowiiii, line 33: I also would like to have the string in the arguments, Its annoying to need add them every time

user622628252416: Hi, could you elaborate? I could inline those consts of course, but they have several advantages:
- IDE auto completion
- typos become compiler errors instead of runtime errors
- actual value can be changed very quickly, eg if we migrate from using string keys to enums or such
- you can see all arguments at a glance

Also, for this command, the work is already done. I totally see why you would want to avoid that work for new commands tho, especially if they are less complex.

user622628252416:
> I would prefer to have all arguments in a folder like the commands, makes it much cleaner

I've already been working on that actually, but before I can commit that I need another issue to be resolved: IMO ArgumentConsumer's consume function should just return an Option<CommandArgument> instead of a Result<String, Option<String>>. As discussed in the comment in CommandDispatcher, we can really use these optional error strings anyways, at least not without a lot of complexity that will slow us down.

See [dispatcher.rs](https://github.com/Snowiiii/Pumpkin/blob/3fda191f0666967ea049a65e16e7ef724027c74d/pumpkin/src/command/dispatcher.rs)
> is_fitting_path being Err only means that THIS path does not fit Do not break/return here or else other paths aren't even tried! I don't really know what to do with this _error string, because there will be one error for every path that was tried. Since there are sometimes many possible paths, aggregating these errors and showing them all to the user probably does not make sense. We can't really know which path the user intended, and just show the error for that one path, either. (we could guess but that'd be hard). IMO is_fitting_path should therefore just be a bool.

Removing this Error in favor of just using an option makes the Argument Consumers much much cleaner too.

Additionally, the aforementioned CommandArgument would be an enum that has a variant for each arguments type: This way we can simply store parsed data in the HashMap, instead of converting consumed data back to a String and having a second parse step.

Snowiiii:
> > I would prefer to have all arguments in a folder like the commands, makes it much cleaner
> 
> I've already been working on that actually, but before I can commit that I need another issue to be resolved: IMO ArgumentConsumer's consume function should just return an Option instead of a Result<String, Option>. As discussed in the comment in CommandDispatcher, we can really use these optional error strings anyways, at least not without a lot of complexity that will slow us down.
> 
> See [dispatcher.rs](https://github.com/Snowiiii/Pumpkin/blob/3fda191f0666967ea049a65e16e7ef724027c74d/pumpkin/src/command/dispatcher.rs)
> 
> > is_fitting_path being Err only means that THIS path does not fit Do not break/return here or else other paths aren't even tried! I don't really know what to do with this _error string, because there will be one error for every path that was tried. Since there are sometimes many possible paths, aggregating these errors and showing them all to the user probably does not make sense. We can't really know which path the user intended, and just show the error for that one path, either. (we could guess but that'd be hard). IMO is_fitting_path should therefore just be a bool.
> 
> Removing this Error in favor of just using an option makes the Argument Consumers much much cleaner too.
> 
> Additionally, the aforementioned CommandArgument would be an enum that has a variant for each arguments type: This way we can simply store parsed data in the HashMap, instead of converting consumed data back to a String and having a second parse step.

Sure, We can use an Option. But i would have proper "Error Handling" in some way. I don't want to just give the player back "Wrong Argument", This is confusing.

pumpkin/src/command/commands/cmd_teleport.rs
Snowiiii: I don't really see the points, I have the feeling you don't understand what i want. We already have like 4 times `ARG_TARGETS` constant. **Which is annoying and can be inconsistent**. So Lets just go to the `ArgumentConsumer` trait and add a const there like `NAME: String`. Then we automatically can use the names from everywhere. I would also make working with translations easier

user622628252416:
We could go through all possible paths, and find the path that we can traverse the furthest before consuption of an argument fails. I believe that's what Minecraft does. Then we could print the syntax of that path up to that point. This will have to wait tho, as it depends on being able to print possible continuations from any node in a path. This might also be related to command suggestions

pumpkin/src/command/commands/cmd_teleport.rs
user622628252416: > I have the feeling you don't understand what i want

Yes, that's why I asked you to elaborate :)

I don't think we should translate these at all (and I don't think Minecraft does either), as they are mostly internal identifiers, i.e. their primary use is as keys in a HashMap. That also means that for any given path, each arg needs a unique name. Adding them to ArgumentConsumer won't work, because some commands have paths where several arguments use the same consumer. For example /fill needs a start Pos3d and an end Pos3d and they can't have the same name.

Snowiiii: 1. I also don't think Minecraft does that but Pumpkin wants to have more Features than Vanilla, I think it could be cool to have the args translated, I don't say that this had to be now or in the future but its just an idea, Because there are efforts going on to add Translations to Pumpkin (https://github.com/Snowiiii/Pumpkin/issues/120).
2. Your right, I was thinking of having a default name in the trait and let the user also give the option to choose a custom name, Example:

Will use default:
`parse_arg_position_3d(args)?;`
Will use custom name:
`parse_arg_position_3d("my_cool_custom_position", args)?;`

Because its way more common to use the default than an custom name

Snowiiii:
> We could go through all possible paths, and find the path that we can traverse the furthest before consuption of an argument fails. I believe that's what Minecraft does. Then we could print the syntax of that path up to that point. This will have to wait tho, as it depends on being able to print possible continuations from any node in a path. This might also be related to command suggestions

I belive this is what we did before i made custom error reporting. Im not totally against this. But for "not experienced players". An proper error message like "Player not Found". Is more clear than "<player> wrong Argument" or something like that

pumpkin/src/command/commands/cmd_teleport.rs
user622628252416: I actually just wrote a macro to replace those parse functions, so this might be very doable 👍

user622628252416: @Snowiiii 

Args can be retrieved in executor using `let distance = get_parsed_arg!(args, ARG_DISTANCE, Arg::F64(v), *v)?;`.
If default name from Consumer should be used: `let Vector2 { x, z } = get_parsed_arg_default!(args, Position2DArgumentConsumer, Arg::Pos2D(vec), *vec)?;`

e.g. `argument_default(&Position2DArgumentConsumer)` can be used in init_command_tree, too.

(All examples are taken from cmd_worldborder.rs)

user622628252416:
> But for "not experienced players". An proper error message like "Player not Found". Is more clear than " wrong Argument" or something like that

I honestly think when #94 is implemented this is a non-issue

Snowiiii:
btw you have a typo you wrote [arg_postition_2d.rs](https://github.com/Snowiiii/Pumpkin/pull/227/files#diff-b66f18b635fe82382e1ea923756c77e401a5743fc5813e7a5db6e65508bd2b74)

pumpkin/src/command/commands/cmd_teleport.rs
Snowiiii: I like `argument_default(&Position2DArgumentConsumer)`. I find the other macros really confusing

user622628252416: I added some more documentation to those. All they do is retrieve the parsed argument from the HashMap and if it matches with the given enum variant, return the enum variants content. The same can be done explicitly, but that'd be several more lines of code for each argument.

user622628252416: So basically
```
let targets = get_parsed_arg!(args, ARG_TARGETS, Arg::Entities(e), e)?;
```
is the same as 
```
let targets = match args.get(ARG_TARGETS) {
    Some(Arg::Entities(e)) => Ok(e),
    _ => Err(crate::command::dispatcher::InvalidTreeError::InvalidConsumptionError(Some(ARG_TARGETS.into())))
}?;`
```

Snowiiii: Im still not a big fan of that, The Whole args API just looks pain in the ass and definitely not fun. I want to a have simple to get things.
I could put the args parsing the to arg files again and have it much simpler. Example
`let entites = EntitiesArgumentConsumer::parse(args, ARG_TARGETS)`

user622628252416: I previously discarded that option because parse cannot be part of the ArgumentConsumer trait as its return type would be generic and that's not allowed for trait objects. Having parse implemented on the structs directly would work tho, I'll look into that.

user622628252416: @Snowiiii I created a FindArg trait and implemented it for all current cosumers and a FindArgDefaultName trait that is automatically implemented for every consumer that implements both DefaultNameArgConsumer and FindArg

Snowiiii: Better already

Snowiiii:
Is this ready for merge ?

user622628252416:
I just did some last testing, I think it's ready 👍

Snowiiii:
Big Thanks @user622628252416. Really appreciate your work :D

Issue #230, opened 2024-11-04
Infinite block placing + incorrect block placing


enthermo:
### I've searched existing issues and couldn't find a duplicate.

- [X] I confirm this is not a duplicate.

### Operating System

Windows 11, Minecraft Version 1.21.3 (Latest)

### Server Software Version/Commit

87aa5ec

### What happened?

The server copies the first placed block by the player and infinitely places it, regardless of what the player has in their hand or if they have enough of that item to place. This works with an empty inventory and in survival.

I have seen the server be 'fine' for ~15 seconds and then remove all blocks (regen the chunk?) and then do this bug. Clearly the server is preserving the first block the player interacts with and not updating, should be an easy fix.

### To Reproduce

```
git clone (this repo)
cargo build -r
cargo run -r
```
Join server running on localhost, use the server console to change gamemode to creative and add blocks to inventory. Place any block, then switch to a different block. The first placed block will be placed instead of the selected block.

### Expected behavior

The server should verify if the player has an item in hand and place the appropriate block.

enthermo:
Found the issue [here](https://github.com/Snowiiii/Pumpkin/blob/87aa5ec49305b7fdafc2e737cd18a6cee24f5403/pumpkin/src/client/player_packet.rs#L592). When set `item_allowed_override` is set to false the server behaves normally. Leaving this issue open as this is likely not intended behavior.

Snowiiii:
Currently the item count is not decreased when a block is being placed by a player in survival mode. I will take a look on this tomorrow

PR #235, opened 2024-11-06
Fix: Use correct packet for set_help_item

duszmox:
*

## Description
A Detailed Description:  According to the wiki we need to use (~~0x2F~~) 0x31 for set held item, this fixes the bug where if you have something in your 1st slot you would be able to place it with any other slot selected. 
This issue is partly referenced in #230

## Testing
A description of the tests performed to verify the changes: Put a block in the first slot, select said slot, then change the selected slot to any other, then try placing the block

## Checklist

### DIFF
--- a/pumpkin-protocol/src/server/play/s_set_held_item.rs
+++ b/pumpkin-protocol/src/server/play/s_set_held_item.rs
@@ -2,7 +2,7 @@ use pumpkin_macros::server_packet;
 use serde::Deserialize;
 
 #[derive(Deserialize)]
-#[server_packet("play:pick_item")]
+#[server_packet("play:set_carried_item")]
 pub struct SSetHeldItem {
     pub slot: i16,
 }


lukas0008:
The packet IDs changed as of the lastest version, so the packet ids you see on wiki.vg are outdated (since they are for version 1.21.1). If you want to see up-to-date packet ids, go into the packets.json file in the assets folder of Pumpkin. (note that the ones in the json file are decimal, not hex, so if you'd like to compaire, convert first)

duszmox:
That explains why they didn't match. But we don't use the IDs directly in the packet class declarations, we use their names. I've fixed the PR description, but the commit is still correct.

Snowiiii:
> That explains why they didn't match. But we don't use the IDs directly in the packet class declarations, we use their names. I've fixed the PR description, but the commit is still correct.

I used wiki.vg for a long time. But i now recommend against using it as packet reference because it sometimes can have wrong/outdated information. Looking directly at Vanilla code is always safer

Snowiiii:
But the fix makes sense. I should add the pick_item packet as well. 
Thank you @duszmox

PR #239, opened 2024-11-07
Fix: Set air block state when breaking a block

duszmox:
* Fixes the dirt block particles when breaking a block

## Description
The server broadcasted a block breaking event with the block state `11`. Which meant that it set the new block's (air block)
state to 11, which isn't a valid state for that block.

## Testing
Place a block, which is not dirt or grass and then break it, no more dual particles

## Checklist

- [ ] Maybe consider using a constant 0 instead of importing the static var

### DIFF
--- a/pumpkin/src/client/player_packet.rs
+++ b/pumpkin/src/client/player_packet.rs
@@ -498,7 +498,6 @@ impl Player {
                         return;
                     }
                     // Block break & block break sound
-                    // TODO: currently this is always dirt replace it
                     let entity = &self.living_entity.entity;
                     let world = &entity.world;
                     world.break_block(location).await;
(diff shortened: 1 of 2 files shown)


duszmox:
I've imported the BlockState class in case there would be a change in the air block's block state. Not sure if it's that important, maybe a constant 0 would be enough as well

Snowiiii:
These are the sounds and particles being broadcasted. It should be not air, It should be the block broken

duszmox:
So it should be the old block's state id rather than the new one's?

Snowiiii:
> So it should be the old block's state id rather than the new one's?

Of course, Air has no sound and no Breaking Particles. We want to use the ID of the broken block

PR #245, opened 2024-11-07
Add Command Suggestions and Tab Completion

user622628252416:
## Description
Add command tab completion and visual suggestions

## Testing
- observed suggestions in game

## Checklist

- [x] add packet for client-side suggestions
- [x] add packets for server-side suggestions
- [x] target selectors
- [x] don't accept numbers as entities (note to myself: maybe the order argument nodes are sent to the client in matters?)
---


Snowiiii:
I was waiting for this :smile:

user622628252416:
@Snowiiii Not sure whether you got a notification already, but this pr is ready for review now 🚀

pumpkin-protocol/src/client/play/c_command_suggestions.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+impl<'a> ClientPacket for CCommandSuggestions<'a> {
+    fn write(&self, bytebuf: &mut crate::bytebuf::ByteBuffer) {
+        bytebuf.put_var_int(&self.id);
+        bytebuf.put_var_int(&self.start);
+        bytebuf.put_var_int(&self.length);
+
+        bytebuf.put_list(&self.matches, |bytebuf, suggestion| {
+            bytebuf.put_string_len(&suggestion.suggestion, 32767);
```
Snowiiii, line 39: 32767 equals `i16::MAX` this is the default max string length. So you can just use `bytebuf.put_string()`

pumpkin/src/main.rs
```diff
@@ -193,6 +194,10 @@ async fn main() -> io::Result<()> {
             {
                 let (player, world) = server.add_player(client).await;
                 world.spawn_player(&BASIC_CONFIG, player.clone()).await;
+
+                // inform client about commands so that command suggestions and tab completion work
+                send_c_commands_packet(player.clone(), &server.command_dispatcher).await;
```
Snowiiii, line 199: We definitely don't want to have this in the main. It would be much nicer and make more sense to put this into the Player or World. There would be also no need for cloning then

pumpkin-protocol/src/client/play/c_commands.rs
```diff
@@ -0,0 +1,382 @@
...
+                    let suggestion_type = &override_suggestion_type.expect("ProtoNode::FLAG_HAS_SUGGESTION_TYPE should only be set if override_suggestion_type is not None.");
+                    bytebuf.put_string(suggestion_type.identifier());
+                },
+                _ => unimplemented!("ProtoNode::FLAG_HAS_SUGGESTION_TYPE is only implemented for ProtoNodeType::Argument"),
+            }
+        }
+    }
+}
+
+/// see https://wiki.vg/Command_Data for meaning of flags/enums/etc.
```
Snowiiii, line 127: I recommend against using wiki.vg

pumpkin-protocol/src/client/play/c_commands.rs
```diff
@@ -0,0 +1,382 @@
...
+        identifier: &'a str,
+    },
+    ResourceKey {
+        identifier: &'a str,
+    },
+    TemplateMirror,
+    TemplateRotation,
+    Heightmap,
+    Uuid,
+}
```
Snowiiii, line 208: This Enum is quiet large and can change from time to time trough updates, I think generating and JSON of these Types in our Extractor would be an better soultion

pumpkin-protocol/src/client/play/c_commands.rs
```diff
@@ -0,0 +1,382 @@
...
+
+        // redirect node
+        if flags & Self::FLAG_HAS_REDIRECT != 0 {
+            bytebuf.put_var_int(&1.into());
+        }
+
+        // name
+        match self.node_type {
+            ProtoNodeType::Argument { name, .. } | ProtoNodeType::Literal { name, .. } => {
+                bytebuf.put_string_len(name, 32767)
```
Snowiiii, line 98: You can use `bytebuf.put_string` here as well

pumpkin-protocol/src/client/play/c_commands.rs
```diff
@@ -0,0 +1,382 @@
...
+        is_executable: bool,
+        parser: ProtoCmdArgParser<'a>,
+        override_suggestion_type: Option<ProtoCmdArgSuggestionType>,
+    },
+}
+
+impl<'a> ProtoNode<'a> {
+    const FLAG_IS_EXECUTABLE: i8 = 4;
+    const FLAG_HAS_REDIRECT: i8 = 8;
+    const FLAG_HAS_SUGGESTION_TYPE: i8 = 16;
```
Snowiiii, line 52: Im not sure. but i think we just can make all these const's unsigned

pumpkin-protocol/src/client/play/c_command_suggestions.rs
```diff
@@ -0,0 +1,61 @@
...
+    start: VarInt,
+    length: VarInt,
+    matches: Vec<CommandSuggestion<'a>>,
+}
+
+impl<'a> CCommandSuggestions<'a> {
+    pub fn new(
+        id: impl Into<VarInt>,
+        start: impl Into<VarInt>,
+        length: impl Into<VarInt>,
```
Snowiiii, line 20: This is against the design of Pumpkin, We use just VarInt's as args

pumpkin/src/command/commands/cmd_pumpkin.rs
```diff
@@ -35,5 +38,7 @@ impl CommandExecutor for PumpkinExecutor {
 }
 
 pub fn init_command_tree<'a>() -> CommandTree<'a> {
-    CommandTree::new(NAMES, DESCRIPTION).execute(&PumpkinExecutor)
+    CommandTree::new(NAMES, DESCRIPTION)
+        .with_child(literal("test").with_child(pumpkin_test_client_side_arg_parsers()))
```
Snowiiii, line 42: I really don't like this.
1. "test" does really say nothing what is being tested.
2. I think just rust unit testing would be enough

pumpkin/src/command/commands/cmd_pumpkin_test_client_side_arg_parsers.rs
```diff
@@ -0,0 +1,112 @@
...
+static RESOURCEORTAGKEY: LazyLock<ClientSideArgParserTester<'static>> = LazyLock::new(|| ClientSideArgParserTester(ProtoCmdArgParser::ResourceOrTagKey { identifier: "item" }));
+static RESOURCE: LazyLock<ClientSideArgParserTester<'static>> = LazyLock::new(|| ClientSideArgParserTester(ProtoCmdArgParser::Resource { identifier: "item" }));
+static RESOURCEKEY: LazyLock<ClientSideArgParserTester<'static>> = LazyLock::new(|| ClientSideArgParserTester(ProtoCmdArgParser::ResourceKey { identifier: "item" }));
+static TEMPLATEMIRROR: LazyLock<ClientSideArgParserTester<'static>> = LazyLock::new(|| ClientSideArgParserTester(ProtoCmdArgParser::TemplateMirror));
+static TEMPLATEROTATION: LazyLock<ClientSideArgParserTester<'static>> = LazyLock::new(|| ClientSideArgParserTester(ProtoCmdArgParser::TemplateRotation));
+static HEIGHTMAP: LazyLock<ClientSideArgParserTester<'static>> = LazyLock::new(|| ClientSideArgParserTester(ProtoCmdArgParser::Heightmap));
+static UUID: LazyLock<ClientSideArgParserTester<'static>> = LazyLock::new(|| ClientSideArgParserTester(ProtoCmdArgParser::Uuid));
+
+pub(super) fn pumpkin_test_client_side_arg_parsers() -> NonLeafNodeBuilder<'static> {
```
Snowiiii, line 59: Also like said these Types can change. So we would have to also change the code all the time

pumpkin/src/command/commands/cmd_give.rs
```diff
@@ -17,7 +17,7 @@ const DESCRIPTION: &str = "Give items to player(s).";
 
 const ARG_ITEM: &str = "item";
 
-static ITEM_COUNT_CONSUMER: BoundedNumArgumentConsumer<u32> =
+static ITEM_COUNT_CONSUMER: BoundedNumArgumentConsumer<i32> =
```
Snowiiii, line 20: Why should the item count be negative ?

pumpkin-protocol/src/client/play/c_command_suggestions.rs
```diff
@@ -0,0 +1,61 @@
...
+            if let Some(tooltip) = &suggestion.tooltip {
+                bytebuf.put_slice(&tooltip.encode());
+            }
+        })
+    }
+}
+
+#[derive(PartialEq, Eq, Hash, Debug)]
+pub struct CommandSuggestion<'a> {
+    pub suggestion: Cow<'a, str>,
```
Snowiiii, line 50: I don't think this need to be an Cow, We usally just use an an &str for all Clientbound Packets

pumpkin/src/main.rs
user622628252416: I put it in spawn_player now and removed some .clone()s (tho it's just Arcs so these clones are cheap). I'm pretty sure I put it in the correct place for following vanilla packet order, too. 

And omg why I didn't start using Intellij for browsing vanilla source code earlier and instead relied on text search only. It's soo much easier this way

pumpkin-protocol/src/client/play/c_commands.rs
user622628252416: It was pretty helpful actually, but as I said above, I for some reason limited myself to text search when browsing vanilla source code so it was a struggle finding the relevant spots. I'll remove that comment

pumpkin-protocol/src/client/play/c_commands.rs
user622628252416: Yes, but I'd need you to extract the "command_argument_type" registry for that. This would also mean losing type safety regarding the properties tho.

I'm not sure how you'd want to extract these properties anyways, since the registry just provides objects that implement an interface for reading/writing these. Could you maybe do something with that Template interface?

pumpkin-protocol/src/client/play/c_commands.rs
user622628252416: I don't think that's a good idea, because 1. I'm not sure either and 2. it does work this way and we don't really have a reason to change it and 3. it's consistent with vanilla which doesn't have unsigned integers

pumpkin-protocol/src/client/play/c_command_suggestions.rs
user622628252416: Thats quite unergonomic but I changed it

pumpkin/src/command/commands/cmd_pumpkin.rs
user622628252416: it's not just test, there are more literal arguments that are more specific. What this is for can't be done in a unit test because it's about seeing how the client handles each argument type in game. This way I could check whether I got all the argument type properies correct for example. I think I am going to remove it for now tho because it was just a really low effort way of doing this

pumpkin/src/command/commands/cmd_give.rs
user622628252416: because minecraft does not have a parser for unsigned ints because java doe not have unisgned ints

I apparently forgot to add a min bound tho when I changed it so I'll do that rq

pumpkin-protocol/src/client/play/c_command_suggestions.rs
user622628252416: iirc there was a case where I wanted to use player names in command suggestions, and I could not send the packet to a player while its name was borrowed. I'll change it tho, entity arguments should be handled client side anyways.

pumpkin-protocol/src/client/play/c_commands.rs
Snowiiii: 1. Its should work.
2. When we are talking about unsigned, I don't want to keep consistency with Mojang. Because first there is no unsigned in Java :D. And Secound there are alot of cases which do not make sense to be signed and can actually break the game. For example the render distance is signed, Pumpkin just converts it into unsigned

user622628252416: You're right actually. I for some reason completely ignored the fact that all that matters for flags is the binary representation

Snowiiii:
Currently testing this. I wonder why on every command there is "execute" ?
image

pumpkin-protocol/src/client/play/c_commands.rs
```diff
@@ -0,0 +1,353 @@
...
+    Dimension,
+    Gamemode,
+    Time { min: i32 },
+    ResourceOrTag { identifier: &'a str },
+    ResourceOrTagKey { identifier: &'a str },
+    Resource { identifier: &'a str },
+    ResourceKey { identifier: &'a str },
+    TemplateMirror,
+    TemplateRotation,
+    Heightmap,
```
Snowiiii, line 177: loot_table, loot_predicate and loot_modifier are missing. They are below Heightmap

user622628252416:
> Currently testing this. I wonder why on every command there is "execute" ? image

image

What version of Minecraft are you using? I'm not seeing this at all

Snowiiii:
> > Currently testing this. I wonder why on every command there is "execute" ? image
> 
> image
> 
> What version of Minecraft are you using? I'm not seeing this at all

I sue 1.21.3. I still see "execute" on every Command, It goes away when pressing ESC

pumpkin-protocol/src/client/play/c_commands.rs
user622628252416: image

I even looked at this, no idea how I could miss that. But they're added now

user622628252416:
> > > Currently testing this. I wonder why on every command there is "execute" ? image
> > 
> > 
> > image
> > What version of Minecraft are you using? I'm not seeing this at all
> 
> I sue 1.21.3. I still see "execute" on every Command, It goes away when pressing ESC

That's extremely weird, but I can't reproduce it. Maybe I could try running pumpkin on windows, but I don't really know what could cause this

user622628252416:
When I run on windows it works fine aswell

Snowiiii:
> When I run on windows it works fine aswell

Im on Linux

user622628252416:
I usually develop on ubuntu using wsl2 and it works there. Ill ask on discord whether someone else could try to reproduce this

> > When I run on windows it works fine aswell
> 
> Im on Linux

user622628252416:
@Snowiiii could you try again using the latest commit? Every suggestions the server sends should be printed to the console

Snowiiii:
LGTM.
Big Thanks @user622628252416 :heart:

PR #251, opened 2024-11-10
Check collision before block place

Tiiita:
## Description
Added bounding boxes to every Entity, where a block has a width: 1.0 and a height: 1.0. A player has a width: 0.6 and a height: 1.8.
When placing a block, pumpkin now creates a temporary bounding box from that target position and checks whether the player's box collides with it or not. This way player are not able to place blocks in themself.

## Testing
Tried placing blocks under myself and beside myself.

## Checklist
- Check collision with bounding box of air block and other entities standing in that block, not only with player's own [ ]

### DIFF
--- a/pumpkin-core/src/math/boundingbox.rs
+++ b/pumpkin-core/src/math/boundingbox.rs
@@ -1,5 +1,7 @@
 use super::{position::WorldPosition, vector3::Vector3};
 
+
+#[derive(Clone, Copy)]
 pub struct BoundingBox {
     pub min_x: f64,
     pub min_y: f64,
--- a/pumpkin-core/src/math/boundingbox.rs
+++ b/pumpkin-core/src/math/boundingbox.rs
@@ -8,30 +8,26 @@ pub struct BoundingBox {
     pub max_x: f64,
     pub max_y: f64,
     pub max_z: f64,
-    pub width: f64,
-    pub height: f64,
 }
 
 impl BoundingBox {
-    pub fn new_default(width: f64, height: f64) -> Self {
-        Self::new_from_pos(0., 0., 0., width, height)
+    pub fn new_default(size: &BoundingBoxSize) -> Self {
+        Self::new_from_pos(0., 0., 0., size)
     }
 
-    pub fn new_from_pos(x: f64, y: f64, z: f64, width: f64, height: f64) -> Self {
-        let f = width / 2.;
+    pub fn new_from_pos(x: f64, y: f64, z: f64, size: &BoundingBoxSize) -> Self {
+        let f = size.width / 2.;
         Self {
             min_x: x - f,
             min_y: y,
             min_z: z - f,
             max_x: x + f,
-            max_y: y + height,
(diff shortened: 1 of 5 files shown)


pumpkin-core/src/math/boundingbox.rs
```diff
@@ -1,23 +1,47 @@
...
+#[derive(Clone, Copy)]
 pub struct BoundingBox {
     pub min_x: f64,
     pub min_y: f64,
     pub min_z: f64,
     pub max_x: f64,
     pub max_y: f64,
     pub max_z: f64,
+    pub width: f64,
+    pub height: f64,
```
Snowiiii, line 13: Please don't store this here

pumpkin/src/client/player_packet.rs
```diff
@@ -577,12 +577,19 @@ impl Player {
                         }
                     }
 
-                    world
+                    let world_pos = WorldPosition(location.0 + face.to_offset());
+                    let block_bounding_box = BoundingBox::from_block(&world_pos);
+                    let bounding_box = self.living_entity.entity.bounding_box.load();
+
+                    //TODO: Make this check for every entity in that posistion
+                    if !BoundingBox::bounding_boxes_intersect(&block_bounding_box, &bounding_box) {
```
Snowiiii, line 585: `block_bounding_box.bounding_boxes_intersect(bounding_box)` would be better

pumpkin-core/src/math/boundingbox.rs
Tiiita: Where else?
In the entity itself?

Tiiita:
Changed the things u said. Testing rn if everything still works

Tiiita:
Seems to still work :D

pumpkin-core/src/math/boundingbox.rs
```diff
@@ -33,10 +50,25 @@ impl BoundingBox {
         }
     }
 
+    pub fn bounding_boxes_intersect(&self, other: &BoundingBox) -> bool {
```
Snowiiii, line 53: using the struct/modules name in functions is a bad design, You should always keep it simple, Also its wrong, You wrote intersect. Just name it `intersects`

Snowiiii:
You also only check if the block collision interects with the player who placed the block. We should do it (and Vanillas does it) that way, that is goes trough all Entities nearby. And performs collision checks.
This solves the Problem that you could place blocks where entities are

pumpkin-core/src/math/boundingbox.rs
Tiiita: oh bruh i actually know, dont know why i did that, lemme fix

Tiiita:
Ik, i wrote that into the checklist. I did not figure out how to get the entities in a WorldPosition so..

Snowiiii:
Thanks @Tiiita

Alvin-HZ:
You aren't able to place non-full blocks like fences or buttons on the same block you're standing on

PR #254, opened 2024-11-11
Implemented /msg, /tell and /w commands

KaliemSB:
## Description

- Added `/msg` command.
- Added `/tell` command.
- Added `/w` command.

All commands were implemented according to the [wiki](https://minecraft.fandom.com/wiki/Commands/msg).

## Testing
- [x] Msg command on self, player, players.
- [x] This pull request has been tested and compared against both Vanilla and Paper versions.

## Checklist

### DIFF
--- a/pumpkin/src/command/commands/cmd_msg.rs
+++ b/pumpkin/src/command/commands/cmd_msg.rs
@@ -1,0 +1,71 @@
+use async_trait::async_trait;
+use pumpkin_core::text::{color::NamedColor, TextComponent};
+
+use crate::command::{
+    args::{arg_message::MsgArgConsumer, arg_players::PlayersArgumentConsumer, Arg, ConsumedArgs},
+    tree::CommandTree,
+    tree_builder::{argument, require},
+    CommandExecutor, CommandSender, InvalidTreeError,
+};
+use InvalidTreeError::InvalidConsumptionError;
+
+const NAMES: [&str; 3] = ["msg", "tell", "w"];
+
+const DESCRIPTION: &str = "Sends a private message to one or more players.";
+
+const ARG_TARGET: &str = "target";
+const ARG_MESSAGE: &str = "message";
+
+struct MsgExecutor;
+
+#[async_trait]
+impl CommandExecutor for MsgExecutor {
+    async fn execute<'a>(
+        &self,
+        sender: &mut CommandSender<'a>,
+        _server: &crate::server::Server,
+        args: &ConsumedArgs<'a>,
+    ) -> Result<(), InvalidTreeError> {
+        let sender_name = match sender {
+            CommandSender::Console => "Console",
+            CommandSender::Rcon(_) => "Rcon",
+            CommandSender::Player(player) => &player.gameprofile.name.clone(),
+        };
+
+        let Some(Arg::Players(targets)) = args.get(&ARG_TARGET) else {
+            return Err(InvalidConsumptionError(Some(ARG_TARGET.into())));
+        };
+
+        let Some(Arg::Msg(msg)) = args.get(ARG_MESSAGE) else {
+            return Err(InvalidConsumptionError(Some(ARG_MESSAGE.into())));
+        };
+
+        for target_player in targets {
+            let recipient_message = format!("{sender_name} whispers to you: {msg}");
+            let sender_message =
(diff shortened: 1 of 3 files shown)


pumpkin/src/command/commands/cmd_msg.rs
```diff
@@ -0,0 +1,69 @@
...
+            CommandSender::Player(player) => &player.gameprofile.name.clone(),
+        };
+
+        let Some(Arg::Players(targets)) = args.get(&ARG_TARGET) else {
+            return Err(InvalidConsumptionError(Some(ARG_TARGET.into())));
+        };
+
+        let Some(Arg::Msg(msg)) = args.get(ARG_MESSAGE) else {
+            return Err(InvalidConsumptionError(Some(ARG_MESSAGE.into())));
+        };
```
user622628252416, line 41: You should be able to just use something like ```let msg = MsgArgConsumer.find_arg(args, ARG_MESSAGE)?;```. Im pretty sure FindArg is implemented for all argument consumers, so same goes for PlayerArgConsumer

pumpkin/src/command/commands/cmd_msg.rs
```diff
@@ -0,0 +1,69 @@
...
+            let sender_message =
+                format!("you whisper to {}: {msg}", target_player.gameprofile.name);
+
+            let recipient_text = TextComponent::text(&recipient_message)
+                .color_named(NamedColor::Gray)
+                .italic();
+
+            let sender_text = TextComponent::text(&sender_message)
+                .color_named(NamedColor::Gray)
+                .italic();
```
user622628252416, line 54: You could lift string formatting out of the loop so it isn't done again for every single target

The sender should receive only one "you whisper to [...]" message even if there are multiple targets, so you should move that out of the loop too

user622628252416 (changes requested):
Looks good IMO, I just noticed some minor things

pumpkin/src/command/mod.rs
```diff
@@ -84,7 +84,7 @@ pub fn default_dispatcher<'a>() -> Arc<CommandDispatcher<'a>> {
     dispatcher.register(cmd_teleport::init_command_tree());
     dispatcher.register(cmd_give::init_command_tree());
     dispatcher.register(cmd_list::init_command_tree());
-    dispatcher.register(cmd_clear::init_command_tree());
```
user622628252416, line 87: I'm not sure whether this just happens because the clear command was merged after you created this pr, but this line shouldn't be removed

user622628252416 (changes requested):

pumpkin/src/command/mod.rs
KaliemSB: Fixed ✅

pumpkin/src/command/commands/cmd_msg.rs
KaliemSB: Thanks, i'm using the `find_arg` now 🙏

pumpkin/src/command/commands/cmd_msg.rs
KaliemSB: Thanks, I hadn't even thought about that, everything changed ✅, I don't know if I did it in the best way, I'm still new to Rust 💪

Snowiiii:
You should use Chat Types. They also include a message (e.g. .... whispers to ...).
There are Chat Types parsed from the registry. I think it would be cool to automaticlly/manuelly make an Enum for Chat  Types

user622628252416 (approved):

KaliemSB:
> You should use Chat Types. They also include a message (e.g. .... whispers to ...). There are Chat Types parsed from the registry. I think it would be cool to automaticlly/manuelly make an Enum for Chat Types

I researched and read about these Chat Types, and it seems to be the correct way to send these messages. However, I have no idea how to use or implement it ☠️☠️☠️

KaliemSB:
Waiting this PR to be merged #324

lucas11222:
fix conflicts and make for the chat types

Snowiiii:
> Waiting this PR to be merged #324

Merged

Snowiiii:
Hey @KaliemSB, I'm sorry but i'm closing this for now. Feel free to reopen when your code uses Chat Types

PR #255, opened 2024-11-12
fix(Docker): improve build and compose files

Yimura:
## Description

While trying to deploy Pumpkin through Docker I noticed that it's impossible to do so without having the pre-generated `configuration.toml` and `features.toml` files.
The current docker configuration mounts these files into the docker container, which is possible and as long as person that's deploying keeps this in mind they can prepare for this by doing the following:
```bash
touch configuration.toml
touch features.toml
```
This process creates the two files required to bind into the container, however the problem that I found out later on was that Pumpkin does not populate these files when they're empty. Meaning that someone who doesn't look into the source code and the lack of documentation is unable to modify any of the settings without relying on someone else to generate these files for them.

- Adapted `Dockerfile` so the PWD will point to the `/config` folder in the image
- Modified the `docker-compose.yml` to follow the Dockerfile logic
- Added a LABEL to the docker image so it points to the Snowiiii/Pumpkin repository

This docker deployment doesn't need to be a fully working production environment, however it should at the least be a correct one for people to base themselves of off.

## Testing

- **OS:** Arch Linux
- **Docker Version:** Docker version 27.3.1, build ce1223035a
- **Docker Compose Version:** Docker Compose version 2.30.2

Build command used to test this deployment:
```bash
docker compose up --build
```

docker example deployment

## Checklist

No Rust code was modified in this PR, left the check boxes just in-case.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,31 @@
 FROM rust:1-alpine3.20 AS builder
 ENV RUSTFLAGS="-C target-feature=-crt-static -C target-cpu=native"
 RUN apk add --no-cache musl-dev
+
 WORKDIR /pumpkin
 COPY . /pumpkin
+
+# build release
 RUN --mount=type=cache,sharing=private,target=/pumpkin/target \
     --mount=type=cache,target=/usr/local/cargo/git/db \
     --mount=type=cache,target=/usr/local/cargo/registry/ \
     cargo build --release && cp target/release/pumpkin ./pumpkin.release
+
+# strip debug symbols from binary
 RUN strip pumpkin.release
 
 FROM alpine:3.20
-WORKDIR /pumpkin
+
+# Identifying information for registries like ghcr.io
+LABEL org.opencontainers.image.source=https://github.com/Snowiiii/Pumpkin
+
 RUN apk add --no-cache libgcc
+
 COPY --from=builder /pumpkin/pumpkin.release /pumpkin/pumpkin
+
+# set workdir to /config 
+WORKDIR /config
+
 ENV RUST_BACKTRACE=1
 EXPOSE 25565
-ENTRYPOINT ["/pumpkin/pumpkin"]
+ENTRYPOINT [ "/pumpkin/pumpkin" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,9 @@ RUN apk add --no-cache libgcc
 
 COPY --from=builder /pumpkin/pumpkin.release /pumpkin/pumpkin
 
-# set workdir to /config 
+# set workdir to /config, this is required to influence the PWD environment variable
+# it allows for bind mounting the server files without overwriting the pumpkin
+# executable (without requiring an `docker cp`-ing the binary to the host folder)
 WORKDIR /config
 
 ENV RUST_BACKTRACE=1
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,18 +1,31 @@
...
+
+# Identifying information for registries like ghcr.io
+LABEL org.opencontainers.image.source=https://github.com/Snowiiii/Pumpkin
+
 RUN apk add --no-cache libgcc
+
 COPY --from=builder /pumpkin/pumpkin.release /pumpkin/pumpkin
+
+# set workdir to /config 
+WORKDIR /config
```
Bryntet, line 27: I do not like this comment. The code is self explanatory.

Yimura: Sure I can agree on that one.

To be honest this `/config`, is still up for debate as it doesn't quite fit the logic for future dynamic content (world, logs, datapacks, etc...).

I'm open for discussion to get a more appropriate name (ex. `/app` or `/server`).

Bryntet: I'm not too familiar with docker, but I think if I've understood the issue correctly is that for some reason when launching docker, it doesn't automatically generate the config files like it does when you run the binary normally? I think `/config` sounds appropriate due to the files being configuration files

Yimura: > I'm not too familiar with docker, but I think if I've understood the issue correctly is that for some reason when launching docker, it doesn't automatically generate the config files like it does when you run the binary normally? I think `/config` sounds appropriate due to the files being configuration files

I'd like to add those are the files the server creates implicitly, these two files are made in the root. In the future the world folder, datapack folders, logs folders, ... will live next to those two files. Making the `/config` folder a less ideal option IMO. As such I'm trying to thing a bit further into the future and generalize the root folder name.

Yimura: Comment should be resolved, IMO. Open for further discussion.

Yimura:
Works as intended:
image

Server Binary: `/bin/pumpkin`
Root Path: `/pumpkin`

Dockerfile
```diff
@@ -1,18 +1,33 @@
...
+LABEL org.opencontainers.image.source=https://github.com/Snowiiii/Pumpkin
+
 RUN apk add --no-cache libgcc
-COPY --from=builder /pumpkin/pumpkin.release /pumpkin/pumpkin
+
+COPY --from=builder /pumpkin/pumpkin.release /bin/pumpkin
+
+# set workdir to /pumpkin, this is required to influence the PWD environment variable
+# it allows for bind mounting the server files without overwriting the pumpkin
+# executable (without requiring an `docker cp`-ing the binary to the host folder)
```
Yimura, line 28: This comment was written prior to moving the pumpkin executable to `/bin`, if required I can move it.

Bryntet:
I think this PR looks good to me, LGTM

Snowiiii:
> Why is it using Alpine though, why not Arch Linux?

Im not familiar with docker or what the best "distro" for docker is. But when searching Online is looks like Alpine is considered to be smaller in Size than Arch

Snowiiii:
LGTM.
Thanks @Yimura

Yimura:
We don't need all the bloat that comes with Arch. All pumpkin needs is the bare minimum to operate within the container.

PR #262, opened 2024-11-13
Implemented Sign Editing Functionalities

whiteout314:
## Description
Currently signs cannot be edited. Added the following:
* Clientbound packet to open the screen editor when client places a sign block or interacts with a sign block. 
* Added serverbound packet that is sent after clicking done on the open screen editor window.
* Added an interactive folder in blocks to hold interactive blocks such as signs, chests, etc...
* Implemented a file that holds sign structure for NBT data and enums associated with valid sign blocks.
* Added get_block_by_state_id function.
* Fixed structure world's get_block function to search for state ids.

## Testing
* Tested using Minecraft version 1.21.3 placing signs and tested placing sign and able to see the sign editor and tested as well that an already placed sign can modified (front and back).

## Checklist


pumpkin-world/src/block/interactive/sign.rs
```diff
@@ -0,0 +1,146 @@
...
+    CrimsonWallHangingSign = 238,
+    WarpedWallHangingSign = 239,
+    MangroveWallHangingSign = 240,
+    BambooWallHangingSign = 241,
+
+    CrimsonSign = 849,
+    WarpedSign = 850,
+    CrimsonWallSign = 851,
+    WarpedWallSign = 852,
+}
```
Snowiiii, line 59: This looks horrible, Also the the ID's will change with every Update

pumpkin/src/entity/player.rs
```diff
@@ -871,6 +873,9 @@ impl Player {
             SSwingArm::PACKET_ID => {
                 self.handle_swing_arm(SSwingArm::read(bytebuf)?).await;
             }
+            SUpdateSign::PACKET_ID => {
+                //TODO
```
Snowiiii, line 877: So this is empty, So Im predicting that Signs currently do not work right ?.
I recommend to make this a draft PR then when its still WIP

pumpkin-world/src/block/interactive/sign.rs
whiteout314: Now looking at it again, it is. I will make sure to fix this using the block.json.

pumpkin/src/entity/player.rs
whiteout314: In the game the sign updates with the text. I believe that this packet is suppose to update the server's nbt file--from what I have been reading. Couldn't test to see if another client sees the modified sign since I don't have another account.

whiteout314: Figured it out, my bad. My latest commit implements the function that changes the sign for all clients by sending a client bound Block Entity Data packet.

pumpkin-world/src/block/interactive/sign.rs
whiteout314: Added a macro that looks up the block entity id instead of the actual id and eliminated the mess.

whiteout314:
Here is a demo:
https://github.com/user-attachments/assets/a7e0ac08-88f2-404e-a31a-01aa283bcd6d

whiteout314:
demo

pumpkin-macros/src/block_entity.rs
```diff
@@ -0,0 +1,35 @@
...
+    name: String,
+}
+static BLOCK_ENTITIES: LazyLock<HashMap<String, u32>> = LazyLock::new(|| {
+    serde_json::from_str::<TopLevel>(include_str!("../../assets/blocks.json"))
+        .expect("Could not parse blocks.json registry.")
+        .block_entity_types
+        .into_iter()
+        .map(|value| (value.ident, value.id))
+        .collect()
+});
```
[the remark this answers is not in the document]
Snowiiii: I also would love to have the `BlockState::new(<name>).unwrap()` replaced with a macro. We had a macro for this before i moved the JSON's to our own Extractor. Feel free to add this to this PR, But you not forced

pumpkin-protocol/src/server/play/s_update_sign.rs
```diff
@@ -0,0 +1,14 @@
...
+
+#[derive(Deserialize)]
+#[server_packet("play:sign_update")]
+pub struct SUpdateSign {
+    pub location: WorldPosition,
+    pub is_front_text: bool,
+    pub line_1: String,
+    pub line_2: String,
+    pub line_3: String,
+    pub line_4: String,
```
Snowiiii, line 13: Every line has the max length of 384. We currently have no way of automatically Deserializing the length. You would need to have to implement the Packet manually and use `buf.read_string_len(384)*

DataM0del: If it's `384` characters per line, then why not just
```suggestion
    pub line_1: [char; 384],
    pub line_2: [char; 384],
    pub line_3: [char; 384],
    pub line_4: [char; 384],
```
?

DataM0del: I forgot that the client is probably going to try and send something more than 384 characters and crash the server... :skull:

Snowiiii: > If it's 384 characters per line, then why not
> 
> ?

That's the wrong approche, It will be also a pain to actually get the lines.

DataM0del: > > If it's 384 characters per line, then why not
> > ?
> 
> That's the wrong approche, It will be also a pain to actually get the lines.

You do save some memory by storing it as a slice though. Also, what do you mean "get the lines"? If you're using it like it's a `String` then you just use `String::from_slice` or `line_n.into()`. Although I do think that we probably shouldn't trust that the client is going to send 384 characters or below for a line.

pumpkin-protocol/src/server/play/s_update_sign.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl ServerPacket for SUpdateSign {
+    fn read(bytebuf: &mut ByteBuffer) -> Result<Self, DeserializerError> {
+        let location = WorldPosition(Vector3::new(
+            bytebuf.get_i32()?,
+            bytebuf.get_i32()?,
+            bytebuf.get_i32()?,
```
leobeg, line 26: This won't work. Look at https://wiki.vg/Protocol#Position. My solution:

pumpkin-core/src/math/position impl WorldPosition add
```
pub fn from_i64(encoded_position: i64) -> Self {
        WorldPosition(Vector3 {
            x: (encoded_position >> 38) as i32,
            y: (encoded_position << 52 >> 52) as i32,
            z: (encoded_position << 26 >> 38) as i32,
        })
    }
```
And in your SUpdateSign you just do: 

`let location = WorldPosition::from_i64(bytebuf.get_i64()?);`

leobeg:
There is also an error when the block gets updatet trough the CAcknowledgeBlockChange packet. Then the popup instantly goes away

DataM0del:
oof

whiteout314:
Sorry about that looks like I had an issue last night updating and pushing to my branch

pumpkin-protocol/src/server/play/s_update_sign.rs
whiteout314: Added this, thanks for the solution!

pumpkin-macros/src/block.rs
```diff
@@ -0,0 +1,17 @@
+use proc_macro::TokenStream;
+use quote::quote;
+
+pub(crate) fn block_entity_impl(item: TokenStream) -> TokenStream {
+    let input_string = item.to_string();
+    let block_entity_name = input_string.trim_matches('"');
+
+    quote! {
+        pumpkin_world::block::block_registry::BLOCKS
```
whiteout314, line 9: @Snowiiii I was wondering if this is fine and in another PR, I could do all the changes to bring the methods from block_registry into this macro

pumpkin/src/client/player_packet.rs
```diff
@@ -601,10 +626,34 @@ impl Player {
...
+                            .send_packet(&COpenSignEditor::new(world_pos, face.to_offset().z == 1))
+                            .await;
+                    }
+                }
+            } else {
+                let block = world.get_block(location).await;
+                if let Some(block) = block {
+                    if block.states.iter().any(|state| {
+                        state.block_entity_type == Some(block_entity!("minecraft:sign"))
+                            || state.block_entity_type == Some(block_entity!("minecraft:sign"))
```
whiteout314, line 649: I need to change this to hanging_sign.

leobeg:
Will this be merged soon? It would be a base for following block entities.

OfficialKris:
Block entities are also required for many things. I wanted to add jukebox support and even that requires it (storing disc). Will this be merged soon? I tested and it works for me.

leobeg:
I think it would be best to fit this PR to the new Block system in #333

leobeg:
@whiteout314 Do you want to do this or should I implement this?

whiteout314:
> @whiteout314 Do you want to do this or should I implement this?

@leobeg Go ahead and implement it since you are already working on blocks that can be interacted with and plus is up to date.

Snowiiii:
@whiteout314 https://github.com/Snowiiii/Pumpkin/pull/333 just got merged

PR #271, opened 2024-11-14
fix(combat): kick players attacking dead players (health <= 0.0)

DataM0del:
This doesn't need a description... the 4 line (5 if you count the extra one that the formatter added) diff explains it all.

## Description

```rs
if victim.living_entity.health.load() <= 0.0 {
    // you can trigger this from a non-modded / innocent client client,
    // so we shouldn't kick the player
    return;
}
```

## Testing

Hit a dead player (e.g. another player, for testing this, you would usually just open another client and hit that client's player) and see if any damage occurs (it doesn't).

## Checklist

### DIFF
--- a/pumpkin/src/client/player_packet.rs
+++ b/pumpkin/src/client/player_packet.rs
@@ -454,7 +454,11 @@ impl Player {
                         .await;
                     return;
                 };
-
+                if victim.living_entity.health.load() <= 0.0 {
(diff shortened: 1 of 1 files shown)


Snowiiii:
I was actually wondering if this would happen. So it makes sense to check if the entity is alive. But for the future please still use better descriptions.
Thanks @DataM0del

PR #279, opened 2024-11-14
Logging system info

Commandcracker:
## Description
Add current git commit hash to version info and print system info at startup for better bug traceability.
Removed ms in log timestamp, because it is unnecessary.

## Testing
Ran inside docker and on windows11

## Checklist


Snowiiii:
LGTM.
Thanks @Commandcracker :+1:

PR #280, opened 2024-11-14
Textcomponent Fix and commands

Commandcracker:
## Description

- Fix Style click_event and hover_event
- Add RGB support to Textcomponent
- Add me and seed command

## Checklist


pumpkin-core/src/text/color.rs
```diff
@@ -59,7 +95,8 @@ impl Color {
                 NamedColor::Yellow => text.bright_yellow(),
                 NamedColor::White => text.white(),
             },
-            Color::Rgb(_) => todo!(),
```
Snowiiii, line 62: I will not agree to making the colors not work in the terminal, This would also then just panic

Snowiiii:
Please split the RGB Support and the new Command into 3 different PRs. I prefer to have a new PR for every new Command which is added

Commandcracker:
Will do that

PR #287, opened 2024-11-15
Add RGB support to TextComponent, Fix some style

Commandcracker:
## Description
Add RGB support to TextComponent
Fix click_event, hover_event, hover ShowText

## Testing
used TextComponent with click_event(coppy) and hover_event(show) in console and client

## Checklist
Things need to be done before this Pull Request can be merged.

- [x] Code is well-formatted and adheres to project style guidelines: `cargo fmt`
- [x] Code does not produce any clippy warnings `cargo clippy`

PR #289, opened 2024-11-15
Add seed Command

Commandcracker:
## Description
Requires #287 

## Checklist


pumpkin/src/command/commands/cmd_seed.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[async_trait]
+impl CommandExecutor for PumpkinExecutor {
+    async fn execute<'a>(
+        &self,
+        sender: &mut CommandSender<'a>,
+        _server: &crate::server::Server,
+        _args: &ConsumedArgs<'a>,
+    ) -> Result<(), CommandError> {
+        let seed = &BASIC_CONFIG.seed;
```
Snowiiii, line 27: That's only the seed from the Config, Which can be empty. If the seed is empty will will generate a random one later and store it in the `world`.

Commandcracker: It's not in the World, it's in GeneratorInit which I can't really access.
IG ill make this into a draft until world gen is ready and has a seed.

Snowiiii: You should now be able to get the seed in the level. See https://github.com/Snowiiii/Pumpkin/commit/b5a8fee243c6a4e57238bf863b57f716453e3c12

Snowiiii:
Looks good now.
Thanks @Commandcracker

PR #294, opened 2024-11-16
Fix packet side in extractor

Alvsch:
## Description
Previously in `packets.json`, `clientbound` packets where called `serverbound`, vice versa.
For example the `Login (Play)` packet was in `serverbound` when it's actually sent by the server to the client

## Testing
Joining the server works as before

## Checklist


Snowiiii:
Looks good now.
Thanks @Alvsch :D

Issue #286, opened 2024-11-15
Falling through world


Commandcracker:
### I've searched existing issues and couldn't find a duplicate.

- [X] I confirm this is not a duplicate.

### Operating System

Windows 11

### Server Software Version/Commit

29d203c8a9899ed9e9e6b147f79d67bc7feca504

### What happened?

I'm **consistently** falling through the world when I join and I when I respawn I'm still 1 inside a block.
Must be some coalition calculation issue.

Maybe also threading because I have 32 core.

https://github.com/user-attachments/assets/2e592645-73c0-4523-9c07-b330be9c0ee1

### To Reproduce

IDK

### Expected behavior

Not falling through the World :D

Commandcracker:
Only happens in debug build tho 🤔

DataM0del:
Also, sometimes it spawns you 2 or 1 block(s) inside the ground.

DataM0del:
> An in release, it sometimes works as intended 🤔

I think it's because the chunk generation or something is faster and makes it so you don't fall because the client hasn't gotten a chunk yet and is assuming all air.

PR #295, opened 2024-11-16
fix respawn position being inside top block

user622628252416:
## Description
Add one to the y coordinate of the respawn position so players spawn on top of the top block rather than iside of it.

## Testing
- use /kill
- observe not being inside of a block

## Checklist


user622628252416:
This should fix the respawning part of #286, but not the falling through the world part

Snowiiii:
Looks good.
Thanks @user622628252416 :D

PR #298, opened 2024-11-16
Fix block replace

user622628252416:
## Description
this fixes two related bugs:
- irreplaceable blocks are no longer replaced when players place blocks inside of them (eg. by clicking the top face of a block below a chest)
- replaceable blocks are now replaced when players try to place blocks on them (eg. by clicking grass)

## Testing
- start server
- try the examples mentioned above
- observe correct behavior

## Checklist

### DIFF
--- a/pumpkin/src/client/player_packet.rs
+++ b/pumpkin/src/client/player_packet.rs
@@ -609,12 +609,21 @@ impl Player {
                         }
                     }
 
-                    let world_pos = WorldPosition(location.0 + face.to_offset());
+                    let clicked_world_pos = WorldPosition(location.0);
+                    let clicked_block_state = world.get_block_state(clicked_world_pos).await?;
 
-                    let previous_block_state = world.get_block_state(world_pos).await?;
-                    if !previous_block_state.replaceable {
-                        return Ok(());
-                    }
+                    let world_pos = if clicked_block_state.replaceable {
+                        clicked_world_pos
+                    } else {
+                        let world_pos = WorldPosition(location.0 + face.to_offset());
+                        let previous_block_state = world.get_block_state(world_pos).await?;
+
+                        if !previous_block_state.replaceable {
+                            return Ok(());
+                        }
+
+                        world_pos
+                    };
 
                     let block_bounding_box = BoundingBox::from_block(&world_pos);
                     let bounding_box = entity.bounding_box.load();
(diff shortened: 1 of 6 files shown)


pumpkin/src/client/player_packet.rs
```diff
@@ -32,13 +33,26 @@ use pumpkin_protocol::{
...
+use thiserror::Error;
 
 use super::PlayerConfig;
 
 fn modulus(a: f32, b: f32) -> f32 {
     ((a % b) + b) % b
 }
 
+#[derive(Debug, Error)]
+pub enum PlayerError {
```
Snowiiii, line 45: This is a horrible name. This can be literally everything. I think something like "BlockPlaceError" would be much better suiteted

pumpkin/src/world/mod.rs
```diff
@@ -53,6 +54,18 @@ type ChunkReceiver = (
     Receiver<Arc<RwLock<ChunkData>>>,
 );
 
+#[derive(Debug, Error)]
+pub enum WorldError {
```
Snowiiii, line 58: Also horrible name

pumpkin/src/error.rs
```diff
@@ -65,3 +67,39 @@ impl PumpkinError for DeserializerError {
...
+        }
+    }
+
+    fn client_kick_reason(&self) -> Option<String> {
+        match self {
+            Self::BlockOutOfReach => None,
+            Self::InvalidBlockFace => Some("Invalid block face".into()),
+        }
+    }
+}
```
Snowiiii, line 105: No. Don't implement the PumpkinError trait here. Do it below the structs

user622628252416: InventoryError's impl was here, should I move that too?

user622628252416: And DeserializerError as well, all PumpkinError impls were here actually

pumpkin/src/world/mod.rs
user622628252416: These errors were meant to be used for other Errors as well so we woudln't end up with infinitely many different error enums per file, but I guess they could be a bit more fine-grained

Snowiiii:
Just tested and very happy the grass can now be easily and other blocks can be replaced. 
Good job @user622628252416 :+1:

PR #299, opened 2024-11-16
Add lan broadcast

Commandcracker:
## Description
See net.minecraft.client.network.LanServerPinger if you have questions

## Testing
Screenshot 2024-11-17 001256

## Checklist


pumpkin/src/lan_broadcast.rs
```diff
@@ -0,0 +1,39 @@
...
+use tokio::net::UdpSocket;
+use tokio::time;
+
+const BROADCAST_ADDRESS: SocketAddr =
+    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(224, 0, 2, 60)), 4445);
+
+pub async fn start_lan_broadcast(bound_addr: SocketAddr) {
+    let port = ADVANCED_CONFIG.lan_broadcast.port.unwrap_or(0);
+
+    let socket = UdpSocket::bind(SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), port))
```
neeleshpoli, line 13:
```suggestion
    let socket = UdpSocket::bind(format!("0.0.0.0:{port}"))
```
This looks cleaner and easier to read.

pumpkin-config/src/lan_broadcast.rs
```diff
@@ -0,0 +1,15 @@
...
+
+#[serde_inline_default]
+#[derive(Deserialize, Serialize, Default)]
+pub struct LANBroadcastConfig {
+    #[serde_inline_default(false)]
+    pub enabled: bool,
+    // We use an extra motd because this only supports one line
+    #[serde_inline_default("A Blazing fast Pumpkin Server!".to_string())]
+    pub motd: String,
+    // No idea why you would want to change this
```
neeleshpoli, line 12:
```suggestion
    // Allow users to specify port so the port is predictable
    // There are many reasons why the port might need to be predictable
    // One reason is docker containers, where specific ports need to be allowed
```
I would update this comment so that people looking at this in the future don't remove it, as the need to specify the port is not very obvious.

pumpkin/src/lan_broadcast.rs
```diff
@@ -0,0 +1,39 @@
...
+        &ADVANCED_CONFIG.lan_broadcast.motd,
+        bound_addr.port()
+    );
+
+    log::info!(
+        "LAN broadcast running on {}",
+        socket
+            .local_addr()
+            .expect("Unable to find running address!")
+    );
```
neeleshpoli, line 31:
```suggestion
```
I don't think printing this info is necessary at all.

neeleshpoli (commented):
I happened to be working on LAN broadcast when you made this PR lol. I have a few suggestions; no pressure and I won't be offended if you choose to not take any of my suggestions 👍.

pumpkin/src/lan_broadcast.rs
```diff
@@ -0,0 +1,39 @@
+use pumpkin_config::ADVANCED_CONFIG;
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::time::Duration;
+use tokio::net::UdpSocket;
+use tokio::time;
+
+const BROADCAST_ADDRESS: SocketAddr =
+    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(224, 0, 2, 60)), 4445);
```
DataM0del, line 8:
```suggestion
// https://www.wikiwand.com/en/articles/Multicast_address

const BROADCAST_ADDRESS: SocketAddr =
    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(224, 0, 2, 60)), 4445);
```

Why not?

pumpkin/src/lan_broadcast.rs
Commandcracker: The query does the same
https://github.com/Snowiiii/Pumpkin/blob/00af2b98683b11946cddb4f08de3b0ce04c1fddd/pumpkin/src/query.rs#L43-L48
I think this should be kept because the default port is set to 0 for LAN broadcast, this means the system will give you some open port and printing this assigned port sounds like a good Ida to me.

pumpkin-config/src/lan_broadcast.rs
```diff
@@ -0,0 +1,15 @@
...
+use serde_inline_default::serde_inline_default;
+
+#[serde_inline_default]
+#[derive(Deserialize, Serialize, Default)]
+pub struct LANBroadcastConfig {
+    #[serde_inline_default(false)]
+    pub enabled: bool,
+    // We use an extra motd because this only supports one line
+    #[serde_inline_default("A Blazing fast Pumpkin Server!".to_string())]
+    pub motd: String,
```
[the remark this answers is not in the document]
Commandcracker: Sounds good!

pumpkin/src/lan_broadcast.rs
Commandcracker: Isn't SocketAddr::new more performant ?

pumpkin/src/lan_broadcast.rs
neeleshpoli: But the thing is query requires clients to send packets to that specific port and address, but LAN broadcast doesn't require clients to send packets to whatever port and address they are bound to at all. Thats why I feel like printing this is unnecessary.

pumpkin/src/lan_broadcast.rs
neeleshpoli: No because the compiler should optimize this and make it performant (it should compile to the same assembly at the end).

pumpkin/src/lan_broadcast.rs
Commandcracker: It's still an open port, IDK. I would rather keep it and remove it when it is decided that It's too much log clutter.

pumpkin/src/lan_broadcast.rs
```diff
@@ -0,0 +1,52 @@
...
+    let motd: String;
+    let advanced_motd = &ADVANCED_CONFIG
+        .lan_broadcast
+        .motd
+        .clone()
+        .unwrap_or_default();
+
+    if advanced_motd.is_empty() {
+        motd = BASIC_CONFIG.motd.replace('\n', " ");
+        log::warn!("Using the server MOTD as the LAN broadcast MOTD. Note that the LAN broadcast MOTD does not support multiple lines, so consider defining it accordingly.");
```
neeleshpoli, line 32:
```suggestion
        log::warn!("Using the server MOTD as the LAN broadcast MOTD. Note that the LAN broadcast MOTD does not support multiple lines, RGB colors, or gradients so consider defining it accordingly.");
```

neeleshpoli (commented):
Looks all good to me! Just this small change in the warning and it should be good to merge.

neeleshpoli (approved):
Looks all good to me!

DataM0del (approved):

Snowiiii:
Just tested and everything works. Code looks good as well.
Thank you @Commandcracker :+1:

PR #303, opened 2024-11-17
Cleanup pumpkin-config

Commandcracker:
## Description
Most changes are because serde_inline_default is no longer needed, see [here](https://github.com/rust-lang/rfcs/pull/3681)

## Checklist

### DIFF
--- a/pumpkin-config/Cargo.toml
+++ b/pumpkin-config/Cargo.toml
@@ -9,4 +9,3 @@ serde.workspace = true
 log.workspace = true
 
 toml = "0.8"
-serde-inline-default = "0.2.2"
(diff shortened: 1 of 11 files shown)


Snowiiii:
The only reason we use `serde_inline_default` is to have default values for config fields we can't parse. I tested your changes and removed a random config value. It gave me an error `Couldn't parse config at "configuration.toml". Reason: missing field `hardcore`. This is is proberbly caused by an Config update, Just delete the old Config and start Pumpkin again`

Commandcracker:
> The only reason we use `serde_inline_default` is to have default values for config fields we can't parse. I tested your changes and removed a random config value. It gave me an error `Couldn't parse config at "configuration.toml". Reason: missing field `hardcore`. This is is proberbly caused by an Config update, Just delete the old Config and start Pumpkin again`

Now it should work

Snowiiii:
Just tested again. Works fine now and looks much better.
Thanks @Commandcracker :D

PR #305, opened 2024-11-17
Add option to disable commands

Commandcracker:
## Checklist


Snowiiii:
You usually don't create field named "disabled". If you look around the codebase we use "enabled" and its also more commonly used

Snowiiii:
I think it would be better to make this an string array. then loop trough the command registry and check if it matches

Snowiiii:
Lets discuss this in an issue or on our discord. Because your solution with a bunch of booleans is really not pretty

PR #306, opened 2024-11-17
Implemented the Bossbar

leobeg:
## Description
This PR adds the Bossbar. It implements the following:

-  /bossbar command
- Implementing all bossbar features
- Keeping track of active bossbars

It does this by sending the CBossEvent packet.
This PR does not try to add the bossbar spawning when starting a fight with the enderdragon/wither

## Testing
Start the server an join. Add a bossbar with `/bossbar add namespace:id Title`
Add the bossbar to the player. `/bossbar set namespace:id players @a`
You should now see the bossbar :D.
You can now test every other feature of the bossbar.

## Checklist

- [x] Code does not produce any clippy warnings `cargo clippy` (Kinda... transfer command produces error?!)

## ToDo

- [x] Adding bossbar
- [x] Removing bossbar
- [x] Updating bossbar
- [x] Command handling
- [x] UUID storing of active bossbars


leobeg:
@Snowiiii What is the correct way of storing such information like UUIDs of the bossbars and which players can see them?

pumpkin/src/world/bossbar.rs
```diff
@@ -0,0 +1,149 @@
...
+    custom_bossbars: HashMap<String, CustomBossbar>,
+}
+
+impl CustomBossbars {
+    pub fn new() -> CustomBossbars {
+
+        let mut example_data: HashMap<String, CustomBossbar> = HashMap::new();
+        let mut players: Vec<Uuid> = Vec::new();
+        players.push(Uuid::from_u128(0x1DCC7E94EA424A0394D0752889039383));
+        example_data.insert("minecraft:123".to_string(), CustomBossbar {namespace: "minecraft:123".to_string(), bossbar_data: Bossbar::default(), player: Vec::new()});
```
Snowiiii, line 88: Don't forget to remove this when your done

Snowiiii:
> @Snowiiii What is the correct way of storing such information like UUIDs of the bossbars and which players can see them?

Maybe using a HashMap ?. Something like `HashMap<UUID, Bossbar>`

leobeg:
> Maybe using a HashMap ?. Something like `HashMap<UUID, Bossbar>`

Well I figured it out how the vanilla server stores it. It's in the level.dat. Afaik it's not read rn. Maybe that implementation of the level.dat is worth it's own PR. My plan rn is to just store to bossbars in memory and not performing read/write to this file. And as soon we have the data from the level file we can easily extend it.

Snowiiii:
> > Maybe using a HashMap ?. Something like `HashMap<UUID, Bossbar>`
> 
> Well I figured it out how the vanilla server stores it. It's in the level.dat. Afaik it's not read rn. Maybe that implementation of the level.dat is worth it's own PR. My plan rn is to just store to bossbars in memory and not performing read/write to this file. And as soon we have the data from the level file we can easily extend it.

Ah you mean how to permanently storing it. Don't worry we currently don't use `level.dat` and you don't have to for now. I plan to support that later

leobeg:
@Snowiiii It should be ready for review. Not gonna lie I really underestimated how big the bossbar feature is. As I'm quite new to rust I'm happy to recieve feedback. The only difference rn to the vanilla bossbar is that the title is a plain string and not a TextComponent as the cmd protocol has not implemented JSON Raw Text yet.

pumpkin-protocol/src/client/play/c_command_suggestions.rs
```diff
@@ -45,12 +45,12 @@ impl<'a> ClientPacket for CCommandSuggestions<'a> {
 
 #[derive(PartialEq, Eq, Hash, Debug)]
 pub struct CommandSuggestion<'a> {
-    pub suggestion: &'a str,
+    pub suggestion: String,
```
Snowiiii, line 48: why did you make this a String ?

leobeg: I had lifetime issues in the command suggestion because the str variable would be dropped after the function.
Could this be implemented a other way?

Snowiiii:
Just tested. Everything works really Good. I also have no real issues with the Code.
Thanks @leobeg :heart:

PR #307, opened 2024-11-17
Add Server Links

Commandcracker:
## Checklist


pumpkin/src/world/mod.rs
```diff
@@ -203,6 +204,68 @@ impl World {
...
+                links.push(Link::new(&Label::BuiltIn(LinkType::News), news))
+            }
+
+            let announcements = &ADVANCED_CONFIG.server_links.announcements;
+            if !announcements.is_empty() {
+                links.push(Link::new(&Label::BuiltIn(LinkType::Announcements), announcements))
+            }
+
+            for (key, value) in &ADVANCED_CONFIG.server_links.custom {
+                // add
```
DataM0del, line 256:
```suggestion
                todo!(format!("Setting custom links hasn't been implemented yet! K: {key} V: {value}"))
```
Makes clippy not scream at you for not using `key` and `value`.

DataM0del:
See Commandcracker#1

pumpkin/src/world/mod.rs
DataM0del: I implemented it in this PR: Commandcracker#1

DataM0del (approved):
I see why you didn't make them `Option`s now.

pumpkin/src/world/mod.rs
```diff
@@ -203,6 +204,68 @@ impl World {
         player.send_permission_lvl_update().await;
         client_cmd_suggestions::send_c_commands_packet(&player, command_dispatcher).await;
 
+        if ADVANCED_CONFIG.server_links.enabled {
```
Commandcracker, line 207: Where should this logic be placed ? I don't think this is the right place.

pumpkin/src/world/mod.rs
Commandcracker, line 256: Got some lifetime and borrowing issue here, would be nice if a rust pro could explain it to me 👉👈

Commandcracker:
```rs
for (key, value) in &ADVANCED_CONFIG.server_links.custom {
    links.push(Link::new(&Label::TextComponent(TextComponent::text(key)), value))
}
```

```
error[E0716]: temporary value dropped while borrowed                                                                                                                                                        
   --> pumpkin\src\world\mod.rs:297:39
    |
297 |                 links.push(Link::new(&Label::TextComponent(TextComponent::text(key)), value))
    |                 -----                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
    |                 |
    |                 borrow later used here
298 |             }
    |             - temporary value is freed at the end of this statement
    |
    = note: consider using a `let` binding to create a longer lived value

For more information about this error, try `rustc --explain E0716`.
error: could not compile `pumpkin` (bin "pumpkin") due to 1 previous error
```

pumpkin/src/world/mod.rs
```diff
@@ -14,6 +14,7 @@ use crate::{
     error::PumpkinError,
     server::Server,
 };
+use itertools::Itertools;
```
DataM0del, line 17:
```suggestion
```
What is this supposed to be here for?

Commandcracker: Because I tried around some stuff

Snowiiii:
There are 2 Server link Packets, One in the Config State and one in the Play State. You can keep the Play one, it will be useful later. But i think it would generally be better if we send this at the Config state

Snowiiii:
LGTM.
Thanks @Commandcracker

PR #310, opened 2024-11-17
feat: schema generator for basic & advanced configs

DataM0del:
## Description

Adds a schema generator so you can have autocomplete.
To use this in configs, run the generator script and add:
- `#:schema ./configuration.schema.json` for basic configs
- `#:schema ./features.schema.json` for advanced configs
This uses [the schema directive](https://taplo.tamasfe.dev/configuration/directives.html#the-schema-directive) in [Even Better TOML](https://open-vsx.org/extension/tamasfe/even-better-toml) (linking to the Open VSX registry instead of the normal extension registry because [it's outdated on the normal / default Visual Studio marketplace](https://github.com/tamasfe/taplo/issues/624)) / [Taplo](https://taplo.tamasfe.dev/)

## Testing

Procedure:
- Add the schema directive for each of the config files
- comment out everything below that schema directive so we can test the autocomplete
- Ctrl + Space to trigger the autocomplete.
- Look at the results.

## Checklist

- [x] Initial schemas (not auto-generated)
- [x] Test the schemas
- [x] Make a script to automatically generate these schemas
- [x] Test it
- [ ] Change the titles for Basic Configuration and Advanced Configuration to not look like it was generated


Snowiiii:
Im not sure if this is really necessary, We already have to add config values in the docs it would be really annoying to have to add them in 3 different places

DataM0del:
> Im not sure if this is really necessary, We already have to add config values in the docs it would be really annoying to have to add them in 3 different places

We could auto-generate the docs from these schemas. Obviously you don't get auto complete from the docs, so it's probably better if we do that instead.

Snowiiii:
> > Im not sure if this is really necessary, We already have to add config values in the docs it would be really annoying to have to add them in 3 different places
> 
> We could auto-generate the docs from these schemas. Obviously you don't get auto complete from the docs, so it's probably better if we do that instead.

I like to have examples and explain in-depth how things work. Which can be done in the docs

Snowiiii:
I meant larger and more fancy examples using markdown. Im sorry but i don't really see the point in this PR. We now have split the docs to its eaiser to maintain

DataM0del:
> I meant larger and more fancy examples using markdown. Im sorry but i don't really see the point in this PR. We now have split the docs to its eaiser to maintain

We could pull the descriptions and etc. from the docs and insert it into the schemas, and maybe insert a link to the docs.
Also, I think most things consuming the JSON schemas are using markdown parsers and then rendering the output HTML anyways.

DataM0del:
Also, I went through all of the config docs, and it seems to just be a few lists here, maybe a info, warning, tip etc. there, and I don't see any images or anything fancy.

Snowiiii:
I wonder if it is possible to generate Vitepress sites using json schemas

DataM0del:
> I wonder if it is possible to generate Vitepress sites using json schemas

We can use https://github.com/tomcollins/json-schema-static-docs to generate markdown docs, also, konflicts!!!!!

DataM0del:
Why is CI failing now...

Snowiiii:
I'm still not sure about this and i'm pretty sure this would make much more sense to add when we are close to release or already released since the config will change many times, and this will result in a bunch of conflicts. Feel free to open a discussion about this so other people can also share their thoughs

PR #312, opened 2024-11-18
fix so that dev on linux doesn't rebuild entire project

Bryntet:
## Description
On linux currently, debug builds rebuilds everything due to the contents of `pumpkin/build.rs`

## Testing
Tested so that the issue no longer occurs.

## Checklist


pumpkin/build.rs
```diff
@@ -2,7 +2,8 @@ use git_version::git_version;
 use std::env;
 
 fn main() {
-    if cfg!(target_os = "windows") {
+    #[cfg(target_os = "windows")]
```
Commandcracker, line 5: Also add `#[cfg(not(debug_assertions))]` because we don't need icons on debug build, and we have the same issue with recompiling on widows if we don't add it.

Commandcracker (changes requested):

pumpkin/build.rs
Bryntet: good catch!

Commandcracker:
I just found out, that tauri-winres uses embed-resource instead of rc.exe or windres.exe and ar.exe
Switching to it will probably speed up adding the icon and metadata. It should also make Cross-compiling easier again.

But I think this is something for another PR :D

Snowiiii:
LGTM.
Thanks @Bryntet 👍

PR #315, opened 2024-11-18
Use tauri-winres

Commandcracker:
## Description
tauri-winres is way faster because it is using embed_resource and this also means it does not rely on rc.exe or windres.exe and ar.exe, making cross-compile possible.

(See https://github.com/Snowiiii/Pumpkin/issues/312#issuecomment-2483389125)

## Checklist


Snowiiii:
Looks good to me.
Thanks @Commandcracker :D

PR #318, opened 2024-11-19
Update introduction.md

DirefulBelt6803:
## Checklist

### DIFF
--- a/docs/config/introduction.md
+++ b/docs/config/introduction.md
@@ -11,7 +11,7 @@ Pumpkin's Configuration is split into a basic Configuration made for quick chang
 
 ### Server Version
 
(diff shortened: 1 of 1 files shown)


Snowiiii:
Im not sure about "Pumpkin only supports the latest version of Minecraft". Pumpkin does not instantly support the latest Minecraft version when it comes out, We still need some time porting over. Something like "Pumpkin aims to support the latest Minecraft version" would be better

DirefulBelt6803:
Is there anything else?

Snowiiii:
> Is there anything else?

Sorry that i don't mentioned it directly. But i also think you should better mention that ViaProxy is a tool/plugin which runs on a 3th party Proxy.

Snowiiii:
Looks good now. 
Thanks @DirefulBelt6803

PR #319, opened 2024-11-20
"populate noise" step of chunk generation

kralverde:
This PR implements (most of) the "populate noise" stage of chunk generation

The blender (for smoothing (i think)) and the beardifier (for leaving space for world structures) are not implemented in this PR

**This PR changes the default world gen to this noise generation**

## Checklist
- [x] Re-implement chunk noise
- [x] Optimize chunk noise references
- [x] Look into hot code if chunks take too long to load
- [x] Add tests for actual chunk noise


kralverde:
The chunk noise generation and population each take like 0.5 seconds, so 1 second per chunk. I'll now start a deep dive into finding optimizations

kralverde:
Chunk gen times are now ~0.1sec which i think is good enough for now

OfficialKris:
This is a very large PR. I'm just curious, what was the methodology of this implementation?

kralverde:
What do you mean by methodology?

kralverde:
But basically I would look at the source code, write an equivalent in rust, and then write a bunch of unit tests for correctness. It was not fun lol

kralverde:
If this PR is merged as-is, the world generation would become that of the populate noise

kralverde:
Final bench is ~80msec per chunk on my machine

pumpkin-world/assets/no_bench_no_blend_0_0.java
```diff
@@ -0,0 +1,158 @@
+package com.example.mixin;
```
Snowiiii, line 1: What the hell is this class and why is this in pumpkin-world ?

kralverde: Its code to generate the no_bench_no_blend_0_0.chunk file

kralverde:
~60msec/chunk on my machine with the unsafe code

Q2297045667:
Report BUG
["populate noise" step of chunk generation](https://github.com/Snowiiii/Pumpkin/issues/340)

kralverde:
> Report BUG
> 
> ["populate noise" step of chunk generation](https://github.com/Snowiiii/Pumpkin/issues/340)

I don't think this is necessarily relevant to this PR

kralverde:
~50msec/chunk on my machine

kralverde:
currently benching at ~55msec/chunk now

pumpkin-world/assets/no_bench_no_blend_0_0.java
Snowiiii: Why you don't put it into the extractor/ ?

kralverde: I didn't think about that, I'll do that

extractor/src/main/kotlin/de/snowii/extractor/extractors/Tests.kt
```diff
@@ -0,0 +1,146 @@
...
+				return result
+			}
+		}
+		System.err.println("No valid method found for block state sampler!");
+		return null;
+	}
+
+	override fun gson_builder(): Gson {
+		return GsonBuilder().create()
+	}
```
Snowiiii, line 113: Is the only reason your using this is that you don't want pretty printing ?. Feel free to remove pretty printing from everything

kralverde:
@Snowiiii should be good to go now I think

OfficialKris:
Runs great on my computer, very impressive. But why does it have more loc than Minecraft itself?

kralverde:
> Runs great on my computer, very impressive. But why does it have more loc than Minecraft itself?

Loc?

user622628252416:
> > Runs great on my computer, very impressive. But why does it have more loc than Minecraft itself?
> 
> Loc?

lines of code

OfficialKris:
How are the asset files generated? Are they from minecraft? Is it possible to generate them at compile time? each of them are 3MB and that could be a problem to the git repo if they change often (like version updates).

kralverde:
> How are the asset files generated? Are they from minecraft? Is it possible to generate them at compile time? each of them are 3MB and that could be a problem to the git repo if they change often (like version updates).

They're generated from Java, not really able to generate them thru rust... maybe I could write a script to generate and run the tests? But then they couldn't be run with the PR tests.

extractor/src/main/kotlin/de/snowii/extractor/extractors/Tests.kt
kralverde: @Snowiiii removed

PR #323, opened 2024-11-20
feat: add envvar for config file paths

kekrs:
## Description
Docker or Kubernetes deployments would need a method to explicitly set configuration file paths. This PR adds functionality to set these.

## Checklist

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -5,6 +5,7 @@ use query::QueryConfig;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use std::{
+    env,
     fs,
     net::{Ipv4Addr, SocketAddr},
     path::Path,
@@ -159,7 +160,10 @@ trait LoadConfiguration {
 
 impl LoadConfiguration for AdvancedConfiguration {
     fn get_path() -> &'static Path {
-        Path::new("features.toml")
+        match env::var("FEATURES") {
+            Ok(v) => Path::new(v.leak()),
+            Err(_) => Path::new("features.toml"),
+        }
     }
 
     fn validate(&self) {
@@ -169,7 +173,10 @@ impl LoadConfiguration for AdvancedConfiguration {
 
 impl LoadConfiguration for BasicConfiguration {
     fn get_path() -> &'static Path {
-        Path::new("configuration.toml")
+        match env::var("CONFIG") {
+            Ok(v) => Path::new(v.leak()),
+            Err(_) => Path::new("configuration.toml"),
+        }
     }
 
     fn validate(&self) {
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -123,7 +123,8 @@ trait LoadConfiguration {
     where
         Self: Sized + Default + Serialize + DeserializeOwned,
     {
-        let path = Self::get_path();
+        let path_string = Self::get_path();
+        let path = Path::new(&path_string);
 
         let config = if path.exists() {
             let file_content = fs::read_to_string(path)
@@ -131,7 +132,7 @@ trait LoadConfiguration {
 
             toml::from_str(&file_content).unwrap_or_else(|err| {
                 panic!(
-                    "Couldn't parse config at {:?}. Reason: {}. This is is proberbly caused by an Config update, Just delete the old Config and start Pumpkin again",
+                    "Couldn't parse config at {:?}. Reason: {}. This is probably caused by a Config update, just delete the old Config and start Pumpkin again",
                     path,
                     err.message()
                 )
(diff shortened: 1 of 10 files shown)


Yimura:
Can you give me a scenario in which this is required? This env var only applies to the path inside of the docker container and I don’t see a reason for why this needs to be configurable.

I’ve already made the effort in #255 to give the deployer all the required flexibility, I’m trying to understand and compare this configurability with other docker containers to understand your reasoning and I can’t.

Sure you’re saying for “what” it’s useful in this PR, just not “why” it’s useful. This is lacking the motivation to convince the main contributor, especially since he may not be as experienced in Docker and Kubernetes.

kekrs:
> Sure you’re saying for “what” it’s useful in this PR, just not “why” it’s useful. This is lacking the motivation to convince the main contributor, especially since he may not be as experienced in Docker and Kubernetes.

Most sensible scenario for me is when you want to store the configuration away from the data. Right now it's not an issue since we don't have world saving yet. In more professional deployments you want to store the configuration in a version-controlled store, which can be accessed by the process only in ReadOnly mode. Good example here is Kubernetes ConfigMap. The data is where we store our state, and configuration shouldn't be treated as such.

Using the current WORKDIR and mount the volume there is absolutely fine for home or small-scale use cases.

Yimura:
So a possible scenario for this would be (just a docker compose as an example):
```yml
services:
  pumpkin:
    image: snowiiii/pumpkin:1.21-latest
    environment: 
      - CONFIG_PATH=/config/configuration.toml # I understand these keys are wrong though I prefer to be specific to avoid conflicts with future environment variables
      - FEATURES_PATH=/config/features.toml 
    ports:
      - 25565:25565
    volumes:
      - config_volume:/config:ro
      - pumpkin_data_volume:/pumpkin

volumes:
  config_volume:
    external: true
  pumpkin_data_volume:
```

Wouldn't you be able to mount those files directly under `/pumpkin/features.toml` and `/pumpkin/configuration.toml`? Since these more elaborate deployment environments are usually hosted on Linux, this should be possible.

kekrs:
> Wouldn't you be able to mount those files directly under `/pumpkin/features.toml` and `/pumpkin/configuration.toml`? Since these more elaborate deployment environments are usually hosted on Linux, this should be possible.

In a nutshell your example is correct. However:

```
    environment: 
      - CONFIG_PATH=/pumpkin/config/configuration.toml
      - FEATURES_PATH=/pumpkin/config/features.toml 
    volumes:
      - config_volume:/pumpkin/config:ro
      - pumpkin_data_volume:/pumpkin
```

its possible to do this as well. But to be frank, I wouldn't trust it with my life. This example is bad practice though.

In a hypothetical scenario there's a software/user/config error during the mounting process, we'll end up with two different and possibly diverging configurations. When modding or other world changing settings are introduced, this could lead to inconsistencies, even world corruption.

Let's say I have a configuration that alters worldgeneration. I forgot to mount the file to it's correct place. The software generates an empty "vanilla" configuration, and starts up, now it will generate chunks that way.

Another example is I have custom authentication configured. File fails to mount and now unauthorized players can join.

pumpkin-config/src/lib.rs
```diff
@@ -159,7 +160,10 @@ trait LoadConfiguration {
 
 impl LoadConfiguration for AdvancedConfiguration {
     fn get_path() -> &'static Path {
-        Path::new("features.toml")
+        match env::var("FEATURES") {
```
Yimura, line 163: Personally I'd like to have this env var be more descriptive and as such less likely to cause a name collision with out other environment variables.

Possible suggestion would be to change it to `FEATURES_PATH`.

pumpkin-config/src/lib.rs
```diff
@@ -169,7 +173,10 @@ impl LoadConfiguration for AdvancedConfiguration {
 
 impl LoadConfiguration for BasicConfiguration {
     fn get_path() -> &'static Path {
-        Path::new("configuration.toml")
+        match env::var("CONFIG") {
```
Yimura, line 176: Personally I'd like to have this env var be more descriptive and as such less likely to cause a name collision with out other environment variables.

Possible suggestion would be to change it to `CONFIG_PATH`.

pumpkin-config/src/lib.rs
```diff
@@ -159,7 +160,10 @@ trait LoadConfiguration {
 
 impl LoadConfiguration for AdvancedConfiguration {
     fn get_path() -> &'static Path {
-        Path::new("features.toml")
+        match env::var("FEATURES") {
+            Ok(v) => Path::new(v.leak()),
```
Yimura, line 164: I'm assuming usage of `.leak()` is fine since `.get_path()` is called only once during the lifetime of the application.

It's also why I left the `match` condition alone, if the `match` condition were to exist on a hot code path I'd rather have seen a different solution.

Yimura (changes requested):

pumpkin-config/src/lib.rs
kekrs: Renamed.

pumpkin-config/src/lib.rs
kekrs: Renamed it to CONFIGURATION_PATH to match filename and be consistent with FEATURES_PATH.

Yimura (approved):

Snowiiii:
Your force pushed already existing changes

PR #324, opened 2024-11-21
Consistent registry values and player changes

Alvsch:
## Description
- Now using IndexMap instead of HashMap for registry values so they have consistent ordering between runtime.
- `player.world()` can now be used instead of `&player.living_entity.entity.world`.
- Moved `player.respawn()` to `world.respawn_player()` as `spawn_player` is on world. (might reconsider)
- Preperations for multi-world support where the registry is involved.

## Testing
- same biome every time
- no spawning in the ground
- multiple players work as before

## Checklist


Snowiiii:
LGTM.
Thanks @Alvsch

PR #326, opened 2024-11-22
Replace Bruteforce search with Hashmap lookup in block_registry

PrintedScript:
## Description
Multiple functions in `block_registry` would do a brute force search by iterating through all blocks until it found a match. This PR would add multiple hashmaps to improve lookup times and improve performance

## Testing
- Tested world generation, setblock command, block placing and destroying
- Compared memory usage before and after changes on windows, added `~3.2MB` of memory usage

## Checklist

### DIFF
--- a/pumpkin-world/src/block/block_registry.rs
+++ b/pumpkin-world/src/block/block_registry.rs
@@ -1,4 +1,5 @@
 use std::sync::LazyLock;
+use std::collections::HashMap;
 
 use serde::Deserialize;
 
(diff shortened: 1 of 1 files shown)


Alvsch:
I feel like this could run into the same problem we had with CChunkData, with the massive amount of accesses/hashing that has to be done it could cause overhead.

Snowiiii:
LGTM. I ofc still would love to have the block! macro back (I removed it when porting to own Extractor). But still looks good.
Thanks @PrintedScript

PR #330, opened 2024-11-23
Parse blocks.json at compile time

user622628252416:
## Description
This pr experiments with parsing blocks.json at compile time instead of packaging the json string into the binary and parsing it at runtime. 

If no major downsides to this approach are found, it can in theory be used for other registries as well.

- Block/State/etc. struct definitions moved to pumpkin-core
- added pumpkin-data crate to make sure data is only recompiled when absolutely necessary
- build.rs is used to generate static arrays from json data

## Testing
run server, place/destoy/pick blocks

## Checklist


user622628252416:
## Obervations
- clippy is configured to skip the macro that generates data, and is therefore not impacted
- pumpkin-data takes ~30s on my machine the first time it's compiled
- cargo check appearantly does not share its cache with cargo run and also takes ~30s when checking for the first time
- incremental compilation is much faster ofc

user622628252416:
Binary size: 11 263 504 Bytes
Before: 17 724 240 Bytes
~36 % reduction

both were compiled in release mode

user622628252416:
Soo, what do we think about this?

1. Is the trade-off of the first compilation taking longer acceptable in the first place? 
2. Should I try to benchmark runtime performance changes? (should be faster, but change might be insignificant) RAM usage should be lower too, but I haven't yet tested that either
3. Should I do this for the rest of blocks.json (shapes and block entity kinds are still missing) and the other registries (items.json, ...) as well? If a decision in favor of this approach is made I'd be ready to invest the time and do this properly

DataM0del (approved):
nooooooo!!! 0.00001% increase in compile time!!! bad 100% would not recommend because 0.00001% increase in compileering time
###### /j

DataM0del (approved):

user622628252416:
@Snowiiii I think this is ready, lmk if you have any other concerns

assets/block_entities.json
```diff
@@ -1,227 +1 @@
...
-  }
-]
\ No newline at end of file
+{"minecraft:furnace":0,"minecraft:chest":1,"minecraft:trapped_chest":2,"minecraft:ender_chest":3,"minecraft:jukebox":4,"minecraft:dispenser":5,"minecraft:dropper":6,"minecraft:sign":7,"minecraft:hanging_sign":8,"minecraft:mob_spawner":9,"minecraft:creaking_heart":10,"minecraft:piston":11,"minecraft:brewing_stand":12,"minecraft:enchanting_table":13,"minecraft:end_portal":14,"minecraft:beacon":15,"minecraft:skull":16,"minecraft:daylight_detector":17,"minecraft:hopper":18,"minecraft:comparator":19,"minecraft:banner":20,"minecraft:structure_block":21,"minecraft:end_gateway":22,"minecraft:command_block":23,"minecraft:shulker_box":24,"minecraft:bed":25,"minecraft:conduit":26,"minecraft:barrel":27,"minecraft:smoker":28,"minecraft:blast_furnace":29,"minecraft:lectern":30,"minecraft:bell":31,"minecraft:jigsaw":32,"minecraft:campfire":33,"minecraft:beehive":34,"minecraft:sculk_sensor":35,"minecraft:calibrated_sculk_sensor":36,"minecraft:sculk_catalyst":37,"minecraft:sculk_shrieker":38,"minecraft:chiseled_bookshelf":39,"minecraft:brushable_block":40,"minecraft:decorated_pot":41,"minecraft:crafter":42,"minecraft:trial_spawner":43,"minecraft:vault":44}
```
DataM0del, line 1: Minified :vomiting_face:

user622628252416: Im not fully convinced either, but if you want to discuss this maybe open a separate issue or do so on discord

DataM0del:
*dies*

PR #332, opened 2024-11-23
Make auth urls optional, fix issues in prevent proxy connections

neeleshpoli:
## Description
Currently the auth URLs are not optional in features.toml. However, 99% of users will not change these values in the config file. This PR makes these URLs optional in the config file and falls back to the default auth URLs when not specified. By making them optional, they will be supported config options but not required in the config.

While making the URLs optional, I realized that there was a small issue in the code where the auth URL is still used even when `prevent-proxy-connections` is enabled. I fixed that and also fixed the IP address not being added to the URL correctly.

## Testing
Tested with building and joining the server 👍 

## Checklist

- [x] Make auth URLs optional
- [x] Fix issues with `prevent-proxy-connections`

### DIFF
--- a/pumpkin-config/src/auth.rs
+++ b/pumpkin-config/src/auth.rs
@@ -6,9 +6,9 @@ use serde::{Deserialize, Serialize};
 pub struct AuthenticationConfig {
     /// Whether to use Mojang authentication.
     pub enabled: bool,
-    pub auth_url: String,
+    pub auth_url: Option<String>,
     pub prevent_proxy_connections: bool,
-    pub prevent_proxy_connection_auth_url: String,
+    pub prevent_proxy_connection_auth_url: Option<String>,
     /// Player profile handling.
     pub player_profile: PlayerProfileConfig,
     /// Texture handling.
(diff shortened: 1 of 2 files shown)


Snowiiii:
I'm not sure if we should begin making Config fields optional. Why not just directly get them out of the Config and putting in a default?

neeleshpoli:
> I'm not sure if we should begin making Config fields optional. 

Some parts of the config are already optional, this PR aims to maintain that consistency for values that aren't changed by most users. Both Query config and LAN broadcast config have optional values, which aren't required to be specified explicitly. This is also the case with Vanilla's `server.properties`; some config fields are omitted by the default but can be added by the user.

> Why not just directly get them out of the Config and putting in a default?

Doing so would leave the values in the configuration, introducing unnecessary verbosity. Custom authentication servers are rarely used—I haven't encountered any public servers using a custom authentication setup.

This change won't break existing configs and studies show that users get more frustrated/turned away from projects with complex/verbose configurations. I see no harm in making the configuration simpler and less verbose, while still allowing for uncommon config options to be used by advanced users.

PR #333, opened 2024-11-23
Interactive blocks

leobeg:
## Description
This PR aims to add interactive blocks such as crafting tables, furnances and many more.

- It will add an easy expandable platform to register new block implementations

### How to use:

- Use the macro `#[pumpkin_block("minecraft:<your_block_id>")]` and specify your block id
- Implement the async trait `PumpkinBlock` for your block
- Include the block in the blocks mod
- Go into the mod file of the block folder
- Add your block to the manager with `manager.register(<YourBlock>);`
- Now your all set

### Currently implemented block events
`on_use`: Gets called when clicked with an empty hand
`on_use_with_item`: Gets called when clicked with an item
`on_placed`: Gets called after placing a block
`on_broken`: Gets called after the block is removed

**Note**: The `on_use_with_item` needs a `BlockActionResult` which can be `Continue` and `Consume`

## Testing
Start up the server, place a jukebox, insert a disc and start vibing

## Checklist

- [X] Code is well-formatted and adheres to project style guidelines: `cargo fmt`
- [X] Code does not produce any clippy warnings `cargo clippy`


OfficialKris:
very impressive! would be nice to get this merged so additional interactive blocks can be added!

leobeg:
> very impressive! would be nice to get this merged so additional interactive blocks can be added! Feature works before the master merge

Yeah I know it's currently broken. The shrinking of the json completly destroyed my code :(
But Ill fix it now

Snowiiii:
Works pretty good. Only problem is when the Jukebox is destroyed the Music is still playing. I think you should implement a function which gets called when a block is destroyed

leobeg:
> Works pretty good. Only problem is when the Jukebox is destroyed the Music is still playing. I think you should implement a function which gets called when a block is destroyed

Sounds good. Kinda forgot about that. Will add that soon

leobeg:
@Snowiiii Done

Snowiiii:
Tested again and everything including the block destruction works fine :D.
Big thanks @leobeg, really good work :+1:

PR #334, opened 2024-11-23
Update block extractor to get blockstate id using actual code and not an index in an array

kralverde:
## Description
This PR uses `Block.getRawIdFromState` instead of counting indices of a where the state is. This is more future proof.

This PR also includes the gradle wrapper scripts for the extractor which are commonly included in java/kotlin projects. It also adds a gitignore rule to allow the generated eula.txt to avoid needing to update it each time the extractor is ran.

## Testing
I extracted the blocks and confirmed there were no changes vs the original blocks.json


extractor/src/main/kotlin/de/snowii/extractor/extractors/Blocks.kt
```diff
@@ -52,9 +52,7 @@ class Blocks : Extractor.Extractor {
             val statesJson = JsonArray()
             for (state in block.stateManager.states) {
                 val stateJson = JsonObject()
-                val id = stateIdCounter
-                stateIdCounter++
-                stateJson.addProperty("id", id)
+                stateJson.addProperty("id", Block.getRawIdFromState(state))
```
Snowiiii, line 55: I think you can then remove the id counter vars right ?

.gitignore
```diff
@@ -125,3 +125,6 @@ docs/.vitepress/cache
 node_modules
 
 run/
+
+# Don't set EULA each time
+!run/eula.txt
```
Snowiiii, line 130: No please

extractor/src/main/kotlin/de/snowii/extractor/extractors/Blocks.kt
kralverde: I'm not sure what you mean, they are removed arent they?

PR #336, opened 2024-11-24
Added Named SoundEvents to Sound Effect Packets

OfficialKris:
## Description
Removed comments in favor of supporting sound events by name. Renamed sound_id to sound_event as it is in MC (can be either ID or SoundEvent). Allows for sounds with custom fixed distances.

## Testing
Ran the following code:
```rust
world.broadcast_packet_all(&CSoundEffect::new(
        VarInt(-1), // needed to send 0x00 to client to allow sound_name
        Some(SoundEvent {
            sound_name: String::from("ambient.cave"),
            range: Some(30.0),
        }),
        SoundCategory::Records,
        emit_sound_loc.x,
        emit_sound_loc.y,
        emit_sound_loc.z,
        1.0,
        1.0,
        seed,
    ))
    .await;
```

## Checklist

### DIFF
--- a/pumpkin-protocol/src/client/play/c_entity_sound_effect.rs
+++ b/pumpkin-protocol/src/client/play/c_entity_sound_effect.rs
@@ -1,12 +1,12 @@
 use pumpkin_macros::client_packet;
 use serde::Serialize;
 
-use crate::{SoundCategory, VarInt};
+use crate::{IDOrSoundEvent, SoundCategory, SoundEvent, VarInt};
 
 #[derive(Serialize)]
 #[client_packet("play:sound_entity")]
 pub struct CEntitySoundEffect {
-    sound_event: VarInt,
+    sound_event: IDOrSoundEvent,
     sound_category: VarInt,
     entity_id: VarInt,
     volume: f32,
@@ -16,7 +16,8 @@ pub struct CEntitySoundEffect {
 
 impl CEntitySoundEffect {
     pub fn new(
-        sound_event: VarInt,
+        sound_id: VarInt,
+        sound_event: Option<SoundEvent>,
         sound_category: SoundCategory,
         entity_id: VarInt,
         volume: f32,
@@ -24,7 +25,10 @@ impl CEntitySoundEffect {
         seed: f64,
     ) -> Self {
         Self {
-            sound_event: VarInt(sound_event.0 + 1),
+            sound_event: IDOrSoundEvent {
+                id: VarInt(sound_id.0 + 1),
+                sound_event,
+            },
             sound_category: VarInt(sound_category as i32),
             entity_id,
             volume,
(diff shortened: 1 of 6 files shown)


pumpkin-protocol/src/client/play/c_entity_sound_effect.rs
```diff
@@ -1,16 +1,30 @@
 use pumpkin_macros::client_packet;
-use serde::Serialize;
 
-use crate::{SoundCategory, VarInt};
+use crate::{ClientPacket, IDOrSoundEvent, SoundCategory, SoundEvent, VarInt};
+
+impl ClientPacket for CEntitySoundEffect {
```
Snowiiii, line 5: We usually always implement ClientPacket or ServerPacket at the under the new impl. You should do that as well so the Code follows the existing code style

pumpkin-protocol/src/client/play/c_sound_effect.rs
```diff
@@ -1,16 +1,32 @@
 use pumpkin_macros::client_packet;
-use serde::Serialize;
 
-use crate::{SoundCategory, VarInt};
+use crate::{ClientPacket, IDOrSoundEvent, SoundCategory, SoundEvent, VarInt};
+
+impl ClientPacket for CSoundEffect {
```
Snowiiii, line 5: Here as well

pumpkin-protocol/src/lib.rs
```diff
@@ -68,6 +68,18 @@ pub enum SoundCategory {
     Voice,
 }
 
+#[derive(Serialize)]
+pub struct IDOrSoundEvent {
+    pub id: VarInt,
+    pub sound_event: Option<SoundEvent>,
+}
```
Snowiiii, line 75: Im not sure what this is Struct is. Looking at Vanilla code the Packets do just need the SoundEvent

pumpkin-protocol/src/client/play/c_entity_sound_effect.rs
OfficialKris: I was following the docs, it makes it seem to replace #[derive(serialize)] with the impl ClientPacket. Maybe a note in the documentation could be added?

pumpkin-protocol/src/lib.rs
OfficialKris: This Struct is an implementation from the wiki.vg page. However, the wiki.vg page seems to be wrong in multiple locations for these packets, they did update it recently but it still wrong. Maybe this part is wrong as well. I will give it a look.

OfficialKris: Just looked at the vanilla code and it seems like it only sends SoundEvent but it must be adding an additional zero hex while encoding the object. Without the extra 0x00 sent, the client does not accept the additional string bytes. The vanilla codebase must only use the strings and thus the additional id parameter is not necessary for them to implement. We can do the same and just remove this struct in favor of a encoded 0x00.

Snowiiii:
Overall looks good to me.
Thanks @OfficialKris

PR #338, opened 2024-11-24
Store introduction-preview and use srcset  

Commandcracker:
## Description
Optimize image loading with responsive `<picture>` and srcset for better performance and accessibility.


Snowiiii:
Im not sure if we really want to sure 3 different or even more images now. The "old" solution worked well and the images where still stored on github

Commandcracker:
My problem is, when I access the page, then the image takes very long to load. And it makes no sense to have a high-res image when the display size is that small and it makes even less sense on mobile devices.

Snowiiii:
Mhh i see, Never encountered that Problems myself but i can see that it can lead to uncomfortable image loading speeds. It also may worth looking at https://squoosh.app/ which can easily compress these images

PR #339, opened 2024-11-24
Plugin system

vyPal:
Time spent on this PR: [wakatime](https://wakatime.com/badge/user/f464de95-0614-429a-a5db-86630ea3fed3/project/ff0097e0-55c2-4b76-8623-802dc34cbc33)

## Description
A plugin system using dynamic library loading. Library metadata is stored directly in the library file (`.so`, `.dylib`, or `.dll`) so all the server admin has to do is put the os-specific file into the plugin directory. Plugin metadata is pulled from the `Cargo.toml` file at compile time.

This system is still in an early stage of development, I need to implement a lot of things and split up the code to make it easier to work with in the future, as well as add tests, and some other things.

## TODO List
- [x] Dynamic library loading
- [x] On load/unload
- [x] Event priorities
- [x] Async support (if anyone could help with this I would appreciate it)
- [x] Custom commands
- [x] Command to manage plugins
- [x] List plugins in query
- [ ] Events (see below)
- [ ] ~~Extism (WASM) support~~ - might not implement
- [ ] ~~Lua support~~ - will be a separate pr

### Event types
- [ ] Block events (2%)
- [ ] Enchantment events (0%)
- [ ] Entity events (0%)
- [ ] Inventory events (0%)
- [ ] Server events (18%)
- [ ] Player events (0%)
- [ ] Vehicle events (0%)
- [ ] Weather events (0%)
- [ ] World events (0%)

(event types  based on this list: https://bukkit.org/threads/directory-list-of-events.112493/)

## Usage
A separate crate named `pumpkin-api-macros` defines macros to simplify plugin development:
```rs
#[plugin_method]
fn on_load(&mut self, server: &dyn PluginContext) -> Result<(), String> {
    server.get_logger().info("Plugin loaded!");
    Ok(())
}

#[plugin_event(blocking = true, priority = Highest)]
async fn on_player_join(
    &mut self,
    server: &dyn PluginContext,
    player: &PlayerEvent,
) -> Result<bool, String> {
    server.get_logger().info(
        format!(
            "Player {} joined the game. Config is {:#?}",
            player.gameprofile.name, self.config
        )
        .as_str(),

    let _ = player
        .send_message(
            TextComponent::text_string(format!(
                "Hello {}, welocme to the server",
                player.gameprofile.name
            ))
            .color_named(NamedColor::Green),
        )
        .await;
    Ok(true)
}

#[plugin_impl]
pub struct MyPlugin {}
```

There is an example plugin available at `plugins/hello-plugin-source/`

## Testing

Works on my machine :tm:
Tested on:
- Arch linux
- Windows 11


Commandcracker:
A method for generating plugin IDs is needed. I would like to have something decentralized. Using uuid4 could be a good option ?

To explain the importance of plugin IDs: plugins need a reliable way to declare dependencies on one another, and using names alone is not a robust long-term solution. Additionally, the plugin manager must be able to handle dependencies efficiently. It would also be worth investigating whether we can support scenarios where two plugins have circular dependencies, requiring each other.

We need a way to detect if a plugin is outdated. I propose that plugin_metadata should include the current plugin API version being used by the plugin. Then, the plugin loader can simply check whether the plugin's API version is compatible with the supported API versions (using semantic versioning).

I want to implement side-loading for plugins, specifically one for WASM and another for Lua. For this to work, we would need to ensure that the plugin stack can handle side-loaded plugins in a way that allows a normally loaded plugin to require a side-loaded one. This might involve allowing a plugin to load another "plugin". I don't know how we can make this work, but it would be really nice.

Commandcracker:
Btw: The system you made looks really good!

Forgot to say this: We also need to implement something like events or hooks.

vyPal:
> plugins need a reliable way to declare dependencies on one another, and using names alone is not a robust long-term solution.
One of my main targets is to make the plugins similair to spigot plugins, they use names as the plugin identifiers and for dependencies, but having some unique id system would probably be better.
> It would also be worth investigating whether we can support scenarios where two plugins have circular dependencies, requiring each other.
Depends on what functionality they would require from each other. Another good question is how exactly would the plugins interface with each other. I'd probably recommend a system where plugins can register certain methods or structs with the plugin manager, so that other plugins can access them.
> We need a way to detect if a plugin is outdated. I propose that plugin_metadata should include the current plugin API version being used by the plugin. Then, the plugin loader can simply check whether the plugin's API version is compatible with the supported API versions (using semantic versioning).
This is definetly a good idea. It would also be nice if the plugin manager could either periodically or on startup check if any plugins have newer versions available, but this would require implementing a plugin marketplace. The plugin marketplace could also help with providing unique IDs to plugins, but this would no longer be a decentralized system. Maybe if IDs were prefixed with the provisioner of the ID (like 'market:abcd1234' or something similair)?
> I want to implement side-loading for plugins, specifically one for WASM and another for Lua. For this to work, we would need to ensure that the plugin stack can handle side-loaded plugins in a way that allows a normally loaded plugin to require a side-loaded one. 
If I were to implement the functionality where plugins can register their methods on the manager and expose them to other plugins, this would allow us to use WASM and Lua as full plugins that could interface with other plugins as well
> Forgot to say this: We also need to implement something like events or hooks.
I am working on those now, I am primarily thinking up some efficient system for distributing these events between plugins, and a system for plugins to register events they will be listening for, so that all events dont need to be sent to all plugins

Snowiiii:
First of all, Good Job so far this PR looks pretty promising and you put a good amount of effort into it already. For now one think i would change is, to remove the plugin_metadata and require it in the plugins `Cargo.toml`.

Tohjuler:
Hi, I have a idea for both the id handling and update.
You could make a public plugin database of sorts, not for holding the plugins just info about the plugins, a id, versions, file hash (for verifying the file) and more.
It could also handle some telemetry, of course optional.
It would also help on security as the plugin files are verified, you could just send a warning if the plugin is not in the database, or have a option for a secure mode, where it will not load plugin that is not verified.
I'm sure there a holes in this idea, but I also just come op with it, have a good day :)

Snowiiii:
I saw you put 
Extism (WASM) support and Lua support into TODO. Im not a fan of that, and also i would make sense to split this into multiple PRs since this is something which will take a long time. Also See https://github.com/Snowiiii/Pumpkin/issues/110

DataM0del:
@vyPal If you want to add support for [Lua](https://lua.org/) (I would prefer [Luau](https://luau-lang.org) because it's basically Lua 5.1 but [typings](https://luau.org/syntax#type-annotations), [compound assignments](https://luau.org/syntax#type-annotations), and [some other features](https://luau.org/syntax), while still preserving support for Lua 5.1)

Commandcracker:
> @vyPal If you want to add support for [Lua](https://lua.org/) (I would prefer [Luau](https://luau-lang.org) because it's basically Lua 5.1 but [typings](https://luau.org/syntax#type-annotations), [compound assignments](https://luau.org/syntax#type-annotations), and [some other features](https://luau.org/syntax), while still preserving support for Lua 5.1), you should consider using [WebIDL](https://www.wikiwand.com/en/articles/Web_IDL) so that people can generate bindings for other languages.

Don't know about Web IDL, something like [wasmer](https://wasmer.io/) or [extism](https://extism.org/) looks better. But ill do Lua support with [mlua](https://crates.io/crates/mlua) it will be a separate plugin that allows sideloading Lua.

EricApostal:
> @vyPal If you want to add support for [Lua](https://lua.org/) (I would prefer [Luau](https://luau-lang.org) because it's basically Lua 5.1 but [typings](https://luau.org/syntax#type-annotations), [compound assignments](https://luau.org/syntax#type-annotations), and [some other features](https://luau.org/syntax), while still preserving support for Lua 5.1), you should consider using [WebIDL](https://www.wikiwand.com/en/articles/Web_IDL) so that people can generate bindings for other languages.
 
It should be noted that Luau is not backwards compatible with Lua 5.1. There are some small technical differences at runtime (naturally), but it isn't compatible with [some Lua 5.1 features](https://luau.org/compatibility) by default. Roblox built Luau to be backwards compatible only with the parts of 5.1 _they_ used, not the entire stack.

That being said, Luau should definitely be a consideration, especially over Lua. Furthermore, Luau is much [easier to sandbox](https://luau.org/sandbox), which is quite nice for securely running mods. I have used [mlua](https://github.com/mlua-rs/mlua) and it is quite nice.

Commandcracker:
I think we can support all Lua versions with mlua. Your plugin just needs to define what it wants to use.

lokka30:
I personally prefer sticking with the legacy plugin names rather than using any other identifier such as UUID. I don't think it solves anything and adds extra complexity. However, a way of handling circular dependencies is certainly worth looking into.

suprohub:
I want it

DataM0del:
> I want it

then make it!

samolego:
If I add my 2 cents, coming from fabric modding world; I'd vote for simple name ids, perhaps `author:plugin_name`, as suggested already. User will get confused if their server didn't start because of missing `637rhud7282j`.

lokka30:
> If I add my 2 cents, coming from fabric modding world; I'd vote for simple name ids, perhaps `author:plugin_name`, as suggested already. User will get confused if their server didn't start because of missing `637rhud7282j`.

Big +1 for a namespaced key type solution if we have to do anything beyond a simple plugin name.

0xnim:
I'm not very involved in pumkin, but i'd say there should be a standard way to package them for distribution, possibly a archive/zip file with each platform in there. so/dylib/dll. sounds simple enough

vyPal:
> I'm not very involved in pumkin, but i'd say there should be a standard way to package them for distribution, possibly a archive/zip file with each platform in there. so/dylib/dll. sounds simple enough

It would be useful to have something like that, but packing all three binaries in an archive seems like a waste of space. There are plans for an official pluign website (something like modrinth), which could be integrated directly into the server, so maybe there would be the possibility of just specifying a list of plugins in a configuration file somewhere, and the server would handle downloading and updating the version of the plugin specific to the OS that the server is running on (so you'd specify that you want to use plugin `a` and the server will recognize that it is on windows, so it will download `a.dll` from the plugin website and place it in the plugins folder

kralverde:
My +1 is that you should never ever ever ever ever ever ever ever just plug and play a dll. Have instructions for users to compile their own plugins and have the source code avaliable

OfficialKris:
Maybe the plugins folder could just contain metadata info about each plugin/settings through a config file. Then when pumpkin runs, it could download and install the requisite binaries in the system with manual input confirmation and hashing validation. That way, the binaries could be secured under administrator privages and be protected against erroneous/malicious writes (which I assume is the security issue here). Otherwise, Lua plugins could be used and be installed alongside in the plugins folder.

suprohub:
Are we on a finish line?

vyPal:
Well I'd say that it's near to a spot where it could potentially be merged. I still haven't implemented many things including world events (like block broken, placed, etc.) and there could still be quite a bit of optimization done, but otherwise I could potentially see this ready to merge soon

smuu:
Can we merge this and work on the different events and optimizations in other PRs?
@vyPal

vyPal:
I guess, it should be pretty stable. I'll just go over the codebase one more time and check if there aren't any things I meant to change but forgot to

vyPal:
But then again I am currently in the process of reworking the entire event system, so maybe I'd wait for that

Snowiiii:
I would prefer to move example plugins out of the main repo and into a new repo which will contain multiple plugin examples

Snowiiii:
I would also like to have the `get` and `set` functions which do only get or set a value and do not perform any more logic to follow the rust conventions and just have the name. e.g.
```diff
- fn get_join_message
+ fn join_message 
```

suprohub:
Final merge?

vyPal:
I think it is about ready to merge

Snowiiii:
Wow @vyPal. Unbelievable, over 100+ commits. Really appreciate the work you put in into this. I'm sure until the first release many things will change and i hope when we release we can provide a stable and consistent plugin API.
Thank you @vyPal, Great work :heart:

PR #346, opened 2024-11-27
feat(protocol): Switch from using flate2 to libdeflater to (de)compress zlib

HookWoods:
## Description
This merge request change the lib flate2 used to (de)compress packet to using directly [libdeflater](https://github.com/adamkewley/libdeflater) which is a Rust binding of the C heavily optimized libdeflate

Some change have been made:
- first we only create one instance of Decompressor and Compressor since the new constructor can be heavily to be created at each packet encoding / decoding.
- added validation on decompression to ensure the decompressed_size is at the good size, this will avoid some CPU attack when the zlib encoding is wrong

Some benchmark directly from the lib
https://github.com/adamkewley/libdeflater?tab=readme-ov-file#benchmarks

## Testing
I have created unit test based on previous system with flate2, run them with the new system to ensure we got the same result
And just launched the server and connected to ensure all was working and that I can play

## Checklist

### DIFF
--- a/pumpkin-protocol/Cargo.toml
+++ b/pumpkin-protocol/Cargo.toml
@@ -21,8 +21,9 @@ num-derive.workspace = true
 
 bytes = "1.8"
 
-flate2 = "1.0"
-
 # encryption
 aes = "0.8.4"
 cfb8 = "0.8.1"
+
+# decryption
+libdeflater = "1.22.0"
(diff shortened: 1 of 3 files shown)


Commandcracker:
I Would like to see benchmarks before and after.
Shouldn't we use streaming, it should be way faster ? But only flate2 can do streaming, right?
Benchmarks with differing compression level would also be nice so we can see what we should set as default.

Also regarding benchmarks, we should not benchmark on package. We should record all packets from a vanilla minecraft session for like 3 or 10 minutes, so we can see how it differs with different packets.

PR #347, opened 2024-11-27
Added compile time BlockState

Alvsch:
## Description
Added `block_state!` macro for compile time `BlockState`

## Testing
Loading into a pre-generated minecraft world, moving around and generating new chunks

## Checklist


kralverde:
Do you think you can also add a static reference to state to block state and also add the block id to state?

Snowiiii:
We now have duplicate code and also load the JSON twice which is a waste of RAM

Alvsch:
> We now have duplicate code and also load the JSON twice which is a waste of RAM

I would like to fix that, but cyclic dependencies caused a bit of problems

Snowiiii:
> > We now have duplicate code and also load the JSON twice which is a waste of RAM
> 
> I would like to fix that, but cyclic dependencies caused a bit of problems

Whats are the Problems?. Why you can't just remove it in pumpkin-world ?

Alvsch:
> Whats are the Problems?. Why you can't just remove it in pumpkin-world ?

We can't remove it from pumpkin-world because some functions require it at runtime as the values aren't known at compile time, such as commands.

> > We now have duplicate code and also load the JSON twice which is a waste of RAM
> 
> I would like to fix that, but cyclic dependencies caused a bit of problems

I realized that pumpkin-macro's static variables are only compile time so what I said before is irrelevant

kralverde:
> We now have duplicate code and also load the JSON twice which is a waste of RAM

No, in the macro the json is loaded at compile time.

Snowiiii:
Also looks good and works.
Thanks @Alvsch :heart:

Commandcracker:
Awesome 👍

Issue #273, opened 2024-11-14
Players can stop server


Rotstein007:
### I've searched existing issues and couldn't find a duplicate.

- [X] I confirm this is not a duplicate.

### Operating System

Linux manjaro

### Server Software Version/Commit

public server

### What happened?

Every player can /stop the server

### To Reproduce

/stop

### Expected behavior

server shouldn't be stoppable from any player

Snowiiii:
Yep, We probably will need to introduce a the Permission system to fix that

PR #348, opened 2024-11-28
Added native operator permission management 

KairuDeibisu:
This commit adds the ops.json configuration found in vanilla servers, and updates the basic configuration to handle the default permission level.

Server now uses ops file and defaults players to op level 0

The /op command has been added.

## Testing

Before starting the server:

1. If you had a previously existing ops.json file in the data folder (e.g., from testing this pull request), delete it.

> Note: ops.json is not present by default. It is automatically generated when the server is started for the first time, and it stores the list of server operators.

Start server:

4. Attempted to change the game mode to creative; this should fail.
5. Attempted to use  `/op JustATempest` ; this should fail.
6. Used `/op JustATempest` in the console; this should work.
7. In-game, attempted to change the game mode to creative; this should work.

### Testing Results

image
image
image
image

# Design challenges

Operator Level Management

To maintain parity with Minecraft, the operator level must be configurable within the basic configuration. Initially, the idea of introducing a new operator level type and conversions was explored. However, this approach led to duplicated code and complexity. Instead, it was suggested to centralize permission levels within the core, as the configuration system already depends on the core module.

Operator List (ops.json)

For parity with Minecraft, an ops.json file is required to store the list of operators on the server. This file is loaded at startup, and its data is referenced whenever a player is constructed. If a player’s UUID is not found in the list, they are assigned the default permission level of 0, which matches Minecraft’s behavior and cannot be changed.

/op Command

The /op command is also necessary for parity. This command accepts a single argument: the target player's name. A player must have a permission level of 3 or higher to use this command. Executing /op <player> adds the target player to the operator list with the default permission level specified in the basic configuration.

Edge Case: Overlapping Permission Levels

An edge case exists where a player with permission level 3 could use /op to grant another player a higher permission level (e.g., 4). To resolve this, the command checks the sender’s permission level against the default operator level in the basic configuration and applies the lower value.

Session Update

After a player is added to the operator list, their current session must reflect the updated permission level. This required adding a mutex to synchronize access to the permission level within the player structure.

Command Builder Limitation

This change introduced a side effect: the require function in the command builder does not support async functions. To address this, we opted to use the parking_lot crate instead of tokio, avoiding the need for async and await in the builder. 

# Fixes

- #273 

## Checklist


Snowiiii:
You added a bunch of deps and put OP into pumpkin-config. **This is not a Config**. It is used as Runtime data which can be changed. Im not sure if we should handle OPs per Server or per World base.

pumpkin/src/command/commands/cmd_op.rs
```diff
@@ -0,0 +1,77 @@
...
+        let Some(Arg::Players(targets)) = args.get(&ARG_TARGET) else {
+            return Err(InvalidConsumption(Some(ARG_TARGET.into())));
+        };
+
+        // log each player to the console.
+        for player in targets {
+            let op_entry = Op {
+                uuid: player.gameprofile.id,
+                name: player.gameprofile.name.clone(),
+                level: BASIC_CONFIG.op_permission_level,
```
Commandcracker, line 41: Maybe make it possible to set what level of op the player should be with cmd args.

pumpkin/src/command/commands/cmd_op.rs
```diff
@@ -0,0 +1,77 @@
...
+                bypasses_player_limit: false,
```
Commandcracker, line 42: Maybe make it possible to set this with cmd args.

pumpkin/src/entity/player.rs
```diff
@@ -218,8 +221,17 @@ impl Player {
             pending_chunks: Arc::new(parking_lot::Mutex::new(HashMap::new())),
             pending_chunk_batch: parking_lot::Mutex::new(HashMap::new()),
             cancel_tasks: Notify::new(),
-            // TODO: change this
-            permission_lvl: PermissionLvl::Four,
+            // Minecraft has no why to change the default permission level of new players.
+            // Minecrafts default permission level is 0
+            permission_lvl: OPERATOR_CONFIG
```
Commandcracker, line 226: It sould not read from the config every time, make it so it is stored in memory

pumpkin/src/server/json_config.rs
```diff
@@ -0,0 +1,92 @@
...
+                "Couldn't write operator config to {:?}. Reason: {}",
+                path, err
+            );
+        }
+    }
+}
+
+impl LoadJSONConfiguration for OperatorConfig {
+    fn get_path() -> &'static Path {
+        Path::new("ops.json")
```
Commandcracker, line 85: We should think about folder structure, maybe store it in ./data/ops.json

docker-compose.yml
```diff
@@ -2,6 +2,8 @@ services:
   pumpkin:
     build: .
     ports:
-      - 25565:25565
+      - "25565:25565"
     volumes:
       - ./data:/pumpkin
+    stdin_open: true
+    tty: true
```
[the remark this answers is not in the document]
Commandcracker: We should not have duplicate code, just move the enum to pumpkin-core, config already depends on it (and gamemodes are also in there).

pumpkin-config/src/op.rs
```diff
@@ -0,0 +1,49 @@
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use uuid::Uuid;
+
+#[derive(Clone, Copy)]
+#[repr(u8)]
+pub enum OpLevel {
```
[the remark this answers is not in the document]
Commandcracker: We should not have duplicate code, just move the enum to pumpkin-core, config already depends on it (and gamemodes are also in there).

pumpkin/src/server/json_config.rs
Commandcracker: I did not mean the data volume. (The volume should be renamed to pumpkin) All things that changed should be a separate folder like data and not inside the root folder.

The goal is to create a robust folder structure. Maybe I should make this an issue so it can be discussed on properly.

- pumpkin(.exe) (can be read and execute only)
- config (you could now make this folder red only)
  - configuration.toml
  - features.toml
- plugins (you could now make this folder read and execute only)
  - example(.dll)
  - example2(.dll)
  - example3(.dll)
- data (needs to be rw to work)
  - world
  - ops.json
  - (maby) perms.db

A folder structure like this is awesome if you use docker because now you can create volumes with restricting permissions.

OfficialKris:
Is there a way to change default op? for instance, on the test server, if we want to go to creative mode it would no longer be possible without an admin to give op permission. Maybe the setting could configure the default permission to be op level three rather than zero?

lukas0008 (commented):
dont have many complaints, maybe @Bryntet check if somethings amiss

PS: i started doing something else and forgot about this, so it took a while

Snowiiii:
Thanks @KairuDeibisu for your work :D

PR #349, opened 2024-11-29
Text component children + fix nbt serialization for sequences of maps

vyPal:
## Description
This PR contains 2 main changes:
### Fix NBT serialization of map inside sequence
When serializing a sequence to NBT, it is necessary to add a header containing the type of the elements and the length of the sequence. While the `serialize_seq` method sets the state to `FirstListElement`, this state is never parsed when the elements are maps (this issue could possibly exist even for other element types, however I havent been able to test this). In this change I have modified the `serialize_map` function to handle the cases where the state is either a `FirstListElement` or a `ListElement`.
```diff
fn serialize_map(self, _len: Option<usize>) -> Result<Self::SerializeMap> {
+   if let State::FirstListElement { .. } = self.state {
+       self.parse_state(COMPOUND_ID)?;
+   } else if let State::ListElement = self.state {
+       return Ok(self);
+   } else {
        self.output.put_u8(COMPOUND_ID);
+   }
    Ok(self)
}
```
### Add extra field to text components
Based on the [wiki](https://wiki.vg/Text_formatting#Content_fields), text components can have a `extra` field which contains other text components, which inherit the styling of their parent text component, but can also modify thier own style (and other attributes). This change adds a `add_child` method which accepts a `TextComponent` and returns `Self`:
```rs
TextComponent::text("Some parent")
    .add_child(TextComponent::text("A child component"))
    .add_child(TextComponent::text("Colored").color_named(NamedColor::Red))
```

## Testing
Existing test pass without an issue, and trying to send a message with extra text components works as expected.

## Checklist

### DIFF
--- a/pumpkin-core/src/text/mod.rs
+++ b/pumpkin-core/src/text/mod.rs
@@ -187,7 +187,15 @@ impl<'a> TextComponent<'a> {
             #[serde(rename = "extra")]
             extra: Vec<&'a TempStruct<'a>>,
         }
-        let temp_extra: Vec<TempStruct> = self.extra.iter().map(|x| TempStruct { text: &x.content, style: &x.style, extra: vec![] }).collect();
+        let temp_extra: Vec<TempStruct> = self
+            .extra
+            .iter()
+            .map(|x| TempStruct {
+                text: &x.content,
+                style: &x.style,
+                extra: vec![],
+            })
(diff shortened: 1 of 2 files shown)


Snowiiii:
Looks good to me.
Thanks @vyPal :+1:

PR #350, opened 2024-11-29
Added Spawn Egg Functionality

OfficialKris:
## Description
Adds initial mob support through spawn eggs. All spawn eggs work. No random characteristics are applied (ex. horse color or slime size). Serves as a basis for future work such as mob AI.

- Added entities.json and extractor
- Added entities registry
- Added support for right clicking spawn egg and spawning entity
- Added add_living_entity function to server and world and corresponding structures
- Living entities spawn facing the player and at the location of the player's cursor

## Testing
Tested both block placing and egg spawning
Tested spawning many different types of eggs

Future work on separate PRs:
- Entities_Registry Enum as a macro
- Random unique characteristics (horses color, slime size)
- Mob AI
- Mob attributes (health, affects)
- Mob spawning

## Checklist

---

TODO:
- [x] Fix constant player kicking after interacting with mobs (only attacks between players are allowed currently).
- [x] Add mobs to server implementation
- [x] Spawn rotation faces player


Snowiiii:
Any progress on this ?

OfficialKris:
I Was thinking that this should come in after an item manager. But maybe this could be merged in before that. However, I am finishing up the interactive blocks pr before this.

Snowiiii:
For some reason the entity ids seem to be wrong at some type, For example spawning a Zombie will result in spawning a Zombie Horse (Which is the next Type), And Spawning a Zombie Horse will result in Spawning a Zombie Villager (Which is the next type)

OfficialKris:
First 30 entities were correct. The `creaking_transient` entity was removed from the list so the elements after were off by one. Fixed

Snowiiii:
Just made an entity extractor my own (wanted to do that for a long time already), It also has the max_health for living entities, (The health is always random so its the max_health and not a static number as of my understanding). See https://github.com/Snowiiii/Pumpkin/commit/29ed502aeaa892ae06497f8dbf941b28bc715b6b

OfficialKris:
Fixed with new entities extractor support. All good to go. Idk why the cargo formatting check failed. `cargo fmt --check` is all good on my end.

pumpkin-world/src/entity/entity_registry.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct Entity {
+    pub id: u16,
+    pub max_health: Option<f64>,
+    pub attackable: bool,
+    pub summonable: bool,
+    pub fire_immune: bool,
+    pub dimension: Vec<f32>,
```
Snowiiii, line 34: Lets make this a `[f32; 2]`.  So we know its only 2 values and can also avoid a Vec

pumpkin-world/src/entity/entity_registry.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+pub fn get_entity_by_id<'a>(entity_id: u16) -> Option<&'a Entity> {
+    ENTITIES.values().find(|&entity| entity.id == entity_id)
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct Entity {
+    pub id: u16,
+    pub max_health: Option<f64>,
```
Snowiiii, line 30: its a float so `Option<f32>`

pumpkin/src/net/packet/play.rs
```diff
@@ -906,4 +894,106 @@ impl Player {
...
+        cursor_pos: Vector3<f32>,
+        face: &BlockFace,
+    ) -> Result<bool, Box<dyn PumpkinError>> {
+        // checks if spawn egg has a corresponding entity name
+        if let Some(spawn_item_name) = get_entity_id(&item_t) {
+            let head_yaw = 10.0;
+            let world_pos = WorldPosition(location.0 + face.to_offset());
+
+            // TODO: this should not be hardcoded
+            let (mob, _world, uuid) = server.add_living_entity(EntityType::Chicken).await;
```
Snowiiii, line 912: Why Chicken ?

Snowiiii:
This PR is exactly 1 Month old :sweat_smile:, Great work @OfficialKris like always. Thank you :heart:

PR #360, opened 2024-12-01
fix(protocol): Don't reserve memory on packet decoding error

HookWoods:
## Description
This MR fix https://github.com/Snowiiii/Pumpkin/issues/354 by returning in client poll function to avoid reserving 4096 bytes for each wrong packet

## Testing
Using the small program provided in https://github.com/Snowiiii/Pumpkin/issues/354 and see that ram never went up

## Checklist

### DIFF
--- a/pumpkin/src/client/mod.rs
+++ b/pumpkin/src/client/mod.rs
@@ -503,7 +503,10 @@ impl Client {
                     return true;
                 }
                 Ok(None) => (), //log::debug!("Waiting for more data to complete packet..."),
-                Err(err) => log::warn!("Failed to decode packet for: {}", err.to_string()),
+                Err(err) => {
+                    log::warn!("Failed to decode packet for: {}", err.to_string());
+                    return true; // return to avoid reserving additional bytes
+                }
             }
 
             dec.reserve(4096);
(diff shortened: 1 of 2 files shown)


Commandcracker:
Works 👍
Now there is only rate limiting left (for another pr) so we don't burn the CPU :P

Normal
image
Spam
Screenshot 2024-12-01 020343

Commandcracker:
Wait, shouldn't it return false ?

pumpkin/src/client/mod.rs
```diff
@@ -503,7 +503,10 @@ impl Client {
                     return true;
                 }
                 Ok(None) => (), //log::debug!("Waiting for more data to complete packet..."),
-                Err(err) => log::warn!("Failed to decode packet for: {}", err.to_string()),
+                Err(err) => {
+                    log::warn!("Failed to decode packet for: {}", err.to_string());
+                    return true; // return to avoid reserving additional bytes
```
Commandcracker, line 508: We should drop if the packet is incorrect 
```suggestion
                    return false; // return to avoid reserving additional bytes
```

Alvsch: Maybe even close the connection. Returning false just doesn't process packets without closing the connection

HookWoods: Ok good idea on that, now implemented thanks !

Snowiiii:
Looks good.
Thank you @HookWoods

PR #364, opened 2024-12-01
Lefthand and sneaking fixes

### LINKED ISSUE #361: Left hand model isnt rendered on server

### Operating System

Arch

latest

### What happened?

The left hand model does not render on the server instead right hand model is present

### To Reproduce

1.Go to settings
2.Skin Customization
3.Select left hand model

### Expected behavior

its expected for a left hand to be rendered and to appear rather then right hand.

Alvsch:
## Description
* Fixes #361 
* Fixed sneaking not working

## Testing
* Sneaking in-front of other players work.
* Switching between left and right handed is shown on the client.
* Hit animation swings with right arm for both self and others.

## Checklist

### DIFF
--- a/pumpkin/src/client/player_packet.rs
+++ b/pumpkin/src/client/player_packet.rs
@@ -438,11 +438,11 @@ impl Player {
             Hand::from_i32(client_information.main_hand.into()),
             ChatMode::from_i32(client_information.chat_mode.into()),
         ) {
-            let config = self.config.lock().await;
+            let mut config = self.config.lock().await;
             let update =
                 config.main_hand != main_hand || config.skin_parts != client_information.skin_parts;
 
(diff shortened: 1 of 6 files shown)


pumpkin/src/client/player_packet.rs
```diff
@@ -442,6 +449,7 @@ impl Player {
                 text_filtering: client_information.text_filtering,
                 server_listing: client_information.server_listing,
             };
+            self.update_client_information().await;
```
Snowiiii, line 452: why here

Alvsch: - - This is where we receive the new client information from the player, like skin layers and which hand they are using. So the function is called every time we get new information

Snowiiii:
Looks good.
Thanks @Alvsch

PR #366, opened 2024-12-03
Make asset extraction a local action

kralverde:
## Description
With changes with minecraft versions, asset files can have massive changes causing a lot of bloat on the git repo that can lead to MBs of extra un-needed data.

This PR introduces a script to generate and move asset files from the extractor to their locations in the project root as well as a workflow artifact for users to download rather than extract if they prefer

## Checklist

- [x] Update workflows to add a compressed archive of extracted files for those who don't want to extract them locally themselves
- [x] Update workflows to extract assets before building
- [x] Add windows scripts
- [x] Add bash scripts


DataM0del:
If you want to save space without setting the `depth` to `1`, you'll probably need to use a tool to remove the `assets` directory entirely from the git history.

kralverde:
I think this is more a forward looking solution. Re-writing got history is a pain with minimal benefits rn imo

OfficialKris:
Also, it is no longer just the assets folder. There are extracted JSON object files under `pumpkin-world/assets/`. The world assets folder is currently 50MiB and that is with just one revision. Any changes would bump that up significantly since the files are one line, minus any git optimizations.

OfficialKris:
Is there a reason to use os dependent scripts rather than running a rust program?

kralverde:
> Is there a reason to use os dependent scripts rather than running a rust program?

I suppose not... but I think it's a bit overkill for what boils down to running commands in a sequence

DataM0del (approved):

Snowiiii:
[extractor moved to:](https://github.com/Pumpkin-MC/Pumpkin/commit/b40c27444586cade80b560653ac443d0d76c5115) https://github.com/Pumpkin-MC/Extractor

PR #371, opened 2024-12-04
Add shadow_color

Commandcracker:
## Description
Add shadow_color, new in 1.21.4

## Testing
image
image

## Checklist


Snowiiii (commented):
I don't think you need to rename it

Commandcracker:
> I don't think you need to rename it

You need to, because it is shadow_color and not shadowColor
see here
image
Mojang with insanity consistent naming conventions -_-

Snowiiii:
Ah i though it just uses the field name, I didn't saw camelCase is used.

Snowiiii:
Looks good then.
Thanks @Commandcracker

PR #378, opened 2024-12-05
Fix chunk loading

kralverde:
## Description
This PR fixes the "but no one was watching spam", removes the need to wait for chunks when removing a player and removes the concept of pending tasks from the player, adds perma-loaded spawn chunks, drastically speeds up loading screen time when joining, and leverages rayon rather than async to stream chunks.

## Testing
Joined a world

## Checklist

### DIFF
--- a/pumpkin-world/src/level.rs
+++ b/pumpkin-world/src/level.rs
@@ -5,7 +5,10 @@ use num_traits::Zero;
 use pumpkin_config::BASIC_CONFIG;
 use pumpkin_core::math::vector2::Vector2;
 use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
-use tokio::sync::{mpsc, RwLock};
+use tokio::{
+    runtime::Handle,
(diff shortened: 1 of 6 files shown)


kralverde:
That's not a memory leak that's just holding chunks in memory...

kralverde:
> Works great. There seems to be a memory leak somewhere because I got up to 700MB ram usage after roaming around at 32 chunks then going back to spawn and turning the setting back to 10. It continued to increase whenever generating chunks and didn't go down. Current master maxes at ~300MB  then immediately drops to 51MB after not generating chunks. This was with release build.

I'm unable to reproduce. Are you sure it's a memory leak and not rust just allocating memory for the map?

kralverde:
@OfficialKris can you check your benchmarks again? I'm curious how this latest commit affects it

Snowiiii:
Just tested and it works pretty nice. Finally pumpkin is usable in debug builds :D.
Thanks @kralverde

PR #381, opened 2024-12-07
Use all cpus for intensive tasks

kralverde:
## Description
https://github.com/Snowiiii/Pumpkin/commit/cd6ac147ec49e47bd3722680511dc55c26b397b0 introduced a regression where CPU intensive tasks are not performed by all CPUs. This is because the rayon pool does not queue tasks until the current task is complete and because the tokio runtime never ends we effectively lose a CPU.

Additionally, the root cause of the tokio runtime choking is because we were waiting for the results of the rayon pool directly from the tokio runtime, blocking it, which obviously defeats the purpose. 

This PR fixes these issues, sorts initial spawn chunks so that rayon processes them in rings rather than slices, and adds a note in CONTRIBUTING.md on how to avoid issues like this in the future.

## Testing
Joined the server in debug mode and ran commands while chunks loaded

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,17 @@ There are several ways you can contribute to Pumpkin:
   Write clear and concise commit messages that describe your changes.
 
 ### Docs
-The Documentation of Pumpkin can be found at https://snowiiii.github.io/Pumpkin/
+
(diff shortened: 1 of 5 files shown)


Snowiiii:
LGTM.
Thanks @kralverde

PR #382, opened 2024-12-07
Added Initial Time Packet on Login

OfficialKris:
## Description
Fixes a bug where when a player logs in, the time is shown to be incorrect for one second.

## Testing
- Logged in
- Set time to 18000
- Logout
- Login and notice time is always correct 

## Checklist


Snowiiii:
For some reason this does not work. If i set the time to 0 and rejoin it shows me the a different time

OfficialKris:
I cannot reproduce, are there any additional steps you took?

pumpkin/src/world/mod.rs
```diff
@@ -404,6 +404,9 @@ impl World {
         // Spawn in initial chunks
         player_chunker::player_join(self, player.clone()).await;
 
+        // Sends initial time
+        player.send_time(self).await;
```
Snowiiii, line 408: Also i think we should send this before sending chunks, like vanilla does it

Snowiiii:
> I cannot reproduce, are there any additional steps you took?

mhh, nope

pumpkin/src/entity/player.rs
```diff
@@ -396,6 +399,18 @@ impl Player {
         self.permission_lvl
     }
 
+    // sends the world time to just the player
```
Snowiiii, line 402: Doc comments are written using ///. Normal comments are just //.

Snowiiii:
Thanks @OfficialKris

PR #383, opened 2024-12-07
Added Basic Attack Modifiers

OfficialKris:
## Description
Added basic attack modifiers for all items (ex. swords, axes, maces):
- base_attack_damage - modifies the base attack damage
- base_attack_speed - modifies the attack speed and resulting damage mid way

## Testing
Tested in offline mode
Counted health damage and compared to wiki

TODO:
- There may be calculated values which are incorrect. I read that MC goes 2 decimal places while Pumpkin uses f32.
- Each hit requires the comparison of many strings. This could be improved by parsing the strings at compile time into enums. 

## Checklist

### DIFF
--- a/pumpkin-world/src/item/item_registry.rs
+++ b/pumpkin-world/src/item/item_registry.rs
@@ -33,9 +33,30 @@ pub struct ItemComponents {
     pub max_stack_size: u8,
     #[serde(rename = "minecraft:jukebox_playable")]
     pub jukebox_playable: Option<JukeboxPlayable>,
+    #[serde(rename = "minecraft:damage")]
+    pub damage: Option<u16>,
+    #[serde(rename = "minecraft:max_damage")]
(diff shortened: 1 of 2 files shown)


pumpkin-world/src/item/item_registry.rs
```diff
@@ -33,9 +33,30 @@ pub struct ItemComponents {
...
+pub struct AttributeModifiers {
+    pub modifiers: Vec<Modifier>,
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct Modifier {
+    #[serde(rename = "type")]
+    pub type_val: String,
+    pub id: String,
+    pub amount: f32,
```
Snowiiii, line 59: amount is a double (f64)

Snowiiii:
Code looks good and damage now finally works, Weapons are now not useless now :smile: 
Thank you @OfficialKris :+1:

PR #387, opened 2024-12-08
Don't break blocks with swords in creative

Commandcracker:
## Description
You can't break blocks in creative while holding those items.

## Testing
tested with those items.

## Checklist

### DIFF
--- a/pumpkin/src/client/player_packet.rs
+++ b/pumpkin/src/client/player_packet.rs
@@ -1,4 +1,5 @@
-use std::sync::Arc;
+use std::collections::HashSet;
+use std::sync::{Arc, LazyLock};
 
 use super::PlayerConfig;
 use crate::block::block_manager::BlockActionResult;
@@ -37,7 +38,7 @@ use pumpkin_protocol::{
     },
 };
 use pumpkin_world::block::{block_registry::get_block_by_item, BlockFace};
-use pumpkin_world::item::item_registry::get_item_by_id;
+use pumpkin_world::item::item_registry::{get_item, get_item_by_id};
 use thiserror::Error;
 
 fn modulus(a: f32, b: f32) -> f32 {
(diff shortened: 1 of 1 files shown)


pumpkin/src/client/player_packet.rs
```diff
@@ -84,6 +85,22 @@ impl PumpkinError for BlockPlacingError {
...
 }
 
+static CREATIVE_DONT_BREAK_WITH: LazyLock<HashSet<u16>> = LazyLock::new(|| {
+    [
+        get_item("wooden_sword").unwrap().id,
+        get_item("stone_sword").unwrap().id,
+        get_item("iron_sword").unwrap().id,
+        get_item("golden_sword").unwrap().id,
+        get_item("diamond_sword").unwrap().id,
+        get_item("netherite_sword").unwrap().id,
```
DataM0del, line 95: Why not just check if it's a sword instead of having a hardcoded list of items?

DataM0del, line 95: Why not just check if it's a sword instead of having a hardcoded list of items?

Commandcracker: is_sword is also hardcoded, and it is missing the trident and mace

Snowiiii:
This is something we want to have abstractions to work with. I would consider an System simliar to the current Interactive blocks System. In the future we will have to add more logic to items and for example a method like Mojang uses `canMine` would be great

OfficialKris:
#350 would also benefit from an "item manager." This seems to be pretty important, especially with the rise in added functionality to individual items. I can start work on it once I finish #388 or someone else can give it a shot. Currently, the handle functions have too much going on. Those function should only find out which item is being used and then run the corresponding category, modifier, and specific item code (pickaxes, enchantments, spawn eggs).

pumpkin/src/client/player_packet.rs
DataM0del: > is_sword is also hardcoded, and it is missing the trident and mace

:facepalm:
something filtering over all items could filter using `contains` and checking if the item name contains `sword`
```rs
itemName.contains("sword")
```
Also, doesn't Minecraft have metadata for which category or what type of item this is (e.g. swords)?

PR #384, opened 2024-12-08
Added Chest and Furnace as Interactive Blocks

OfficialKris:
## Description
Added chest and furnace as interactive blocks (furnace does not produce any items)
Also added on_close function for interactive blocks for when client closes a container

## Testing
Place chest, crafting table, and furnace
Place items in interactive blocks
Close interactive blocks and no client errors

## Checklist
Things need to be done before this Pull Request can be merged.

- [x] Code is well-formatted and adheres to project style guidelines: `cargo fmt`
- [x] Code does not produce any clippy warnings: `cargo clippy`
- [x] All unit tests pass: `cargo test`
- [ ] I added new unit tests, so other people don't accidentally break my code by changing other parts of the codebase. [How?](https://doc.rust-lang.org/book/ch11-01-writing-tests.html)

PR #388, opened 2024-12-09
Correct Interactive Block Behavior

### LINKED ISSUE #275: Items placed in the crafting table are shared globaly

### Operating System

Linux manjaro

public server

### What happened?

I placed Items in the /craft crafting table. Another player on the Server saw, what I put in and could "edit" the position of the Items.

### To Reproduce

/craft and put items in the crafting table. Another player should have acces to the items.

### Expected behavior

Crafting tables should stay independent for each player.

OfficialKris:
## Description
Fixes #275 and many other interactive block issues:
- All interactive blocks are per block now (for instance, chests and furnaces do not share containers globally)
- Crafting table containers are both unique per player AND per block
- Breaking interactive blocks removes all contents (can drop items once item drop is implemented)
- All active container users are forcibly closed when another `player` removes a block
- Removed artificial cap on container screen limit which kicked players after opening a container 15 times (we may not be handling container ids correctly)
- Added a `standard` set of container functions to reduce code duplication for interactive blocks

This PR requires #384 

## Testing
Two clients
- Place a crafting table, a furnace, and a chest
- Open each and place items in container
- Open crafting table again, notice no items are stored
- Open all other blocks and see items in both clients
- Open crafting table in both and notice containers are not shared

Internally, this change maintains the OpenContainer HashMap and a new AtomicU32 which tracks the current index for the generation of uuids. On large persistent servers, unique containers could be optimized to reduce memory usage.  Every user who opens a unique container will generate and cache a new data structure (think 10,000 users opening a crafting table, enderchest, etc). Future work could include separating unique containers (crafting table) and persistent unique containers (ender chest) into separate data structures (ex. stacks instead of maps).

Also, another optimization could be made to client interactive block interactions. Currently, this requires that the server iterate through the entirety of the OpenContainer HashMap. It may be better to create a sort of registry which maps block position to OpenContainerID in the long term (again large server optimizations). Otherwise, the algorithm is randomized brute force with a time complexity of O(n + m).

## Checklist


pumpkin/src/block/blocks/crafting_table.rs
```diff
@@ -6,52 +6,112 @@ use async_trait::async_trait;
...
-                    Some(location),
-                    get_block("minecraft:crafting_table").cloned(),
-                );
-                open_containers.insert(1, open_container);
+                player
+                    .open_container
+                    .store(Some(id_to_use.try_into().unwrap()));
             }
         }
+        drop(open_containers);
```
DataM0del, line 114: not sure what's the need for manually managing memory as if it was C

leobeg: > not sure what's the need for manually managing memory as if it was C

I think it's not necassary here but it's more like freeing the guard then managing the memory. 
For example if you call an sub function that trys to lock the variable too you create an deadlock.
Which results in freezing

OfficialKris: I think this is a draft and in progress :D I think I was trying to fix a different issue that I thought was from a deadlock. I will probably refactor this code

Snowiiii:
When opening a Chest, the server freezes and i get a timeout

OfficialKris:
Fixed, tried to optimize code like this was C.

Snowiiii:
Okay works now, But i found a bug. If you shift click a item from a chest into your inventory and reopen the chest the item will be in the chest again

OfficialKris:
I checked comparing with master and it seems shift clicking is just not supported yet. Shift clicking into containers does the same thing. I believe that should be a separate PR since it would be modifying other interaction container code.

Snowiiii:
Waiting for @Bryntet review, but for some reason crafting does not work anymore :c

Bryntet:
@Snowiiii IIRC crafting is currently broken on master too? or am I misremembering

Snowiiii:
> @Snowiiii IIRC crafting is currently broken on master too? or am I misremembering

yep, its broken on master

Snowiiii:
Merged! Thanks for your hard work on this, @OfficialKris :heart:

PR #390, opened 2024-12-09
check closed status before encrypting and sending assets

kralverde:
## Description
This PR makes it so the server does not attempt to encrypt and send a packet if the connection is closed

## Testing
N/A

## Checklist


Snowiiii:
I was thinking of that, I wonder how much of an Performance impact this may has due to the AtomicBool we have to load every time.
But it makes sense to check if the connection is closed there.
So thanks @kralverde 👍

PR #402, opened 2024-12-21
Implement level.dat reading and load the seed for use in world generation

neeleshpoli:
## Description
Adds support for reading the `level.dat` file in the anvil world format. With this support, the seed in the `level.dat` file is now used for chunk generation.

## Testing
Create a world in vanilla minecraft, and load the world into pumpkin. The newly generated chunks should seamlessly blend with the already generated chunks. A sample `level.dat` was added to unit tests for further testing.

## Checklist
- [x] Implement `level.dat` loading


Snowiiii:
This should be placed not in the chunk dir.

neeleshpoli:
I placed it in the chunk dir because this is specific thing to anvil world format; hence why it's in anvil file. Do you want me to separate the level.dat loading and chunk loading?

Snowiiii:
I would like to separate chunk loading and "general world data loading". It would also make no sense to put it into the chunk folder since it has nothing to do with chunks. Also later someone can may use a different world format while having anvil chunks or the other way around

Snowiiii:
Looks good now. It so impressive how well everything plays together now. Our generation exactly matches the vanilla. Still have to say great job @kralverde 
image
Thanks @neeleshpoli :+1:

PR #403, opened 2024-12-21
Fix seed command output

kralverde:
## Description
Prints the seed as a i64 instead of a u64

## Testing
Ran /seed with a negative seed

## Checklist

### DIFF
--- a/pumpkin/src/command/commands/cmd_seed.rs
+++ b/pumpkin/src/command/commands/cmd_seed.rs
@@ -24,11 +24,9 @@ impl CommandExecutor for PumpkinExecutor {
         _args: &ConsumedArgs<'a>,
     ) -> Result<(), CommandError> {
         let seed = match sender {
-            CommandSender::Player(player) => {
-                player.living_entity.entity.world.level.seed.0.to_string()
-            }
+            CommandSender::Player(player) => player.living_entity.entity.world.level.seed.0,
             _ => match server.worlds.first() {
-                Some(world) => world.level.seed.0.to_string(),
+                Some(world) => world.level.seed.0,
                 None => {
                     return Err(CommandError::GeneralCommandIssue(
                         "Unable to get Seed".to_string(),
@@ -36,6 +34,7 @@ impl CommandExecutor for PumpkinExecutor {
                 }
             },
         };
+        let seed = (seed as i64).to_string();
 
         sender
             .send_message(


Snowiiii:
I'm pretty sure that https://github.com/Snowiiii/Pumpkin/blob/cc0f30848ea597ef702558cb1e5fc57ed2a20372/pumpkin-world/src/world_gen/seed.rs#L4 should be also an i64 then

kralverde:
> I'm pretty sure that https://github.com/Snowiiii/Pumpkin/blob/cc0f30848ea597ef702558cb1e5fc57ed2a20372/pumpkin-world/src/world_gen/seed.rs#L4 should be also an i64 then

No, internally our seed is a u64, it doesn't make sense for it to be signed. Converting from u64 to i64 is a no-op anyway because it has the same bits underneath

Snowiiii:
> > I'm pretty sure that https://github.com/Snowiiii/Pumpkin/blob/cc0f30848ea597ef702558cb1e5fc57ed2a20372/pumpkin-world/src/world_gen/seed.rs#L4
> >  should be also an i64 then
> 
> No, internally our seed is a u64, it doesn't make sense for it to be signed. Converting from u64 to i64 is a no-op anyway because it has the same bits underneath

But why it should be unsigned ?. We may also want to use negative seeds internally, Or save them

kralverde:
> > > I'm pretty sure that https://github.com/Snowiiii/Pumpkin/blob/cc0f30848ea597ef702558cb1e5fc57ed2a20372/pumpkin-world/src/world_gen/seed.rs#L4
> 
> > >  should be also an i64 then
> 
> > 
> 
> > No, internally our seed is a u64, it doesn't make sense for it to be signed. Converting from u64 to i64 is a no-op anyway because it has the same bits underneath
> 
> 
> 
> But why it should be unsigned ?. We may also want to use negative seeds internally, Or save them

IMO It doesn't make sense to me conceptually for it to be signed. There's not concept of positive or negative for what is effectively a bit string for our RNG. I think having it signed was a limitation of Java.

We can always add a serialize method to the seed struct for writing and stuff to do to no-op `as i64` to minimize confusion.

Snowiiii:
Mhh Okay.
I also wonder if https://github.com/Snowiiii/Pumpkin/blob/9b184f73b99c6b7b97cf4ee0fe08be84d497b33a/pumpkin-config/src/lib.rs#L70 could may be a i64 directly so we don't have to parse it from a String

kralverde:
> Mhh Okay.
> 
> I also wonder if https://github.com/Snowiiii/Pumpkin/blob/9b184f73b99c6b7b97cf4ee0fe08be84d497b33a/pumpkin-config/src/lib.rs#L70 could may be a i64 directly so we don't have to parse it from a String

Users can input strings as seeds tho

Snowiiii:
> > Mhh Okay.
> > I also wonder if https://github.com/Snowiiii/Pumpkin/blob/9b184f73b99c6b7b97cf4ee0fe08be84d497b33a/pumpkin-config/src/lib.rs#L70
> >  could may be a i64 directly so we don't have to parse it from a String
> 
> Users can input strings as seeds tho

I'm not sure what you mean, a seed is always a number isn't it ?

kralverde:
> > > Mhh Okay.
> 
> > > I also wonder if https://github.com/Snowiiii/Pumpkin/blob/9b184f73b99c6b7b97cf4ee0fe08be84d497b33a/pumpkin-config/src/lib.rs#L70
> 
> > >  could may be a i64 directly so we don't have to parse it from a String
> 
> > 
> 
> > Users can input strings as seeds tho
> 
> 
> 
> I'm not sure what you mean, a seed is always a number isn't it ?, If you mean to just using " ", this isn't really an argument for me, I think its nicer to just force the user to put a raw number instead of using a String and then parsing is back into a number

But like, you can have a string be the seed and it gets hashed into a numerical value. So you are right in that they are ultimately numbers, but if you go into Minecraft you can type a sentence into the set seed and it will work.

Snowiiii:
Overall looks good.
Thanks @kralverde

PR #407, opened 2024-12-24
Create configs in a folder

Snowiiii:
## Description
Creates configuration files in a separate folder which then can be in a read-only mode.
Addresses https://github.com/Snowiiii/Pumpkin/issues/377

## Testing
- [x] Linux: Changed config values and saw that the server detected the new values

- [x] Docker

## Checklist

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -5,7 +5,7 @@ use query::QueryConfig;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use std::{
-    fs,
+    env, fs,
     net::{Ipv4Addr, SocketAddr},
     num::NonZeroU8,
     path::Path,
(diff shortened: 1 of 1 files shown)


pumpkin-config/src/lib.rs
```diff
@@ -126,26 +128,33 @@ trait LoadConfiguration {
     where
         Self: Sized + Default + Serialize + DeserializeOwned,
     {
-        let path = Self::get_path();
+        let exe_path = env::current_exe().unwrap();
```
Commandcracker, line 131: In docker, the config folder is created in `/bin`, something is wrong!
It should not use current_exe as base path, instead use CWD see env::current_dir()

Commandcracker:
Maybe we should make it possible tho change the path via CLI args or/and env vars.
see `java -jar .\paper.jar --help` (--plugins --config etc.)

pumpkin-config/src/lib.rs
Snowiiii: Makes sense, did not saw that. Can you may test it again with the latest change

Commandcracker:
Now it works how it should be, config folder inside data volume

PR #421, opened 2024-12-27
Added basic chest opening and closing

OfficialKris:
## Description
Added chest opening and closing animation and only trigger sound when the first or last player does the correct action.

## Testing
Two clients:

- Place one chest, open and close, see animation in both and notice sound
- Open in second client, close in first, stays open and no sound
- Close in both, it closes and plays a sound

This does not add the constant ticks and also sends all chest packets to all players instead of just when the respective chunks are loaded. Both of these issues would be separate PRs.

## Checklist


Snowiiii:
Merged, Big thanks for your work @OfficialKris :+1:

PR #425, opened 2024-12-27
Add egg-pumpkin.json

Commandcracker:
## Description
Add egg for Pterodactyl

## Testing

### DIFF
--- a/egg-pumpkin.json
+++ b/egg-pumpkin.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2024-12-27T22:08:25+01:00",
+    "exported_at": "2024-12-27T23:40:43+01:00",
     "name": "pumpkin",
     "author": "unknown@unknown.com",
     "description": "Pumpkin is a Minecraft server built entirely in Rust, offering a fast, efficient, and customizable experience. It prioritizes performance and player enjoyment while adhering to the core mechanics of the game. Requires around 2GB RAM to build, adjust docker.installer_limits.memory in your wing accordingly! Currently, you need to adjust the port in the config manually!",
@@ -16,13 +16,13 @@
     "startup": ".\/pumpkin",
     "config": {
         "files": "{}",
-        "startup": "{}",
+        "startup": "{\r\n    \"done\": \"Started Server took\"\r\n}",
         "logs": "{}",
         "stop": "stop"
     },
     "scripts": {
         "installation": {
-            "script": "#!\/bin\/ash\r\nset -eux\r\n\r\napk add --no-cache musl-dev git\r\nmkdir -p \/mnt\/server\r\n\r\nREPO_URL=\"${GIT_URL:-https:\/\/github.com\/Snowiiii\/Pumpkin.git}\"\r\n\r\nif [ -n \"$GIT_BRANCH\" ]; then\r\n    git clone --depth 1 --single-branch --branch \"$GIT_BRANCH\" \"$REPO_URL\"\r\nelse\r\n    git clone --depth 1 --single-branch \"$REPO_URL\"\r\nfi\r\n\r\nif [ -n \"$GIT_COMMIT\" ]; then\r\n    git checkout \"$GIT_COMMIT\"\r\nfi\r\n\r\ncd Pumpkin\r\n\r\nif [[ \"$BUILD_RELEASE\" == \"1\" || \"$BUILD_RELEASE\" == \"true\" ]]; then\r\n    cargo build --release\r\n    strip target\/release\/pumpkin\r\n    cp target\/release\/pumpkin \/mnt\/server\/pumpkin\r\nelse\r\n    cargo build\r\n    cp target\/debug\/pumpkin \/mnt\/server\/pumpkin\r\nfi",
+            "script": "#!\/bin\/ash\r\nset -eux\r\n\r\napk add --no-cache musl-dev git\r\nmkdir -p \/mnt\/server\r\n\r\nREPO_URL=\"${GIT_URL:-https:\/\/github.com\/Snowiiii\/Pumpkin.git}\"\r\n\r\nif [ -n \"$GIT_BRANCH\" ]; then\r\n    git clone --depth 1 --single-branch --branch \"$GIT_BRANCH\" \"$REPO_URL\"\r\nelse\r\n    git clone --depth 1 --single-branch \"$REPO_URL\"\r\nfi\r\n\r\nif [ -n \"$GIT_COMMIT\" ]; then\r\n    git checkout \"$GIT_COMMIT\"\r\nfi\r\n\r\ncd Pumpkin\r\n\r\n#export RUSTFLAGS=\"-C target-feature=-crt-static -C target-cpu=native\"\r\n\r\nif [[ \"$BUILD_RELEASE\" == \"1\" || \"$BUILD_RELEASE\" == \"true\" ]]; then\r\n    cargo build --release\r\n    strip target\/release\/pumpkin\r\n    cp target\/release\/pumpkin \/mnt\/server\/pumpkin\r\nelse\r\n    cargo build\r\n    cp target\/debug\/pumpkin \/mnt\/server\/pumpkin\r\nfi",
             "container": "docker.io\/rust:1-alpine3.21",
             "entrypoint": "ash"
         }
@@ -76,6 +76,16 @@
             "user_viewable": true,
             "user_editable": true,
             "rules": "nullable|string",
+            "field_type": "text"
+        },
+        {
+            "name": "Cargo build jobs",
+            "description": "Amount of jobs that cargo spawns to compile. Try to reduce them if you encounter a deadlock.\r\nScope: installation",
+            "env_variable": "CARGO_BUILD_JOBS",
+            "default_value": "",
+            "user_viewable": true,
+            "user_editable": true,
+            "rules": "nullable|integer",
             "field_type": "text"
         }
     ]
--- a/egg-pumpkin.json
+++ b/egg-pumpkin.json
@@ -1,0 +1,82 @@
+{
+    "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
+    "meta": {
+        "version": "PTDL_v2",
+        "update_url": null
+    },
+    "exported_at": "2024-12-27T22:08:25+01:00",
+    "name": "pumpkin",
+    "author": "unknown@unknown.com",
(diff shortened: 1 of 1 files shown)


egg-pumpkin.json
```diff
@@ -0,0 +1,82 @@
...
+        "stop": "stop"
+    },
+    "scripts": {
+        "installation": {
+            "script": "#!\/bin\/ash\r\nset -eux\r\n\r\napk add --no-cache musl-dev git\r\nmkdir -p \/mnt\/server\r\n\r\nREPO_URL=\"${GIT_URL:-https:\/\/github.com\/Snowiiii\/Pumpkin.git}\"\r\n\r\nif [ -n \"$GIT_BRANCH\" ]; then\r\n    git clone --depth 1 --single-branch --branch \"$GIT_BRANCH\" \"$REPO_URL\"\r\nelse\r\n    git clone --depth 1 --single-branch \"$REPO_URL\"\r\nfi\r\n\r\nif [ -n \"$GIT_COMMIT\" ]; then\r\n    git checkout \"$GIT_COMMIT\"\r\nfi\r\n\r\ncd Pumpkin\r\n\r\nif [[ \"$BUILD_RELEASE\" == \"1\" || \"$BUILD_RELEASE\" == \"true\" ]]; then\r\n    cargo build --release\r\n    strip target\/release\/pumpkin\r\n    cp target\/release\/pumpkin \/mnt\/server\/pumpkin\r\nelse\r\n    cargo build\r\n    cp target\/debug\/pumpkin \/mnt\/server\/pumpkin\r\nfi",
+            "container": "docker.io\/rust:1-alpine3.21",
+            "entrypoint": "ash"
+        }
+    },
+    "variables": [
```
Yimura, line 30: Please add a variable to configure the amount of jobs that cargo spawns to compile to final executable.

I've had this process deadlock on my server with too many jobs being spawned by cargo, having lowered the amount of jobs made it go through successfully.

egg-pumpkin.json
```diff
@@ -0,0 +1,82 @@
...
+    "startup": ".\/pumpkin",
+    "config": {
+        "files": "{}",
+        "startup": "{}",
+        "logs": "{}",
+        "stop": "stop"
+    },
+    "scripts": {
+        "installation": {
+            "script": "#!\/bin\/ash\r\nset -eux\r\n\r\napk add --no-cache musl-dev git\r\nmkdir -p \/mnt\/server\r\n\r\nREPO_URL=\"${GIT_URL:-https:\/\/github.com\/Snowiiii\/Pumpkin.git}\"\r\n\r\nif [ -n \"$GIT_BRANCH\" ]; then\r\n    git clone --depth 1 --single-branch --branch \"$GIT_BRANCH\" \"$REPO_URL\"\r\nelse\r\n    git clone --depth 1 --single-branch \"$REPO_URL\"\r\nfi\r\n\r\nif [ -n \"$GIT_COMMIT\" ]; then\r\n    git checkout \"$GIT_COMMIT\"\r\nfi\r\n\r\ncd Pumpkin\r\n\r\nif [[ \"$BUILD_RELEASE\" == \"1\" || \"$BUILD_RELEASE\" == \"true\" ]]; then\r\n    cargo build --release\r\n    strip target\/release\/pumpkin\r\n    cp target\/release\/pumpkin \/mnt\/server\/pumpkin\r\nelse\r\n    cargo build\r\n    cp target\/debug\/pumpkin \/mnt\/server\/pumpkin\r\nfi",
```
Yimura, line 25: This `cargo build` command will build a generic image, since this will most likely be deployed on server with one specific set of hardware we can add additional instructions to optimize for this server's hardware. Through setting the following environment variables:
```bash
RUSTFLAGS="-C target-feature=-crt-static -C target-cpu=native"
```

Yimura (changes requested):
Adding the installation script here to make it readable for everyone:
```bash
#!/bin/ash
set -eux

apk add --no-cache musl-dev git
mkdir -p /mnt/server

REPO_URL="${GIT_URL:-https://github.com/Snowiiii/Pumpkin.git}"

if [ -n "$GIT_BRANCH" ]; then
    git clone --depth 1 --single-branch --branch "$GIT_BRANCH" "$REPO_URL"
else
    git clone --depth 1 --single-branch "$REPO_URL"
fi

if [ -n "$GIT_COMMIT" ]; then
    git checkout "$GIT_COMMIT"
fi

cd Pumpkin

if [[ "$BUILD_RELEASE" == "1" || "$BUILD_RELEASE" == "true" ]]; then
    cargo build --release
    strip target/release/pumpkin
    cp target/release/pumpkin /mnt/server/pumpkin
else
    cargo build
    cp target/debug/pumpkin /mnt/server/pumpkin
fi
```

egg-pumpkin.json
```diff
@@ -0,0 +1,82 @@
...
+    "description": "Pumpkin is a Minecraft server built entirely in Rust, offering a fast, efficient, and customizable experience. It prioritizes performance and player enjoyment while adhering to the core mechanics of the game. Requires around 2GB RAM to build, adjust docker.installer_limits.memory in your wing accordingly! Currently, you need to adjust the port in the config manually!",
+    "features": null,
+    "docker_images": {
+        "Alpine": "ghcr.io\/pterodactyl\/yolks:alpine"
+    },
+    "file_denylist": [],
+    "startup": ".\/pumpkin",
+    "config": {
+        "files": "{}",
+        "startup": "{}",
```
Yimura, line 19: This can be changed to:
```json
        "startup": "{\r\n    \"done\": \"Started Server took\"\r\n}",
````
So Pterodactyl can properly detect startup of the server.

**Comment by Yimura** [comment]
What have you done exactly to test it? I don't see anything specified in regards to that.

Even though docker is supposed to allow for one and the same environment to be replicate able across machines, I've found out that this may not always be the case.

**Comment by Commandcracker** [comment]
> What have you done exactly to test it? I don't see anything specified in regards to that.
> 
> Even though docker is supposed to allow for one and the same environment to be replicate able across machines, I've found out that this may not always be the case.

I ran it on Windows 11 WSL Debian bookworm.

File: egg-pumpkin.json (hunk 2 above)
Comment (Commandcracker) [reply egg-pumpkin.json]: Probably need to tweak the images a bit

Error loading shared library libgcc_s.so.1: No such file or directory (needed by ./pumpkin)
Error relocating ./pumpkin: _Unwind_Resume: symbol not found
Error relocating ./pumpkin: _Unwind_GetTextRelBase: symbol not found
Error relocating ./pumpkin: _Unwind_GetIPInfo: symbol not found
Error relocating ./pumpkin: _Unwind_GetIP: symbol not found
Error relocating ./pumpkin: _Unwind_SetGR: symbol not found
Error relocating ./pumpkin: _Unwind_GetLanguageSpecificData: symbol not found
Error relocating ./pumpkin: _Unwind_Backtrace: symbol not found
Error relocating ./pumpkin: _Unwind_GetRegionStart: symbol not found
Error relocating ./pumpkin: _Unwind_SetIP: symbol not found
Error relocating ./pumpkin: _Unwind_RaiseException: symbol not found
Error relocating ./pumpkin: _Unwind_GetDataRelBase: symbol not found
Error relocating ./pumpkin: _Unwind_DeleteException: symbol not found

Comment (Yimura) [reply egg-pumpkin.json]: Ah yes, I was honestly wondering on if the pumpkin executable could run without libgcc. I guess this answers that.

What we'd need to do is provide our own runtime container with libgcc installed. I had done that in my own repository for a pumpkin egg however I didn't bother building an ARM64 variant. Which the alpine yolk has.

@Snowiiii would you be fine with adding another container image for Pumpkin to run on Pterodactyl?
Probably tagged along the lines of `ghcr.io/snowiiii/pumpkin:yolk`.

File: egg-pumpkin.json [hunk 4]
```diff
@@ -0,0 +1,92 @@
+{
+    "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
+    "meta": {
+        "version": "PTDL_v2",
+        "update_url": null
+    },
+    "exported_at": "2024-12-27T23:40:43+01:00",
+    "name": "pumpkin",
+    "author": "unknown@unknown.com",
```
Snowiiii, line 9: Should this contain my E-Mail ?. When yes you can use lilalexmed@proton.me

Yimura:
Ran into the following issue after setting the `CARGO_BUILD_JOBS` environment value to a value and then emptying out the field.
image

```
error: could not parse ``. Number of parallel jobs should be `default` or a number.
```

You may need to setup better fallbacks with the following scripting syntax:
```bash
# takes CARGO_BUILD_JOBS env variable, if empty assigns "default".
CARGO_BUILD_JOBS=${CARGO_BUILD_JOBS:-default}
```

DataM0del:
> > What have you done exactly to test it? I don't see anything specified in regards to that.
> > Even though docker is supposed to allow for one and the same environment to be replicate able across machines, I've found out that this may not always be the case.
> 
> I ran it on Windows 11 WSL Debian bookworm.

Windows (*dies inside*)

Commandcracker:
> > > What have you done exactly to test it? I don't see anything specified in regards to that.
> > > Even though docker is supposed to allow for one and the same environment to be replicate able across machines, I've found out that this may not always be the case.
> > 
> > 
> > I ran it on Windows 11 WSL Debian bookworm.
> 
> Windows (_dies inside_)

Need to play league of legends and I don't want to clutter my Debian server. I also have a dual boot with fedora, but I have not used it for a while.

Commandcracker:
> > What have you done exactly to test it? I don't see anything specified in regards to that.
> > Even though docker is supposed to allow for one and the same environment to be replicate able across machines, I've found out that this may not always be the case.
> 
> This is in docker's nature, because its repeatable and not reproducible, there is no way around this while using docker, i suggest we at least test it on macos, linux and windows before pushing this into master

We don't need to because Pterodactyl only supports Linux, see https://pterodactyl.io/wings/1.0/installing.html#supported-systems

egg-pumpkin.json
Snowiiii: So as of my understanding, We then have 2 images. One without libgcc, And one with libgcc packaged ?. Isn't there a way to just install libgcc in the script ?

Yimura:
I've tested this on my server and as long as the installation memory set in your Pterodactyl wing to a sufficient high number 4GB is probably best. It'll install just fine from there.

egg-pumpkin.json
Commandcracker: At best 3 images

1. normal pumpkin on alpine
2. yolk based on ghcr.io/pterodactyl/yolks:alpine
3. installer based on ghcr.io/pterodactyl/installers:alpine

(and maybe add a scratch and Debian image later on)

Yimura (approved):
For me this is ready to ship, haven't had the `libgcc` issues that @Commandcracker has experienced.

Only slightly confusing thing is that there's no feedback of the installation process having finished (though the install banner disappears in Pterodactyl).
So perhaps a small echo to mention that the install is finished is in order.

Q2297045667:
Very good

Q2297045667:
The test runs normally after merging into master.
I think ojbk（I think it's absolutely fine.）
link: https://github.com/Q2297045667/PumpkinX/commits/Add-egg-pumpkin/
image

image
 An item was used(SUseItem), but the packet is not implemented yet

Snowiiii:
Looks good, Thank you @Commandcracker :+1:

PR #426, opened 2024-12-27
Added default op level

OfficialKris:
## Description
Added default op level so server admins can set it for test servers.

## Testing

Add `default_op_level=3` under `commands` section in `features.toml`

Can now do everything except stop server!

## Checklist


OfficialKris:
This only adds an option for server administrators to change the default behavior. The configuration.toml remains unchanged.

Snowiiii:
Oh, I tough this was already implemented but it was`op_permission_level` which only specifies the level you get from OP.
Thanks @OfficialKris for implementing the default level, especially useful for the test server :D

Issue #424, opened 2024-12-27
Fix Command permission message


Snowiiii:
### I've searched existing issues and couldn't find a duplicate.

- [X] I confirm this is not a duplicate.

### Operating System

Linux Manajro

### Server Software Version/Commit

a03f554

### What happened?

When you don't have the permission to execute a command you will just get a invalid syntax message instead of a no permission message

### To Reproduce

Execute a command you don't have a high enough permission at

### Expected behavior

A clear message explaining you don't have the permission

PR #428, opened 2024-12-29
fix: command permission handling

smuu:
## Description

This PR introduced the following:

- When registering a new command, one must specify the permission level
- If a player tries to execute a command without having permission, it will show that to the player
- When a player wants to autocomplete a command without having permission, it will not show in the autocomplete

Closes https://github.com/Pumpkin-MC/Pumpkin/issues/424

## Checklist

- [ ] Code is well-formatted and adheres to project style guidelines: `cargo fmt`
- [ ] Code does not produce any clippy warnings: `cargo clippy`


Snowiiii:
Works great, Great work. Thanks for your first Pull request @smuu :rocket:

PR #430, opened 2024-12-29
fix(combat): kick for self-attacking

DataM0del:
## Description

Kicks the player when the entity they're attacking has the same ID as their ID.

## Testing

Use the kick module with self-hurt mode and expect to be kicked.

https://github.com/user-attachments/assets/95f1becc-f18a-44d5-b293-34202b9fa5fd

## Checklist


Snowiiii:
Looks good. Thank you @DataM0del :D

Issue #422, opened 2024-12-27
Setblock not working after breaking block manually


vyPal:
### I've searched existing issues and couldn't find a duplicate.

- [X] I confirm this is not a duplicate.

### Operating System

Arch linux

### Server Software Version/Commit

0.1.0-dev 2a3a6ac-modified

### What happened?

When the /setblock command is executed to set a block at some position, and then the player manually breaks the block, and tries to run the command again with the same position, it no longer sets the block

### To Reproduce

1. run the setblock command
2. break the block that you set
3. run the setblock command again using the same coordinates
4. see the issue

### Expected behavior

Sets the block again

PR #434, opened 2024-12-30
add the missing Acknowledge Block Change packet when the player break…

aure31:
## Description
fix the issue #422 with the setblock by adding the Acknowledge Block Change packet when block is break in creative

## Testing
follow the bug instruction

## Checklist

### DIFF
--- a/pumpkin/src/net/packet/play.rs
+++ b/pumpkin/src/net/packet/play.rs
@@ -723,6 +723,10 @@ impl Player {
                                 .on_broken(block, self, location, server)
                                 .await;
                         }
+
+                        self.client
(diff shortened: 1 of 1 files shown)


Snowiiii:
Actually looking at vanilla this should be always send, I think it would make sense to move below 
https://github.com/Pumpkin-MC/Pumpkin/blob/a1b7f3aca8129342029f142b15bd76a2293e82f9/pumpkin/src/net/packet/play.rs#L698
If i remember correctly the packet was always send, someone probably changed it

aure31:
your right i modified the code

Snowiiii:
Looks good now. Good work. Thank you @aure31 :D :+1:

PR #439, opened 2025-01-01
feat: ignore casing when getting player by name

smuu:
## Checklist


Snowiiii:
I don't think this is really needed, This could be useful since then devs could decide their own if they want to have it case sensitive or not

PR #441, opened 2025-01-01
feat: ArgumentComsumer return result with err

smuu:
## Description

With the ability to return errors with the `ArgumentConsumer`, we can provide proper feedback to the command sender.
For example, when specifying an invalid player, we can provide a proper message instead of sending a message that the command syntax is incorrect.

## Checklist


Snowiiii:
Found out that `/gamemode invalid_gamemode` and an empty player (e.g. `/kill ` does produce and "Internal Error"

smuu:
Thinking about this again, we don't need this (yet)

PR #444, opened 2025-01-02
Pumpkin saving format

suprohub:
I think will be cool, if now we write a simple file format for saving, like a temp solution, because anvil is not working now.
When in future, we can modify and remove java & mojang & skill issues.
Also will be cool make post in docs

## Testing
cargo run
join -> leave -> join
when say "Yaaa yooo its works wooow"


Snowiiii:
Does this actually even work ?. If you modify the terrain and restart the Server will the new placed blocks be saved ?. I also wonder how we should get the exact chunk from the region file which contains multiple chunks in one file because there is no location table. I also already can guess that these files will be really large since there is no compression (or any optimization at all)

I also want to mention that making our own "File Format" or a "Protocol" or any kind of thing you or others want to use in the future, you should write an **good** Specification before doing that.

I'm sorry to say but this just seems really error prone and i'm sure this will have many issues, Lets just wait until the anvil saving thing will be ready, It will be one day. It would be more useful if you would try to help with that than making an own Format

suprohub:
SPecificatiob
Its works not correctly now, because its depends on palette system from anvil, but i fix it (i try create a normal system with using bytes instead of longs).  Also i not want to use nbt.

So how it works: i want no use region files, and rely on os filesystem, because file systems are designed to handle large numbers of files. So we can just get chunk by name.
Also separating chunks by what can increase performance in theory, because we not need to compress & decompress large data every second what player places block or other things.

Compression

Compression is difficult, its need talks. We need to use many different techniques, like palette or rle (RleVec), and on output use standart compression like lz4. For regions optimization, we can create a settings what defines a region size (like 2 chunks, 4 chunks, etc..), because some people wants a wery big performance, and other wants a big compression ratio. For region compression we can use zip.

suprohub:
also, user can set custom filesystem to increase performace or stability

suprohub:
https://voxel.wiki/wiki/palette-compression/

suprohub:
https://docs.rs/bitvec/latest/bitvec/

bitvec enables you to write code for bit-addressed memory that is simple, easy, and fast. It compiles to the same, or even better, object code than you would get from writing shift/mask instructions manually. It leverages Rust’s powerful reference and type systems to create a system that seamlessly bridges single-bit addressing, precise control of in-memory layout, and Rust-native ownership and borrowing mechanisms.

suprohub:
@Snowiiii

suprohub:
saving works very good

suprohub:
btw, this is isze
изображение

samolego:
Endcrystalme has spent quite some time on developing alternative file format to anvil, https://github.com/xymb-endcrystalme/LinearRegionFileFormatTools

Might be worth looking into

Snowiiii:
Great Anvil is merged and i will close this now. 

So first you have to understand that the file-system is really slow, It also involves multiple [System Calls](https://en.wikipedia.org/wiki/System_call) which produce overhead. There will be also [File metadata](https://en.wikipedia.org/wiki/File_system#Metadata) created for every file you create, things like

- the file name
- [size](https://en.wikipedia.org/wiki/File_size) which may be stored as the number of blocks allocated or as a [byte](https://en.wikipedia.org/wiki/Byte) count
- [when](https://en.wikipedia.org/wiki/System_time) created, last accessed, last backed-up
    owner [user](https://en.wikipedia.org/wiki/User_ID) and [group](https://en.wikipedia.org/wiki/Group_ID)
- [access permissions](https://en.wikipedia.org/wiki/File_system_permissions)
- [file attributes](https://en.wikipedia.org/wiki/File_attribute) such as whether the file is read-only, [executable](https://en.wikipedia.org/wiki/Executable), etc.
 - [device type](https://en.wikipedia.org/wiki/Device_file) (e.g. [block](https://en.wikipedia.org/wiki/Block_devices), - [character](https://en.wikipedia.org/wiki/Character_devices), [socket](https://en.wikipedia.org/wiki/Internet_socket), [subdirectory](https://en.wikipedia.org/wiki/Subdirectory), etc.)

You don't want to have a bunch of duplicated meta data for all the chunks, Instead you can just use one larger file.
The Region format may seems Complex but it often makes sense, But like yourself said you don't want to look into it and just call everything an "skill-issue"

I also don't like the way your documentation is structured, Making a new file-format is something serious. You usually create a "more professional" PDF or markdown file, Including a table of data types and things like version identification numbers.

This format is also **very** incomplete, It only saves blocks. No entities. No light data... Yes ofc you can then later add more things but this would require to Update the file format and documentation and potentially break things others have implemented

There is also no vanilla support ofc, and may tools (e.g. Axiom) will become useless atleast when they are not decide to implement our own Format (Which is incomplete and will soon break then :wink:)

I'm sorry @suprohub i can see you put a good amount of work into this PR, but i often wish there would be a better communication. Like yourself said this should be an "temp" solution until Anvil works, Here we are
> like a temp solution, because anvil is not working now.

suprohub:
> Great Anvil is merged and i will close this now. 
> 
> So first you have to understand that the file-system is really slow, It also involves multiple [System Calls](https://en.wikipedia.org/wiki/System_call) which produce overhead. There will be also [File metadata](https://en.wikipedia.org/wiki/File_system#Metadata) created for every file you create, things like
> 
> - the file name
> - [size](https://en.wikipedia.org/wiki/File_size) which may be stored as the number of blocks allocated or as a [byte](https://en.wikipedia.org/wiki/Byte) count
> - [when](https://en.wikipedia.org/wiki/System_time) created, last accessed, last backed-up
>     owner [user](https://en.wikipedia.org/wiki/User_ID) and [group](https://en.wikipedia.org/wiki/Group_ID)
> - [access permissions](https://en.wikipedia.org/wiki/File_system_permissions)
> - [file attributes](https://en.wikipedia.org/wiki/File_attribute) such as whether the file is read-only, [executable](https://en.wikipedia.org/wiki/Executable), etc.
>  - [device type](https://en.wikipedia.org/wiki/Device_file) (e.g. [block](https://en.wikipedia.org/wiki/Block_devices), - [character](https://en.wikipedia.org/wiki/Character_devices), [socket](https://en.wikipedia.org/wiki/Internet_socket), [subdirectory](https://en.wikipedia.org/wiki/Subdirectory), etc.)
> 
> You don't want to have a bunch of duplicated meta data for all the chunks, Instead you can just use one larger file.
> The Region format may seems Complex but it often makes sense, But like yourself said you don't want to look into it and just call everything an "skill-issue"
> 
> I also don't like the way your documentation is structured, Making a new file-format is something serious. You usually create a "more professional" PDF or markdown file, Including a table of data types and things like version identification numbers.
> 
> This format is also **very** incomplete, It only saves blocks. No entities. No light data... Yes ofc you can then later add more things but this would require to Update the file format and documentation and potentially break things others have implemented
> 
> There is also no vanilla support ofc, and may tools (e.g. Axiom) will become useless atleast when they are not decide to implement our own Format (Which is incomplete and will soon break then :wink:)
> 
> I'm sorry @suprohub i can see you put a good amount of work into this PR, but i often wish there would be a better communication. Like yourself said this should be an "temp" solution until Anvil works, Here we are
> > like a temp solution, because anvil is not working now.

Yeah i know

suprohub:
Oke

PR #446, opened 2025-01-02
feat: add whitelist command

smuu:
## Description

- This PR adds the `whitelist `command and the corresponding `whitelist.json` file to keep track of the whitelisted users.

## Testing

### DIFF
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -85,14 +85,15 @@ fn init_logger() {
             logger = logger.without_timestamps();
         }
 
-        if ADVANCED_CONFIG.logging.env {
-            logger = logger.env();
-        }
-
         logger = logger.with_level(convert_logger_filter(ADVANCED_CONFIG.logging.level));
 
         logger = logger.with_colors(ADVANCED_CONFIG.logging.color);
         logger = logger.with_threads(ADVANCED_CONFIG.logging.threads);
+
+        if ADVANCED_CONFIG.logging.env {
+            logger = logger.env();
+        }
+
         logger.init().unwrap();
     }
 }


pumpkin-config/src/lib.rs
```diff
@@ -101,6 +103,10 @@ pub struct BasicConfiguration {
     pub use_favicon: bool,
     /// Path to server favicon
     pub favicon_path: String,
+    /// Whether to enable the whitelist
+    pub white_list: bool,
+    /// Whether to enforce the whitelist
+    pub enforce_whitelist: bool,
```
Snowiiii, line 109: One boolean would be enough, If the white list is enabled it will be automatically enforced

pumpkin-config/src/player_profile.rs
```diff
@@ -0,0 +1,14 @@
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+#[derive(Serialize, Deserialize, Clone, Default, Debug)]
+pub struct PlayerProfile {
```
Snowiiii, line 5: Why this is here in `pumpkin-config`?

pumpkin/src/main.rs
```diff
@@ -233,6 +235,21 @@ async fn main() {
...
+                        .whitelist
+                        .iter()
+                        .any(|p| p.uuid == player.gameprofile.id);
+                    let is_op = player.permission_lvl.load() >= PermissionLvl::Three;
+                    if !is_whitelisted && !is_op {
+                        player
+                            .kick(TextComponent::text("You are not whitelisted!"))
+                            .await;
+                    }
+                }
```
Snowiiii, line 252: Lets not clutter our main (even more), A new method in the `Server` which gets called in `Server::add_player` would be probably ideal

pumpkin-config/src/player_profile.rs
smuu: I placed it where the op files were and moved both to the data crate.

pumpkin/src/main.rs
smuu: hat makes sense. Moved it

pumpkin-config/src/lib.rs
smuu: Both variables are used in the vanilla `server.properties` file. You can enable the whitelist but not enforce it. If you enforce it, players will be kicked when you reload the whitelist if they are not whitelisted.
From the Minecraft wiki:

**white-list**
> Whether the whitelist is enabled.
> 
> With a whitelist enabled, users not on the whitelist cannot connect. Intended for private servers, such as those for real-life friends or strangers carefully selected via an application process, for example.
> 
>     false - No white list is used.
>     true - The file whitelist.json is used to generate the white list.
> 
> Note: Ops are automatically whitelisted, and there is no need to add them to the whitelist. 

**enforce-whitelist**
> Whether to enforce changes to the whitelist.
> 
> When this option as well as the whitelist is enabled, players not present on the whitelist get kicked from the server after the server reloads the whitelist file.
> 
>     false - Online players that are not on the whitelist are not kicked.
>     true - Online players that are not on the whitelist are kicked.

Issue #197, opened 2024-10-27
Slabs dont stack


Z-fentom:
- [x] I've searched existing issues and couldn't find a duplicate.

**Describe the bug**
Upon placing 2 slabs on top of each other they don't stack it appears so for a split second as they do but they don't

**To Reproduce**
1.Take any variant of slabs
2.Place 1 and place another 1 on top

**Expected behavior**
Slabs are expected to stack

**Environment (please complete the following information):**

- Operating System: Arch linux
- Server Software Version/Commit: [e.g., 1.0.0/39b4cb3]

**Additional context**

Snowiiii:
Fixed in https://github.com/Pumpkin-MC/Pumpkin/commit/6c4eb88dd3bae7c4989bdbd77b676fadb632a110

PR #447, opened 2025-01-02
Add properties support [WIP]

Bafran:
## Description
Issue #197 

Add support for block properties. Uses a hashmap on server startup to build a property mapping.

## Testing
image

## Checklist

- [ ] Code is well-formatted and adheres to project style guidelines: `cargo fmt`
- [ ] Code does not produce any clippy warnings: `cargo clippy`

### DIFF
--- a/pumpkin-world/src/block/block_registry.rs
+++ b/pumpkin-world/src/block/block_registry.rs
@@ -101,11 +101,10 @@ pub struct Block {
     pub default_state_id: u16,
     pub states: Vec<State>,
 }
-#[expect(dead_code)]
 #[derive(Deserialize, Clone, Debug)]
 pub struct Property {
-    name: String,
-    values: Vec<String>,
+    pub name: String,
+    pub values: Vec<String>,
 }
 #[derive(Deserialize, Clone, Debug)]
 pub struct State {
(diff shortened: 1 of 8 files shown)


OfficialKris:
Works good. Only issue I have is to add an additional slab block, you have to click on a placed block (slab). I'm sure the client sends the position along the block the player clicked, so you can do a comparison with that. Also, If you use the offset math function in WorldPosition, you can check if the block is a slab and respond accordingly.

Q2297045667:
Perhaps we could add a vertical step?This is definitely a groundbreaking development😂

Bafran:
Hey guys, sorry I haven't made progress on this, just started a new job so been pretty busy. @KairuDeibisu if you're planning to build out those other features please use this branch if it helps or as reference. The core function is implemented here but could definitely use clean up. Let me know if you have questions about my method

Snowiiii:
This was Now merged in https://github.com/Pumpkin-MC/Pumpkin/commit/6c4eb88dd3bae7c4989bdbd77b676fadb632a110

Snowiiii:
Thank you again @Bafran. Great work :+1:

PR #451, opened 2025-01-04
feat: add whitelist feature

smuu:
## Description

I accidentally closed the other PR. Recreated it here.

- This PR adds the `whitelist `command and the corresponding `whitelist.json` file to keep track of the whitelisted users.

## Testing


pumpkin-config/src/lib.rs
```diff
@@ -101,6 +101,10 @@ pub struct BasicConfiguration {
     pub use_favicon: bool,
     /// Path to server favicon
     pub favicon_path: String,
+    /// Whether to enable the whitelist
+    pub white_list: bool,
+    /// Whether to enforce the whitelist
+    pub enforce_whitelist: bool,
```
smuu, line 107: Both variables are used in the vanilla `server.properties` file. You can enable the whitelist but not enforce it. If you enforce it, players will be kicked when you reload the whitelist if they are not whitelisted.
From the Minecraft wiki:

**white-list**
> Whether the whitelist is enabled.
> 
> With a whitelist enabled, users not on the whitelist cannot connect. Intended for private servers, such as those for real-life friends or strangers carefully selected via an application process, for example.
> 
>     false - No white list is used.
>     true - The file whitelist.json is used to generate the white list.
> 
> Note: Ops are automatically whitelisted, and there is no need to add them to the whitelist. 

**enforce-whitelist**
> Whether to enforce changes to the whitelist.
> 
> When this option as well as the whitelist is enabled, players not present on the whitelist get kicked from the server after the server reloads the whitelist file.
> 
>     false - Online players that are not on the whitelist are not kicked.
>     true - Online players that are not on the whitelist are kicked.

Snowiiii:
You can always reopen PR's, I would prefer this next time instead of making a new one

smuu:
> You can always reopen PR's, I would prefer this next time instead of making a new one

I tried. First, it said I can only do that when adding a new commit. Then I added a new commit but could still not reopen the PR. I'm unsure if this was a permission issue or if I missed something.

pumpkin-config/src/lib.rs
Snowiiii: Mhh okay, I mean, I'm fine with leaving it then

Snowiiii:
Okay @smuu can you may fix Ci and conflicts

Q2297045667:
Is there any progress?

Snowiiii:
Feel free to reopen

Issue #438, opened 2025-01-01
Keep track of offline users that joined previously into a word


smuu:
### I've searched existing issues and couldn't find a duplicate.

- [X] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

Currently, you can only address online players. For example, I can only add a player as a server operator when the player is online.

### Describe the solution you'd like

Be able to retrieve an offline player object like in craftbukkit-based servers. (`Bukkit.getOfflinePlayer`)

PR #453, opened 2025-01-04
feat: allow using offline player

smuu:
## Description

- You will be able to address an offline player (for example, for `op` and `whitelist`)
- With `player.is_online()`, you can check if a fetched player is online
- If the player is not online, fields that only an online player has, is an empty `Option`
- Main file to review is: `pumpkin/src/entity/player.rs`
- I was considering different approaches but ended up with this one
- Happy to discuss the design I decided on

Closes https://github.com/Pumpkin-MC/Pumpkin/issues/438

## Testing


Snowiiii:
Instead of making everything Optional you probably want to create a new OfflinePlayer struct

Snowiiii:
Feel free to reopen

PR #455, opened 2025-01-05
fix: Usage of TextContent::Translate resulted in a bad packet

jonasborsch:
## Description
_As a heads up, I've never developed in rust before, so please give me the hardest feedback you can._

The usage of `TextContent::Translate` resulted in a bad packet. This was due to the implementation causing the `TextComponent`'s from the `with` vector to be serialized into a byte array. I would guess, that for the same reason the temporary struct was used.

I tried different solutions but the only viable for me seemed to be a wrapper. With that the whole temporary struct could be removed. The `TextComponent` now handles JSON-serialization and the `TextComponentNbt` is the NBT binary representation.

## Testing

Now we can edit the `assets/synced_registries.json` file and we see that the `/say` command uses the correct translation.
image
<br />
image


jonasborsch:
Feel free to review, but still a draft because there are still two things open:
1. Messages from the `/say` command do not appear in the console
2. Need to as tests for the TextComponent
3. My format is probably shit

pumpkin-core/src/text/mod.rs
```diff
@@ -30,6 +30,38 @@ pub struct TextComponent {
     pub extra: Vec<TextComponent>,
 }
 
+pub struct TextComponentNbt {
+    pub wrappee: TextComponent,
+}
```
Snowiiii, line 35: I think you can just make this an inline struct `pub struct TextComponentNbt(pub TextComponent)`

pumpkin-core/Cargo.toml
```diff
@@ -13,3 +13,4 @@ num-traits = "0.2"
 
 colored = "2"
 md5 = "0.7.0"
+serde_json.workspace = true
```
Snowiiii, line 16: Please put this at the top

pumpkin-protocol/src/client/play/c_player_chat_message.rs
```diff
@@ -49,11 +49,11 @@ impl<'a> CPlayerChatMessage<'a> {
             salt,
             previous_messages_count: previous_messages.len().into(),
             previous_messages,
-            unsigned_content,
+            unsigned_content: unsigned_content.map(TextComponent::to_nbt),
             filter_type,
             chat_type,
-            sender_name,
-            target_name,
+            sender_name: sender_name.to_nbt(),
```
Snowiiii, line 55: I'm thinking of may implementing something as `AsRef`

Snowiiii:
Really impressive for someone who just Started with Rust :+1:. Your approche is similar to [valence](https://github.com/valence-rs/valence/blob/main/crates/valence_text/src/lib.rs) (which is not a bad thing). One thing you have to fix is the CI, In particulier some clippy checks are failing, make sure to fix every warning produced when running `cargo clippy --all-targets`

Overall good work, Thank you @jonasborsch

pumpkin-protocol/src/client/play/c_player_chat_message.rs
jonasborsch: I'll take a look into that to see if it's possible

jonasborsch:
> Really impressive for someone who just Started with Rust 👍. Your approche is similar to [valence](https://github.com/valence-rs/valence/blob/main/crates/valence_text/src/lib.rs) (which is not a bad thing). One thing you have to fix is the CI, In particulier some clippy checks are failing, make sure to fix every warning produced when running `cargo clippy --all-targets`
> 
> Overall good work, Thank you @jonasborsch

Thanks for the feedback! If I knew there was already an implementation of that I could've saved couple hours of my time. 🥲
The implementations do seem to differ though, because valence uses `with: Vec<Text>`, but the [wiki](https://minecraft.fandom.com/wiki/Raw_JSON_text_format#Translated_Text) states that any text component can be used.

Anyway, I'll work on the fixes and then mark the PR as ready for review

pumpkin-core/src/text/mod.rs
jonasborsch: Done! Also stole the name from valence, `Text` sounded better then `TextComponentNbt` or something like `TextComponentWrapper`

pumpkin-core/Cargo.toml
jonasborsch: Fixed

pumpkin-protocol/src/client/play/c_player_chat_message.rs
jonasborsch: Was researching and at first sight `Into` (or actually `From`) seemed to be a better trait, open for other opinions

Snowiiii:
All files in the assets are coming from our Extractor. Can you may modify the extractor so it produces the latest en-US.json

jonasborsch:
> All files in the assets are coming from our Extractor. Can you may modify the extractor so it produces the latest en-US.json

That‘s definitely is more convenient. I will work on it in the next couple of days. Thanks for the tip

jonasborsch:
> All files in the assets are coming from our Extractor. Can you may modify the extractor so it produces the latest en-US.json

See https://github.com/Pumpkin-MC/Extractor/pull/2

Snowiiii:
@jonasborsch Just added the translations my self

Snowiiii:
So @jonasborsch @DarkMrMelther just fixed the issue. But still thank you for your time. Your code is really code for the first time of coding with rust 👍

PR #457, opened 2025-01-05
Made getting players by name case-insensitive

RealVaprant:
## Description

Made the player name and the parameter lower case when matching them. This way, things like player arguments in commands don't need to be case-sensitive. Overall a revolutionary change.

## Testing

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -626,7 +626,7 @@ impl World {
     /// Gets a Player by username
     pub async fn get_player_by_name(&self, name: &str) -> Option<Arc<Player>> {
         for player in self.current_players.lock().await.values() {
-            if player.gameprofile.name == name {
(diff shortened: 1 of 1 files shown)


Snowiiii:
I'm still not sure about this, There was an PR for this recently https://github.com/Pumpkin-MC/Pumpkin/pull/439. But i'm thinking about plugin devs may wanting to have the case checked

RealVaprant:
wdym? not sure about making player arguments case insensitive in commands? or altering the function used to get players?

I believe the player arguments in commands should be case-insensitive one way or another, as it is in Vanilla, otherwise, it would be an annoying inconvenience.

Snowiiii:
I mean for arguments it's probably fine, just was thinking that someone many wants to match with case in the future. anyway. Thank you @RealVaprant 👍

RealVaprant:
ig on the next update you could just make a new function for not ignoring cases. Though I think for now, ignoring cases would be better for default.

PR #463, opened 2025-01-06
Add a proper error message for a non-existent command

NGSpace:
## Description

Add a more descriptive message when a player sends a non-existent command in chat.

Message format: "Command {command} does not exist"


Snowiiii:
Thank you @NGSpace :+1:

PR #466, opened 2025-01-07
feat: support receiving player_loaded packet

MasterCash:
## Description
server bound packet id 43 (`player_loaded`) was not being handled. now the player object tracks if the client has reported that it has loaded

This adds two new fields to the `Player` struct:
client_loaded:         indicates that the client for this player has reported it has loaded its initial chunks.
client_loaded_timeout: indicates (in ticks) how long the client is given to report it has loaded before it starts treating it as if it has loaded. This seems to mostly affect listening to a few of the client's packets.

There is one place in the source code I couldn't figure out where it would match up in the codebase but I think that is due to it not being implemented yet. It is in an invulnerability check. Part of the check makes sure the client is loaded if it isn't then it is invulnerable.

## Testing
The console no longer warns about unhandled packet id 43. player movement and actions still work, debugged to ensure packet was being sent and received. 


Snowiiii:
Looks good, i think we may should do something about that many checks if the player is loaded, but overall this is good. Thank you @MasterCash

PR #467, opened 2025-01-08
Consider using rle vec instead of simple array of blocks

suprohub:
## Description
What?
So, after adding a subchunks will be cool to add a first alghoritm for optimization:
Run len encoding.
Its converts data like aabbbbbxxxs to a2b5x3s1

## Testing
`cargo run`
see what all works, check mem usage
`git checkout -b main`
see what without rle main eats more mem

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,1 +1,0 @@
-blank_issues_enabled: false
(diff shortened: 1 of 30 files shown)


Snowiiii:
I actually think that other compression algorithms are may better to use here than RLE. I think lz4 would be an better option and is also already included in `pumpkin-world`

PR #468, opened 2025-01-08
Fix misspelled word "probably"

dotnetian:
## Description
The word "probably" was mispronounced in a couple of places in the codebase, and I fixed them.

## Testing
The logic of the code is unchanged, only a typo was fixed.


Snowiiii:
Looks good. Thank you @dotnetian

PR #470, opened 2025-01-10
Separate chunk saving

suprohub:
## Description
WHat??? what the separation? u sily?
..

No.
My idea its separate chunk reading and writing. I want make 2 steps.

First we serialize a chunk in bytes. Its can be a simple vanilla anvil serialization.
Second we need write chunk to disk.
But how? Mojang for java now uses a location table. its a simple approach and not very good.

Will be cool add a db like in [bedrock](https://github.com/Mojang/leveldb), or some other solutions. And also will be cool to use custom more optimized format.

But what, if server owner wants very fast chunk saving and not prioritizes space on disk what chunks are consuming? Or if owner wants another solution - dont waste disk space and have slow chunk saving? Or he wants a better vanilla compalibitle but want fast chunk access?

So, for all these questions i make solution - separated chunks.
Now its looks like this (example) :

db:
 - InformativeTable (Location + Timestamp)
 - RocksDB
 - SurrealDB
 - LevelDB
 - iWantUseMultipleFiles

format:
 - Anvil
 - MCRegion
 - Pumpkin
 - StupidSolutionsInc

Also its not a final version, and i want to discuss with us. And yeah, plugins can also make impls simplier, because for plugins dont need to be recreate location table functions.

## Testing
`cargo run`
see inital works, see errors in console
see whats all savings and chunks working


Snowiiii:
1. This is called abstraction, using that term is more clear than "Separate chunk saving".
2. The `ChunkStatus` enum is gone from `pumpkin-world` 

I don't link the abstraction you made, You just put in 4 different types into `level` now

pumpkin-world/src/chunk/format/anvil.rs
```diff
@@ -0,0 +1,372 @@
...
+    #[serde(rename = "zPos")]
+    z_pos: i32,
+    status: ChunkStatus,
+    #[serde(rename = "sections")]
+    sections: Vec<ChunkSection>,
+    heightmaps: ChunkHeightmaps,
+}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+enum ChunkStatus {
```
Snowiiii, line 61: This needs to be gone since we parse that automaticlly

suprohub: Ok, remove tomorrow

suprohub: Pr little outdated

pumpkin-world/src/chunk/format/anvil.rs
```diff
@@ -0,0 +1,344 @@
...
+            status: ChunkStatus::Full,
+            heightmaps: chunk_data.heightmap.clone(),
+            sections,
+        };
+
+        fastnbt::to_bytes(&nbt).map_err(|e| ChunkWritingError::ChunkSerializingError(e.to_string()))
+    }
+}
+
+/*#[cfg(test)]
```
Snowiiii, line 266: Why you commented out the test

suprohub: Idk.
Maybe because i waited opinions, amd forgor fix test

suprohub: You can uncomment it. Or i can fix it when can

Snowiiii:
Looks like you just copy pasted the anvil implementation into InformativeTable. Also we should use Anvil as default

suprohub:
> Looks like you just copy pasted the anvil implementation into InformativeTable. Also we should use Anvil as default

Yeah, but like.. How i can name it?? "Anvil raw saving implementation?"
I named it with short name for simplicity.

suprohub:
And also we use anvil by default

pumpkin-world/src/level.rs
```diff
@@ -77,8 +78,8 @@ impl Level {
             world_gen,
             world_info_writer: Arc::new(AnvilLevelInfo),
             level_folder,
-            chunk_reader: Arc::new(AnvilChunkFormat),
-            chunk_writer: Arc::new(AnvilChunkFormat),
+            chunk_format: Arc::new(AnvilChunkFormat),
+            chunk_storage: Arc::new(InformativeTable),
```
Snowiiii, line 82: You changed default

pumpkin-world/src/chunk/mod.rs
```diff
@@ -1,79 +1,23 @@
...
-};
+use crate::{coordinates::ChunkRelativeBlockCoordinates, WORLD_HEIGHT};
 
-pub mod anvil;
+pub mod compression;
+pub mod db;
+pub mod format;
+
+// 1.21.4
+const WORLD_DATA_VERSION: i32 = 4189;
```
Snowiiii, line 14: Why this is here ?, This is anvil specific

suprohub: Ok, i move

pumpkin-world/src/level.rs
suprohub: But informative table its default saver

suprohub: Ah, i think maybe i can name AnvilSaver?

Snowiiii:
So i don't like this implementation tbh, And the current Chunk system will also change in the future, @Mili-ssm wants to add a new Abstraction so this becomes unecessary

PR #471, opened 2025-01-11
Consider using blockstates instead of simple blocks

suprohub:
## Description
i maked blockstates working in chunks, but needs fix anvil loading
## Testing
`cargo run`
see what all works, without anvil loading

### DIFF
--- a/pumpkin-protocol/src/client/play/c_chunk_data.rs
+++ b/pumpkin-protocol/src/client/play/c_chunk_data.rs
@@ -58,7 +58,7 @@ impl ClientPacket for CChunkData<'_> {
 
                     palette.iter().for_each(|id| {
                         // Palette
-                        data_buf.put_var_int(&VarInt(*id as i32));
+                        data_buf.put_var_int(&VarInt(*id));
                     });
                     // Data array length
                     let data_array_len = subchunk.len().div_ceil(64 / block_size as usize);
--- a/pumpkin-world/src/block/block_state.rs
+++ b/pumpkin-world/src/block/block_state.rs
@@ -61,12 +61,12 @@ impl BlockState {
 
     #[inline]
     pub const fn as_u32(&self) -> u32 {
-        ((self.block_id as u32) << 16) | (self.state_id as u32)
+        (self.block_id as u32) | (self.state_id as u32)
     }
 
     #[inline]
     pub const fn as_i32(&self) -> i32 {
-        ((self.block_id as i32) << 16) | (self.state_id as i32)
+        (self.block_id as i32) | (self.state_id as i32)
     }
 }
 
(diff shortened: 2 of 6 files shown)


kralverde:
State ids are all unique, so I don't know what you are doing bit shifting the block id in there. That is useless information.

kralverde:
It seems you're misunderstanding something because multi block and multi block state are the same thing conceptually. The u16s are the state ids which are unique across all blocks

kralverde:
From_block and from_block_props are already implemented somewhere I believe. The block macro?

kralverde:
I don't like the to i32 and u32 implementations, if anything, it should only be the state_id

suprohub:
> From_block and from_block_props are already implemented somewhere I believe. The block macro?

I didnt found any block props creator

suprohub:
And from block too, so if you find make me know

kralverde:
> > From_block and from_block_props are already implemented somewhere I believe. The block macro?
> 
> I didnt found any block props creator

Looks like it was removed when the ill-fated runtime migration happened, but it should be implemented as a part of: https://github.com/Pumpkin-MC/Pumpkin/blob/master/pumpkin-macros/src/block_state.rs#L19

suprohub:
> It seems you're misunderstanding something because multi block and multi block state are the same thing conceptually. The u16s are the state ids which are unique across all blocks

No, blockstate structure its u32, while simple block its u16

suprohub:
Or you mean what its u16 its a block state?

suprohub:
If that, why anywhere whe uses block id, instead of state id?

suprohub:
Its a different things

kralverde:
> > It seems you're misunderstanding something because multi block and multi block state are the same thing conceptually. The u16s are the state ids which are unique across all blocks
> 
> No, blockstate structure its u32, while simple block its u16

That's because that's how we define a var int. We use a u16 simply because there is less that 65535 unique block states and we can save memory

kralverde:
> If that, why anywhere whe uses block id, instead of state id?

This is just for a quick check if two states are of the same block. It is not used in chunk serialization

suprohub:
Ok, but why all comments and vars says what its a block id not a state?

suprohub:
> > If that, why anywhere whe uses block id, instead of state id?
> 
> This is just for a quick check if two states are of the same block. It is not used in chunk serialization

Can you more explain?

kralverde:
It should be "block state id" I'm pretty sure the comments were made before all this complexity came in

suprohub:
> It should be "block state id" I'm pretty sure the comments were made before all this complexity came in

No, before this its also be block id.
Its confusing

PR #484, opened 2025-01-20
Open block GUI on empty hand interaction

yarml:
## Description
Trying to open the crafting table/furnace/etc before with an empty hand was not possible.

Slightly change the logic in `handle_use_item_on` (which probably should be renamed since the packet is sent on empty hands as well), so that it calls `PumpkinBlock::on_use` (previously never called).

Introduce a new `BlockPlacingError`: `NoBaseBlock`, since the use_item_on packet should (usually) have a base block.

## Testing
Not sure how I can do that tbh, but existing tests still run.

### DIFF
--- a/pumpkin/src/net/packet/play.rs
+++ b/pumpkin/src/net/packet/play.rs
@@ -57,6 +57,7 @@ pub enum BlockPlacingError {
     InvalidBlockFace,
     BlockOutOfWorld,
     InventoryInvalid,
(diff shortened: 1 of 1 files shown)


Q2297045667:
Very good
After testing,he indeed effectively fixed the bug.
image

Snowiiii:
Works great. Thank you @yarml :D :+1:

PR #485, opened 2025-01-20
Adds support for slabs and stairs

KairuDeibisu:
This merge adds support for slabs and stairs.

The framework was made by @Bafran and expanded to support stairs.

Levers can be added in a latter pull request. For now, I want to get in the basics.

https://github.com/user-attachments/assets/b128ebb0-839f-46b6-b428-a9a9b8068800


Snowiiii:
So whats next @KairuDeibisu. Please use a better PR description than this

Snowiiii:
I wonder if this supports properties which are loaded directly from anvil, and also the other way. So is this saved properly ?

pumpkin-world/src/entity/mod.rs
```diff
@@ -1 +1,9 @@
 pub mod entity_registry;
+
+#[derive(Debug, PartialEq, Clone, Copy)]
+pub enum FacingDirection {
```
Snowiiii, line 4: This is already defined in `pumpkin-world/block/mod`
as `BlockFace`. We may want to rename that to Direction in place it in `pumpkin-util/math`

Snowiiii:
Overall this works fine already. So i think its worth merging and will help #490 and closes #447

Snowiiii:
Thank you @KairuDeibisu and @Bafran

PR #486, opened 2025-01-20
Fixes our NBT crate

### LINKED ISSUE #420: Completely replace `fastnbt` with our own NBT crate

Our own NBT crate currently does work great when serializing values into NBT, But deserialzing seems not to really work

Fix our NBT crate and finally remove `fastnbt`. This should also finally allow https://github.com/Snowiiii/Pumpkin/pull/283 to be merged

asodugf12:
## Description
Replaces the code that consumes the name, because it only consumes 1 byte instead of the full 2 byte length and the name itself. This causes the code to split the string at a position way bigger than the buffer itself could hold.

The new code instead skips the bytes that store the name, however this could be simply changed to return the name if it is needed.

Also updates our serializer to correctly serialize NBT Lists

Fixes #420 
(Partially)


Snowiiii:
Oh very nice, I wonder if you can may replace fastnbt now with our nbt crate ?

Snowiiii:
Fixed in the latest commit

PR #487, opened 2025-01-20
Placement restrictions

DarkMrMelther:
## Description
Added restrictions and errors for when a player tries to place blocks above or below the world or while in the wrong gamemode.

Note: In order to send the maximum height message, it was necessary to modify the type of the variable “with” of TextComponent::Translate, because it crashes the client.
(Theory: The client crashes due to a circular encoding of the “with” value).

## Testing

- [x] Manual test ;) 

### DIFF
--- a/pumpkin-util/src/text/mod.rs
+++ b/pumpkin-util/src/text/mod.rs
@@ -42,7 +42,7 @@ impl TextComponent {
         }
     }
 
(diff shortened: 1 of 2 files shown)


Snowiiii:
Great job!. Thank you @DarkMrMelther. Very happy translate works now. I will try to translate everything now

PR #488, opened 2025-01-21
Add Mobs to Login Sequence

OfficialKris:
## Description
Added mobs to login sequence

## Testing
- Login
- Spawn pig, cow, and zombie
- Logout and login again
- See mobs in same position

For zombies, the AI is correct but there is an issue with players not being removed correctly so the zombie follows the old player. This is a separate issue.


pumpkin/src/entity/mob/mod.rs
```diff
@@ -60,4 +61,24 @@ impl MobEntity {
     pub async fn goal<T: Goal + 'static>(&self, goal: T) {
         self.goals.lock().await.push((Arc::new(goal), false));
     }
+
+    pub fn create_spawn_entity_packet(&self, uuid: Uuid) -> CSpawnEntity {
+        let e = &self.living_entity.entity;
+        let y = e.pos.load();
```
Snowiiii, line 67: why y ?

pumpkin/src/entity/mob/mod.rs
```diff
@@ -60,4 +61,24 @@ impl MobEntity {
...
+            y.x,
+            y.y,
+            y.z,
+            e.pitch.load(),
+            e.yaw.load(),
+            e.yaw.load(), // todo: head_yaw and yaw are swapped, find out why
+            0.into(),
+            0.0,
+            0.0,
+            0.0,
```
Snowiiii, line 81: why don't send the velocity ?

pumpkin/src/entity/player.rs
```diff
@@ -523,6 +523,17 @@ impl Player {
...
+    /// Sends the mobs to just the player.
+    // TODO: This should be optimized for larger servers based on current player chunk
+    pub async fn send_mobs(&self, world: &World) {
+        let mobs = world.current_living_mobs.lock().await.clone();
+        for (uuid, mob) in mobs {
+            self.client
+                .send_packet(&mob.create_spawn_entity_packet(uuid))
+                .await;
+        }
+    }
```
Snowiiii, line 535: this should be in World

OfficialKris: I don't think this should be in world as the code stands now. It only sends the mobs to a single player. Maybe it could be refactored?

Snowiiii:
Just tested and it works great!. Thank you @OfficialKris :+1:

PR #490, opened 2025-01-22
Dustoff Whiteout314 Sign PR

OfficialKris:
## Description
Takes #262 and re-implements based on #333 interactive block PR. Most code is the same with various refactors.

## Testing
- Place any type of sign
- Write text, submit

TODO:
- [ ] Refactor as an "interactive/utility block" see [this link](https://minecraft.wiki/w/Category:Utility_blocks) for a list


Snowiiii:
You should use our not crate instead of fastnbt

OfficialKris:
> You should use our nbt crate instead of fastnbt

When I try to switch to pumpkin-nbt I get UnsupportedType("tuple") on the String array. I do not know how this would be implemented as it is not a tuple but an array.

Snowiiii:
When i try to edit a sign i get the Sign edit Screen, When are done editing and leaving the Screen the Sign is empty. And when rejoining the Entire sign is not there

OfficialKris:
I can reproduce. However, these are two separate issues. The first is pumpkin currently has no block direction detection, so the text is currently on the "back" side. This is fixed in #485 and can be implemented for signs once that pr is merged. The second issue affects all block entities (ex. chest blocks do not render after login). I think that should be a separate PR. It is similar to the mobs not showing after logging in pr.

Snowiiii:
@OfficialKris #485 is merged now

Snowiiii:
Thanks @whiteout314 and @OfficialKris. Great Work :heart:, So nice having Signs finally Working! yay

PR #492, opened 2025-01-24
Text component refactor

DarkMrMelther:
## Description

Changed the whole TextComponent structure to allow nested translated texts and avoid recursion in serialization, but keeping the api unchanged. Also changed all the translate components to fit the new format, and added translation to gamemode command.
Truly happy of beign able to solve this issue, the TextComponent had two problems:
1. It had a loop in its serialization, being that TC::serialize called to_bytes_unnamed and this in turn called TC::serialize again.
2. As it has to be converted to bytes (or so I understand) when serializing, the children and translation arguments were converted twice, which generated problems when reading it in the client.

> After having achieved this achievement (considering that I have just
started programming in rust) I am extremely happy to go to sleep.
-- Melther - 24/1/25 - 5:00

## Testing

image

### DIFF
--- a/pumpkin-nbt/src/serializer.rs
+++ b/pumpkin-nbt/src/serializer.rs
@@ -11,6 +11,12 @@ use crate::{
 
 pub type Result<T> = std::result::Result<T, Error>;
 
+pub trait SerializeChild {
+    fn serialize_child<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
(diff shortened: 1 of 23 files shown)


Snowiiii:
Wow. The rate of your PR's is really impressive, That's also something i already wanted to fix for a long time.
We can also then now close https://github.com/Pumpkin-MC/Pumpkin/pull/455
Thank you @DarkMrMelther :+1:

Issue #118, opened 2024-10-14
Support for multiple World formats


Snowiiii:
I saw there might be interested in multiple custom world formats (e.g. https://github.com/Snowiiii/Pumpkin/pull/95#issuecomment-2385949412)

We currently only support reading Anvil files. My ideas where

- Region
- Slime Format
- Linear Format
- Schematic Format

I also discovered when [zstd](https://github.com/facebook/zstd) which Linear Format uses, It is more modern and promises to be much more faster than the old ZLib

Tell me what do you think

Snowiiii:
One disadvantages of using different formats or different compression (e.g. zstd) is that these world become not usable in vanilla. I recently discovered that actually LZ4 which is supported in vanilla world is even fast than zstd but has a bit lower compression ratio, We definitely should use that

RemineHD:
For the scope of this project, I think we should focus on supporting Anvil and Linear formats by default. Region format is outdated and rarely used, I don't see a real use case for it in Pumpkin. 
Slime is great, but it does not work with larger worlds (Like a Vanilla survival). As it loads the whole world in RAM. 
In my opinion, adding support for Slime would only add more tasks for contributors and is not really necessary to have it out of the box. I think we could leave both Slime and Schematic for plugin developers.

I may work on adding support for Linear, I really like Pumpkin and I want to collaborate in some way, but no promises!

Commandcracker:
Small Note: lz4 compression was added in [24w04a](https://minecraft.wiki/w/Java_Edition_24w04a)

Mili-ssm:
I am willing to help, but i am not used to the minecraft internals, so if you can post links for the descriptions or reference of the algorithms (ie. [linear region format)](https://github.com/xymb-endcrystalme/LinearRegionFileFormatTools) it will be very helpfull <3.

Snowiiii:
> I am willing to help, but i am not used to the minecraft internals, so if you can post links for the descriptions or reference of the algorithms (ie. [linear region implementation](https://github.com/xymb-endcrystalme/LinearRegionFileFormatTools) and [linear region Format Doc](https://gist.github.com/Aaron2550/5701519671253d4c6190bde6706f9f98)) it will be very helpfull <3.

I think the Schematic format is the easiest one. Here are the docs https://minecraft.fandom.com/wiki/Schematic_file_format

Mili-ssm:
[...]

PR #494, opened 2025-01-24
feat: add linear chunk format.

Mili-ssm:
# Description

Introduce a new linear chunk format and integrate zstd compression as a dependency. This enhances the compression capabilities of the project.

Just now Linear format is not integrated on any part of the project but we have the needed APIs (sames as Anvil format) to start using it.

# Testing

It has its owns unit test (same as the Anvil format)
Load a linear format world to ensure the spec is well implemented

# Checklist

- [X] Add an Advance configuration for Chunk File Format.
- [X] Ensure the linear format is correctly implemented loading an external world saved with the same format.

Related: #118


pumpkin-world/src/chunk/linear.rs
```diff
@@ -0,0 +1,446 @@
...
+                compression_level: ADVANCED_CONFIG.chunk.compression.compression_level as u8,
+                chunks_count: 0,
+                chunks_bytes: 0,
+                region_hash: 0,
+            },
+            chunks_headers: [LinearChunkHeader::default(); CHUNK_COUNT],
+            chunks_data: vec![],
+        }
+    }
+    fn is_valid_file(file: &mut File) -> bool {
```
Snowiiii, line 106: You should check if the version is valid as well.

pumpkin-world/src/chunk/linear.rs
```diff
@@ -0,0 +1,446 @@
...
+        }
+
+        Ok(())
+    }
+}
+
+impl LinearChunkFormat {
+    const fn get_region_coords(at: &pumpkin_util::math::vector2::Vector2<i32>) -> (i32, i32) {
+        (at.x >> 5, at.z >> 5) // Divide by 32 for the region coordinates
+    }
```
Snowiiii, line 300: We already have that in `pumpkin-util/math`

pumpkin-world/src/chunk/linear.rs
Mili-ssm: I do some lines under there, 
nevertheless i change the function namings to clarify it.

pumpkin-world/src/chunk/linear.rs
Mili-ssm: In the `Anvil` format file we have similar operation inside that code (these are transformation specifics for this formats (theorically))

suprohub:
Can you use dynamic var instead of consts? Because plugins can save some stuff with this format.

Mili-ssm:
> Can you use dynamic var instead of consts? Because plugins can save some stuff with this format.

I suppoues there is no problem? but exactly about what constants are you talking? (some code ref or file/line_number would be great).

Snowiiii:
Mhh just tried to test but for some reason got it panicked saying me this 
`thread '<unnamed>' panicked at pumpkin-world/src/chunk/linear.rs:245:48:
range end index 2061500416 out of range for slice of length 52582031`.
I converted my anvil files (which worked in pumpkin) into linear and changed linear to the default chunk reader/writer

Mili-ssm:
> Mhh just tried to test but for some reason got it panicked saying me this 
> `thread '<unnamed>' panicked at pumpkin-world/src/chunk/linear.rs:245:48:
> range end index 2061500416 out of range for slice of length 52582031`.
> I converted my anvil files (which worked in pumpkin) into linear and changed linear to the default chunk reader/writer 

with what software do you generate the save? (it is just thr last part i required, the real test hahaha, so this is perfect hehe)

Snowiiii:
> > Mhh just tried to test but for some reason got it panicked saying me this
> > `thread '<unnamed>' panicked at pumpkin-world/src/chunk/linear.rs:245:48: range end index 2061500416 out of range for slice of length 52582031`.
> > I converted my anvil files (which worked in pumpkin) into linear and changed linear to the default chunk reader/writer
> 
> with what software do you generate the save? (it is just thr last part i required, the real test hahaha, so this is perfect hehe)

I used linear's own convert script
`./convert_region_files.py mca2linear my_region_folder my_linear_folder`

Mili-ssm:
> > > Mhh just tried to test but for some reason got it panicked saying me this
> > > `thread '<unnamed>' panicked at pumpkin-world/src/chunk/linear.rs:245:48: range end index 2061500416 out of range for slice of length 52582031`.
> > > I converted my anvil files (which worked in pumpkin) into linear and changed linear to the default chunk reader/writer
> > 
> > 
> > with what software do you generate the save? (it is just thr last part i required, the real test hahaha, so this is perfect hehe)
> 
> I used linear's own convert script `./convert_region_files.py mca2linear my_region_folder my_linear_folder`

I add the `Advance Configuration`:
```
[chunk]
file_format = "Anvil"|"Linear"
```
also, i add some extra checks about the files to ensure is compatible (or otherwise get better info at least for debug). @Snowiiii if do you wanna try and give me some feedback about the new message you receive perfect, if not dont worry i will try tomorrow some tests by my self with the **mc converter**.

Mili-ssm:
Okey guys, some good/bad news.

@Snowiiii i have to admit i dont yet try with the `mc_converter` but i think i get the issue (or any ways i found a great issue xd).
Pumpkin reads a lot in parallel, with anvil this is not a problem but with linear many chunks share file and its compress so you need time to read and write, this tend to create race bugs.
 1. Thread 1 is writing a file while Thread 2 is reading.
 2. The data must be compressed, so also rewritten.
 3. The Thread 2 found an `EOF` before it would or make an a wrong parse of the data.

For all of this there is some available fixes ideas but i want/need your thoughts.
 - A `Queue` system.
 To store pending chunks that need to be stored and every X time or X chunks flush everything in bulk (while you swap the queue to store new chunks).

 - A `Map` with many `Muttex`.
 To know if "you" (Thread X) can use the File Z with out any problem (many readers at a time is allowed, but if anyone is writer only the writer. <sub>something like the rustc borrow checker what a life, hahahahah</sub>).

 - Change the `APIs`.
 To manage this type of things inside the core and add the `Bulk API` to the ChunkWriter and ChunkReader.


Any idea or proposition is wellcome 😄.

<sub>pd 1: 
 For my tests and for now i am using aliases (file name with a timestamp, then y write, and when is everything written i rename the files with the correct one but this bring many issues like the possibility of lose chunks on the **races** but at least less o zero crashes.</sub>



<sub>pd 2: 
 good new because i think the spec is well implemented and aboard this type of stuff now probably save us many future headaches.</sub>

Mili-ssm:
As for now the LinearFormat look working well ( still not tested the [oficial implementation](https://github.com/xymb-endcrystalme/LinearRegionFileFormatTools/tree/master) ).

But just for now it has perf issues caused by the deallocation and write of chunks. As talked before Linear format will benefit from a some type of `bulk/batch API` for reading and for writing ( too much compressions and decompressions at a time )

pd: For now i put some annoying `dbg! logs` to watch how many writings/reading are happening at a time.

Mili-ssm:
The mca2linear converter was tested with the implementation (it had some bugs buts now everything is okey).

The last issues is about performance (tend to kick the player when the server gets saturated from IO operation) but i pretend to fix this issues on others PRs (improving the chunk management and improving the `ChunkReader` and `ChunkWritter` APIs

pumpkin-world/src/chunk/linear.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    fn to_bytes(&self) -> [u8; Self::FILE_HEADER_SIZE] {
+        let mut bytes = [0u8; LinearFileHeader::FILE_HEADER_SIZE];
+        bytes[0] = self.version as u8;
+        bytes[1..9].copy_from_slice(&self.newest_timestamp.to_be_bytes());
+        bytes[9] = self.compression_level;
+        bytes[10..12].copy_from_slice(&self.chunks_count.to_be_bytes());
+        bytes[12..16].copy_from_slice(&self.chunks_bytes.to_be_bytes());
+        bytes[16..24].copy_from_slice(&self.region_hash.to_be_bytes());
```
Snowiiii, line 132: Using direct slice access via bytes can be error prone and looks really complicated, Instead i can use function from `bytes::Buf`, I would also like to refactor anvil to use `Buf` but that's something for another time.
Using  `Buf` you can use something like `get_u8` or `get_u32` (big endian by default)

pumpkin-world/src/chunk/mod.rs
```diff
@@ -13,12 +20,15 @@ use crate::{
...
 
 pub mod anvil;
+pub mod linear;
 
 pub const CHUNK_AREA: usize = 16 * 16;
 pub const SUBCHUNK_VOLUME: usize = CHUNK_AREA * 16;
 pub const SUBCHUNKS_COUNT: usize = WORLD_HEIGHT / 16;
 pub const CHUNK_VOLUME: usize = CHUNK_AREA * WORLD_HEIGHT;
 
+// Manejador global para múltiples archivos
```
Snowiiii, line 30: What?

pumpkin-world/src/chunk/mod.rs
```diff
@@ -13,12 +20,15 @@ use crate::{
...
+static FILE_LOCK_MANAGER: LazyLock<FileLocksManager> = LazyLock::new(FileLocksManager::default);
```
Snowiiii, line 31: I'm not sure, but maybe we can use the File locking crate which we also use to lock the `session.lock` ?

Snowiiii:
Overall I like your code and you did a great Job of using abstraction and not cluttering anything :D. I also really like that you write good comments and documentation for almost everything. 

I see some unwraps which may can be replaced but besides of that Great Work @Mili-ssm :+1:

pumpkin-world/src/chunk/mod.rs
Mili-ssm: yea... i will fix that as early as i can (i am from spain, sorry... :( )

pumpkin-world/src/chunk/mod.rs
Mili-ssm: i will check that lib and commit the changes or ask to clarify any doubt that i could have.

pumpkin-world/src/chunk/linear.rs
Mili-ssm: Okey, thats looks correct for me, i have to admit i like the slicing thing as "code as documentation" but is true that can be error prone or as you say complex to follow, i will change de comments also to use a link that i find in one branch of the official repo (the specs are very bad on main and issues but one branch have de V1 and V2 spec correctly explained)

pumpkin-world/src/chunk/mod.rs
Snowiiii: Oh, that's nice, no problem. Amigo :wink:

pumpkin-world/src/chunk/mod.rs
Mili-ssm: I was reading about the lib and its says:
 - IS based on a POSIX spec. (that require the other program/process use the same spec if not it can be ignorable)
 - The windows version require open the file with write permission.
 - Apparently the POSIX spec suffer from write starvation contrary to Tokio RWLock.
  Could happen that the writter never get the lock what in this case include accumulate thousands of threads trying to write that chunk and overheading the system.

I am happy if we want to make our own implementation with RWLock and make it global for all the crates or system. Or also make only inside the level struct.
<sub> its just the thing i make may be little improved but is just a hash table so this dont require neither a lot of develop/LoCs nor complex stuff</sub>

Both cases in my opinion preferable than the POSIX solution.

Snowiiii:
Everything looks Awesome! Thank you so much for your Hard work @Mili-ssm. I like your Code style :+1:. Would love to see more from You :smile: 

Thank you :heart:

PR #495, opened 2025-01-25
Separated server logic from main

paolobettelini:
## Description

The server logic, such as enabling console, rcon, lan, query, ticker and polling and the connection loop, has been moved into a new file `server_manager.rs`.
Now, `main.rs` is more readable and mostly contains logging and signal handling.
The server is started as follows:
```rust
    let server_manager = ServerManager::new().await;
    server_manager.init_plugins().await;

    log::info!("Started Server took {}ms", time.elapsed().as_millis());
    log::info!(
        "You now can connect to the server, Listening on {}",
        server_manager.addr
    );

    server_manager.start().await;
```
I also removed the expectation of `clippy::too_many_lines`.


Snowiiii:
So i guess the work is mostly done so others can build their own Servers using Pumpkin as a framework right ?. I generally don't have anything against this. But i may think we somehow should add more configuration to this. for example accepting bools in `new` or making new functions to enable query or RCON...

paolobettelini:
Yes, this is definitely a step into making it more framework-like, but I believe this PR simplifies the readability a bit with the separation of concerns, without making it an actual framework.
> accepting bools in new or making new functions to enable query or RCON

I think the main thing that would need to be done in order to make it a framework is being able to pass the config struct to the server manager.
```rust
let server_manager = ServerManager::new(basic_config, advanced_config);
```
If I'm using it as a framework, I don't want to load the config from a file, I want to do it in some other way. The same goes for the plugins; I might not want to use the FFI plugins. The internal usage of the framework should read the config from the file, and implement the FFI plugins. However, I would want to be able to register event listeners on the server manager.

If we want to add functions like `enable_rcon` and such, then it would make sense for the internal code not to rely on the config at all. The config would be read from the file and used to setup the server manager accordingly. However, this would require more labor and even simply passing the config to the server manager is enough to make it a bit framework-like.

suprohub:
Maybe inxlude thia to lib.rs?

paolobettelini:
@suprohub I did declare the module in `lib.rs` if that's what you mean

Snowiiii:
I think it would be better to move the new `server_manager.rs` file into `lib.rs` and also rename `ServerManager` to `PumpkinServer`

Snowiiii:
Now it looks nice. Great work @paolobettelini. I hope that others may can use Pumpkin now as a foundation or similar :+1:

PR #497, opened 2025-01-25
Changed hover types

DarkMrMelther:
## Description

On the HoverEvents changed the ShowText type to allow using TextComponent on it, made the same chage to the "name" attribute of ShowEntity. 
> [!CAUTION]
> This solution perhaps is not the most suitable, since it is more a workaround, I explain, since when passing a single TextComponentBase the game is not able to read it, we take advantage of the fact that we can pass an array of them so the error doesn't jump, but this forces us to use Vec instead of Box.

In the seed command, the hover message has been changed so it will be translated as its behavior in vanilla.
Also fixed the "get_text" function from TextComponent to make it possible to use translations on it.

## Testing

**Translated hover text of seed command (In spanish (●'◡'●))**
image

**Showing that translations works on ShowEntity (Using the seed command as base)**
image

### Extra
Cool *(At least for me)* logo I made to ingame because icons with transparecy feels strange on servers menu
image

### DIFF
--- a/pumpkin-protocol/src/lib.rs
+++ b/pumpkin-protocol/src/lib.rs
@@ -191,7 +191,7 @@ impl PositionFlag {
 
 pub enum Label {
     BuiltIn(LinkType),
-    TextComponent(TextComponent),
+    TextComponent(Box<TextComponent>),
 }
 
 impl Serialize for Label {
--- a/pumpkin-util/src/text/hover.rs
+++ b/pumpkin-util/src/text/hover.rs
@@ -1,33 +1,56 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, vec};
 
 use serde::{Deserialize, Serialize};
 
-use super::TextComponent;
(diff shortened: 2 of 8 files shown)


Snowiiii:
For some reason i get a `Segmentation fault (core dumped)` when try to join the Server

DarkMrMelther:
@Snowiiii I just try downloading the version from github actions and it works great for me, I don't know why it is, I'm on windows 11, maybe is a OS thing?

DarkMrMelther:
I just tried it also on Ubuntu 22.04.5, and it doesn't give me any error either, can you check it again and tell me exactly how to replicate it if it fails again?

Snowiiii:
So i found out that this actually breaks plugins which do use TextComponent. Since i had a plugin loaded which modifies the existing Join message it crashes the server

Snowiiii:
But otherwise this works fine and since we don't have any Official release i don't really care about breaking any plugins. So good work @DarkMrMelther :+1:

PR #505, opened 2025-01-26
Command parity to vanilla

DarkMrMelther:
## Description
Added parity with vanilla to texts from the following commands:
- Clear
- Kill
- Give
- Msg

The pumpkin command has also been formatted to make it more readable.
Other minors changes are:
1.  Updated show_entity from HoverEvents to make it easier to use.
2.  On EntityType added to_str and to_string which returns the identifier of the entity. (Resource location)
3.  On item added translated_name which returns the TextComponent::translate suitable for that item.

Note: Many of the changes requires [F3 + H] (Advanced Tooltips) to be shown.

## Update
Added msg command to match with the changes above, for its implementation the function send_message was added to Player.

## Testing

**Pumpkin commands**
image
**Player selection and tell**
image
**Item show**
image
**Msg command**
image

### DIFF
--- a/pumpkin-data/build/entity_type.rs
+++ b/pumpkin-data/build/entity_type.rs
@@ -52,18 +52,6 @@ pub(crate) fn build() -> TokenStream {
         })
         .collect::<TokenStream>();
 
-    let type_to_str = json
-        .iter()
-        .map(|sound| {
-            let name = ident(sound.0.to_pascal_case());
-            let identifier = format!("minecraft:{}", sound.0);
-
-            quote! {
-                Self::#name => #identifier,
-            }
-        })
-        .collect::<TokenStream>();
-
     quote! {
         #[derive(Clone, Copy, PartialEq, Eq, Debug)]
         #[repr(u8)]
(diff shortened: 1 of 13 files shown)


pumpkin-data/build/entity_type.rs
```diff
@@ -52,8 +52,20 @@ pub(crate) fn build() -> TokenStream {
...
+
+            quote! {
+                Self::#name => #identifier,
+            }
+        })
+        .collect::<TokenStream>();
+
     quote! {
-        #[derive(Clone, Copy, Debug)]
+        #[derive(Clone, Copy, PartialEq, Eq, Debug)]
```
Snowiiii, line 68: I don't think its needed since we can just use the Debug derive

DarkMrMelther:
# Bossbar translated!
Added translations for the bossbar command along with other props such as the bossbar id when hovering on its name.
Now if no namespace is specified for the bossbar it automatically assigns "minecraft:" as namespace.
Things I've noticed that are missing to be implemented in the command:
- Get players
- Bossbar not saved on exit
- Autocompletitions for many values

## Bossbar in action!
image

Snowiiii:
Everything works great. Great work like always @DarkMrMelther :+1:

PR #508, opened 2025-01-28
ABS (Another Boring Separation)

suprohub:
## Description
I separated event sending from main plugin struct.
It can help for event developing.
For example i can send chunk load directly from world crate.
## Testing
workfloss

WARNING
I not tested plugins, so maybe it not work, but it must work.


suprohub:
Also maybe i improve something, but now its seems mostly normal

PR #515, opened 2025-01-29
Fixed slabs properties

DarkMrMelther:
## Description
Changed a little bit the properties system to make the slabs work properly and the collision calculation when placing them.
Now depends on the state calculations if the block is updatable or not, also now the calculus knows if the target block was clicked or was other. Also when the block is placed/updated from another block will have inverted the blockface to match the expected face of the block.

Alex: Also added "Layers" Property used for Stacking Snow (Bit buggy tho)

## Testing
**Strange slabs pillars i made testing**
image

### DIFF
--- a/pumpkin-util/src/math/boundingbox.rs
+++ b/pumpkin-util/src/math/boundingbox.rs
@@ -2,12 +2,8 @@ use super::{position::BlockPos, vector3::Vector3};
 
 #[derive(Clone, Copy)]
 pub struct BoundingBox {
-    pub min_x: f64,
-    pub min_y: f64,
-    pub min_z: f64,
-    pub max_x: f64,
-    pub max_y: f64,
-    pub max_z: f64,
+    pub min: Vector3<f64>,
+    pub max: Vector3<f64>,
 }
 
 impl BoundingBox {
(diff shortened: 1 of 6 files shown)


Snowiiii:
Good work but like already said in discord i noticed a bug with the collisions. I also think it would be so much better to save `state_mappings` and `property_mappings` directly using the BlockProperty enum like this

```rust
pub struct BlockProperties {
    // Mappings from property state strings -> offset
    state_mappings: HashMap<Vec<BlockProperty>, u16>,
    // Mappings from offset -> property state strings
    property_mappings: HashMap<u16, Vec<BlockProperty>>,
}
```
This would simplify a lot since we now don't have to work with ugly Strings

DarkMrMelther:
## Update
As neither of us like the change of the margin in the calculation of the intersection I have undone it, I have also made a few adjustments to it, from now on the intersection will be calculated with the players in a radius of 3 blocks from the placed block (not from the player that places it as before), this should be more than enough since there are no blocks of more than 2 spaces, about the error I have no idea how to fix it, but my latest theory is that it is related to the multithreading that is not giving the most updated exact position for the calculations.

Snowiiii:
For this PR its probably enough but i would like to have the bug fixed in the future (definitely before release). If you somehow get an idea how to fix it i would appreciate an PR :D.

Thank you @DarkMrMelther :heart:

PR #520, opened 2025-01-31
Implement DamageType and add damage command

drakeerv:
## Description
I Implemented damage type and the damage command. I also added damage_with_context on the living entity.

## Testing
Cargo fmt & clippy. Also tested on Minecraft.


pumpkin/src/entity/living.rs
```diff
@@ -70,16 +70,22 @@ impl LivingEntity {
...
 
-    // TODO add damage_type enum
-    pub async fn damage(&self, amount: f32, damage_type: u8) {
+    pub async fn damage_with_context(
+        &self,
+        amount: f32,
+        damage_type: DamageType,
+        position: Option<Vector3<f64>>,
+        source: Option<Arc<Player>>,
+        cause: Option<Arc<Player>>,
```
Snowiiii, line 79: source and cause can be non Players (e.g Creeper, Arrow...). It would make more sense to make them just Entity

drakeerv: Oopsie... I missed that. One sec

pumpkin/src/entity/mob/mod.rs
```diff
@@ -49,6 +50,18 @@ impl EntityBase for MobEntity {
     fn get_living_entity(&self) -> Option<&LivingEntity> {
         Some(&self.living_entity)
     }
+
+    fn is_invulnerable_to(&self, damage_type: DamageType) -> bool {
```
Snowiiii, line 54: Isn't this just duplicate code, which already exists in entity ?

drakeerv: Just fixed it :)

pumpkin/src/entity/item.rs
```diff
@@ -35,4 +36,8 @@ impl EntityBase for ItemEntity {
     fn get_living_entity(&self) -> Option<&LivingEntity> {
         None
     }
+
+    fn is_invulnerable_to(&self, damage_type: DamageType) -> bool {
```
Snowiiii, line 40: Why place it inside Entity base ?, Entity base should have almost no code only things which are implemented with different behavior in different Entities. Which is not the case here. So why not just move this to Entity ?

PR #521, opened 2025-01-31
Implement experience and experiene/xp commands

drakeerv:
## Description
I implemented experience and the commands to control it (currently the only way to get experience).

## Testing
Tested it in game and ran cargo fmt & clippy


pumpkin/src/entity/player.rs
```diff
@@ -135,6 +137,12 @@ pub struct Player {
     pub client_loaded: AtomicBool,
     /// timeout (in ticks) client has to report it has finished loading.
     pub client_loaded_timeout: AtomicU32,
+    /// The player's experience level
+    pub experience_level: AtomicI32,
+    /// The player's experience progress (0.0 to 1.0)
+    pub experience_progress: AtomicCell<f32>,
+    /// The player's total experience points
+    pub total_experience: AtomicI32,
```
Snowiiii, line 145: 1. I wonder if `experience_level` and `total_experience` can become unsigned since i'm not sure if you can get negative XP?
2. I don't really get the difference between `experience_level` and `total_experience`, I would suggest to improve the doc on these

Snowiiii:
Looks good. Great work @drakeerv, Keep it up :D.

Thank you :heart:

PR #522, opened 2025-01-31
Allow plugins to unregister a command

vyPal:
### Changes:
- Adds a `pub(crate) fn unregister(&mut self, name: &str)` function to the command dispatcher which removes a command tree (including all aliases) by the primary name
- Adds a `pub async fn unregister_command(&self, name: &str)` function to the `Context` object passed to plugins
- Modifies the `register_command` and `unregister_command` functions to also send `CCommand` packets to all players in all worlds so that client-side command suggestions update to match the server

## Testing


Snowiiii:
Clippy :1st_place_medal:

Snowiiii:
Looks good. Thanks @vyPal :D

PR #524, opened 2025-02-01
Teleport cross dimension

vyPal:
## Description
This PR adds one new public function to the `Player` struct:
- `pub async fn teleport_world(self: Arc<Self>, new_world: Arc<World>, position: Option<Vector3<f64>>, yaw: Option<f32>, pitch: Option<f32>)` - Teleport player to new world, this will also teleport the player to the specific coordinates provided by the caller. It will not check the coordinates for correctness (like if the y isn't below the surface). If no coordinates are provided, it will select the default spawn coordinates for the new_world.

### Important:
- This PR includes a custom command (/tcd) which is only used for debugging / demo purposes, and some extra code (loading a second world in `server/mod.rs`. Please <ins>**DO NOT MERGE**</ins> until this code is removed.

## Testing


Snowiiii:
What about making only one new struct `teleport_world` and providing the position as an `Option`. If its None the default spawn location will be used, Otherwise the provided location

vyPal:
Yeah that sounds like a better option

vyPal:
There are too many merge conflicts for me to resolve, I will delete this branch, make a new one and add the methods again.

PR #525, opened 2025-02-01
Density Function Re-Work

kralverde:
## Description
This PR adds a density function extractor and reworks density functions as a stack of components to allow for easy mutation of state. 

The end goal of this PR is to make the density function code cleaner and more readable, remove unsafe code, and to separate density functions from the chunk noise generator to allow ease of use with other dependencies such as biome generation. (maybe we'll get a performance boost too)

For use with https://github.com/Pumpkin-MC/Extractor/pull/3
## Testing


kralverde:
Final result:
- A bunch of redundant code removed
- Compartmentalized density functions
- 25% speed-up (on my machine)
- Removal of unsafe code
- -60 Sanity

Snowiiii:
Please add a way to extract all the .chunk files in our extractor, i only see one single chunk.json file

kralverde:
> Please add a way to extract all the .chunk files in our extractor, i only see one single chunk.json file

Should be good now

Snowiiii:
Great work @kralverde (like always :upside_down_face:). The new implementation definitely looks cleaner, Hope this also helps with Biomes.
The Speed difference is also significant, For my machine it got down to 12ms from 20ms 

Thank you @kralverde :heart:

PR #526, opened 2025-02-02
Implement DamageType and add damage command (v2)

drakeerv:
## Description
I Implemented damage type and the damage command. I also added damage_with_context on the living entity. I redid everything from the previous pr #520.

## Testing
Cargo fmt & clippy. Also tested on Minecraft.


Snowiiii:
Great work @drakeerv :D. Nothing to complain about :+1: 

Thank you :1st_place_medal:

PR #528, opened 2025-02-02
Optimize projectile

suprohub:
## Description
I added functions in entity for plugins and projectile, because its can be used in other places and more simplier. Optimized some places, and now use `.to_degrees()`
## Testing
`cargo run`


Snowiiii:
You made it worse, Now its less Vanilla

suprohub:
> You made it worse, Now its less Vanilla

Why you think what? Like our target its not make all what in vanilla. I really optinized what, for example now we use .to_degrees(). Also not all like only one fuction. Plugin devs maybe want use linear projectile.

suprohub:
> > > You made it worse, Now its less Vanilla
> > 
> > Why you think what? Like our target its not make all what in vanilla. I really optinized what, for example now we use .to_degrees(). Also not all like only one fuction. Plugin devs maybe want use linear projectile.
> 
> No, The current implementation matches Vanilla, You remove rotation and power, Now it's completely different 

I not remove rotation and pover. Now its just in separated methods for better plugin dev

PR #529, opened 2025-02-02
Reworking block properties

DarkMrMelther:
## Description
Added the following properties:

- Axis (For logs and directed blocks)
- Open
- Age (For crops)
- Attachment (For bells)
- Signal Fire (For campfire smoke)
- 6 connection directions (For fences, walls, chorus)

Note: The blocks are not updated automatically, we should refactor block placement to allow update blocks properties, and custom placement functions (like for doors, beds or the blocks which uses the facing property inverted)

## Testing
**Some of the blocks affected** (The lantern not)
image


Snowiiii:
I like the changes, but i believe we may should rewrite the properties to be inside the block implementations, this would make using them in code easier, That way we can implement Redstone for example

DarkMrMelther:
> I like the changes, but i believe we may should rewrite the properties to be inside the block implementations, this would make using them in code easier, That way we can implement Redstone for example

@Snowiiii I was planning on changing properties and blocks to allow having custom place and update functions, if you like I can start working on that rework and in some days I could have a playable version.

DarkMrMelther:
# Update: Properties rework
Now the block properties are based on the “BlockProperty” trait, which allows better separation and more extensibility, added a proc macro to make the implementation of the properties easier and now you only need to put the necessary information.
These changes have allowed us to separate the updateability of the block from its status update, separating it in the functions “on_place” and “can_update”.
Changed how to detect that the block is replaceable, being part of the “can_update” function now. 

Note: All the previously implemented properties should work correctly.

### Next: on_interact function implementation for properties and block modifiers to "on_place"

DarkMrMelther:
# Update: Levers works now! 🎊 🌟 🥳 
Made the lever work, implementing modifiers to the “on_place” per block, plus the “on_interact” function to handle the state when interacting with a block that has it.

Now it remains to combine the APIs of the blocks and the properties, with respect to interactions.

## Video
https://github.com/user-attachments/assets/0e3a2ac3-782e-494e-bd4d-483130a873a4

pumpkin/src/block/properties/powered.rs
```diff
@@ -0,0 +1,22 @@
...
+use pumpkin_world::block::registry::Block;
+use pumpkin_world::item::ItemStack;
+
+use super::{BlockProperty, BlockPropertyMetadata};
+
+#[block_property("powered")]
+pub enum Powered {
+    True,
+    False,
+}
```
Snowiiii, line 12: This should be a boolean, This would also match Vanilla then.
You can make a direct field
```rust
pub struct Powered(bool)
```

Snowiiii:
Looks awesome. Great work like always @DarkMrMelther :D. But i think we should not only use Enums. I think there should be some kinda implementation for different property types like in Vanilla, Supporting all common types like booleans, integers and enums

DarkMrMelther:
@Snowiiii I think that this is the nearest aproach i can do to implement primitive types (Bool at the moment) without having to rewrite the entire system.

Snowiiii:
Overall Great work @DarkMrMelther!. Thank you so much :heart: always love to see a new PR from you :D

Issue #511, opened 2025-01-28
Teleport player between worlds / dimensions


vyPal:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

I am working on a multiverse-like plugin for pumpkin to allow easy management of multiple worlds and dimensions, and I was working on implementing the logic to teleport the player to either the same world but a different dimension, or a entirely different world, but I noticed there was no functionality to do this easily. I tried to make my own implementation of such a function, but I do not have a large understanding of which packets need to be communicated during such a teleport.

Furthermore, I have found that the current structure of certain structs (like the Player and Entity structs) lack support for changing the dimension (the necessary structs are behind a non-mutable shared reference). 

So in total it looks like a lot of change will have to take place to allow this, and I sadly don't have the knowledge (or time) to implement such large changes. Therefore I will leave this issue here, so that someone might get to this in the future.

### Describe the solution you'd like

Add a method (or multiple methods) to allow moving a player between worlds/dimensions. 
Example:
```rs
impl Player {
    pub async fn teleport_to_world(&self, destination: Arc<World>, position: Vector3<f64>, yaw: f32, pitch: f32) {}
}
```

vyPal:
I guess I'm adding it myself #524

vyPal:
or #532

PR #532, opened 2025-02-07
Teleport to different world

vyPal:
Implements #511 

## Description
This PR adds one new public function to the `Player` struct:
- `pub async fn teleport_world(self: Arc<Self>, new_world: Arc<World>, position: Option<Vector3<f64>>, yaw: Option<f32>, pitch: Option<f32>)` - Teleport player to new world, this will also teleport the player to the specific coordinates provided by the caller. It will not check the coordinates for correctness (like if the y isn't below the surface). If no coordinates are provided, it will select the default spawn coordinates for the new_world.

## Testing


pumpkin/src/entity/player.rs
```diff
@@ -575,6 +575,129 @@ impl Player {
...
+        let uuid = self.gameprofile.id;
+        current_world
+            .broadcast_packet_all(&CRemovePlayerInfo::new(1.into(), &[uuid]))
+            .await;
+        current_world
+            .players
+            .lock()
+            .await
+            .remove(&self.gameprofile.id);
+        self.client
```
Snowiiii, line 619: Can't we just do `world.remove_player` ?

pumpkin/src/entity/player.rs
```diff
@@ -575,6 +575,129 @@ impl Player {
...
+            .send_packet(&CGameEvent::new(GameEvent::StartWaitingChunks, 0.0))
```
Snowiiii, line 620: Also i think we should have some kinda `send_world_info` function like in vanilla which is being send on join, respawn and here. Like this
```java
 public void sendWorldInfo(ServerPlayerEntity player, ServerWorld world) {
        WorldBorder worldBorder = this.server.getOverworld().getWorldBorder();
        player.networkHandler.sendPacket(new WorldBorderInitializeS2CPacket(worldBorder));
        player.networkHandler.sendPacket(new WorldTimeUpdateS2CPacket(world.getTime(), world.getTimeOfDay(), world.getGameRules().getBoolean(GameRules.DO_DAYLIGHT_CYCLE)));
        player.networkHandler.sendPacket(new PlayerSpawnPositionS2CPacket(world.getSpawnPos(), world.getSpawnAngle()));
        if (world.isRaining()) {
            player.networkHandler.sendPacket(new GameStateChangeS2CPacket(GameStateChangeS2CPacket.RAIN_STARTED, GameStateChangeS2CPacket.DEMO_OPEN_SCREEN));
            player.networkHandler.sendPacket(new GameStateChangeS2CPacket(GameStateChangeS2CPacket.RAIN_GRADIENT_CHANGED, world.getRainGradient(1.0f)));
            player.networkHandler.sendPacket(new GameStateChangeS2CPacket(GameStateChangeS2CPacket.THUNDER_GRADIENT_CHANGED, world.getThunderGradient(1.0f)));
        }
        player.networkHandler.sendPacket(new GameStateChangeS2CPacket(GameStateChangeS2CPacket.INITIAL_CHUNKS_COMING, GameStateChangeS2CPacket.DEMO_OPEN_SCREEN));
        this.server.getTickManager().sendPackets(player);
    }
    ```

pumpkin/src/entity/player.rs
vyPal: Well the `world.remove_player` function also emits the `PlayerLeaveEvent` which is meant for when the player leaves the entire server, not just a specific world. I could try to modify the `remove_player` method to take a boolean argument which would decide if the event should be fired

Snowiiii:
Looks good now. Thank you @vyPal :+1:

PR #533, opened 2025-02-07
Fixed item decrementing while block fails to place due to player intersection

RobIux:
## Description
Some code was left out of an if statement causing unexpected behaviour, eg. if a plyer stood in the way of a block it would not place but still decrement the players items.


Snowiiii:
Looks good. Thank you @RobIux :D

PR #535, opened 2025-02-09
Implement Bulk Loading/Writing for Chunks

Mili-ssm:
This PR implement the Bulk API for chunks loading/writing with some function changes that get leverage of the bulk operation (like reducing allocations, locks and more).

Also this includes some Anvil refactor and the creation of a Traits/API that can be re-used on entity storage system and to implements diferents Writers/Loaders (as DB cached, one that use in Cloud Storage or more optimizations).

## Actual Improves
 
 - Don´t write a chunk on disk when loaded from disk.
 - Lock the **Chunk Cache** entry while generating a new chunk.
   <sub>( if more threads wants that same chunks, they can wait without doing any work until the chunk is generated )</sub>

 - Every time a bunch of chunks (of the same player) are deallocated, the file is write only 1 time.
 - The Files that have ongoing writes/reads operations are stored on memory (to avoid disk readings and file lockings).
 - We have a DashMap for File Cache to avoid writing collisions and don`t corrupt files. (not an Anvil issue at least usually but it was a problem with the Linear format that enforce an *FULL FILE WRITE*)
- Chunk writing operation now are made on a `path.tmp` so now the writing operation cant corrupt files <sub>( when the writting finish, the file name is change, so the file is never swap before the full writting )</sub>
- Now the file Lock/Cache implements a `clean_cache` fn that ensure no memory leak along the server run.

## Actual Changes

 - The system behind ChunkReader and ChunkWriter is now a ChunkIO who use ChunkSerializer.
 - ChunkIO is a trait that use a Serializer.
 - ChunkSerializer is a trait that helps to serialize some type of chunk data (like Chunks terrain or Chunks entitis)
 - Now the ChunkIO instead of a Result<Data,Err> it returns a LoadedData<Data,Error> wich helps with:
   - `LoadedData::Error((position,Err))` : Manage errors per chunk loaded on the bulk.
   - `LoadedData::Missing(position)` : Manage missing chunks (not previusly generated) becaouse the return chunks order is not enforced.
   - `LoadedData::Loaded(Data)` : Manage normal returned chunks.
 - The fetch made by clients threads use  par_chunk instead of iter, this helps to group chunks by the asked sorted way but grouped in 64 chunks of Chunks, it take advantage of the file cache and bulk API.
 - Now when cleaning chunks, we check if the chunks is not being watched prev to remove it (to improve the undesirable instant loading of removed chunks) but if we cant remove it, we still write it in the file, this way we make the chunk writing worth and avoid unsaved changes.
- The *File Lock/Cache* is refactored and embedded into the `ChunkFileManager`. (the actual ChunkIO implementation for **Anvil** and **Linear**)


pumpkin-world/src/chunk/anvil.rs
```diff
@@ -4,34 +4,49 @@ use flate2::bufread::{GzDecoder, GzEncoder, ZlibDecoder, ZlibEncoder};
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[repr(u8)]
 pub enum Compression {
     /// GZip Compression
     GZip = 1,
     /// ZLib Compression
     ZLib = 2,
+    /// Uncompressed (since a version before 1.15.1)
+    #[default]
+    Uncompressed = 3,
```
Snowiiii, line 49: I don't like this, We had this before already. But it actually makes no sense to have "No Compression" in an Compression enum, Also now we have to use convert the input data to vec when uncompressed

Mili-ssm: I have no problem to let it as before but, if the code `3`exists and means something specific, even documented officially (https://minecraft.wiki/w/Region_file_format). Dont we should have it well represented on code? this way we dont need to manage where we use the compression struct. And also `NoCompression` can be a "zero runtime cost but no memory footprint improve version". Whatever you choose its good for me, i just thought it was desirable.

Snowiiii: I think it should not contain  `Uncompressed`, I'm not a fan of having to just pass the raw data back and especially turning it into an Vec

Bryntet:
This is dope! I'd love to see some before/after benchmarks :+1:

pumpkin-world/src/chunks_io.rs
```diff
@@ -0,0 +1,307 @@
...
+
+pub enum LoadedData<R>
+where
+    R: Send,
+{
+    /// The chunk data was loaded successfully
+    Loaded(R),
+    /// The chunk data was not found
+    Missing(Vector2<i32>),
+}
```
Snowiiii, line 28: I'm not sure but shouldn't this also handle when there is an error when loading a chunk, when its corrupted for example

Mili-ssm: i wrap that enum with a result, the points is because we now try to retrieve N chunks, but the result can be in a different order so now when a chunk was not yet created we must to return the position of the uninitialized chunk (probable i should change the names).

I wrap that enum with a result to manage that type of things (but now if only one chunk fails, its consider that all chunks on that retrieve failed together).

Snowiiii: I believe it's wrong to just not load all chunks when one fails

Mili-ssm: Fair, (thats the reason i post it as a draft)  one extra value that contains an error sound good?

```Rust
enum LoadedData{
				 Loaded(ChunkData),
				NontFound(Vector2),
				Error(Err),
}
```

Mili-ssm: @Snowiiii tell me if the new enum is indeed better of if you see something more to improve :D.
thanks you.

kralverde:
Are entries ever removed from the file lock when the file is no longer being used?

Mili-ssm:
@kralverde i refactor the "file lock" and yes i create a **clean_cache** fn that executes when the read lr the write finishes that check how much references of the lock are being holded. May be i should add a trace log for that but i checked that actually those entre are removed.

pumpkin-world/src/chunk/anvil.rs
```diff
@@ -140,361 +140,318 @@ impl Compression {
...
-        let chunk_data: Vec<u8> = file_buf.drain(0..size as usize - 1).collect();
+        Ok(AnvilChunkData {
+            length,
+            compression,
+            compressed_data: buffer.to_vec(),
+        })
+    }
+
+    fn to_bytes(&self) -> Vec<u8> {
+        let total_size = self.compressed_data.len() + 5;
```
Snowiiii, line 190: Why + 5 ?, You should add docs or make it a const

pumpkin-world/src/chunk/anvil.rs
```diff
@@ -140,361 +140,318 @@ impl Compression {
...
-            compression
-                .decompress_data(&chunk_data)
-                .map_err(ChunkReadingError::Compression)?
-        } else {
-            chunk_data
-        };
+        let mut bytes = Vec::with_capacity(padded_size);
+
+        bytes.put_u32(self.length);
+        bytes.put_u8(self.compression.map_or(3, |c| c as u8));
```
Snowiiii, line 197: What is 3 ?, Yes i know it represents an compression algorithm. But something like Compression:GZip as u8 would be more clear

pumpkin-world/src/chunk/anvil.rs
```diff
@@ -140,361 +140,318 @@ impl Compression {
...
-                .map(|(i, block)| {
-                    let name = STATE_ID_TO_REGISTRY_ID.get(block).unwrap();
-                    (block, (name, i))
-                })
-                .collect();
+    fn to_bytes(&self) -> Vec<u8> {
+        let mut chunk_data: Vec<u8> = Vec::new();
+
+        let mut location_bytes: Vec<u8> = Vec::with_capacity(SECTOR_BYTES);
+        let mut timestamp_bytes: Vec<u8> = Vec::with_capacity(SECTOR_BYTES);
```
Snowiiii, line 269: Why we can't allocate this on stack like the old code did
```rust
let mut location_table = [0u8; 4096];
let mut timestamp_table = [0u8; 4096];
```

pumpkin-world/src/chunk/anvil.rs
```diff
@@ -140,361 +140,318 @@ impl Compression {
...
+        let compression_method = buffer.get_u8();
+        let compression = Compression::from_byte(compression_method)
             .map_err(|_| ChunkReadingError::Compression(CompressionError::UnknownCompression))?;
 
-        // size includes the compression scheme byte, so we need to subtract 1
-        let chunk_data: Vec<u8> = file_buf.drain(0..size as usize - 1).collect();
+        Ok(AnvilChunkData {
+            length,
+            compression,
+            compressed_data: buffer.to_vec(),
```
Snowiiii, line 185: I don't like this to_vec(), Can't we remove it somehow, Maybe using lifetimes?

pumpkin-world/src/chunk/anvil.rs
Mili-ssm: true, thanks i tried to change the less i could the anvil implementation just using the bytes thing of get_u8 and that type of thing but i should let everything commented anyways.

pumpkin-world/src/chunk/anvil.rs
Mili-ssm: i think that was a limitation of try to use the get_u8 and things like that but not sure so i will try again.

pumpkin-world/src/chunk/anvil.rs
Mili-ssm: that is to simplify the chunk data access (try to isolate every chunk buffer, if you want to overwrite it use a slice ref dont simplify nothing because you are still obligated to manage the *insert new bytes in the middle*.

Snowiiii:
What do you think about using the Minecraft chunk batch packets to may improve network bandwidth and probably also client side performance ?

Snowiiii:
I'm not sure if this is right for this PR, but we can think about saving unfinished chunks when something bad happens like the Server shutdown or is out of memory, Vanilla uses the Chunk Status to have the current state saved so it is cheaper to fully generate when some work is already spend

Mili-ssm:
> I'm not sure if this is right for this PR, but we can think about saving unfinished chunks when something bad happens like the Server shutdown or is out of memory, Vanilla uses the Chunk Status to have the current state saved so it is cheaper to fully generate when some work is already spend

I would prefer to do that on other PR, here we have the new ChunkIO API. Actually i was thinking to include the batch packets system but i think i prefer to do that in other PR too.


TL;DR
when we get this PR merge next work from here should be with non specific order:

 - Entitys Serializer for save its state.
     Now using the ChunkIO Api it should be very easy, at leas on comparision to before were we need not only the serialyzer, the file managing system too.
 - Batch packet to the clients.
    To improve bandwith managment and get more wins too from the chunk Bulk Api)
 - Investigate the Chunk Status system.
    We save the chunks ver fast and early maybe we dont get many improve and in the other hand we lost performance and code flexibility.

kralverde:
Works for me. I think this PR is in a good state as-is. We should merge after doing a few more tests and then work on more functionality from there

kralverde:
I noticed you made some u8 arrays Box<[u8]>, if we know the array size (and it's relatively small) we shouldn't use the heap

Snowiiii:
@Mili-ssm @kralverde 
I remember that writing the actual packet was a big bottleneck, We do quiet a lot of things when sending the chunk packet, maybe we can somehow reduce the load there

Snowiiii:
I also remember there was a crash due to some chunk issue (IIRC) when a large amount of players connect as once on master, Can you may try running https://github.com/Pumpkin-MC/BotMark and test if the issue is fixed here ?

kralverde:
The stack overflow is probably unrelated

Snowiiii:
There is some kind of regression made, This is how master currently looks like

https://github.com/user-attachments/assets/1cacf97c-a832-4ded-8d4f-b9494763b96b

I used Fabric and the [Creative Fly](https://modrinth.com/mod/creative-fly) Mod to get faster Creative speed.

When trying flying at high speed on this PR the console spams erros and the Pumpkin crashes.
```
[ERROR] Failed to load chunk at Vector2 { x: -3, z: -36 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -3, z: -37 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -3, z: -38 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -2, z: -39 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
$
  thread '<unnamed>' panicked at pumpkin-world/src/chunk/format/anvil.rs:222:18:
                                                                                Failed to decompress chunk data: LZ4Error(Custom { kind: Other, error: LZ4Error("ERROR_blockChecksum_invalid") })
```

kralverde:
@Snowiiii could you retest with ur special mod?

Snowiiii:
The Error above is now gone and is not being spammed
```diff
- [ERROR] Failed to load chunk at Vector2 { x: -3, z: -36 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
```
But it still crashes
```
thread '<unnamed>' panicked at pumpkin-world/src/chunk/format/anvil.rs:222:18:
                                                                                Failed to decompress chunk data: LZ4Error(Custom { kind: Other, error: LZ4Error("ERROR_blockChecksum_invalid") })
```

kralverde:
Writing to console that many times is probably the cause of the slow down

kralverde:
> The Crash is gone now, but the Chunks load slower than on master and the console is full of errors
> 
> ```
> [ERROR] Failed to load chunk at Vector2 { x: 5, z: -41 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 3, z: -40 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 5, z: -41 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 2, z: -39 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 3, z: -40 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: -3, z: -36 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: -2, z: -37 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: -1, z: -37 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: -1, z: -37 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: -6, z: -28 }: Compression error Error while working with LZ4 compression: LZ4 error: ERROR_blockChecksum_invalid (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 0, z: -38 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 1, z: -39 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 2, z: -39 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 2, z: -39 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 0, z: -38 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: 0, z: -38 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: -36, z: 0 }: Compression error Error while working with LZ4 compression: LZ4 error: ERROR_blockChecksum_invalid (regenerating)
> [ERROR] Failed to load chunk at Vector2 { x: -40, z: -6 }: Compression error Error while working with LZ4 compression: LZ4 error: ERROR_blockChecksum_invalid (regenerating)
> [ERROR] Expected chunk at -45:-4, but got -39:-11
> [ERROR] Failed to load chunk at Vector2 { x: -37, z: 16 }: Compression error Error while working with LZ4 compression: LZ4 error: ERROR_blockChecksum_invalid (regenerating)
> [ERROR] Expected chunk at -39:17, but got -36:22
> ```

Was this with anvil or linear?

kralverde:
@Snowiiii could you retest?

Mili-ssm:
@Snowiiii @kralverde i archive to improve the performance while generating world, also simplified the code avoiding some overhead with channel that now is not needed anymore thanks to the performance imrpoves.

Snowiiii:
After testing again the Chunks seem to load faster!, But i still get 
`[ERROR] Failed to load chunk at Vector2 { x: 23, z: -19 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)`
Multiple times, and i'm not sure about handling the "Expected at ... but got ..." error like this @kralverde, The error should not happen when loading the chunk correctly and usually means that something is wrong internally

Mili-ssm:
@Snowiiii i saw that is an error from parsing the chunks, when the chunk is not detected as ***Full*** (the chunk status != ***Full***).

>`[ERROR] Failed to load chunk at Vector2 { x: 23, z: -19 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)`

When i use this branch to generate a world i didn't see that error, neither creating a world with master and opening with the branch.
Could this be an error that comes from using a pre-existent save or when loading many semi-generated chunks?

Snowiiii:
> @Snowiiii i saw that is an error from parsing the chunks, when the chunk is not detected as _**Full**_ (the chunk status != _**Full**_).
> 
> > `[ERROR] Failed to load chunk at Vector2 { x: 23, z: -19 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)`
> 
> When i use this branch to generate a world i didn't see that error, neither creating a world with master and opening with the branch. Could this be an error that comes from using a pre-existent save or when loading many semi-generated chunks?

Mhh weird, Can't really reproduce the error message again :sweat_smile:. But i sometimes got an other error
```
[ERROR] Failed to load chunk at Vector2 { x: -22, z: 9 }: Compression error Error while working with LZ4 compression: LZ4 error: ERROR_blockChecksum_invalid (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -22, z: 9 }: Compression error Error while working with LZ4 compression: LZ4 error: ERROR_blockChecksum_invalid (regenerating)
```

Snowiiii:
Also when flying at max speed (which you can set using [Mod Menu](https://www.curseforge.com/minecraft/mc-mods/modmenu)). The Chunks are lagging behind, it seems like it first tries to load the old ones, It would make more sense to just cancel the old chunk loading "Job" and just load the new one

https://github.com/user-attachments/assets/eacff781-c03c-4e10-a230-ba55378c5d50

Snowiiii:
Also when quiting using CTRL+C it crashes
```
thread 'main' panicked at pumpkin-world/src/chunk/io/chunk_file_manager.rs:309:18:
                                                                                  We initialize the once cells immediately
```

Mili-ssm:
@Snowiiii that's little hard because you don't know from the bunch of chunks:
 - which ones  you want.
 - which ones you want to drop. 

for that, we should fetch all of the chunks the player can watch but send only needed ones. Otherwise cancel a job could cancel chunks that will be in your field of view. 
This means change a lot of logic in the client connection logic i think, also is not entirely bad, if you decide to go backwards, those chunk will be cached some fractions of a second.
Also with this PR, once you load the file and semi-parse it, retrieve chunks are very fast so cancel the job could be faulty and hurt the performance. (this is my overall opinion for now)

Mili-ssm:
> Also when quiting using CTRL+C it crashes
> 
> ```
> thread 'main' panicked at pumpkin-world/src/chunk/io/chunk_file_manager.rs:309:18:
>                                                                                   We initialize the once cells immediately
>                                                                                                                           note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

That issue is *"known"*, i will push now a change for that. 
This error happens because while you are closing the server some chunks were still loading. i will print warns about that, not crash.

pumpkin-world/src/level.rs
```diff
@@ -219,147 +224,197 @@ impl Level {
...
+                send_chunk(false, value).await;
             });
-        });
+
+            futures::future::join_all(tasks).await;
+        }
+
+        // Finally generate any chunks that are missing
+        if !to_generate.is_empty() {
+            let chunks = to_generate
```
kralverde, line 399: We cannot use rayon function directly from an async function because it will block the async thread until it is done. Can you re-add the streaming you removed in https://github.com/Pumpkin-MC/Pumpkin/pull/535/commits/cf2d2a4c1c1b603b7fda29f166e9b5d831e4e7fa

kralverde (changes requested):

Snowiiii:
> @Snowiiii i saw that is an error from parsing the chunks, when the chunk is not detected as _**Full**_ (the chunk status != _**Full**_).
> 
> > `[ERROR] Failed to load chunk at Vector2 { x: 23, z: -19 }: Failed to parse Chunk from bytes: The chunk isn't generated yet (regenerating)`
> 
> When i use this branch to generate a world i didn't see that error, neither creating a world with master and opening with the branch. Could this be an error that comes from using a pre-existent save or when loading many semi-generated chunks?

And yes I used a pre generated world

pumpkin-world/src/chunk/format/mod.rs
```diff
@@ -42,10 +42,10 @@ impl ChunkData {
             .map_err(|e| ChunkParsingError::ErrorDeserializingChunk(e.to_string()))?;
 
         if chunk_data.x_pos != position.x || chunk_data.z_pos != position.z {
-            return Err(ChunkParsingError::ErrorDeserializingChunk(format!(
+            panic!(
```
Snowiiii, line 45: Not sure about this panicking. A log would be definitely safer

kralverde: Wouldn't this be an implementation error though?

Snowiiii: Normally yes, but this is still risky. A corrupted or broken Chunk would then crash the entire server. Generally we want to avoid panic or unwraps or expect etc. Since a Server is a critical infrastructure. We only want to panic when it's really necessary, for example when failing to bind an TCP address

pumpkin-world/src/level.rs
Mili-ssm: Done with a performance gains actually, thanks for it :D

Snowiiii:
Can someone may create a benchmark how long it takes to load/save a Chunk/Chunk group for master. So we can compare this branch with master

Mili-ssm:
> Can someone may create a benchmark how long it takes to load/save a Chunk/Chunk group for master. So we can compare this branch with master

image

This benchmark is on ***master***, the change is against BulkAPI PR (this PR).

kralverde:
More cpu usage is better (as long as it's not inefficient) because that means we are doing more work in parallel

kralverde:
You need to create a new Level from the root folder for each test, or else it will cache. 

Also you uploaded the test folder, maybe add it to git ignore?

kralverde:
Also you re added the temp dir... read the comment right above it

Mili-ssm:
> You need to create a new Level from the root folder for each test, or else it will cache.
> 
> Also why are you sorting stuff?
> 
> Also you uploaded the test folder, maybe add it to git ignore?

This is because the return order from the fetch is random, so its possible if you only retrieve a part of all of generated chunks that the ones that we use in the benchmark are grouped (for example all of them being part of the same file).

About the cache thing that's the reason i use the `level.clean_memory();`, if not we need to test with more chunks or the benchmark will be more about level instantiation.

Any ways i will fix the thing about temps dirs and test folder thanks, actually this explain somethings i was seeing in my experiments <3 .

kralverde:
Clean memory only happens once if I'm reading it correctly. Also level initialization has no overhead

Mili-ssm:
> You need to create a new Level from the root folder for each test, or else it will cache.
> 
> Also why are you sorting stuff?
> 
> Also you uploaded the test folder, maybe add it to git ignore?

This is because the return order from the fetch is random, so its possible if you only retrieve a part of all of generated chunks that the ones that we use in the benchmark are grouped (for example all of them being part of the same file).

About the cache thing thats the reason i use the `level.clean_memory();`, if not we need to test with more chunks or the benchmark will be more about level instantiation.

> Clean memory only happens once if I'm reading it correctly. Also level initialization has no overhead

Sorry thats true, i missunderstand something in the previus code, the level initialization has almost 0 overhead, and also doing the clean_memory is not enough to avoid some tipe of caching.

Mili-ssm:
These are the result with the updated/fixed benchmark (and running the equivalent one on master).

<details>
    <summary>Master Bench</summary>

image

</details>


<details>
    <summary>BulkApi Bench</summary>

image

</details>

---


<details>
    <summary>Master Bench code</summary>

```Rust

use std::{fs, num::NonZeroU8, path::PathBuf, sync::Arc};

use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
use pumpkin_util::math::vector2::Vector2;
use pumpkin_world::{
    GlobalProtoNoiseRouter, GlobalRandomConfig, NOISE_ROUTER_ASTS, bench_create_and_populate_noise,
    chunk::ChunkData, cylindrical_chunk_iterator::Cylindrical, global_path, level::Level,
};
use tokio::sync::RwLock;

fn bench_populate_noise(c: &mut Criterion) {
    let seed = 0;
    let random_config = GlobalRandomConfig::new(seed);
    let base_router =
        GlobalProtoNoiseRouter::generate(NOISE_ROUTER_ASTS.overworld(), &random_config);

    c.bench_function("overworld noise", |b| {
        b.iter(|| bench_create_and_populate_noise(&base_router, &random_config));
    });
}

async fn test_reads(root_dir: PathBuf, positions: Vec<Vector2<i32>>) {
    let level = Level::from_root_folder(root_dir);
    let (send, mut recv) = tokio::sync::mpsc::channel(positions.len());

    let rt = tokio::runtime::Handle::current();
    level.fetch_chunks(&positions, send, &rt);

    while let Some(x) = recv.recv().await {
        // Don't compile me away!
        let _ = x;
    }
}

async fn test_writes(root_dir: PathBuf, chunks: Vec<(Vector2<i32>, Arc<RwLock<ChunkData>>)>) {
    let level = Level::from_root_folder(root_dir);
    for item in chunks {
        level.write_chunk(item).await;
    }
}

// Depends on config options from `./config`
fn bench_chunk_io(c: &mut Criterion) {
    // System temp dirs are in-memory, so we cant use temp_dir

    let root_dir = global_path!("./bench_root_tmp");
    let _ = fs::remove_dir_all(&root_dir);
    fs::create_dir(&root_dir).unwrap();

    let async_handler = tokio::runtime::Builder::new_current_thread()
        .build()
        .unwrap();

    println!("Initializing data...");
    // Initial writes
    let mut chunks = Vec::new();
    let mut positions = Vec::new();

    async_handler.block_on(async {
        let (send, mut recv) = tokio::sync::mpsc::channel(1024);
        // Our data dir is empty, so we're generating new chunks here
        let level = Level::from_root_folder(root_dir.clone());
        tokio::spawn(async move {
            let cylindrical = Cylindrical::new(Vector2::new(0, 0), NonZeroU8::new(32).unwrap());
            let chunk_positions = cylindrical.all_chunks_within();
            let rt = tokio::runtime::Handle::current();
            level.fetch_chunks(&chunk_positions, send, &rt);
            level.clean_chunks(&chunk_positions).await;
            level.clean_memory(&chunk_positions);
        });
        while let Some((chunk, _)) = recv.recv().await {
            let pos = chunk.read().await.position;
            chunks.push((pos, chunk));
            positions.push(pos);
        }
    });

    chunks.sort_unstable_by_key(|chunk| chunk.0.x * chunk.0.x + chunk.0.z * chunk.0.z);
    positions.sort_unstable_by_key(|pos| pos.x * pos.x + pos.z * pos.z);

    // These test worst case: no caching done by `Level`
    {
        let mut write_group = c.benchmark_group("write_chunks");
        for n_chunks in vec![16, 64, 256] {
            let chunks = &chunks[..n_chunks];
            println!("Testing with {} chunks", n_chunks);
            write_group.bench_with_input(
                BenchmarkId::from_parameter(n_chunks),
                &chunks,
                |b, chunks| {
                    b.to_async(&async_handler)
                        .iter(|| test_writes(root_dir.clone(), chunks.to_vec()))
                },
            );
        }
    }
    {
        let mut read_group = c.benchmark_group("read_chunks");
        for n_chunks in vec![16, 64, 256] {
            println!("Testing with {} chunks", n_chunks);
            let positions = &positions[..n_chunks];
            read_group.bench_with_input(
                BenchmarkId::from_parameter(n_chunks),
                &positions,
                |b, positions| {
                    b.to_async(&async_handler)
                        .iter(|| test_reads(root_dir.clone(), positions.to_vec()))
                },
            );
        }
    }

    fs::remove_dir_all(&root_dir).unwrap();
}

criterion_group!(benches, bench_populate_noise, bench_chunk_io);
criterion_main!(benches);
```

</details>

Snowiiii:
Setting `view_distance` to 32 crashes
```
$
  thread 'rayon-worker' panicked at pumpkin-world/src/chunk/format/anvil.rs:219:42:
                                                                                   range end index 8188 out of range for slice of length 8187
```

kralverde:
Should have a fix for that coming up, I'm also working on a bit of a speed-up

kralverde:
> > Setting `view_distance` to 32 crashes
> > ```
> > thread 'rayon-worker' panicked at pumpkin-world/src/chunk/format/anvil.rs:219:42:
> > range end index 8188 out of range for slice of length 8187
> > ```
> 
> @kralverde i think this is caused by the -1 +1 on the anvil chunk bytes length (its a thing about align/padding, if you have bad luck and the chunk have an exact padding size, with +1 it will add 4096 empty bytes, that could make the broken slice). [ resume: anvil format specify the last byte of every chunk to be _empty_, and then make an align of all of the chunk data to `4096 bytes` ]
> 
> if every thing is okey i can commit that fix.
> 
> (also sorry for the delay i am doing some benchmarks, there are thing making me uncomfortable (about inconsistences on the benches, i am trying to let everything well done and documented with out finetune to the BulkAPI)

I'd prefer you don't push anything because I have a fix + something else im working on right now

kralverde (approved):

kralverde:
Alright @Snowiiii what do you think?

Snowiiii:
Something makes the client really laggy when loading a huge amount of Chunks at once
image
I also got 
`[ERROR] Failed to load chunk at Vector2 { x: -18, z: -54 }: Compression error Error while working with LZ4 compression: LZ4 error: ERROR_blockChecksum_invalid (regenerating)` Once

Snowiiii:
Also when quitting i'm getting warns
```
[WARN] Closing FileManager while the File ./world/region/./r.-25.-1.mca is being loaded
[WARN] Closing FileManager while the File ./world/region/./r.-25.-2.mca is being loaded
[WARN] Closing FileManager while the File ./world/region/./r.-26.-2.mca is being loaded
```

kralverde:
@Snowiiii could you retest?

pumpkin/src/world/mod.rs
```diff
@@ -112,6 +109,7 @@ impl PumpkinError for GetBlockError {
 pub struct World {
     /// The underlying level, responsible for chunk management and terrain generation.
     pub level: Arc<Level>,
+    single_chunk_lfu: Mutex<[Vector2<i32>; 16]>,
```
Snowiiii, line 112: This looks weird tbh, I would like at least some documentation

Snowiiii:
Looks good. I'm fine with this being merged

PR #536, opened 2025-02-10
Add new plugin events / change plugin organization

ThomasTartrau:
## Description

- Added new plugin events to enhance 
- Refactored the plugin structure for better organization and 
- Updated documentation and adjusted internal logic to reflect these changes.

**This PR improves modularity and allows for easier integration of new features in the future.**

## Testing


pumpkin/src/plugin/api/events/player/player_join.rs
```diff
@@ -0,0 +1,44 @@
...
+            cancelled: false,
+        }
+    }
+}
+
+impl PlayerEvent for PlayerJoinEvent {
+    fn get_player(&self) -> &Arc<Player> {
+        &self.player
+    }
+}
```
Snowiiii, line 44: Why you have a player getter here ?

Snowiiii:
> Honestly, I'd almost be against making a single file for each event, I understand that it is one way to have better organization, but considering the number of events that need to be implemented, this would result in a huge resulting number of files. Personally, I would prefer grouping similar events together into a single file, for example the BlockPlace, BlockCanBuild and such could be put together.

Maybe we can group some packets into one file with the same category, (e.g. Block Events). But i definitely not want to have everything in an mod.rs file, We also create a new files for every packet, I would much rather prefer to do this here as well.

Snowiiii:
Overall the refactor seems fine and i like the new structure more than the old one . Thanks @ThomasTartrau :+1:

PR #538, opened 2025-02-10
Terminal and exit improvements

DarkMrMelther:
## Description
Solved the bad handling of Ctrl + C to exit as intended, also combined all the stop behaviors into "handle_stop".
Added history to terminal, currently only saved on Ctrl +C/D exits.

### DIFF
--- a/pumpkin/src/server/mod.rs
+++ b/pumpkin/src/server/mod.rs
@@ -12,6 +12,7 @@ use pumpkin_util::math::boundingbox::{BoundingBox, EntityDimensions};
 use pumpkin_util::math::position::BlockPos;
 use pumpkin_util::math::vector2::Vector2;
 use pumpkin_util::math::vector3::Vector3;
+use pumpkin_util::text::color::NamedColor;
 use pumpkin_util::text::TextComponent;
 use pumpkin_world::block::registry::Block;
 use pumpkin_world::dimension::Dimension;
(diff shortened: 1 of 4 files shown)


Yimura:
Which stop signals have you tested? I see you've removed them and instead are relying on `ReadlineError` to handle SIGINT and end of input stream (CTRL + D).

I don't see how `SIGTERM` or `SIGKILL` are captured at all?
For example when you're stopping a docker container it'll send a `SIGTERM`, which I do not see handled in your code.

pumpkin/src/server/mod.rs
```diff
@@ -476,4 +484,28 @@ impl Server {
...
+        log::warn!(
+            "{}",
+            TextComponent::text("Received interrupt signal; stopping server...")
+                .color_named(NamedColor::Red)
+                .to_pretty_console()
+        );
+
+        // TODO: Gracefully stop
+        let kick_message = kick_msg
+            .unwrap_or_else(|| TextComponent::text("Server stopped").color_named(NamedColor::Red));
```
Snowiiii, line 498: you should use translations, `commands.stop.stopping`

DarkMrMelther: That translation is for the command, it should not be used for the kick message, because the server is already stopped, not "stopping".
image

Snowiiii: Upsi, it was 
`multiplayer.disconnect.server_shutdown`

DarkMrMelther:
@Snowiiii as you see I added the translated kick message, but I was wondering if we should change to [rustyline-async](https://github.com/zyansheep/rustyline-async) to prevent the cursor from being left behind the system messages. What do you think?

Snowiiii:
> @Snowiiii as you see I added the translated kick message, but I was wondering if we should change to [rustyline-async](https://github.com/zyansheep/rustyline-async), or any async alternative to rustyline to prevent the cursor from being left behind the system messages. What do you think?
> 
> Edit: I've found [this](https://github.com/kkawakam/rustyline/pull/200/commits/b496a2b2616209ce679abd289567c34782275b53) commit of rustyline implementing it on a async context.

I'm fine with using a different crate, But i would generally prefer to use the main rustyline crate due to faster Updates. So if its really necessary you can use an other crate if not i would like to see an implementation with the existing rustyline crate

Snowiiii:
Fixed in https://github.com/Pumpkin-MC/Pumpkin/commit/a0d3a71ba8b0beba68c007ec09c8a7602936d0c0

PR #539, opened 2025-02-10
Fix HoverEvent::show_entity calls

yarml:
## Description
Give valid entity type to HoverEvent::show_entity calls.

Previously, executing a command such as /give, /msg, etc would cause the client to close connection because of an invalid entity type in HoverEvent within the chat message sent to the player.

Modified pumkin-data to expose the resource name of entities.
Also made the `kind` parameter not `Option`, because a value of `None` also causes the client to crash.

## Previous Behavior
https://github.com/user-attachments/assets/4f514ad9-27da-478a-b311-881bd84936cb

## Behavior with Fix
https://github.com/user-attachments/assets/b0a64455-6c03-4f95-a984-976155aaed5c

### DIFF
--- a/pumpkin-data/build/entity_type.rs
+++ b/pumpkin-data/build/entity_type.rs
@@ -18,7 +18,7 @@ pub struct EntityType {
 
 pub struct NamedEntityType<'a>(&'a str, &'a EntityType);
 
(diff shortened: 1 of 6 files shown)


Snowiiii:
Awesome!. Thank you @yarml :heart:

PR #550, opened 2025-02-13
Make the players field in World use RwLock instead of Mutex

vyPal:
## Description
This PR changes the `players` field in the `World` struct:
```rs
pub struct World {
    /// A map of active players within the world, keyed by their unique UUID.
    pub players: Arc<Mutex<HashMap<uuid::Uuid, Arc<Player>>>>,
}
```
from a `Mutex` to a `RwLock`.

This fixes a couple deadlocks which can cause the player to time out and be unable to join the server again. Specifically any function called from within the `player.tick()` method which also broadcasted packets to players on the server would cause a deadlock as the `world.broadcast_packet_all` method (and similar) require a lock to the `players` mutex, but this mutex is already locked by the `world.lock()` method which calls `player.tick()`.

This change also theoretically helps prevent other similar deadlocks, as the only cases where mutable access to the `players` mutex is necessary are when a player join, leaves, or teleports between worlds, all other cases can be called on a `read()` lock.

## Testing

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -635,7 +635,7 @@ impl Player {
         let uuid = self.gameprofile.id;
         current_world.remove_player(self.clone(), false).await;
         *self.living_entity.entity.world.write().await = new_world.clone();
-        new_world.players.lock().await.insert(uuid, self.clone());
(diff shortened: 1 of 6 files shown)


Snowiiii:
Looks good. Thank you @vyPal

PR #554, opened 2025-02-16
Fix Item Tool Components and Partially Implement Tag Updates

kralverde:
## Description
This PR partially implements tag updates to allow for block tags to be sent to the client. It also uses tags for tool components instead of a list of what the tags reference. With this, tools properly break blocks with different speeds.

With a `Fluid` registry extractor, this will also fix the "no swimming" issue in water.

## Testing


pumpkin-data/src/tag.rs
```diff
@@ -38,15 +54,15 @@ pub fn get_tag_values(tag_category: RegistryKey, tag: &str) -> Option<&Vec<Optio
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
 pub enum TagType {
-    Item(String),
+    Block(String),
```
Snowiiii, line 57: Why you renamed it block ?. Item does not mean an actually Minecraft item, but rather just a non Tag

kralverde: You are right, I'll change that now

kralverde:
Also added a fluid extractor with https://github.com/Pumpkin-MC/Extractor/pull/6 so swimming works

Snowiiii:
Nice. Great work @kralverde :+1:

PR #557, opened 2025-02-17
Gracefully exit on stop, new logger and rustyline

kralverde:
## Description
This PR gracefully exits on a `SIGXXX` signal or `stop` command. 

Rustyline needed to be replaced with an async version because the originally was blocking and could not be interrupted, leaving the terminal in raw mode after the server terminated. 

The logger needed to be updated to a logger that supports files to work with the new rusty line to pretty print logs while still having the user input.

## Testing

### DIFF
--- a/pumpkin-world/src/level.rs
+++ b/pumpkin-world/src/level.rs
@@ -97,6 +97,7 @@ impl Level {
         log::info!("Saving level...");
 
         // chunks are automatically saved when all players get removed
+        // TODO: Await this ^
 
         // then lets save the world info
         self.world_info_writer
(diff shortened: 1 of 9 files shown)


Snowiiii:
Would be great if you could somehow get spawn chunks to get saved as well

kralverde:
Spawn chunks should be spawned now @Snowiiii

Snowiiii:
I noticed that log level is not colorful anymore, Any way to get the color back ?

Snowiiii:
I'm also not fully sure about simplelog, Are there any other alternatives we can use ?

kralverde:
> I noticed that log level is not colorful anymore, Any way to get the color back ?

Yeah, it has color settings

kralverde:
> I'm also not fully sure about simplelog, Are there any other alternatives we can use ?

Any logger that supports writing to file should work, but this is the only one I've found

kralverde:
Colors should work now @Snowiiii

PR #559, opened 2025-02-17
Differentiate between single and stack drops

4lve:
## Description
Add support for `DropItem` and `DropItemStack`. (Also prevents item dupe glitch where it would always drop the stack but only subtract a single item)

## Testing


Snowiiii:
LGTM. Thank you @4lve :rocket:

PR #560, opened 2025-02-18
Permissions support

zp3x:
## Description
Adds support for traditional permissions.

> `"example.command"`

For larger servers, a robust permission system is essential, ideally one that supports existing solutions already using the traditional permissions format.

### Changes

- Allows setting, removing, and retrieving player permissions.
- Validates permissions upon command execution.

## Testing


zp3x:
Just found out that when no commands are registered the Server Crashes on /help

Messed the permission Level and Operator Status on /help.

Fixing that

vyPal:
One thing I'd maybe consider implementing is namespaces for these permissions. Specifically for use with plugins so that there aren't any collisions. Each plugin has a plugin name, which has to conform to the Cargo naming system.
 I'd recommend using this as either the base node of the permission tree for each plugin (so permissions registered by each plugin would start with `plugin_name.some.permission`) or as a standard namespace (so each permission registered by a plugin would look more like `plugin_name:some.permission`)

In this way we can prevent conflicts in permissions (like if two plugins both wanted to register a `user.teleport` permission and use it for different things)

vyPal:
One thing I'd maybe consider implementing is namespaces for these permissions. Specifically for use with plugins so that there aren't any collisions. Each plugin has a plugin name, which has to conform to the Cargo naming system.
 I'd recommend using this as either the base node of the permission tree for each plugin (so permissions registered by each plugin would start with `plugin_name.some.permission`) or as a standard namespace (so each permission registered by a plugin would look more like `plugin_name:some.permission`)

In this way we can prevent conflicts in permissions (like if two plugins both wanted to register a `user.teleport` permission and use it for different things)

zp3x:
That's actually a very good idea. Will definitely implement and leave a option to enable it

DataM0del (changes requested):
default_dispatcher's code, is... :vomiting_face:

Snowiiii:
Why you created this AtomicLinkedList thing ?, I see you use it for save new Permissions, Can't you just use a RwLock or a Mutex Vec ?

zp3x:
I could try without but it works.

kralverde:
I would recommend you go without the atomic linked list

Snowiiii:
Feel free to reopen

Issue #546, opened 2025-02-12
Picking up items doesn't put them into your hotbar first


Pookachu:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch OS 

### Server Software Version/Commit

Pumpkin 0.1.0-dev (73d3fe0)

### What happened?

I broke a stone block with my fists and it dropped, and was placed into the top left of my inventory instead of the first open inventory slot.

### To Reproduce

Pickup any item.

### Expected behavior

It should have been placed in my first available hotbar slot instead of my first available inventory slot.

Issue #534, opened 2025-02-07
Item duplication glitch


RobIux:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows

### Server Software Version/Commit

2e2fe83

### What happened?

Items can be duplicated within minutes to an above max_stack_number and have multiple thousands of items dropped on ground causing lag issues.

## Steps to reproduce
Give yourself 64 stone
Use ctrl +q to drop said stone
Pick up stone and u should have 127
Hold down ctrl + q and your stone will infinitly duplicate>

## Unrelated issue that might be helpfull for fixing this
Give yourself 2 totems
Drop 1 totem
Totem that gets picked up ignore max_stack_size and turns 1 totem into 2 totems
Press q on the totem
You have now duplicated the totem

### Expected behavior

Item giving correct amount back after pickup

RobIux:
> The title should rather be: "Picking up items does not respect the maximum amount of material being picked up"

no it duplciates the items

4lve:
#559 Should fix this

PR #562, opened 2025-02-18
Implement item stacking

4lve:
## Description

What's added:
 - Picking up items prioritizes your hotbar
 - Stacking works when picking up items. (Before it would just add to your current stack allowing you to have more than stack limit  items)

Fixes:
#546 
#534 

## Testing

### DIFF
--- a/pumpkin/src/entity/item.rs
+++ b/pumpkin/src/entity/item.rs
@@ -1,6 +1,10 @@
-use std::sync::{atomic::AtomicI8, Arc};
+use std::sync::{
+    atomic::{AtomicI8, AtomicU8},
+    Arc,
+};
 
 use async_trait::async_trait;
+use pumpkin_data::item::Item;
 use pumpkin_protocol::{
     client::play::{CTakeItemEntity, MetaDataType, Metadata},
     codec::{slot::Slot, var_int::VarInt},
@@ -13,7 +17,7 @@ pub struct ItemEntity {
     entity: Entity,
     item: Slot,
     id: u16,
-    count: u8,
+    count: AtomicU8,
     pickup_delay: AtomicI8,
 }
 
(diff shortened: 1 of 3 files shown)


pumpkin/src/entity/item.rs
```diff
@@ -49,23 +53,56 @@ impl EntityBase for ItemEntity {
             // Check if we have space in inv
             if let Some(slot) = inv.collect_item_slot(self.id) {
                 let mut item = self.item.clone();
+                let max_stack = Item::from_id(self.id)
+                    .unwrap_or(Item::AIR)
+                    .components
+                    .max_stack_size;
```
Snowiiii, line 59: We should rather save the Item/Itemstack somewhere instead of this

4lve: That was just a catch case, a dropped item not having a valid item id should never occur

4lve: Ye, good idea

4lve: Not sure from where though, as storing the item itself on the entity would take up memory. Having the item being created runtime might be the best option

4lve: Even `Slot::to_item` uses `Item::from_id`

Snowiiii:
I changed some of the code, this is definitely faster now. Only found a bug that item picking is now broken, Currently don't have time maybe you can fix that ?

4lve:
Thanks, gonna take a look

4lve:
You can item pick now but there's still something odd about it, will have to take another look later

4lve:
Pickblock is now fixed

Snowiiii:
In Vanilla the Item stack you currently holding is prioritized, this means when you have a free slot at the first hotbar slot but already holding stone in an other slot while picking up stone you get the stone in the current held slot

4lve:
Fixed

Snowiiii:
Looks good now. Thank you @4lve! :+1:

PR #565, opened 2025-02-19
Fix deadlock when "packet ready" buffer is full

kralverde:
## Description
When the send packet function is invoked on the client, the lock on the encrypter was unnecessarily held for the remainder of the function. When a notification was sent to the `server_packets_channel`, the encrypter lock was still held, preventing the serialized packet from being taken and blocking the next call to the send packet function

## Testing
You can verify this worked by teleporting to the world border.
Before: Deadlock
Now: Normally chunk loading


Snowiiii:
Nice. Thank you @kralverde 🚀

PR #583, opened 2025-02-23
fix: encode Uuid correctly (fixes #454)

### LINKED ISSUE #454: Resource packs broken

### Operating System

Client is MacOS Sequoia, server is Ubuntu

0.1.0-dev f2cfad0

### What happened?

the client could not connect because of a bad resource pack packet, if the resource pack is enabled, otherwise everything* works fine

### To Reproduce

[resource_pack]
enabled = true
resource_pack_url = "https://memetech.ca/minecraft/pack.zip"
resource_pack_sha1 = "cedaa3d98df8e37a71dd960e23496c0dba660610"
prompt_message = "You should accept this"
force = true

### Expected behavior

As a vanilla server would with resource packs enabled - the client prompts to install the resource pack, installs it successfully (unless it's a 1.20 pack for some reason, but that's not a Pumpkin bug so I won't report it here), and then I get blue hearts in the health bar or whatever I put in that pack

LooFifteen:
## Description
Fixes #454 by properly configuring `uuid::Uuid` serialization. This also removes the need for `#[serde(with = "uuid::serde::compact")]` across the codebase.

## Testing
`features.toml`
```toml
[resource_pack]
enabled = true
url = "http://localhost:8080/resourcepack.zip"
sha1 = "abc123"
message = "test message"
force = true
```

### DIFF
--- a/pumpkin-config/src/resource_pack.rs
+++ b/pumpkin-config/src/resource_pack.rs
@@ -5,25 +5,32 @@ use serde::{Deserialize, Serialize};
 pub struct ResourcePackConfig {
     pub enabled: bool,
     /// The path to the resource pack.
-    pub url: String,
+    pub resource_pack_url: String,
     /// The SHA1 hash (40) of the resource pack.
-    pub sha1: String,
+    pub resource_pack_sha1: String,
     /// Custom prompt Text component, Leave blank for none
-    pub message: String,
+    pub prompt_message: String,
     /// Will force the Player to accept the resource pack
     pub force: bool,
 }
 
 impl ResourcePackConfig {
     pub fn validate(&self) {
+        if !self.enabled {
+            return;
+        }
+
         assert_eq!(
-            !self.url.is_empty(),
-            !self.sha1.is_empty(),
+            !self.resource_pack_url.is_empty(),
+            !self.resource_pack_sha1.is_empty(),
             "Resource Pack path or Sha1 hash is missing"
         );
+
+        let hash_len = self.resource_pack_sha1.len();
         assert!(
-            self.sha1.len() <= 40,
-            "Resource pack sha1 hash is too long (max. 40)"
+            hash_len == 40,
+            "Resource pack sha1 hash is the wrong length (should be 40, is {})",
+            hash_len
         )
     }
 }
--- a/pumpkin-protocol/src/bytebuf/serializer.rs
+++ b/pumpkin-protocol/src/bytebuf/serializer.rs
@@ -219,6 +219,9 @@ impl<B: BufMut> ser::Serializer for &mut Serializer<B> {
         self.output.put_var_int(&variant_index.into());
         Ok(())
     }
+    fn is_human_readable(&self) -> bool {
+        false
+    }
 }
 
 impl<B: BufMut> ser::SerializeSeq for &mut Serializer<B> {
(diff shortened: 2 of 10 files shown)


pumpkin-protocol/src/client/play/spawn_entity.rs
```diff
@@ -9,7 +9,6 @@ use crate::VarInt;
 #[packet(PLAY_ADD_ENTITY)]
 pub struct CSpawnEntity {
     entity_id: VarInt,
-    #[serde(with = "uuid::serde::compact")]
```
kralverde, line 12: I think we should keep the compact, from what I understand this only works because of the underlying UUID data is bytes and our current implementation of serializer skips newtype structs. If that were to change this would break. I think its good to be explicit about how we want to serialize stuff from crates out of our control. Unless there's something I'm not seeing here?

pumpkin-protocol/src/client/play/player_chat_message.rs
```diff
@@ -9,7 +9,6 @@ use crate::{VarInt, codec::bit_set::BitSet};
 #[derive(Serialize)]
 #[packet(PLAY_PLAYER_CHAT)]
 pub struct CPlayerChatMessage<'a> {
-    #[serde(with = "uuid::serde::compact")]
```
kralverde, line 12: Here too

kralverde (changes requested):

pumpkin-protocol/src/bytebuf/serializer.rs
```diff
@@ -219,6 +219,9 @@ impl<B: BufMut> ser::Serializer for &mut Serializer<B> {
         self.output.put_var_int(&variant_index.into());
         Ok(())
     }
+    fn is_human_readable(&self) -> bool {
+        false
+    }
```
LooFifteen, line 224: `#[serde(with = "uuid::serde::compact")]` does the equivalent of what's been changed here. As all data typed written by the serializer are not 'human readable', this removes the need to specify 'compacting' any of them.

pumpkin-protocol/src/client/play/spawn_entity.rs
LooFifteen: https://github.com/Pumpkin-MC/Pumpkin/pull/583#discussion_r1966848656

pumpkin-protocol/src/bytebuf/serializer.rs
kralverde: Uuid's deserialization does not depend on `is_human_readable`. I agree that it is good to set false, but I'm saying right now Uuid is defined as `Uuid(Bytes)`. Right now, our serializer skips newtype structs making it serialize bytes. Its a happy accident that our serialization matches the `serde::compact` serialization. If we change our serializer or Uuid changes it's internal implementation for some reason it will break.

This is all uuid::serde::compact is doing:
```
        serde::Serialize::serialize(u.as_bytes(), serializer)
```

Explicitly doing what we are doing implicitly based on our and uuid's current implementation

pumpkin-config/src/resource_pack.rs
```diff
@@ -5,25 +5,32 @@ use serde::{Deserialize, Serialize};
...
     pub enabled: bool,
     /// The path to the resource pack.
-    pub url: String,
+    pub resource_pack_url: String,
     /// The SHA1 hash (40) of the resource pack.
-    pub sha1: String,
+    pub resource_pack_sha1: String,
     /// Custom prompt Text component, Leave blank for none
-    pub message: String,
+    pub prompt_message: String,
```
Snowiiii, line 12: I don't like this, The Category is already named Resource so this is just duplicate naming

LooFifteen: cc @kralverde

kralverde: I'll revert

PR #588, opened 2025-02-24
Add /defaultgamemode

hthug06:
## Description
Add /defaultgamemode

## Testing
With force-gamemode: false in configuration.toml:
- Do /defaultgamemode creative, exit the server and log in. You are now in creative

With force-gamemode: true in configuration.toml:
- Do /defaultgamemode creative, the gamemode change for every player in the server. Change your gamemode to survival, exit the server and log in. Your are now in creative


rasgajag:
I really like this command, but I have a few thoughts on it. Within the idea of Vanilla Parity, this does it very well. But I also believe the permission level of this command in Vanilla is too low, and should be placed in permission level 3 instead of 2. This would bring it inline with other multiplayer management commands usually reserved for this permission level. For example:
[/setidletimeout](https://minecraft.wiki/w/Commands/setidletimeout) which is another command used for player/world management.

kralverde:
> I really like this command, but I have a few thoughts on it. Within the idea of Vanilla Parity, this does it very well. But I also believe the permission level of this command in Vanilla is too low, and should be placed in permission level 3 instead of 2. This would bring it inline with other multiplayer management commands usually reserved for this permission level. For example:
> 
> [/setidletimeout](https://minecraft.wiki/w/Commands/setidletimeout) which is another command used for player/world management.

I disagree; in my eyes this is no different than just running `gamemode` for every player

Snowiiii:
Looks good to me. Thank you @hthug06 :+1:

PR #591, opened 2025-02-25
 Fix dupe, doubble click and shift click (re-PR)

kralverde:
## Description
I don't know what I did, but see https://github.com/Pumpkin-MC/Pumpkin/pull/577

## Testing


kralverde:
@4lve alternatively, if you're okay with it, we can just use this PR

kralverde:
@Snowiiii can you dummy check my work?

4lve:
Ye, we could use this

Snowiiii:
Seems like something is wrong, Now the hotbar is not being prioritized

4lve:
Maybe he messed up some constants

Snowiiii:
Also it would be great to have Tags parsed instead of strings

kralverde:
> Seems like something is wrong, Now the hotbar is not being prioritized

In what case was it not prioritized?

Snowiiii:
> > Seems like something is wrong, Now the hotbar is not being prioritized
> 
> In what case was it not prioritized?

If picking up items it will first try to put it in your current selected horbar slot, and if can't in an other hotbar slot

kralverde:
@Snowiiii Think its all good now, could you retest?

OfficialKris:
When picking up items (/give) with multiple stacks it always skips hotbar 1-4 and even goes to the off hand after the hotbar is full. Also trying to open chests gives "Failed to get Block Properties mapping for chest."

4lve:
> When picking up items (/give) with multiple stacks it always skips hotbar 1-4 and even goes to the off hand after the hotbar is full. Also trying to open chests gives "Failed to get Block Properties mapping for chest."

Could you check if these issues exsist in my latest commit in this PR?

OfficialKris:
> Could you check if these issues exsist in my latest commit in this PR?

Are there new commits? I tested on both master and this branch. Works on master not on this branch.

4lve:
> > Could you check if these issues exsist in my latest commit in this PR?
> 
> 
> 
> Are there new commits? I tested on both master and this branch. Works on master not on this branch.
> 
> 

There aren't any new commits but Check the old commit by me here, iirc everything worked as it should

OfficialKris:
Yes, your latest commit works well. The latest in this PR has a few bugs (maybe due to other changes in master).

4lve:
> Yes, your latest commit works well. The latest in this PR has a few bugs (maybe due to other changes in master).

Hmm okay, maybe we should revert to my commit and reimplement the constants from there

kralverde:
> > Yes, your latest commit works well. The latest in this PR has a few bugs (maybe due to other changes in master).
> 
> Hmm okay, maybe we should revert to my commit and reimplement the constants from there

I'd rather not, I've done a lot of refactoring + fixing bad code to redo it all, though I do admit that this PR has definitely ballooned out of the original scope. The `/give` and container deadlocks should be fixed, the prior was due to a hidden bug that was brought to light where item stacks arent set to none when their count is 0 and the latter was a faux pas on my part. 

I really, really want to get rid of all of these mutexs but that will have to be a different PR i think.

4lve:
> > > Yes, your latest commit works well. The latest in this PR has a few bugs (maybe due to other changes in master).
> > 
> > 
> > Hmm okay, maybe we should revert to my commit and reimplement the constants from there
> 
> I'd rather not, I've done a lot of refactoring + fixing bad code to redo it all, though I do admit that this PR has definitely ballooned out of the original scope. The `/give` and container deadlocks should be fixed, the prior was due to a hidden bug that was brought to light where item stacks arent set to none when their count is 0 and the latter was a faux pas on my part.
> 
> I really, really want to get rid of all of these mutexs but that will have to be a different PR i think.

For sure, as long as we can fix the issues

OfficialKris:
Works well now, maybe there will be some other bug reports but already significantly better.

pumpkin/src/entity/item.rs
```diff
@@ -1,37 +1,38 @@
...
-    pickup_delay: AtomicI8,
+    // These cannot be atomic values because we mutate their state based on what they are; we run
+    // into the ABA problem
+    item_count: Mutex<u32>,
+    pickup_delay: Mutex<u8>,
 }
 
 impl ItemEntity {
-    pub fn new(entity: Entity, stack: ItemStack) -> Self {
+    pub fn new(entity: Entity, item_id: u16, count: u32) -> Self {
```
Snowiiii, line 25: Not sure about this since we later want to apply components and stuff like enchatments

Snowiiii:
Now the Selected slot and hotbar are prioritised, Nice

pumpkin/src/entity/item.rs
kralverde: It can be changed later, I just want to differentiate the entity and item stacks because the entities can have a much larger item count

DataM0del:
doubble :fire:

kralverde:
@Snowiiii if I fix conflicts, are you good with this going to master?

Snowiiii:
> @Snowiiii if I fix conflicts, are you good with this going to master?

Yes. LGTM

PR #600, opened 2025-03-01
feat(player-data): Support saving and loading player NBT Data

HookWoods:
## Description

This PR adds the must-have feature to save and load player NBT Data !

This code actually does:
- Add a util class to pumpkin-nbt to compress and decompress NBT to GZip which is useful in multiple places in Minecraft code.
- Load or create player data on join
- Cache player data to avoid having to read the player file again
- Save player data on leave
- Save every player's data on server shutdown
- Save player data in a scheduler to avoid massive data loss in case of a crash
- Add a config to control if player data saving should be enabled
- Add some working content to player NBT

This code also contains a rework of the experience packet and health packet with `tick_experience` and `tick_health`; this is needed with how NBT data are handled before player spawning in the world. Like that, we send experience, health, ... directly on the first player tick and send them when there is an update like before.

I also added some data to NBT, like gamemode, or food, and impl them in the code to use them now instead of default data. Now the future part needed for this code is the implementation of other NBT values accessible from [https://minecraft.wiki/w/Player.dat_format](https://minecraft.wiki/w/Player.dat_format)

## Testing

I have tested to move on the map, give experience, change gamemode, ... stopping the server and joining again and all is working good. May need more test from others !


pumpkin-config/src/player_data.rs
```diff
@@ -0,0 +1,22 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize, Serialize)]
+#[serde(default)]
+pub struct PlayerDataConfig {
+    /// Is Player Data saving enabled?
+    pub save_player_data: bool,
+    /// Is Player Data should be cached?
+    pub cache_player_data: bool,
```
kralverde, line 9: I don't think we need to cache player data, just read it back from disk when they rejoin

pumpkin-nbt/src/nbt_compress.rs
```diff
@@ -0,0 +1,330 @@
...
+    // Decompress the data
+    let mut decompressed_data = Vec::new();
+    decoder
+        .read_to_end(&mut decompressed_data)
+        .map_err(Error::Incomplete)?;
+
+    // Read the NBT data
+    let nbt = Nbt::read(&mut deserializer::ReadAdaptor::new(&decompressed_data[..]))?;
+    Ok(nbt.root_tag)
+}
```
kralverde, line 29: This should take an `impl Read`, then you can:
```
Nbt::read(&mut ReadAdaptor::new(GzDecoder::new(input)))
```
and skip having to buffer the data.

pumpkin-nbt/src/nbt_compress.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    // Then compress it with GZip
+    let mut compressed_data = Vec::new();
+    {
+        let mut encoder = GzEncoder::new(&mut compressed_data, Compression::default());
+        encoder.write_all(&serialized).map_err(Error::Incomplete)?;
+        encoder.finish().map_err(Error::Incomplete)?;
+    }
+
+    Ok(compressed_data)
```
kralverde, line 55: You should add a `write_to(impl Write` for Nbt and chain the writers like the readers above to avoid having to write to a buffer.

pumpkin-nbt/src/nbt_compress.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    // Decompress the data
+    let mut decompressed_data = Vec::new();
+    decoder
+        .read_to_end(&mut decompressed_data)
+        .map_err(Error::Incomplete)?;
+
+    // Deserialize the NBT data
+    deserializer::from_bytes(&decompressed_data[..])
+}
```
kralverde, line 84: Don't write to a buffer here, just chain the readers

pumpkin-nbt/src/nbt_compress.rs
```diff
@@ -0,0 +1,330 @@
...
+    {
+        let mut encoder = GzEncoder::new(&mut compressed_data, Compression::default());
+        encoder
+            .write_all(&uncompressed_data)
+            .map_err(Error::Incomplete)?;
+        encoder.finish().map_err(Error::Incomplete)?;
+    }
+
+    Ok(compressed_data)
+}
```
kralverde, line 116: Don't write to a buffer here, just chain the writers.

pumpkin/src/data/player_data.rs
```diff
@@ -0,0 +1,492 @@
...
+
+/// Manages the storage and retrieval of player data from disk and memory cache.
+///
+/// This struct provides functions to load and save player data to/from NBT files,
+/// with a memory cache to handle player disconnections temporarily.
+pub struct PlayerDataStorage {
+    /// Path to the directory where player data is stored
+    data_path: PathBuf,
+    /// In-memory cache of recently disconnected players' data
+    cache: Mutex<HashMap<Uuid, (NbtCompound, Instant)>>,
```
kralverde, line 29: Does this really need to be a mutex? Can you have `PlayerDataStorage` be mutable when you need to modify it?

pumpkin/src/data/player_data.rs
```diff
@@ -0,0 +1,492 @@
...
+}
+
+/// Helper for managing player data in the server context.
+///
+/// This struct provides server-wide access to the `PlayerDataStorage` and
+/// convenience methods for player handling.
+pub struct ServerPlayerData {
+    storage: Arc<PlayerDataStorage>,
+    save_interval: Duration,
+    last_cleanup: Mutex<Instant>,
```
kralverde, line 351: Do these need to be mutexs? can server player data be mutable when you need to modify them?

pumpkin/src/error.rs
```diff
@@ -65,3 +66,21 @@ impl PumpkinError for ReadingError {
...
+
+    fn severity(&self) -> log::Level {
+        log::Level::Warn
+    }
+
+    fn client_kick_reason(&self) -> Option<String> {
+        match self {
+            Self::Io(err) => Some(format!("Failed to load player data: {err}")),
+            Self::Nbt(err) => Some(format!("Failed to parse player data: {err}")),
+            Self::NotFound(uuid) => Some(format!("Player data not found for UUID: {uuid}")),
```
kralverde, line 83: If the player data is not found, shouldn't a new file be created?

pumpkin/src/server/mod.rs
```diff
@@ -137,6 +139,12 @@ impl Server {
             server_listing: Mutex::new(CachedStatus::new()),
             server_branding: CachedBranding::new(),
             bossbars: Mutex::new(CustomBossbars::new()),
+            player_data_storage: ServerPlayerData::new(
+                "./world/playerdata",      // TODO: handle world name in config
```
kralverde, line 143: You should make a helper function for getting the world name and feed that here and where the chunk loading gets the world name

kralverde (changes requested):

Snowiiii:
Why not move the nbt data into the hunger manager

pumpkin-nbt/src/nbt_compress.rs
```diff
@@ -0,0 +1,381 @@
...
+    #[test]
+    fn test_direct_file_io() {
+        use std::fs::File;
+        use std::path::Path;
+
+        let mut compound = NbtCompound::new();
+        compound.put_int("test_value", 12345);
+
+        // Create a temporary file path
+        let path = Path::new("test_nbt_file.dat");
```
kralverde, line 363: You should use the temp file crate for this

pumpkin/src/server/mod.rs
```diff
@@ -137,6 +141,10 @@ impl Server {
             server_listing: Mutex::new(CachedStatus::new()),
             server_branding: CachedBranding::new(),
             bossbars: Mutex::new(CustomBossbars::new()),
+            player_data_storage: ServerPlayerData::new(
+                format!("./{world_name}/playerdata"),
```
kralverde, line 145: I'm not comfortable with this being a solo relative path. I know that our world is relative to where we execute the server, but having "magic string" like this, even if is correct now can cause issues in the future if there are refactors. 

Could you make a `get_world_path` or something on the config instance, then pass that here and to our world loading?

kralverde (changes requested):
Why do we now save player data periodically? Isn't saving when players leave enough?

DataM0del:
> Why do we now save player data periodically? Isn't saving when players leave enough?

What if the program panics / crashes, and we can't handle leaving or the server shutting down?
What if the power is unexpectedly unplugged while the server is running?
Etc.

kralverde (approved):

pumpkin-world/src/data/player_data.rs
```diff
@@ -0,0 +1,179 @@
...
+            return Ok(());
+        }
+
+        let path = self.get_player_data_path(uuid);
+
+        // Run disk I/O in a separate tokio task
+        let uuid_copy = *uuid;
+        let data_clone = data;
+
+        match tokio::spawn(async move {
```
kralverde, line 134: Why are we spawning a tokio task here? No async code is used

pumpkin-world/src/data/player_data.rs
```diff
@@ -0,0 +1,179 @@
...
+        // If not in cache, load from disk
+        let path = self.get_player_data_path(uuid);
+        if !path.exists() {
+            log::debug!("No player data file found for {}", uuid);
+            return Ok(NbtCompound::new());
+        }
+
+        // Offload file I/O to a separate tokio task
+        let uuid_copy = *uuid;
+        let nbt = tokio::task::spawn_blocking(move || -> Result<NbtCompound, PlayerDataError> {
```
kralverde, line 81: Why are we spawning a tokio task here? No async code is used

kralverde (changes requested):
Looks good really good! Other than a couple spots I mentioned, its easy to follow too.

Maybe also include a Java player nbt file in an assets folder and test if we can de serialize that?

Snowiiii:
Seems like the player position is wrong

PR #601, opened 2025-03-01
Chunk io benchmarking

kralverde:
## Description
Benchmarks chunk IO based on the current `./config` settings. 

Removed spawning an async task for saving chunks because otherwise the benchmark would not be accurate


Snowiiii:
Looks good to me :+1:

PR #602, opened 2025-03-02
feat(eating): initial version of food eating

HookWoods:
## Description
This PR is the initial version of food eating, the PR is still in draft cause some small bugs are present but this is working in a general way.

I implemented food and consumable item properties at the json loading so like that we have a generic way of handling food data

Some system will need to be handled later like effects ticking when eating some food but we currently send effect data to clients.

The system I implemented is very basic, I created a trait of PumpkinItem to handle all food function like eating duration, start eating, common on use, ...
Then there is a basic eating ticking that will send packet to client, check if eating is finished or if player has cancel the eating effect
## Testing


pumpkin/src/item/items/food.rs
```diff
@@ -0,0 +1,251 @@
...
+pub struct PufferfishItem;
+
+#[async_trait]
+impl PumpkinFood for PufferfishItem {}
+
+#[pumpkin_item("glow_berries")]
+pub struct GlowBerriesItem;
+
+#[async_trait]
+impl PumpkinFood for GlowBerriesItem {}
```
Snowiiii, line 251: Why implementing all foods yourself, We have a data driven design to not hard code such things

Snowiiii:
any progress on this?.

Snowiiii:
Feel free to reopen

PR #607, opened 2025-03-03
Abstract networking to streamline code and avoid creating a new buffer for (de)compression.

kralverde:
## Description
This PR removes our `ByteBuf` and `ByteBufMut` code and abstracts the network serializer and deserializer to streamline code and remove a write-to-new-buffer when data is compressed/decompressed.

## Testing


Snowiiii:
Looks nice, I was thinking of writing/deserialzing packets when created via `MyPacket::new(...)` instead of when doing `client.send_packet`. this would be useful when broadcasting packets to multiple clients since when then only need to write into one buffer

Snowiiii:
> Looks nice, I was thinking of writing/deserialzing packets when created via `MyPacket::new(...)` instead of when doing `client.send_packet`. this would be useful when broadcasting packets to multiple clients since we then only need to write into one buffer

Was thinking of how this can be done, I think the easiest way would be to to add something like `send_raw_packet` to the client which only takes in a packet id and a &buffer, and then writing the packet into an buffer one before the player loop in the world `broadcast_packet` function. But we would have to optimize then packet broadcasting in commands for example  then as well.

kralverde:
I was thinking of giving ownership of the stream to the client instead of doing this whole polling thing

Snowiiii:
Chunks seem to load really slow now

kralverde:
> Chunks seem to load really slow now

That should be fixed now with my latest commit

kralverde:
@Snowiiii thoughts?

Snowiiii:
Chunks load faster than before but it still feels slower than master tbh

Snowiiii:
Also when flying into void and waiting for chunks to load i got
```
[ERROR] Failed to load chunk at Vector2 { x: -70, z: -36 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -70,-36 but got it for -72,-36! (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -72, z: -35 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -72,-35 but got it for -74,-35! (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -73, z: -34 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -73,-34 but got it for -75,-34! (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -71, z: -36 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -71,-36 but got it for -73,-36! (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -74, z: -33 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -74,-33 but got it for -76,-33! (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -72, z: -36 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -72,-36 but got it for -74,-36! (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -73, z: -35 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -73,-35 but got it for -75,-35! (regenerating)
[ERROR] Failed to load chunk at Vector2 { x: -74, z: -34 }: Failed to parse chunk from bytes: Error deserializing chunk: Expected data for chunk -74,-34 but got it for -76,-34! (regenerating)
```

kralverde:
@Snowiiii could you re-test?

PR #609, opened 2025-03-05
Nix flake

Eveeifyeve:
## Description

* Followup from pr https://github.com/Pumpkin-MC/Pumpkin/pull/283 to take over from [@nicholaschiasson comment](https://github.com/Pumpkin-MC/Pumpkin/pull/283#issuecomment-2693886889) 
* Adds nix flake 
* Changed nix flake to use flake parts over utils and flake compact for old nix compatibility
* Adds back the cargo lock file which is needed for nix run package and recommended by rust team.

## Testing

N/A since there is no code change.
I simply ran nix run locally to test that it is working.


Eveeifyeve:
I have one more thing to do for this pr and that is adding an github action to update the flake.lock over time.

Eveeifyeve:
Finished. Note this will open a PR every time this updates.

kralverde:
I'm not a nix user but I'm seeing a lot of relative paths in the script, is there away to anchor them to the script's directory?

Also I'm not a big fan of the workflow (as a concept) why do you need periodic automatic updates? Can you just add a contrib/script to re-generate the workflow that's ran as needed?

Eveeifyeve:
> I'm not a nix user but I'm seeing a lot of relative paths in the script, is there away to anchor them to the script's directory?

No sadly 

> Also I'm not a big fan of the workflow (as a concept) why do you need periodic automatic updates? Can you just add a contrib/script to re-generate the workflow that's ran as needed?

Because of dependancy changes from nixpkgs and rust-overlay I wanted to do this so you would have developers worrying about updates to dependancy updates with flakes everything is locked to what is specified in flake.lock like cargo lock but it doesn’t update.

nicholaschiasson:
Isn't it a better idea to only ever intentionally update the flake lock file and manually check that into version control? I would think that having it automatically updated on a cron workflow could introduce breaking dependencies accidentally sometimes, and not to mention maintainers using the flake locally would be forced to move to a new derivation much more often, which is perhaps not a huge problem but a little inconvenient imo. Is this best practice for flakes right now?

Snowiiii:
> Isn't it a better idea to only ever intentionally update the flake lock file and manually check that into version control? I would think that having it automatically updated on a cron workflow could introduce breaking dependencies accidentally sometimes, and not to mention maintainers using the flake locally would be forced to move to a new derivation much more often, which is perhaps not a huge problem but a little inconvenient imo. Is this best practice for flakes right now?

@Eveeifyeve ?

Eveeifyeve:
> > Isn't it a better idea to only ever intentionally update the flake lock file and manually check that into version control? I would think that having it automatically updated on a cron workflow could introduce breaking dependencies accidentally sometimes, and not to mention maintainers using the flake locally would be forced to move to a new derivation much more often, which is perhaps not a huge problem but a little inconvenient imo. Is this best practice for flakes right now?
> 
> @Eveeifyeve ?

I am still not sure, I will try to ask someone in the nix matrix.

Snowiiii:
@Eveeifyeve any news ?

Eveeifyeve:
> @Eveeifyeve any news ?

I did ask, but no replies.

CC @getchoo I've tried all options you my only hope.

getchoo:
> Isn't it a better idea to only ever intentionally update the flake lock file and manually check that into version control?

I'd say it largely depends on the project

Most probably would benefit from frequent-ish bumps, as it would mean you're getting the latest security patches, features, and there would be less to download for people consuming it (assuming they are also up to date)

Other projects that have a *very* specific set of dependency requirements would probably struggle a fair bit more, especially without CI ensuring the updates actually work -- simply because frequent updates to Nixpkgs are likely to change those dependencies

.github/workflows/update-flake-lock.yml
```diff
@@ -0,0 +1,27 @@
+name: update-flake-lock
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * 0"
+
+jobs:
+  lockfile:
+    runs-on: ubuntu-22.04
```
Snowiiii, line 9: why not using ubuntu latest ?

Eveeifyeve: Well considering that, I will be removing this soon, since I think there is no dependancy update requirements, this review comment is no longer needed.

Snowiiii:
> Well considering that, I will be removing this soon, since I think there is no dependancy update requirements, this review comment is no longer needed.

Hey @Eveeifyeve, I really would like to have this merged or closed. considering your comment saying you want to remove the CI file and change some things, can you give an time estimate when you can finish this PR?

Snowiiii:
Feel free to reopen

PR #613, opened 2025-03-06
Generate Blocks and Properties Revamp

4lve:
## Description
This PR makes it easier to work with block properties.
image

Ram usage and file size are also reduced. In my quick testing on dev build mode I got a 70% decrease in ram usage
image
image
{90138B5E-911A-4A4C-A8F1-B01A5E539643}
{048CBE59-12B5-40AE-AE8C-17791C4A47E3}

## Testing


pumpkin-data/build/block.rs
```diff
@@ -0,0 +1,1439 @@
...
+
+include!("../src/tag.rs");
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct BlockProperty {
+    pub name: String,
+    pub values: Vec<String>,
+}
+
+fn get_enum_name(props: Vec<String>) -> String {
```
kralverde, line 14: This is kind of messy, can it be cleaned up?

kralverde (commented):
Looks good to me other than the one request

OfficialKris:
How does this increase binary size? Currently, pumpkin stores several megabytes of json data for blocks. Doesn't this optimize that?

pumpkin-data/build/block.rs
4lve: Ye, any suggestions on how you could clean it up?

4lve:
> How does this increase binary size? Currently, pumpkin stores several megabytes of json data for blocks. Doesn't this optimize that?

Yes, look at the images I provided, the smallest ones is this PR

pumpkin-data/build/block.rs
kralverde: Honestly, I have no idea what is going on here

kralverde:
Also: `async_trait` is only for when you need to do something like `Box<dyn AsyncTrait>` if you don't need to do that, you can have traits with async functions just fine

pumpkin-data/build/block.rs
4lve: Basically Minecraft's default property names suck cause one name can resolve to diffrent properties, that's why I'm mapping values to custom enum names instead

4lve:
> Also: `async_trait` is only for when you need to do something like `Box<dyn AsyncTrait>` if you don't need to do that, you can have traits with async functions just fine

Oh alr will clean that up then

pumpkin-data/build/block.rs
kralverde: Yeah I'm saying you can probably define these all as enums and then let serde sort through it when deserializing

4lve: Ye, not sure if that would make things better though, it would take up even more lines

4lve: Also there isn't really any point of defining the enums in the build code cause then each enum would need a ToTokens and would also need to be implemented in the generated code along with the EnumVariant trait

Snowiiii:
I really don't like the properties parsing here, It would make much more sense to use an enum, I'm not sure if we then even need the values since they are them predefined. 

I know it would may nice to have Block properties parsed from json files but i don't it makes sense here, First because there are not that many, and second because it would be much harder

4lve:
> I really don't like the properties parsing here, It would make much more sense to use an enum, I'm not sure if we then even need the values since they are them predefined. 
> 
> 
> 
> I know it would may nice to have Block properties parsed from json files but i don't it makes sense here, First because there are not that many, and second because it would be much harder

Not sure how it would be done with enums, could you give an example?

Snowiiii:
I think is already gave and example on discord. For example if we have an simple boolean property, we don't need the values and just can parse it from the name
```
"properties": [
        {
          "name": "snowy",
          "values": [
            "true",
            "false"
          ]
        }
      ],
```

```rust
pub enum BlockProperty {
     Snowy(bool),
     Powered(bool),
     ....
}
```
You can them implement a default for example and should be able to use #[serde(default]. I also recommed you this page for see what options there are when working with enums in serde https://serde.rs/enum-representations.html

4lve:
Hmm ok, how would you do for non boolean properties, reference the property enum?

Snowiiii:
> Hmm ok, how would you do for non boolean properties, reference the property enum?

```rust
pub enum BlockProperty {
    Facing(DirectionEnum),
    Layers(u8)
     ....
}
```
You mean this ?

4lve:
> > Hmm ok, how would you do for non boolean properties, reference the property enum?
> 
> 
> 
> ```rust
> 
> pub enum BlockProperty {
> 
>     Facing(DirectionEnum),
> 
>     Layers(u8)
> 
>      ....
> 
> }
> 
> ```
> 
> You mean this ?

Yes, and should we really use u8? It could cause invalid states which was the whole point to prevent in this pr

Snowiiii:
> > > Hmm ok, how would you do for non boolean properties, reference the property enum?
> > 
> > 
> > ```rust
> > pub enum BlockProperty {
> > 
> >     Facing(DirectionEnum),
> > 
> >     Layers(u8)
> > 
> >      ....
> > 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You mean this ?
> 
> Yes, and should we really use u8? It could cause invalid states which was the whole point to prevent in this pr

I'm not sure why it should cause any problems, We know that Layers is a number from 0-8, So an u8 would make sense ig

Snowiiii:
I also think that the `vaules` are not needed actually

4lve:
> > > > Hmm ok, how would you do for non boolean properties, reference the property enum?
> > > 
> > > 
> > > ```rust
> > > pub enum BlockProperty {
> > > 
> > >     Facing(DirectionEnum),
> > > 
> > >     Layers(u8)
> > > 
> > >      ....
> > > 
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You mean this ?
> > 
> > 
> > Yes, and should we really use u8? It could cause invalid states which was the whole point to prevent in this pr
> 
> I'm not sure why it should cause any problems, We know that Layers is a number from 0-8, So an u8 would make sense ig

Ye, but what for like redstone power?

Snowiiii:
> Ye, but what for like redstone power?
Afaik redstone power is 0-15, So why not use an u8 there as well ?

4lve:
> > Ye, but what for like redstone power?
> 
> Afaik redstone power is 0-15, So why not use an u8 there as well ?

Cause it can cause invalid states, what if someone sets the power to 100, the point of this system is to make it impossible to compile a program with invalid states. Also I think structs are a better fit cause it's easier to work with and it's equates to how it's implemented in minecraft better.

Also, you can work with Booleans as primative boolean by using the `::to_bool` and `::from_bool`

Snowiiii:
Setting the power to 100 would be pointless since the client would than just ignore it or may crash/do unexpected behavior. Minecraft often uses larger primitives then they need to and there is not unsigned in Java as well.

4lve:
> Setting the power to 100 would be pointless since the client would than just ignore it or may crash/do unexpected behavior. Minecraft often uses larger primitives then they need to and there is not unsigned in Java as well.

That's not how properties work, all properties are converted to a state id before being sent to the client

4lve:
But my point was was don't get any benefit from making them into enums, I would even say it would be harder to work with

4lve:
I updated it so you can use it generically now, check the fences for an example

kralverde (approved):

kralverde:
@Snowiiii I think this is good to go now, could you take a look

Snowiiii:
Looks good now. Thank you, Good job @4lve :+1:

PR #616, opened 2025-03-09
Show player respawn screen if the chat or menu is open

### LINKED ISSUE #541: The respawn screen does not appear if you have the chat open

### Operating System

Win 11

Pumpkin 0.1.0-dev (8755b6c) for Minecraft 1.21.4 (Protocol 769)

### What happened?

I died with the chat open, and the respawn screen only appear when i close the chat

### To Reproduce

Kill the player with the chat open

### Expected behavior

Close automatically and display the respawn screen

FlorianLang06:
## Description

This PR fixes #541. After a player was killed the PLAY_PLAYER_COMBAT_KILL packet will be send to the player. This means the player sees the respawn screen even if he has the chat open or is in the menu. The sending of this packet was already implemented in the kill function of the player but this function was never used.

To achieve this I changed that in the kill command use the kill function of the player and after a damage of a player it will be check if the player is death and the packet will also be send.

The only way a player can still be killed without this packet is if the kill method of the living entity is directly called.

### DIFF
--- a/pumpkin/src/command/commands/kill.rs
+++ b/pumpkin/src/command/commands/kill.rs
@@ -33,7 +33,7 @@ impl CommandExecutor for Executor {
         let target_count = targets.len();
         let mut name = String::new();
         for target in targets {
-            target.living_entity.kill().await;
+            target.kill().await;
             name.clone_from(&target.gameprofile.name);
         }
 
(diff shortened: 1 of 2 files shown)


Snowiiii:
Looks good. Thank you @FlorianLang06

Issue #574, opened 2025-02-21
All torches place in a wall state, example: redstone_wall_torch


Z-fentom:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

Pumpkin 0.1.0-dev (2af2fbb)

### What happened?

While i placed the torches they kept placing a wall state 

### To Reproduce

1.Take any type of a torch
2.Place it

### Expected behavior

Its expected for torches to only place in a walled state when player is looking at a wall and placing it

Snowiiii:
This was fixed in https://github.com/Pumpkin-MC/Pumpkin/pull/621

PR #618, opened 2025-03-09
Fix issue #574 "All torches place in a wall state"

lwalton101:
I have created a trait for blocks which either attach vertically to a wall or the ground. For example, a torch block can either be wall_torch or just torch.

The trait will place the standing torch if it is placed on the right face and if it places a wall attached torch it inverts the face so that the torch faces the correct way.

This issue was reported in Issue #574 

## Testing

All 3 torch types are working


lwalton101:
marked as draft as the new block property stuff has partially fixed this

OfficialKris:
> marked as draft as the new block property stuff has partially fixed this

Maybe it can be reworked using the new block property code?

Snowiiii:
@lwalton101 The properties recently got reworked by @4lve. Can you please use the new system

pumpkin/src/block/blocks/torch.rs
```diff
@@ -0,0 +1,168 @@
...
+                for bd in BlockDirection::horizontal() {
+                    let block_offset = bd.to_offset();
+                    let base_block_pos = _block_pos.offset(block_offset);
+                    let base_block = _world.get_block(&base_block_pos).await.unwrap();
+                    if base_block.id != 0 {
+                        possible_directions.push(bd.to_cardinal_direction());
+                    }
+                }
+
+                return if possible_directions.contains(_player_direction) {
```
Snowiiii, line 153: _player_direction should be player_direction

pumpkin/src/block/blocks/torch.rs
```diff
@@ -0,0 +1,168 @@
...
```
Snowiiii, line 102: _player_direction should be player_direction

pumpkin/src/block/blocks/torch.rs
```diff
@@ -0,0 +1,168 @@
...
```
Snowiiii, line 52: _player_direction should be player_direction

Snowiiii:
Also why you copied the same thing 3 times ?. You could just call one function

pumpkin/src/block/blocks/torch.rs
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+}
+async fn get_wall_block_props(
+    world: &World,
+    block: &Block,
+    block_pos: &BlockPos,
+    face: &BlockDirection,
+    player_direction: &HorizontalFacing,
+    mut block_properties: Vec<(String, String)>,
```
kralverde, line 118: Instead of serializing the block properties, you should mutate the `LikeProperties` struct itself, which contains enums of valid states.

kralverde (changes requested):
all the `...LikeProperties` should be type aliased into a definitive type, because although the changes are relatively low, a minecraft update can change those names. Type aliasing will make this easier to fix when breaking updates happen.

Snowiiii:
This was fixed in https://github.com/Pumpkin-MC/Pumpkin/pull/621

PR #620, opened 2025-03-10
Fix order of explosion block handling that droping and explotion of blocks work

### LINKED ISSUE #595: TNT Currently does not interact with other TNT blocks nor does it drop destroyed blocks

Currently, TNT does not drop any block when it destroys them as in the base game. It also currently does not ignite other TNT when it blows them up also as in the base game.

TNT to interact with blocks as it does in the vanilla game

FlorianLang06:
## Description

In the explode function of a explosion the dropping of the loot and explode function of the block were executed after the block state was set to 0. This was the reason both methods had no impact. I changed this that the setting of the block_state will be done after. Now both the dropping of blocks and explosion of other TNT worked fine.

This PR fixes #595.

## Testing

### DIFF
--- a/pumpkin/src/world/explosion.rs
+++ b/pumpkin/src/world/explosion.rs
@@ -80,14 +80,15 @@ impl Explosion {
 
             let block = world.get_block(&pos).await.unwrap();
             let pumpkin_block = server.block_registry.get_pumpkin_block(&block);
+
+            world.set_block_state(&pos, 0).await;
+
             if pumpkin_block.is_none_or(|s| s.should_drop_items_on_explosion()) {
                 drop_loot(world, &block, &pos, false, block_state.id).await;
             }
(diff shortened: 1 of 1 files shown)


Snowiiii:
TNT is still buggy but nice now item drops do work again. But it would be nicer to match vanilla and first set the blockstate and then call other blocks on explotion
```java
world.setBlockState(pos, Blocks.AIR.getDefaultState(), Block.NOTIFY_ALL);
block.onDestroyedByExplosion(world, pos, explosion);
```

FlorianLang06:
@Snowiiii OK you a right I just need to get the Block before than set the state and after that explode the other Blocks. This was something I didn't thought about.

FlorianLang06:
Hope this variant is now better

Snowiiii:
Looks good now. Thank you @FlorianLang06

PR #624, opened 2025-03-12
ChunkIO api and world saving optimizations

kralverde:
## Description
This PR does several things:
* Adds a `dirty` flag to chunk data such that ChunkIO can see if a chunk has been modified when determining whether or not to write it
* Adds a config option of `write_in_place` for chunks. If this is true (it defaults false), the Anvil chunk serializer will attempt to write chunks within the file instead of writing an entirely new file.
    * Pros: a lot less file I/O and at lot faster
    * Cons: **IF** the server shuts down uncleanly *while* a write is being done, the chunks being written *may* be corrupted rather than rolled back to the original save.
* Adds a method to ChunkSerializers to determine whether or not they should write to file when chunks are sent to them.
* Separates spawn chunks from watched chunks
* Add a type to the ChunkSerializer trait to denote what back-end it uses

New:
* Adds a feature to pumpkin-config to compile helper functions to mutate the config at runtime for tests

## Testing
* Tests have been added for the in-place anvil file IO

### DIFF
--- a/pumpkin-config/Cargo.toml
+++ b/pumpkin-config/Cargo.toml
@@ -9,4 +9,7 @@ serde.workspace = true
 log.workspace = true
 uuid.workspace = true
 
+# So we don't read/write to disk when testing
+tempdir = "0.3.7"
+
 toml = "0.8"
--- a/pumpkin-config/Cargo.toml
+++ b/pumpkin-config/Cargo.toml
@@ -9,7 +9,8 @@ serde.workspace = true
 log.workspace = true
 uuid.workspace = true
 
-# So we don't read/write to disk when testing
-tempdir = "0.3.7"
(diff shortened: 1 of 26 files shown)


Snowiiii:
Not writing to chunks when not necessary is great, Not sure but i think Vanilla does this as well. Just keep in mind that you have to save initial chunks

kralverde:
> Not writing to chunks when not necessary is great, Not sure but i think Vanilla does this as well. Just keep in mind that you have to save initial chunks

They are; newly generated chunks are marked dirty

kralverde:
@Snowiiii do you have any issues with me merging this?

Snowiiii:
Looks good. Thank you

PR #627, opened 2025-03-12
fixed wrongly used coordinates in explosion calculation

### LINKED ISSUE #617: Explosion not in every direction

### Operating System

Fedora 41

0.1.0-dev/26f822ba

### What happened?

If a TNT explodes the explosion is not in every direction. Its not explodes the block above and south of it. 

### To Reproduce

Place a TNT surrounded by blocks and light it up. 

### Expected behavior

It should explode into every direction.

FlorianLang06:
## Description

The calculation in the explosion get_blocks_to_destroy method was wrong. In the first calculation of each axis there was y and z wrongly calculated. For y was z used in the calculation and for z the already calculated y value. I check this with the vanilla code an know its correct. You can view a recording of the fix in the issue.

This fixes #617.

## Testing

### DIFF
--- a/pumpkin/src/world/explosion.rs
+++ b/pumpkin/src/world/explosion.rs
@@ -18,8 +18,8 @@ impl Explosion {
     async fn get_blocks_to_destroy(&self, world: &World) -> Vec<BlockPos> {
         let mut set = Vec::new();
         for x in 0..16 {
-            for z in 0..16 {
-                'block2: for y in 0..16 {
+            for y in 0..16 {
+                'block2: for z in 0..16 {
                     if x != 0 && x != 15 && z != 0 && z != 15 && y != 0 && y != 15 {
                         continue;
                     }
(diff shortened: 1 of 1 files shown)


Snowiiii:
i think the loop order is wrong

FlorianLang06:
I agree with you its still not like in vanilla but its better than before. I will look into this and try to replicate the explosion logic exactly.

FlorianLang06:
Explosion in dirt looks already like in vanilla, the explosion of stone only if I change the power to 5. So I think there is maybe a Problem with the blast resistance or something like that. I look into this tomorrow.

pumpkin/src/world/explosion.rs
```diff
@@ -18,15 +18,15 @@ impl Explosion {
     async fn get_blocks_to_destroy(&self, world: &World) -> Vec<BlockPos> {
         let mut set = Vec::new();
```
teknostom, line 19: This should be a Hashset right? Otherwise, the block removal code will run a lot more.

Only other change needed for that is to add "std::hash::Hash" to BlockPos's "derive's"

teknostom (commented):
Seems like this code works well, except for downwards.

In the java code, this is the startingpoint:
double d5 = this.center.x;
double d6 = this.center.y;
double d7 = this.center.z;

But if you print the pos in this function, the y is not in the center.

Snowiiii:
Should be fixed in https://github.com/Pumpkin-MC/Pumpkin/pull/632 now. Thank you @FlorianLang06

PR #629, opened 2025-03-13
Add Block Properties for Jukeboxes

tjf801:
## Description
This (small) PR just adds the logic for jukebox block properties (which are directly observable in some resource packs).

## Limitations
The jukebox still doesnt actually store the disc in NBT data (or pop out the disc once its removed), since block NBT isn't implemented yet

### DIFF
--- a/pumpkin/src/block/blocks/jukebox.rs
+++ b/pumpkin/src/block/blocks/jukebox.rs
@@ -6,7 +6,7 @@ use crate::entity::player::Player;
 use crate::server::Server;
 use crate::world::World;
 use async_trait::async_trait;
-use pumpkin_data::block::{Block, BlockState};
+use pumpkin_data::block::{Block, BlockProperties, BlockState, Boolean, JukeboxLikeProperties};
 use pumpkin_data::item::Item;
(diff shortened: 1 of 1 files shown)


pumpkin/src/block/blocks/jukebox.rs
```diff
@@ -15,25 +15,53 @@ use pumpkin_util::math::position::BlockPos;
...
 pub struct JukeboxBlock;
 
+impl JukeboxBlock {
+    async fn has_record(&self, location: BlockPos, world: &World) -> bool {
+        let state_idx = world
+            .get_block_state(&location)
+            .await
+            .expect("`location` should be a jukebox")
+            .id;
+        JukeboxLikeProperties::from_index(state_idx).has_record == Boolean::True
```
4lve, line 25: Use from_state_id instead of from_index

kralverde:
Could you make a type alias for the `JukeboxLikeProperties`

Snowiiii:
Everything works fine (Expect the disk dropping out when clicking but that's expected). Thank you @tjf801

PR #632, opened 2025-03-15
Rewrote TNT logic

teknostom:
A lot of small mistakes were in the previous impl, fixed for performance and validity

## Description

- Used HashSet instead of Vector to reduce calculations when going through all blocks to destroy them

- Added @FlorianLang06 's yz-fix Where y and z were used wrong.

- Aligned code more to how Java does it

- Added 0.0625 to Y as java does it. (0.0625 * entity height)

- Added velocity, but this won't look right until velocity is implemented.

## Testing

Only in-game testing

### DIFF
--- a/pumpkin-util/src/math/position.rs
+++ b/pumpkin-util/src/math/position.rs
@@ -1,11 +1,12 @@
 use super::vector3::Vector3;
 use std::fmt;
+use std::hash::Hash;
 
 use crate::math::vector2::Vector2;
 use num_traits::Euclid;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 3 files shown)


Snowiiii:
So does this make TNT behave like in Vanilla and have the same explosion?

teknostom:
Yes, from what i have tested, it seems to work just like vanilla with the exception of the velocity making some explosions different. But that is added here too, so when velocity is implemented, it should work just like vanilla.

Snowiiii:
Looks good now. Thank you @teknostom and @FlorianLang06

PR #638, opened 2025-03-17
add Entity Saving/Reading

Snowiiii:
## Description
Adds Entity Saving/Reading

## Testing


pumpkin-world/src/storage/format/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+    index as usize
+}
+
+pub const fn get_region_coords(at: &Vector2<i32>) -> (i32, i32) {
+    // Divide by 32 for the region coordinates
+    (at.x >> SUBREGION_BITS, at.z >> SUBREGION_BITS)
+}
+
+/// Used for Saving
+pub trait DataToBytes
```
kralverde, line 48: I think this should be named something more specific

pumpkin-world/src/storage/format/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+where
+    Self: Send + Sync,
+{
+    type Data: Send + Sync + Sized;
+
+    fn data_to_bytes(chunk_data: &Self::Data) -> Result<Vec<u8>, ChunkSerializingError>;
+}
+
+/// Used for Reading
+pub trait BytesToData
```
kralverde, line 58: I think this should be named something more specific

pumpkin-world/src/storage/format/anvil/entity.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use super::{AnvilFile, chunk::AnvilChunkFormat};
+
+#[derive(Default)]
+pub struct AnvilEntityFormat {
+    anvil: AnvilFile,
+}
+
+#[async_trait]
+impl ChunkSerializer for AnvilEntityFormat {
```
kralverde, line 21: We probably want to create a brand new serializer instead of using the chunk serializer. `ChunkFileManager` can be probably renamed to `FileManager`. Then we can have `FileManager<EntitySerializer>` instead of `ChunkSerializer`

Snowiiii: Now sure what you mean, Do we want to fully replace `ChunkSerializer` ?. I mean its pretty different then `FileManager` and is a trait

Snowiiii:
I will make a new PR, too many conflicts

PR #643, opened 2025-03-19
Implement Flowing Fluids

teknostom:
## Description
This implementation follows the Minecraft vanilla implementation as closely as possible.

Features
[x] Flow from a source.
[x] Dissapate when source is removed.
[x] form new sources when applicable.

## Testing


Q2297045667:
How does the performance of this compare to vanilla?

Snowiiii:
@teknostom any progress on that ?. Would be super nice if this gets done because fluidstates are used in many places

teknostom:
Yeah, i have had some trouble with understanding the java code, I have gotten most things to work, but the "flow into holes" mechanic is definitely the hardest. Will try to finish this up this weekend though.

Snowiiii:
please move the fluid folder to blocks

Snowiiii:
Also why you made a new trait and "register system" for fluids, i think they are the same as blocks

teknostom:
@Snowiiii This PR should be done now. It includes both lava and water. Currently, it only works with setblock, but I could maybe work on buckets as a next PR.

It also works with neighbor updates now, so behaviour should be 100% match to vanilla.

Snowiiii:
Hey @teknostom, can you may make the width match Vanilla?
image
image

teknostom:
Oh, weird. Will look into that.

Snowiiii:
Also could you please add support for water and lava which is located in the World, would be nice to have that in this PR as well
image

teknostom:
The CI is failing due to the new rust release, but the two bugs are fixed. Water and lava should now flow from generated blocks.

One problem I have noticed is with the set_block_state. It does not always update the blocks for the client. Might be some async stuff there causing updates to be lost. But if you hold right click while changing blocks around water, it should work like vanilla.

Snowiiii:
Hey, Can you please merge master again

pumpkin/src/block/blocks/fluids/flowing_fluid.rs
```diff
@@ -0,0 +1,466 @@
...
+    block::Block,
+    fluid::{Falling, Fluid, FluidProperties, Level},
+};
+use pumpkin_util::math::position::BlockPos;
+use pumpkin_world::block::BlockDirection;
+
+use crate::world::{BlockFlags, World};
+type FlowingFluidProperties = pumpkin_data::fluid::FlowingWaterLikeFluidProperties;
+
+fn level_to_int(level: Level) -> i32 {
```
4lve, line 15: You should be able to use Level.from_index and to_index

pumpkin/src/block/blocks/fluids/flowing_fluid.rs
```diff
@@ -0,0 +1,466 @@
...
+                    .set_block_state(
+                        block_pos,
+                        new_fluid_state.to_state_id(fluid),
+                        BlockFlags::NOTIFY_ALL,
+                    )
+                    .await;
+            }
+            self.spread(world, fluid, block_pos, &new_fluid_state).await;
+        } else {
+            world
```
4lve, line 132: maybe world.break_block instead so it drops items?

pumpkin/src/block/blocks/fluids/lava.rs
```diff
@@ -0,0 +1,64 @@
...
+impl FlowingFluid for FlowingLava {
+    async fn get_drop_off(&self) -> i32 {
+        2
+    }
+
+    async fn get_slope_find_distance(&self) -> i32 {
+        2
+    }
+
+    async fn can_convert_to_source(&self, _world: &Arc<World>) -> bool {
```
4lve, line 60: Lava can't convert to source right?

4lve (commented):
Looks good, just have some questions here

pumpkin/src/block/blocks/fluids/lava.rs
teknostom: Yep, true

pumpkin/src/block/blocks/fluids/flowing_fluid.rs
teknostom: Done

Snowiiii:
Awesome work. Thank you @teknostom! :100:

PR #648, opened 2025-03-21
Biome & surface generation fixes

kralverde:
## Testing


Snowiiii:
You should use the Vanilla biome codec so we have more than weather, iirc  this also gives a list of all features we can use for chunk generation. 

Biomes are also in the sync_registry.json

Snowiiii:
Looks good. Thank you @kralverde

Issue #651, opened 2025-03-22
When falling out of the world (I dug strait down) crash "BlockOutOfWorldBounds"


jkoop:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Linux Mint 22 Wilma

### Server Software Version/Commit

0.1.0-dev (eb587c97)

### What happened?

Server crashes when falling out of the world

```plain
thread 'tokio-runtime-worker' panicked at /home/joek/git/Pumpkin-MC/Pumpkin/pumpkin/src/world/mod.rs:1459:84:
called `Result::unwrap()` on an `Err` value: BlockOutOfWorldBounds
stack backtrace:
0:     0x649b46b9dcd4 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::he089f96442833f67
1:     0x649b465907c3 - core::fmt::write::h2f210ed4c94745cb
2:     0x649b46b9d022 - std::io::Write::write_fmt::h7de08171ab770fb2
3:     0x649b46b9db23 - std::sys::backtrace::BacktraceLock::print::h810fbd31421329e6
4:     0x649b46b9d5bf - std::panicking::default_hook::h24e207139139d40a
5:     0x649b465773fe - pumpkin::main::{{closure}}::{{closure}}::h4ed4c628957b9f77
6:     0x649b46b9cef2 - std::panicking::rust_panic_with_hook::ha9131beeb2ddc506
7:     0x649b46bd9fd8 - std::panicking::begin_panic_handler::{{closure}}::h1bba0eaeb6da506f
8:     0x649b46bd9f39 - std::sys::backtrace::__rust_end_short_backtrace::h1d1ca3eade483f4c
9:     0x649b46bda5fe - rust_begin_unwind
10:    0x649b4641772f - core::panicking::panic_fmt::h896a0727a1a943f9
11:    0x649b46417a45 - core::result::unwrap_failed::h1b5ed8541c7bebd6
12:    0x649b4651ecf3 - pumpkin::world::World::set_block_state::{{closure}}::h58984febf4be8b3a
13:    0x649b465139f8 - pumpkin::world::World::break_block::{{closure}}::h674147d7ff5d77ca
14:    0x649b46540952 - pumpkin::entity::player::Player::handle_play_packet::{{closure}}::h6f35d1bd6adb86a0
[...]

PR #653, opened 2025-03-22
Fixes BlockOutOfWorldBounds server crash when bedrock gets updated

aaneeley:
## Description
Server previously would crash if any player was below y -64 (bedrock).

Fixes https://github.com/Pumpkin-MC/Pumpkin/issues/651

## Changed
ChunkBlocks.get_block now returns the ID for AIR under y level -64

**Note:** This does not implement player damage under y -128. It simply fixes the BlockOutOfWorldBounds error. Player will fall forever in current state.

**EDIT:** The actual issue is the lowest level of bedrock (y=-64) tries to update neighbors on y=-65 which doesn't exist.

### DIFF
--- a/pumpkin-world/src/chunk/mod.rs
+++ b/pumpkin-world/src/chunk/mod.rs
@@ -217,7 +217,6 @@ impl SubchunkBlocks {
 impl ChunkBlocks {
     /// Gets the given block in the chunk
     pub fn get_block(&self, position: ChunkRelativeBlockCoordinates) -> Option<u16> {
-
         match &self {
             Self::Homogeneous(block) => Some(*block),
             Self::Subchunks(subchunks) => subchunks
--- a/pumpkin-world/src/chunk/mod.rs
+++ b/pumpkin-world/src/chunk/mod.rs
@@ -219,7 +219,7 @@ impl ChunkBlocks {
     /// Gets the given block in the chunk
     pub fn get_block(&self, position: ChunkRelativeBlockCoordinates) -> Option<u16> {
         if position.y.0 < -64 {
-            return Some(Block::AIR.id)
+            return Some(Block::AIR.id);
         }
 
         match &self {
--- a/pumpkin-world/src/chunk/mod.rs
+++ b/pumpkin-world/src/chunk/mod.rs
@@ -1,4 +1,3 @@
-use pumpkin_data::block::Block;
 use pumpkin_nbt::nbt_long_array;
 use pumpkin_util::math::{position::BlockPos, vector2::Vector2};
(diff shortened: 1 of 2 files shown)


kralverde:
Thank you for making a PR!

 -64 should not be a fixed value since it can be changed based on chunk parameters. You should change the logic inside of `subchunk.get_block` to return None if it is out of bounds instead.

aaneeley:
Looking further into it, it does appear I misunderstood the issue at hand. subchunk.get_block isn't broken.

The issue is that world.get_block_state_id is called at the player's position multiple times, and it's default behavior is to return BlockOutOfWorldBounds if there is no subchunk at the player's current location.

```rust
// If the player isn't in a subchunk this crashes the server
let Some(id) = chunk.blocks.get_block(relative) else {
    return Err(GetBlockError::BlockOutOfWorldBounds);
};
```

Can someone look into the utility of BlockOutOfWorldBounds? Is it necessary? Why can't it default to AIR?

aaneeley:
Confirmed that https://github.com/Pumpkin-MC/Pumpkin/pull/653/commits/da0703b35fead255a438a0034343e64fd45aabd0 works. Just want to make sure I didn't mess up any edge cases by removing the error.

kralverde:
I think the error should be kept; it's valid error that we aren't querying a block within the world bounds. 

Instead, the error should be properly handled where the crash happens instead of (what I assume to be) unwrapping the error.

aaneeley:
However below bedrock is currently (in pumpkin) considered "out of bounds" even though falling out of the world is a valid vanilla mechanic.

I don't know if get_block returning None always means its out of bounds.

Not sure what the best solution is here.

kralverde:
> However below bedrock is currently (in pumpkin) considered "out of bounds" even though falling out of the world is a valid vanilla mechanic.

Correct. You should find what is trying to get the block that is below the bedrock and gracefully handle the error rather than unwrapping it.

aaneeley:
It appears I misunderstood the issue once again.

The crash is caused by world.replace_with_state_for_neighbor_update it attempts to update all neighbors (including neighbors under the world) whenever a block is updated on y=-64. 

So any update to the lowest bedrock level crashes the server.

I can go ahead an implement a check here to make sure neighbors below the world are ignored.

However is updating all neighbors for every block place/break intended functionality or is this another bug? I thought neighbor updates were exclusive to redstone/stairs/etc.

aaneeley:
@4lve Confirmed that the neighbor update functionality is correct so I handled the error so it doesn't try to update a non-existent neighbor

kralverde:
LGTM, thanks!

PR #654, opened 2025-03-22
fix: view distance goes not above 16

unschlagbar:
The is_within_distance function squared the view distance while it was still in u8 format which prevented it from exceed 16 chunks view distance.

I also added a simple capacity estimation in the all_chunks_within function to prevent reallocation.


kralverde:
Do you think you could use https://youtu.be/hpiILbMkF9w?si=ovX7idmgfe5JDFk5 to avoid floating point operations? Maybe even remove needing to call `is_within_distance` entirely?

Also we have a test case that all chunks returned by `all_chunks_within` are `is_within_distance`. Could you add the inverse; test that all chunks that are not returned by `all_chunks_within` are not `is_within_distance`.

unschlagbar:
> Do you think you could use https://youtu.be/hpiILbMkF9w?si=ovX7idmgfe5JDFk5 to avoid floating point operations? Maybe even remove needing to call `is_within_distance` entirely?
> 
> Also we have a test case that all chunks returned by `all_chunks_within` are `is_within_distance`. Could you add the inverse; test that all chunks that are not returned by `all_chunks_within` are not `is_within_distance`.

I don't think it would make sense to replace 'is_within_distance' with 'all_chunks_within', since it's already very fast.
In my opinion, the inverse test is not necessary, because as a client you can clearly see when a chunk is missing.

Let me know if you want me to change something else

unschlagbar:
I've changed the capacity estimate to not use a floating-point number, and it even looks cleaner now

kralverde:
The integer divisions are truncating; it doesn't make sense to use them in this context.

As for the tests, the goal is to find problems quickly before getting to the client :p. I think we should also iterate over different view distances to make sure it holds.

unschlagbar:
image
image
image

The solution gets better and better, the worst case within 1..32 view distance is 24 more capacity than needed.
I actually dont know where the test you mentioned is, and how could i prove that all chunks that are not returned by `all_chunks_within` are not `is_within_distance` without using the same `is_within_distance` that the fn already uses.

kralverde:
We're basically testing that the values we give to be checked by `is_within` is greater than or equal to the set within `is_within`. 

You should start with a square larger than the radius, remove the values given by all within, and then verify that none of the remaining are within.

The test module should be within the same file. You will need to modify the tests.

kralverde:
What is your logic behind 3450 and 10?

unschlagbar:
its floating point multiplication without floating point number, its more like fixed point number

unschlagbar:
if you want to multiplicate some number with a 5.5 for example you can do that this way: x * 55 / 10

unschlagbar:
And you drew my attention to the fact that you can replace divisions with bitshifting

Snowiiii:
I want to mention that valance uses pre calculated chunk values, maybe that's something we can do aswell?
https://github.com/valence-rs/valence/blob/main/crates/valence_generated/build/chunk_view.rs

unschlagbar:
We could do it but this code is already really optimized. i dont think any circle point algorithm can make this faster. I expandet the test to test with any view distance from 1 to 32.

Snowiiii:
Looks good. Thank you @unschlagbar :+1:

PR #655, opened 2025-03-22
Don't place block if button will be clicked with item in hand

FlorianLang06:
## Description

Implemented the `use_with_item` function for buttons and return `BlockActionResult::Consume` in it to prevent placing of blocks if a player clicks on button with item in hand. This is the same logic as for a switch. 

## Testing


kralverde:
Looks good to me. Thanks!

Issue #649, opened 2025-03-21
Support secure Chat


Snowiiii:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

Currently we don't support Mojangs Cryptographic chat. We should change that

### Describe the solution you'd like

Support Cryptographic chat

Q2297045667:
I think there should be a setting added to allow players to disable this feature.

i-am-called-glitchy:
OK, but OFF by default.

Snowiiii:
resolved by https://github.com/Pumpkin-MC/Pumpkin/commit/a63b49a095e0478fb1a29183ca5569706103fad8

PR #656, opened 2025-03-23
1.21.5 Chat Overhaul + Secure Chat + Chat Formatting

aaneeley:
## Description
WIP to implement Mojang secure chat and formatting

EDIT: Mojang made a change that requires chat messages to be handled completely differently server side so now i'm doing that too

Related to https://github.com/Pumpkin-MC/Pumpkin/issues/649

## Progress
- [x] Add configuration option for enforce secure chat
- [x] Implement player session packet and deserializer
- [x] Store session data for player
- [x] Handle previous messages and message index correctly
- [x] Add configuration option to disable chat reporting
- [x] Add chat formatting
- [x] Add 1.21.5 chat index to chat packets
- [x] Fix weird multiplayer session issues 
- [x] Properly handle acknowledge bitset

## Testing


kralverde:
I think the last maybe should be a definitely

aaneeley:
> I think the last maybe should be a definitely

I agree it should be an option at the very least. Another option is to just disable chat reporting by default in the source code.

I don't know if Pumpkin wants to take a stance on Microsoft's "interesting" business decisions regarding server chats.

Looking for input from other people on this.

suprohub:
> I think the last maybe should be a definitely

I think last should be by default enabled

aaneeley:
I agree that some functionality similar to [FreedomChat](https://github.com/ocelotpotpie/FreedomChat) should be implemented.

This would require a decent amount of modification to the chat system overall because anti-reporting plugins send all player messages as decorated system messages.

Should I make this the default behavior? It would also make configurable chat formatting much easier.

Snowiiii:
Why you first do the configuration first ?. You should focus on getting the feature itself done and then thinking about how to add a way of disabling and and using an other chat system.

aaneeley:
Everything mostly working so far. Having some weird issues when multiple players connect. Some players randomly get chat validation errors depending on who connects to the server first.

aaneeley:
> also 1.21.5 has changed the chat packets, so you may want to take a look if this still works in 1.21.5

Yep looks like they completely changed how chat verification works. Nice.

Snowiiii:
> > also 1.21.5 has changed the chat packets, so you may want to take a look if this still works in 1.21.5
> 
> Yep looks like they completely changed how chat verification works. Nice.

Not sure about the verification, I only remember reading that the new packets better validate the message index so the messages don't get out of order

Snowiiii:
@aaneeley #650 is merged

pumpkin-protocol/src/client/play/player_chat_message.rs
```diff
@@ -9,8 +9,11 @@ use crate::{VarInt, codec::bit_set::BitSet};
 #[derive(Serialize)]
 #[packet(PLAY_PLAYER_CHAT)]
 pub struct CPlayerChatMessage {
+    /// An index that increases for every message sent TO the client
+    pub global_index: VarInt,
```
Snowiiii, line 13: why is this public ?

aaneeley: My current approach is to modify global_index on the packet before sending to each player to avoid cloning CPlayerChatMessage for each player. Is cloning the better option here? Admittedly my memory optimization skills are lacking.

aaneeley:
@Snowiiii looking at your current implementation on master of global_index I think you might be mistaken about the chat update. I believe global_index is supposed to track the number of messages sent TO each player (meaning the packet must be unique to each player). 

In your current configuration it tracks how many messages were sent FROM the player, which happens to work when only one player is on the server, but it will not work with multiple players.

Snowiiii:
> @Snowiiii looking at your current implementation on master of global_index I think you might be mistaken about the chat update. I believe global_index is supposed to track the number of messages sent TO each player (meaning the packet must be unique to each player).
> 
> In your current configuration it tracks how many messages were sent FROM the player, which happens to work when only one player is on the server, but it will not work with multiple players.

yep your right i made an upsi :sweat_smile:. Would be awesome if you could may fix this and we could merge this as soon as possible

aaneeley:
> yep your right i made an upsi 😅. Would be awesome if you could may fix this and we could merge this as soon as possible

Debugging now. Hoping to finish this up within the next couple hours.

Snowiiii:
I get this when trying to join with an new player
image

aaneeley:
Oops. Looks like I did so much testing with secure chat on I forgot to test with it off. Should be a simple fix. Its trying to send a player session that doesn't exist because secure chat is disabled.

aaneeley:
@Snowiiii It should be fixed now

Snowiiii:
Looks good now. Thank you @aaneeley :rocket:

aaneeley:
I plan on fully implementing all the vanilla server-side signature checks over the next few days. Just wanted to get this functional ASAP.

PR #657, opened 2025-03-23
Consider using Cow instead of String or &str

suprohub:
## Description
This pr tries make make our crates use Cow more, and improve functionality & optimization of protocol crate.

Some changes:
- Added serialize and deserialize to all packets. This can help creating client too, not only server (for example botmark).
- Identifier now <'a> and inner fields use Cow.
- Strings and strs now replaced by Cow in some places.
- VarInt now contains simd like optimizations for decoding.
- Bitset now contains more completefull (idk how say) implementation.
- Added tests for varint, bitset and slot.
- Synced registries now have public api for plugins and also use cow.

Why use Cow ?

1 Borrow checker simplicity. When not using Cow borrow checker give big amouth of errors, what deserialize not implemented for &T. Instead we can use cow, and it passes this.

2 Convenience. Plugins, our inner code, clients and etc can change ingoing and outgoing packets. If we use simple borrowing we cant change packets.

3 Optimizations. With cow we can use less memory and speedup more things, specially for plugin `conveyors`. Its combines features of borrowed and owned values.

What i mean by plugin conveyors? Imagine what big server have big amouth of plugins, and more plugins tries change packets, because events not what much or plugin want own control. If we set only owned value some plugin may just not use it, and we clone for nothing -_-. Instead, we can use cow and when if plugin needed full value he can clone it. Also we can use borrowed or owned values where it needed, vecayse some functions like deserialization can give us owned string. It works bidirectonal, so need change all.

## Testing

`cargo run`, see what world is loading an say: fuf, finally


suprohub:
I idk what wrong with actions, because on my machine it runs successfully.

Snowiiii:
> I idk what wrong with actions, because on my machine it runs successfully.

You need to all --all-targets to clippy

suprohub:
> > I idk what wrong with actions, because on my machine it runs successfully.
> 
> You need to all --all-targets to clippy

Oh i forgot, thanks

suprohub:
> > I idk what wrong with actions, because on my machine it runs successfully.
> 
> You need to all --all-targets to clippy

What do you think about pr? Cool? Not?

suprohub:
> > I idk what wrong with actions, because on my machine it runs successfully.
> 
> You need to all --all-targets to clippy

not helped bruh, waiting pr approval from another contributor to fix this

suprohub:
CI fixed

suprohub:
When merge?

suprohub:
@kralverde

pumpkin-protocol/src/client/config/cookie_request.rs
```diff
@@ -1,17 +1,20 @@
...
 
-#[derive(serde::Serialize)]
+#[derive(Serialize, Deserialize)]
+#[serde(bound(deserialize = "'a: 'de"))]
 #[packet(CONFIG_COOKIE_REQUEST)]
 /// Requests a cookie that was previously stored.
 pub struct CCookieRequest<'a> {
-    pub key: &'a Identifier,
+    #[serde(borrow)]
+    pub key: Identifier<'a>,
```
kralverde, line 13: Does `Identifier` need a lifetime bound now that we use Cows?

pumpkin-protocol/src/client/play/command_suggestions.rs
```diff
@@ -10,15 +10,20 @@ use crate::{
 };
 
 #[packet(PLAY_COMMAND_SUGGESTIONS)]
-pub struct CCommandSuggestions {
+pub struct CCommandSuggestions<'a> {
     id: VarInt,
     start: VarInt,
     length: VarInt,
-    matches: Vec<CommandSuggestion>,
+    matches: Vec<CommandSuggestion<'a>>,
```
kralverde, line 17: This should be a boxed slice instead of a `Vec`

pumpkin-protocol/src/client/play/login.rs
```diff
@@ -39,21 +41,21 @@ impl<'a> CLogin<'a> {
     pub fn new(
         entity_id: i32,
         is_hardcore: bool,
-        dimension_names: &'a [Identifier],
+        dimension_names: Vec<Identifier<'a>>,
```
kralverde, line 44: This should be a boxed slice instead of a `Vec`

kralverde (changes requested):
Sorry it took so long to review, it was a big read through. I made some comments, but I also really don't like all of the `#[serde(bound(deserialize = "'a: 'de"))]`. This limits the packets to the lifetime of the buffer, which we don't want (unless you have a good reason why). I'd like those lifetime bounds removed.

pumpkin-protocol/src/client/play/command_suggestions.rs
suprohub: Why boxed slice? Like it really zero difference. Vector is already boxed slice, ibstead vector has just one more pointer.
I can use just slice if borrow checker pass it

suprohub:
> Sorry it took so long to review, it was a big read through. I made some comments, but I also really don't like all of the `#[serde(bound(deserialize = "'a: 'de"))]`. This limits the packets to the lifetime of the buffer, which we don't want (unless you have a good reason why). I'd like those lifetime bounds removed.

Using Cow<'a, str> allows packets to borrow data directly from the deserialization buffer (e.g., when receiving network packets), avoiding unnecessary clones. If we used Cow<'static, str> or String, we'd force a clone every time, even when borrowing is possible.

pumpkin-protocol/src/client/config/cookie_request.rs
suprohub: Yes, because it use cow insternally too.
It must carry a lifetime 'a to represent the potential borrowed data. However, this doesn’t force packets to depend on the buffer’s lifetime:

When deserializing into an owned Identifier (e.g., from a JSON string), Cow::Owned is used, and 'a can be 'static.

When borrowing from a buffer (e.g., zero-copy deserialization with bincode), 'a matches the buffer’s lifetime.

Why not use Cow<'_, Identifier>? Flexibility.
Really big amouth of code can use just minecraft: or pumpkin: or own namespace. in all places. Instead of every time cloning it, we can use borrow and minimize memory consumptiom.

suprohub:
This can also really help memory for projecst with intensive traffic and big amouth of plugins.

pumpkin-protocol/src/client/play/command_suggestions.rs
kralverde: Just because are aren't expected to change the list at all

suprohub: Ok. Can i just use slices?

kralverde: I think it should be owned. We need to create these structs anyway

Snowiiii:
I will close this for now, you should be able to reopen your PR's yourself

PR #668, opened 2025-03-26
use tags with pumpkin items

kralverde:
## Description
Uses item tags for defining logic instead of hardcoded values where applicable

## Testing

### DIFF
--- a/pumpkin/src/item/items/egg.rs
+++ b/pumpkin/src/item/items/egg.rs
@@ -11,7 +11,9 @@ use pumpkin_data::sound::Sound;
 pub struct EggItem;
 
 impl ItemMetadata for EggItem {
-    const IDS: &'static [u16] = &[Item::EGG.id];
+    fn ids() -> Box<[u16]> {
+        [Item::EGG.id].into()
+    }
 }
 
 const POWER: f32 = 1.5;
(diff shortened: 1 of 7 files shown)


Snowiiii:
I was thinking of also making an array of ids for blocks instead of the current approche using just one u16 as an id

Snowiiii:
Currently when implementing a block we are using a macro, like we did also used for items, like this
```rust
#[pumpkin_block("minecraft:cactus")]
pub struct CactusBlock;
```
This is a problem when we want to implement multiple blocks with the same logic, For example Carpets

kralverde:
> Currently when implementing a block we are using a macro, like we did also used for items, like this
> 
> ```rust
> #[pumpkin_block("minecraft:cactus")]
> pub struct CactusBlock;
> ```
> 
> This is a problem when we want to implement multiple blocks with the same logic, For example Carpets

I think this will be best done in a separate PR

Snowiiii:
> > Currently when implementing a block we are using a macro, like we did also used for items, like this
> > ```rust
> > #[pumpkin_block("minecraft:cactus")]
> > pub struct CactusBlock;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This is a problem when we want to implement multiple blocks with the same logic, For example Carpets
> 
> I think this will be best done in a separate PR

can you may refactor this ?

Issue #664, opened 2025-03-25
EntityDamageEvent, EntityDamageByEntityEvent


haldor-dev:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

I want to be able to cancel damage in parts of my map, disallow attacking players outside of the no damage zone while in the no damage zone, while allowing pvp and damage outside of a certain area. 

### Describe the solution you'd like

Plugin events similar to [EntityDamageEvent](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/entity/EntityDamageEvent.html) and [EntityDamageByEntityEvent](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/entity/EntityDamageByEntityEvent.html) implemented in Pumpkin. 

- If this is a good first issue, then please let me know so I can attempt to implement this myself.

Moyettes:
I've recently been working on the plugin API. Adding more player events, similar to spigot. I somewhat have the EntityDamageEvent and EntityDamageByEntityEvent. I just need to verify my reference to the entity to make sure you can get the entity by the event. And then go over the events I already made and make some functions. Should be done hopefully today or tomorrow. Will be working on the plugin api since I want to see if I can implement skywars as a plugin.

TonimatasDEV:
@RoosterBooster007 The same here, it is already in #1609

PR #671, opened 2025-03-27
Add more events to the Plugin API

Moyettes:
## Description
Adding more Bukkit inspired events to the Pumpkin Plugin API, and solving enhancement issue #664 

## Player events:

- [x] PlayerDeathEvent - called when a player dies
- [x] PlayerDropItemEvent- called when a player drops an item
- [x] PlayerPickupItemEvent - called when a player pickups a item
- [x] PlayerRespawnEvent - called when the player respawns
- [x] PlayerToggleSneakEvent - called when the player toggles their sneak state
- [x] PlayerToggleSprintEvent - called when a player toggles their sprint state

## Entity events:
- [x] EntityDamageEvent - called when a entity is damaged
- [x] EntityDamageByEntityEvent - called when a entity is damaged by another entity

## Testing


vyPal:
Looks good, I'm glad to see other people adding events as well! I have one small comment tho, when using the `send_cancellable {{}}` macro, you can omit the `'cancelled: {}` block if you don't plan on putting anything in there (instead of putting a comment or `false` or something in there)

Moyettes:
I looked over it,  I did call ```cancelled``` was to return false because in ```let remove_entity``` for ```PlayerPickupItemEvent``` it needs to return a bool. I also see ```EntityDamageEvent``` in ```living.rs``` but it needs it since the damage function expects a bool return. I tried omitting the ```'cancelled: {}``` block, but it just started throwing errors, saying I need it. I could be completely wrong, it's my first time working in Rust. Maybe I'm missing something? The other events already have the ```'cancelled: {}` block omited but the 2 events that have it need for its return

pumpkin/src/entity/player.rs
```diff
@@ -1251,13 +1303,26 @@ impl Player {
...
+    pub async fn drop_held_item(self: &Arc<Self>, drop_stack: bool) {
         let mut inv = self.inventory.lock().await;
         if let Some(item_stack) = inv.held_item_mut() {
             let drop_amount = if drop_stack { item_stack.item_count } else { 1 };
-            self.drop_item(item_stack.item.id, u32::from(drop_amount))
-                .await;
-            inv.decrease_current_stack(drop_amount);
+
+            // Create an item stack clone for the event
+            let event_stack = ItemStack::new(drop_amount, item_stack.item.clone());
```
Manolo8, line 1312: Idk how rust works, but there's a inventory lock before the event is launched, if some plugin use the inventory when listening to the event, wouldn't it cause a deadlock?

idk how self.clone works, just asking :D

Moyettes: The lock was already there before implementing the new event. The way I believe it works is that it is requesting access to the inventory, and if some piece of code is using it, it will pause that code until it becomes available again and will release the lock at the end of the function. The lock was not implemented by me. The function should work the same with or without the event.

vyPal: Someone should test this out, it definitely has the potential to cause a deadlock, depending on how getting the item stack is implemented. Also that is not exactly how mutex locks work, if any piece of code wants to access a guarded variable, that code has to acquire a lock (using the `.lock()` function), since only one piece of code can hold this lock at a time, other threads trying to lock the same variable have to wait until the first thread is done and releases the lock. This code could cause a deadlock if the inventory lock `let mut inv = self.inventory.lock().await;` isn't dropped, and a plugin tries to lock the inventory from this event. The deadlock will happen because the plugin can't continue execution, as it is waiting to acquire the lock, and the parent thread (the one that fired the event) also can't continue execution because it is waiting for the plugin to return. This could be avoided by making the plugin spawn a new thread for any logic involving the inventory, but that wouldn't be ideal.

Moyettes: Oh I get what you mean now, if no one tries it early today I'll try it out tomorrow (Technically today) after work since its 1am here. And report back what I find

Moyettes: Tested and it did cause a deadlock issue, pushed a update and tested. No more deadlock when trying to access inventory.

haldor-dev:
Sorry for the dumb question, but do `PlayerToggleSneakEvent` and `PlayerToggleSprintEvent` only cancel sprint/sneak serverside, or also clientside? If it does not do both, then will this cause desync? 

I was looking into the implementation as they seemed useful for something I was working on, as these seemed they would be very useful in my use case, but when checking for this I did not see anything re-syncing the clients state after cancellation, which made me concerned about whether this would cause desync or not.

Moyettes:
> Sorry for the dumb question, but do `PlayerToggleSneakEvent` and `PlayerToggleSprintEvent` only cancel sprint/sneak serverside, or also clientside? If it does not do both, then will this cause desync?
> 
> I was looking into the implementation as they seemed useful for something I was working on, as these seemed they would be very useful in my use case, but when checking for this I did not see anything re-syncing the clients state after cancellation, which made me concerned about whether this would cause desync or not.

I can confirm it works exactly how paper does it. I've cancelled the event every 6 times it was called on both Pumpkin and paper and it behaved the same, and resynced after they toggled again. Remember in paper and this implementation it does not stop the player from sprinting it kind of just stops other clients and server from thinking the player is sprinting. Same for sneaking. Does not stop the client from sneaking on their screen but the server and other clients see the client is not sneaking.

haldor-dev:
> > Sorry for the dumb question, but do `PlayerToggleSneakEvent` and `PlayerToggleSprintEvent` only cancel sprint/sneak serverside, or also clientside? If it does not do both, then will this cause desync?
> > I was looking into the implementation as they seemed useful for something I was working on, as these seemed they would be very useful in my use case, but when checking for this I did not see anything re-syncing the clients state after cancellation, which made me concerned about whether this would cause desync or not.
> 
> I can confirm it works exactly how paper does it. I've cancelled the event every 6 times it was called on both Pumpkin and paper and it behaved the same, and resynced after they toggled again. Remember in paper and this implementation it does not stop the player from sprinting it kind of just stops other clients and server from thinking the player is sprinting. Same for sneaking. Does not stop the client from sneaking on their screen but the server and other clients see the client is not sneaking.

So am I correct with this tl;dr? 
the event is fine, but if my use case requires sneak/sprint state to be resynced to the client then I should implement that functionality separately in my plugin?

If so, should that be documented so people like me don't get confused in the future? (for reference, this is a question not a request)

Moyettes:
That i know of there is no way of forcing a client to sneak on their screen. Only to what the server and other players can see. So technically you can stop them sneaking around the PVP arena and other players wont see them sneaking even if they are trying to sneak just on their screen they will see themself sneaking even though they are not sneaking. Its a client side thing. In the end the event is fine and works as intended. The only thing is maybe some documentation explaining that the client will not be affected might be needed

Snowiiii:
> That i know of there is no way of forcing a client to sneak on their screen. Only to what the server and other players can see. So technically you can stop them sneaking around the PVP arena and other players wont see them sneaking even if they are trying to sneak just on their screen they will see themself sneaking even though they are not sneaking. Its a client side thing. In the end the event is fine and works as intended. The only thing is maybe some documentation explaining that the client will not be affected might be needed

I think this should be documented, So a plugin developer known's that this will only cancel the action server side

Snowiiii:
> I think this should be documented, So a plugin developer known's that this will only cancel the action server side

@Moyettes ?, Also you should fix conflicts

Moyettes:
> > I think this should be documented, So a plugin developer known's that this will only cancel the action server side
> 
> @Moyettes ?, Also you should fix conflicts

Yeah ill finish that later today, Has stuff popup recently. Will get on that soon.

kralverde:
@vyPal what do you think?

pumpkin/src/entity/player.rs
```diff
@@ -1540,7 +1612,20 @@ impl EntityBase for Player {
...
-                self.handle_killed().await;
+                if let Some(player_arc) = self
+                    .world()
+                    .await
+                    .get_player_by_uuid(self.gameprofile.id)
+                    .await
+                {
+                    // Fire the death event
+                    let mut event =
+                        PlayerDeathEvent::new(player_arc, TextComponent::text("noob"), damage_type);
```
vyPal, line 1623: It would be nice to have some other default death message here as well, like the one from here:
```rs
let player_name = self.gameprofile.name.clone();
let default_message = TextComponent::translate(
    "death.attack.generic",
    vec![TextComponent::text(player_name)],
);

// Fire the death event
let mut event = PlayerDeathEvent::new(
    self.clone(),
    default_message,
    DamageType::GENERIC, // Use a generic damage type
);

event = PLUGIN_MANAGER.lock().await.fire(event).await;

// Pass the possibly modified death message to handle_killed
self.handle_killed(event.death_message).await;
```

pumpkin/src/net/packet/play.rs
```diff
@@ -666,22 +668,58 @@ impl Player {
...
+                        send_cancellable! {{
+                            PlayerToggleSneakEvent::new(
+                                self.clone(),
+                                false
+                            );
+
+                            'after: {
+                                entity.set_sneaking(false).await;
+                            }
+                        }}
```
vyPal, line 694: The doc comment for `PlayerToggleSneakEvent` speficies this:
> /// This event contains information about the player and whether they are now sneaking.

That would imply that the boolean value `is_sneaking` should be `true` if the player **is already sneaking** and not when they aren't currently sneaking but will be if the event goes through. 

I'd either swap the `true`/`false` values in the `PlayerToggleSneakEvent`s, or change the documentation comments in the event's file to specify that the value of `is_sneaking` refers to the future state if the event were to pass

pumpkin/src/net/packet/play.rs
```diff
@@ -666,22 +668,58 @@ impl Player {
...
+                        send_cancellable! {{
+                            PlayerToggleSprintEvent::new(
+                                self.clone(),
+                                false // is_sprinting
+                            );
+
+                            'after: {
+                                entity.set_sprinting(false).await;
+                            }
+                        }}
```
vyPal, line 722: Same as with sneaking

pumpkin/src/entity/item.rs
```diff
@@ -77,88 +80,116 @@ impl EntityBase for ItemEntity {
...
+                !event_cancelled
+            };
 
-                            slot_updates.push((slot, item_stack));
+            if proceed_with_pickup {
+                let mut inv = player.inventory.lock().await;
+                let mut total_pick_up = 0;
+                let mut slot_updates = Vec::new();
+                let remove_entity = {
+                    let mut stack_size = self.item_count.lock().await;
```
vyPal, line 111: I would potentially use `event.amount` so that plugins can limit the number of items to be picked up (the item stack will still have the same amount of items, but the plugin will be able to specify something like: the player will only pick up 16 of the 32 items)

pumpkin/src/entity/player.rs
```diff
@@ -441,42 +448,71 @@ impl Player {
...
+                // Play the "no damage" sound if the attack was cancelled
+                world
+                    .play_sound(
+                        Sound::EntityPlayerAttackNodamage,
+                        SoundCategory::Players,
+                        &self.living_entity.entity.pos.load(),
+                    )
+                    .await;
+            }
+        }}
```
vyPal, line 515: Maybe there could be an additional field in the `EntityDamageByEntityEvent` that would specify if the plugin wanted to play the `AttackNodamage` sound. It would be on by default, but would allow plugins to sort-of cancel the no damage sound. Just an idea

pumpkin/src/plugin/api/events/entity/entity_damage.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /// Gets the type of damage dealt.
+    ///
+    /// # Returns
+    /// The type of damage dealt.
+    #[must_use]
+    pub fn get_damage_type(&self) -> DamageType {
+        self.damage_type
+    }
```
vyPal, line 75: I believe previously some people (maybe including @Snowiiii not sure tho) have spoken out against having a getter/setter for each field in the struct and that the code should just directly access the field.

pumpkin/src/plugin/api/events/entity/entity_damage_by_entity.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// A new instance of `EntityDamageByEntityEvent`.
+    pub fn new(
+        victim: Arc<dyn EntityBase>,
+        attacker: Arc<dyn EntityBase>,
+        damage: f32,
+        damage_type: DamageType,
+    ) -> Self {
+        Self {
+            base_event: EntityDamageEvent::new(victim, damage, damage_type),
+            attacker: attacker as Arc<dyn EntityBase>,
```
vyPal, line 42: Is this cast necessary?

vyPal (changes requested):
Overall the PR looks good, the events are documented and closely resemble the Bukkit API (which isn't completely necessary, but it's easier to follow), there are some things I'd still change, but they're mostly smaller code bits. I apologize for not reviewing earlier, I didn't have much time recently

pumpkin/src/plugin/api/events/entity/entity_damage.rs
Snowiiii: yeah, this just feels like Java btw. directly field access would be nicer

Snowiiii:
@Moyettes please respond to the reviews or this will be closed

Snowiiii:
@Moyettes feel free to reopen

pumpkin/src/plugin/api/events/entity/entity_damage.rs
tn-lorenz: Commenting so I don't forget: apparently using `&dyn`s requires getters and setters and therefore `entity_damage` and `entity_damage_by_entity` need to use them as well.

pumpkin/src/plugin/api/events/entity/entity_damage_by_entity.rs
tn-lorenz: Commenting so I don't forget: this has been fixed.

Issue #672, opened 2025-03-28
Add PlayerInteractEvent


haldor-dev:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

I want to block all player interactions (right click/left clicks) on blocks, or run custom actions when a specific item is used. Additionally, it could be useful to track right clicks of non-blocks (of the air)

### Describe the solution you'd like

A event that fires every time a player attempts to click a block (left or right), including right clicking air, but only including left clicking when facing a block. Being able to tell if the player is sneaking from data in the event, would simplify usage a lot as this helps define the type of interaction.

Ideally it would be based on spigot - https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/player/PlayerInteractEvent.html

TonimatasDEV:
@RoosterBooster007 This is already in #1609

PR #673, opened 2025-03-28
Add PlayerInteractEvent (#672)

haldor-dev:
## Description
I added the player interact event, using a similar implementation to spigot, by creating the event in `player_interact.rs` adding calls to the event in `play.rs`.  The event is loosely based on `player_teleport.rs`
Please let me know if there are any issues with this pr.

## Testing
It works on my machine, and all unit tests pass.
I am not confident in my ability to write new tests for this specific functionality in this codebase.
There are no related benchmarks I found to test against.

### DIFF
--- a/pumpkin/src/plugin/api/events/player/player_interact.rs
+++ b/pumpkin/src/plugin/api/events/player/player_interact.rs
@@ -1,0 +1,66 @@
+use pumpkin_macros::{Event, cancellable};
+use pumpkin_data::block::Block;
+use std::sync::Arc;
+use pumpkin_world::block::BlockDirection;
+use pumpkin_world::item::ItemStack;
+use crate::entity::player::Player;
+use crate::world::GetBlockError;
+use super::PlayerEvent;
+
+/// An event that occurs when a `Player` interacts with a `Block` using their hand.
+/// This event does not consider interactions through block movement, eg pressure plates, tripwire hooks, sculk sensors etc.
+///
+/// If the event is cancelled, the interaction  will not happen.
+///
+/// This event contains information about the player, whether the player is sneaking or not, the `Block` they are interacting with,
+/// the `ItemStack` they are interacting using, the block face (`BlockDirection`) they are interacting with.
+#[cancellable]
+#[derive(Event, Clone)]
+pub struct PlayerInteractEvent {
+    /// The player who attempted to interact.
+    pub player: Arc<Player>,
+
+    /// Is the player sneaking?
+    pub sneaking: bool,
+
+    /// The block the player is interacting with.
+    pub block: Result<Block, GetBlockError>,
+
+    /// The face of the block the player is interacting with
+    pub block_direction: BlockDirection,
+
+    /// The ItemStack the player is interacting using
+    pub item_stack: Arc<Option<ItemStack>>
+}
+
+impl PlayerInteractEvent {
+    /// Creates a new instance of `PlayerInteractEvent`.
+    ///
+    /// # Arguments
+    /// - `player`: A reference to the player who interacted.
+    /// - `sneaking`: Is the player sneaking?
+    /// - `block`: The block the player is interacting with.
+    /// - `block_face`: The face of the block the player is interacting with.
+    /// - `item`: The `ItemStack` the player is interacting using.
+    /// - `cancelled`: A boolean indicating whether the interaction should be cancelled.
+    ///
+    /// # Returns
+    /// A new instance of `PlayerInteractEvent`.
+    pub fn new(player: Arc<Player>, sneaking: bool, block: Result<Block, GetBlockError>, block_direction: BlockDirection, item_stack: Arc<Option<ItemStack>>, cancelled: bool) -> Self {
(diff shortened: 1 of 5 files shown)


haldor-dev:
This PR addresses #672

pumpkin/src/plugin/api/events/player/player_interact.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl InteractAction {
+    /// Returns whether the player was `sneaking` during the interaction.
+    #[must_use]
+    pub fn is_sneaking(&self) -> bool {
+        match self {
+            Self::RightClick { sneaking } | Self::LeftClick { sneaking } => *sneaking,
+        }
+    }
```
Snowiiii, line 101: Instead of having sneaking in each enum, why not just make a field in PlayerInteractEvent ?

bruhmolol: I was originally planning on having the following types:
- LeftClickBlock(sneaking, block, block_direction, block_position, itemstack)
- RightClickBlock(sneaking, block, block_direction, block_position, itemstack)
- LeftClickEntity(sneaking, entity, entity_position, itemstack)
- RightClickEntity(sneaking, entity, entity_position, interact_position, itemstack)

Then, I decided entity stuff should probably be separate to avoid making an overly complex event, and that doing it with the complex enum would not fit with the codebase, but I added sneaking while expecting to move to this pattern and kind of forgot about it. I guess it does kind of make sense as most other attributes handle when you mutate their state (other than player), and sneaking does NOT. Additionally, when I tried moving the type to an argument, I had to override the two many args lint. Therefore, moving some stuff to be on the action could be beneficial.

I will make sneaking (somewhat) deal with mutations, and move this back into the main struct, and ignore the lint.

If you want me to move to the proposed pattern above, please let me know, but I do not feel confident implementing these entity events in this codebase.

pumpkin/src/plugin/api/events/player/player_interact.rs
```diff
@@ -0,0 +1,92 @@
...
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        player: Arc<Player>,
+        action: InteractAction,
+        block: Result<Block, GetBlockError>,
+        block_direction: Result<BlockDirection, InvalidBlockFace>,
+        item_stack: Arc<Option<ItemStack>>,
+        position: BlockPos,
+        sneaking: bool,
+        cancelled: bool,
```
Snowiiii, line 66: instead of having this as an var here there is also an send_cancellable macro

bruhmolol: In some uses, I have a default value for `cancelled`, to allow people to override default behaviour by setting cancelled to false when it would be true, as well as allowing the behaviour of people cancelling an event. If something needs to be done here because of, or despite this, please let me know.

vyPal: I do realize how the cancelled argument is being used in this case, but it would be nice to have the `new()` method be coherent with the constructors in the other events. Because of this, I would opt to replace using the `new()` method to construct the event in the `play.rs` file with directly constructing the struct.

pumpkin/src/plugin/api/events/player/player_interact.rs
```diff
@@ -0,0 +1,92 @@
...
+        &self.player
+    }
+}
+
+/// Represents the type of interaction performed by the player.
+#[derive(Clone, Debug)]
+pub enum InteractAction {
+    LeftClick,
+    RightClick,
+}
```
vyPal, line 92: I am quite sure that there is also a middle-click event, maybe it could be added to this enum for future implementation

vyPal (changes requested):
Only some minor things I'd maybe change, otherwise lgtm

pumpkin/src/plugin/api/events/player/player_interact.rs
bruhmolol: Removing the new() seems to make sense and should resolve two issues at once. 
On adding the middle click, it is just one line of code to add it, so does it not make more sense to add it at the same time as the middle click event is implemented, maybe I just add a todo for it? This is not my codebase, so if you me to do this anyways, I will, I am just not sure which approach would be better.

bruhmomen69: Sorry, I should have been more clear about that in my initial message. 
Please let me know if these issues are resolved with my latest push.
If they are, what is the recommended way to deal with the merge conflict? Should I click the resolve conflicts button on GitHub and fix them in the browser?

vyPal: I'd say your latest push looks good. As for the merge conflicts, if GitHub offers you the option to resolve them in the browser, I'd probably do that, it's much easier than doing it from the command line or other utilities

Snowiiii:
Hey @bruhmolol, Can you please fix conflicts

Snowiiii:
feel free to reopen

PR #679, opened 2025-03-28
Never strip debug symbols

Commandcracker:
## Description
See https://doc.rust-lang.org/cargo/reference/profiles.html#strip

## Testing


suprohub:
Wait why stri = false

Commandcracker:
Don't we need them, when profiling ?

Snowiiii:
I'm not sure if its worth to strip debug symbols since when the server crashes we have no information why

Commandcracker:
Should I change this PR to make it so we don't strip anywhere?

Snowiiii:
> Should I change this PR to make it so we don't strip anywhere?

I think this would make more sense tbh

Snowiiii:
Thank you @Commandcracker :+1:

PR #680, opened 2025-03-28
Use none root user inside docker container

Commandcracker:
## Description
Running pumpkin as root opens up a larger attack surface if the container has been hijacked, and we don't need root so we can just use a none root user.

## Testing


Snowiiii:
i'm not sure if we should just use `1000:1000`. Can't we just promt the user with an message saying something like "Please run this as non root". Because what if `1000:1000` does not exist or what if someone wants to use an other user

Commandcracker:
## You're questions

> i'm not sure if we should just use 1000:1000

System users are below 1000. We should actually change to something above 2000 so we don't collide with any real users because we actually don't want that.
For example, Microsoft is using 1654 in .NET docker images or postgres is using 999 which is actually `systemd-journal` on my system, so Postgres chose a bad ID.

> Can't we just promt the user with an message saying something like "Please run this as non root".

It is a common security practice (examples: Microsoft or Postgres) to do it, and we should actually do booth (set a default user in docker and tell them if Pumpkin is run as root also outside of docker)

> Because what if 1000:1000 does not exist

It will still work (In the container the user and group actually does not exist, but it can still run as 1000:1000)

> what if someone wants to use an other user

If someone wants to use a different user, they can easily do that by using `--user` on the docker run command or `user:` in docker-compose. (If they are creating an image based on Pumpkin, they can also use the `USER` instruction to change it) - so there should be no problem.

---

## Some other things

The only "problem" that might occur is with mounting a folder to a volume that the user inside the container can't access, but this can easily be fixed by
- changing the folder permission
- changing the user inside the container like already mentioned
- using the best practice: named Volume and bind mounts - this is actually how every mount should be done (it also comes with some other small advantages)

Another thing: as I used USER without creating a user first, the container won't be able to access a home folder and some other user specific folders, and it won't have a name nor a group name. This might cause some problems if we chose to require those.

---

## Finally, my questions :D

What ID should I change to - 2613 ?
Should I actually use adduser and addgroup so we have names and a home folder?
Should I add the warning when running as root ?

Snowiiii:
Hey @Commandcracker, Sorry for the late response

> What ID should I change to - 2613 ?

Reading trough https://www.baeldung.com/linux/user-ids-reserved-values, it makes sense to may use 101 since this is within the range of the application system user IDs


> Should I actually use adduser and addgroup so we have names and a home folder?

Mhh, not sure about that. I think it can be useful to have a shared folder

> Should I add the warning when running as root ?

I'm not fully sure, if we can just switch the user from root something else, there should be no need for this

I also want to note that i have never used Docker in my life, so I'm probably not the best person to review this

Commandcracker:
> I also want to note that i have never used Docker in my life, so I'm probably not the best person to review this

It's ok, I'll see if I can get someone else to join the review. Just think of docker as a VM :P

> Reading trough https://www.baeldung.com/linux/user-ids-reserved-values, it makes sense to may use 101 since this is within the range of the application system user IDs

It won't make sense, it would be the same as doing 1000:1000 or useradd&groupadd because the users are not the same as on the host system, but when mounting they use the same UID and GID so we want to chose some number that is most likely not normally used. _(like Microsoft in .NET with 1654)_

> > Should I add the warning when running as root ?
>
> I'm not fully sure, if we can just switch the user from root something else, there should be no need for this

I think you misunderstood my question because of bad wording by me, I was thinking of a message that will appear in general (not restricted to docker) so if you run pumpkin as administrator on Windows or as admin on Mac or root on Linux you will get a warning because you should never do that!

Dockerfile
```diff
@@ -23,15 +23,16 @@ FROM alpine:3.21
...
 # set workdir to /pumpkin, this is required to influence the PWD environment variable
 # it allows for bind mounting the server files without overwriting the pumpkin
 # executable (without requiring an `docker cp`-ing the binary to the host folder)
 WORKDIR /pumpkin
 
+RUN apk add --no-cache libgcc && chown 2613:2613 .
+
 ENV RUST_BACKTRACE=1
 EXPOSE 25565
+USER 2613:2613
```
Yimura, line 37: Have you had any ownership issues after mounting a directory into the container? If the folder doesn't exist docker will create it before bind mounting it with root ownership (this will cause issues).

Commandcracker: Currently, I can only test with Podman, and it will just error if you don't have a folder.
In general, you should preferably use named volumes because they don't have those kinds of issues, see https://docs.docker.com/engine/storage/volumes/
I'll see if I can test on docker tomorrow

Snowiiii: did you test this ?

Snowiiii:
@Commandcracker can you resolve conflicts

Dockerfile
Yimura: Tbh, this is probably more of a me issue than a developer issue. A lot of docker containers out there do it this way, I've only seen some of them use an intermediate script to take ownership of the files and then transfer execution to the actual non-privileged user.

Though doing it this way means there's is an initial high privilege, though to exploit it you'd probably need to figure out persistence in some kind of way (modify the entrypoint).

A non-issue to get this merged on my end.

Yimura:
@Snowiiii, probs fine to be merged now.

Snowiiii:
Sorry for the delay, looks good. thank you @Commandcracker :+1:

PR #685, opened 2025-03-30
add Chunk features

Snowiiii:
## Description
This adds the Chunk features generation step.

Chunk features are everything which decorates the Chunk, for example Trees, Ores, flowers and grass

There are 2 things which need to be done
1.  Go trough all **Features** and check **if they should be placed**
2. Check how to place them (So the Feature is configured using the step above)

## Testing


Snowiiii:
Not everything is done yet and there are some bugs (e.g. Trees don't generate on other chunks) but its in an good enough state to be merged

PR #688, opened 2025-03-31
Make coarse and rooted dirt work as intended with the hoe

hthug06:
## Description
Rewrite how the hoe work.
What are the changes:
 - Dirt, grass and dirt path --> farmland
- Coarsed dirt --> dirt
- Rooted dirt --> dirt and drop hanging root (and depend on what is the clicked face). Also only the rooted dirt can be clicked from under the block

## Testing
Place dirt, grass, dirt path, coarsed dirt and rooted dirt and use any type of hoe on it

### DIFF
--- a/pumpkin/src/item/items/hoe.rs
+++ b/pumpkin/src/item/items/hoe.rs
@@ -47,13 +47,13 @@ impl PumpkinItem for HoeItem {
             || block == &Block::COARSE_DIRT
             || block == &Block::ROOTED_DIRT
         {
-            let mut future_block = block.clone();
+            let mut future_block = block;
             let world = player.world().await;
 
             //Only rooted can be right-clicked on the bottom of the block
             if face == &BlockDirection::Down {
                 if block == &Block::ROOTED_DIRT {
-                    future_block = Block::DIRT;
+                    future_block = &Block::DIRT;
                 }
             } else {
                 // grass, dirt && dirt path become farmland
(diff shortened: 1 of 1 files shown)


pumpkin/src/item/items/hoe.rs
```diff
@@ -43,19 +47,70 @@ impl PumpkinItem for HoeItem {
             || block == &Block::COARSE_DIRT
             || block == &Block::ROOTED_DIRT
         {
+            let mut future_block = block.clone();
```
Snowiiii, line 50: does this need to be cloned?

pumpkin/src/item/items/hoe.rs
```diff
@@ -43,19 +47,70 @@ impl PumpkinItem for HoeItem {
...
+                    ),
+                    BlockDirection::West => world.create_entity(
+                        location.up().to_f64().add_raw(-1.0, -0.4, 0.0),
+                        EntityType::ITEM,
+                    ),
+                    BlockDirection::East => world.create_entity(
+                        location.up().to_f64().add_raw(1.0, -0.4, 0.0),
+                        EntityType::ITEM,
+                    ),
+                };
```
Snowiiii, line 106: this is way too much code, if you do the same thing for every match case here, but only the location is different why not just returning the location ?

Snowiiii:
This also not seems like a "Redesign". Existing functionality of the farmland is the same. so it would be better to rename the PR i believe.

Snowiiii:
> Is "make coarse and rooted dirt work as intended with the hoe" better?

yes

Snowiiii:
Looks good now, Thank you @hthug06 :D

PR #692, opened 2025-04-02
Various essential chat improvements

aaneeley:
## Description
Chat "works" but it has a lot of errors and its buggy especially with multiple online players. This PR is meant to more accurately mirror vanilla chat handling logic.

## Progress
- [x] Correctly implement message signature cache and last seen messages
- [x] Chat message validation & error handling
- [x] Chat session validation & error handling

## Testing

### DIFF
--- a/pumpkin-util/src/math/mod.rs
+++ b/pumpkin-util/src/math/mod.rs
@@ -172,7 +172,7 @@ pub fn lerp3(
 
 /// Calculates a Polynomial Rolling Hash
 /// Mojang's checksum algorithm for previous messages
-pub fn polynomial_rolling_hash(signatures: &Vec<Box<[u8]>>) -> u8 {
+pub fn polynomial_rolling_hash(signatures: &[Box<[u8]>]) -> u8 {
     let mut i: i32 = 1;
 
     for signature in signatures.iter() {
--- a/pumpkin-util/src/math/mod.rs
+++ b/pumpkin-util/src/math/mod.rs
@@ -177,7 +177,7 @@ pub fn polynomial_rolling_hash(signatures: Vec<Box<[u8]>>) -> u8 {
 
     for signature in signatures.iter() {
         // Ensure hash_code operates on i64
-        i = i.wrapping_mul(31).wrapping_add(hash_code(signature) as i32);
+        i = i.wrapping_mul(31).wrapping_add(hash_code(signature));
     }
 
     let b = (i & 0xFF) as u8; // Take the least significant byte.
--- a/pumpkin-util/src/math/mod.rs
+++ b/pumpkin-util/src/math/mod.rs
@@ -176,19 +176,68 @@ pub fn polynomial_rolling_hash(signatures: &[Box<[u8]>]) -> u8 {
     let mut i: i32 = 1;
 
     for signature in signatures.iter() {
-        i = i.wrapping_mul(31).wrapping_add(hash_code(signature)); // Wrap to prevent multiplication overflow
+        i = i.wrapping_mul(31).wrapping_add(java_array_hash(signature)); // Wrap to prevent multiplication overflow
     }
 
     let b = (i & 0xFF) as u8; // Take the least significant byte.
     if b == 0 { 1 } else { b } // Ensure the checksum is never zero.
 }
 
-/// Basically a 1:1 rewrite of Java's Arrays.hashCode()
-fn hash_code(data: &[u8]) -> i32 {
+/// Arrays.hashCode() and String.hashCode() have similar but different implementations.
+fn java_array_hash(data: &[u8]) -> i32 {
     let mut hash: i32 = 1;
     for &byte in data {
-        let signed_byte = byte as i8 as i32;
-        hash = hash.wrapping_mul(31).wrapping_add(signed_byte); // Wrap to prevent multiplication overflow
+        let signed_byte = byte as i32;
+        hash = hash.wrapping_mul(31).wrapping_add(signed_byte);
     }
     hash
+}
+
+pub fn java_string_hash(string: &str) -> i32 {
+    let mut result = 0i32;
+    for char_encoding in string.encode_utf16() {
+        result = 31i32
+            .wrapping_mul(result)
+            .wrapping_add(char_encoding as i32);
+    }
+    result
+}
+
+#[test]
+fn test_java_hash() {
+    let values = [
+        ("", 0, 1),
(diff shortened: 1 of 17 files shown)


pumpkin-util/src/math/mod.rs
```diff
@@ -169,3 +169,26 @@ pub fn lerp3(
...
+    for signature in signatures.iter() {
+        i = i.wrapping_mul(31).wrapping_add(hash_code(signature)); // Wrap to prevent multiplication overflow
+    }
+
+    let b = (i & 0xFF) as u8; // Take the least significant byte.
+    if b == 0 { 1 } else { b } // Ensure the checksum is never zero.
+}
+
+/// Basically a 1:1 rewrite of Java's Arrays.hashCode()
+fn hash_code(data: &[u8]) -> i32 {
```
kralverde, line 187: I think this is also implemented somewhere in `pumpkin_util/random`. It is used in the legacy random string driver. Maybe add a generic "java hash" for use in both places?

pumpkin/src/entity/player.rs
```diff
@@ -1886,6 +1890,87 @@ impl ChatSession {
...
+
+impl MessageCache {
+    /// Not used for caching seen messages. Only for non-indexed signatures from senders.
+    pub fn cache_signatures(&mut self, signatures: &[Box<[u8]>]) {
+        for sig in signatures.iter().rev() {
+            if self.full_cache.contains(sig) {
+                // Already cached
+                continue;
+            }
+            self.full_cache.truncate(MAX_CACHED_SIGNATURES as usize);
```
kralverde, line 1962: This leaves the `self.full_cache` with a length of `MAX_CACHED_SIGNATURES + 1` right? Also you might want to look into a `VecDeque` for efficient removal of the first element without shifting the elements after.

pumpkin/src/net/packet/play.rs
```diff
@@ -107,6 +117,77 @@ impl PumpkinError for BlockPlacingError {
...
+    InvalidPublicKey,
+}
+
+impl PumpkinError for ChatError {
+    fn is_kick(&self) -> bool {
+        true
+    }
+
+    fn severity(&self) -> log::Level {
+        match self {
```
kralverde, line 146: Can't this just directly be `Level::Warn`?

kralverde (changes requested):
Looks really clean! Just have a couple of nit-picks

pumpkin/src/net/packet/play.rs
aaneeley: Good catch. I wrote it like this initially in case I wanted to add some that weren't `Warn` but I never did.

pumpkin-util/src/math/mod.rs
aaneeley: Interestingly Java implements Arrays.hashCode and String.hashCode differently. Arrays.hashCode has a starting value of 1 vs 0 so the functions are not interchangeable as seen here:

image

I'll see if I can find a clean way to combine the functions

aaneeley: I think combining the functions would just increase complexity rather than decrease it. I'll go ahead a move the both functions/tests to `pumpkin_util/math/mod.rs` and rename them to make it more straightforward.

pumpkin/src/entity/player.rs
aaneeley: I'll go ahead and implement `VecDeque` for the full cache. I think keeping `last_seen` as a `Vec` is a good idea though because it requires frequent contiguous access.

kralverde:
Kind of a nit, but since we control the signature queue, it think it would be best to use `VecDequeue::with_capacity(max_signatures)` and to control insertions such that they only happen after removals to maintain the size of the vec <= max_signatures. This will prevent possible unnecessary overallocation of the vec as it will never need to allocate more memory.

pumpkin/src/net/packet/play.rs
```diff
@@ -841,6 +902,54 @@ impl Player {
...
+    /// Runs vanilla checks for a valid player session
+    pub async fn validate_chat_session(
+        &self,
+        server: &Server,
+        session: &SPlayerSession,
+    ) -> Result<(), ChatError> {
+        // Verify key length
+        if session.key_signature.len() != 512 {
+            return Err(ChatError::InvalidPublicKey);
+        }
```
Snowiiii, line 905: I think this should be done in the packet, so we only read 512 bytes directly

pumpkin/src/server/mod.rs
```diff
@@ -38,12 +42,33 @@ use tokio::sync::{Mutex, RwLock};
...
+    #[serde(rename = "publicKey")]
+    pub public_key: String,
+}
+#[derive(Deserialize, Clone, Debug)]
+pub struct MojangPublicKeys {
+    #[serde(rename = "profilePropertyKeys")]
+    pub profile_property_keys: Vec<JsonPublicKey>,
+    #[serde(rename = "playerCertificateKeys")]
+    pub player_certificate_keys: Vec<JsonPublicKey>,
+    #[serde(rename = "authenticationKeys")]
```
Snowiiii, line 68: this is the wrong place, this should be in a seperate file like the auth or inside the auth

pumpkin/src/world/mod.rs
```diff
@@ -96,6 +96,20 @@ bitflags! {
...
+        const ADD_PLAYER            = 0x01;
+        const INITIALIZE_CHAT       = 0x02;
+        const UPDATE_GAME_MODE      = 0x04;
+        const UPDATE_LISTED         = 0x08;
+        const UPDATE_LATENCY        = 0x10;
+        const UPDATE_DISPLAY_NAME   = 0x20;
+        const UPDATE_LIST_PRIORITY  = 0x40;
+        const UPDATE_HAT            = 0x80;
+    }
+}
```
Snowiiii, line 111: What???. Why you placed this here

aaneeley: You're right. Should probably go in same file as `CPlayerInfo` right?

pumpkin/src/net/packet/play.rs
aaneeley: It looks like the length isn't technically fixed at 512. It's length prefixed and size *could* change. I'll just remove this check so it doesn't break later if/when Mojang changes the size of `key_signature`

Snowiiii:
Looks good now. Thank you @aaneeley :+1:

PR #693, opened 2025-04-02
Add security_opt and new restart policy to docker-compose options

BigPebble123:
## Description

This PR adds YAML-anchor security options and a restart policy:
- `no-new-privileges:true`: This will prevent the container from gaining any new privileges. It will deny any calls to `setuid`, `setgid` and other similar methods. By preventing processes in the container from gaining new privileges, we can mitigate the risk of privilege escalation and container escape attacks if any part of the app is vulnerable
- `apparmor=docker-default`: This is the default option that docker sets, but it makes it clear that such an option will be used. Can be useful to easily change custom profiles. 
- `restart: unless-stopped`: The difference from the default `always` policy is that the container won't auto-start on daemon restart (e.g. on host reboot) if the user previously manually stopped the container ([source](https://docs.docker.com/engine/containers/start-containers-automatically/)).

## Testing

Pumpkin does not and is unlikely to gain additional container privileges, so we can safely disable this option.


docker-compose.yml
```diff
@@ -1,5 +1,12 @@
+x-common-keys-core: &common-keys-core
+  security_opt:
+    - no-new-privileges:true
+    - apparmor=docker-default
```
Commandcracker, line 4: I don't think we should set default values another time.
Just make a comment if you want to emphasize the use of apparmor profiles.

docker-compose.yml
```diff
@@ -1,5 +1,12 @@
+x-common-keys-core: &common-keys-core
```
Commandcracker, line 1: I don't think that we should define an anchor because we are only running one container, and it just makes the compose file unnecessarily more complex.

Commandcracker:
I tested if the option is correctly set by running the following command inside the container:

`grep NoNewPrivs /proc/self/status`
`NoNewPrivs:     1`

docker-compose.yml
Commandcracker: I agree, but I think it only makes sense when there are more settings and I think the people who know how to use it are the ones who would create an anchor them self, so I'm not sure :p

Snowiiii:
Why do we want to use `restart: unless-stopped` because i think it makes sense for a server to auto start

Commandcracker:
It will do that.

Similar to always, except that when the container is stopped (manually or otherwise), it isn't restarted even after Docker daemon restarts.

Snowiiii:
> It will do that.
> 
> Similar to always, except that when the container is stopped (manually or otherwise), it isn't restarted even after Docker daemon restarts.

But I think we do want that it Auto restarts when the deemon restarts ?

Commandcracker:
No it will also do that just not if you stop it like

With always
If you stop the container
Restart Docker daemon
Then the container will spin up

With unless stopped
If you stop the container
Restart Docker daemon
Then the container will not spin up

With always and unless stopped
If you don't stop
Restart Docker daemon
The container will spin up again

Yimura:
@Snowiiii, this should be fine to be merged once the conflicts have been resolved.

@BigPebble123, if you could?

PR #694, opened 2025-04-02
Add docker to dependabot

Commandcracker:
## Testing


Commandcracker:
No because we don't use any other images inside docker-compose, but we should add it when we add some kind of db container alongside pumpkin.

Commandcracker:
Well we could just add it preemptively.

Snowiiii:
Thanks @Commandcracker :+1:

PR #699, opened 2025-04-04
Fix creative block breaking bug

teknostom:
## Description
During fluid flow testing, we discovered that the handling for creative breaking was not sending the updates. This would put the client in a state of assuming the block wasn't broken yet, so it would remain empty on the client side, even when told that it wasn't empty.

## Testing
Manual


Snowiiii:
Looks good. thanks @teknostom :+1:

PR #702, opened 2025-04-05
Dropping impossible but protocol-compliant packets

BigPebble123:
## Description

I added:
- a configuration option (`allow_impossible_actions`) that allows the admin to decide whether they want to allow packets that are protocol-compliant, but the non-modded client won't send them (e.g. sign line above 15 characters)
- several protocol validations that enforce the mentioned setting
- a few TODOs
- a list of Minecraft locales (`locales.json`) to implement locale validation in client information packet

## Testing

Manual testing, everything seems to be fine!


Snowiiii:
Instead of just having one value of either allow or disallow everything we should make a new config value for everything which is possible. it would also make more sense to have a array of string of **banned** locales instead of all the locals which exist, this would also not break updates when a new locale is added.
Also everything which is inside the asset folder is from the extractor, you did not make a PR for that.

This is something we want to do later if everything is done, we first should  add all the features and then configuration. otherwise configuration will block new features

PR #704, opened 2025-04-05
Fix fluid weirdness

teknostom:
## Description
Fixed two issues with water:
- 1-level water not disappearing
- weird spreading behaviour

## Testing
manual


Snowiiii:
Nice works correctly now

PR #712, opened 2025-04-06
Fixed fall damage

akashbeh:
## Description
Changed LivingEntity::update_fall_distance so that fall damage works

## Testing
Verified correct damage in-game.
Issues:
- Water does not stop fall damage
- One time, when I was getting out of water, I took damage. Could not replicate


Snowiiii:
I will try to test this soon™️

Snowiiii:
Looks good. Finally it works again. Thank you @akashbeh

PR #713, opened 2025-04-06
implement Door placement & pumpkin_data changes

unschlagbar:
## Description
I implemented the vanilla door placement. The doors now interact with redstone.
Fences now connect to blocks that are solid and full cube blocks like in vanilla.
the bool properties of blockstates are now in stageflags and can be access with functions. major code cleanup.
Reduced the size of BlockState from 48 bytes to 32, clone & copy this type is now really cheap.
Merging this pr fixes this issue https://github.com/Pumpkin-MC/Pumpkin/issues/729
## Testing


4lve:
The issue with leaves is due to redstone not accounting for transparency atm

pumpkin-data/build/block.rs
```diff
@@ -1098,6 +1098,12 @@ pub(crate) fn build() -> TokenStream {
             pub is_solid: bool,
         }
 
+        impl BlockState {
+            pub fn is_full_cube(&self) -> bool {
```
4lve, line 1102: We already have is_solid and this calculation seems off

pumpkin/src/block/blocks/doors.rs
```diff
@@ -56,14 +61,93 @@ async fn toggle_door(world: &Arc<World>, block_pos: &BlockPos) {
...
+
+#[inline]
+async fn get_hinge(
+    world: &World,
+    block: &Block,
+    block_pos: &BlockPos,
+    use_item_on: &SUseItemOn,
+    player_direction: &HorizontalFacing,
+) -> DoorHinge {
+    let lv4 = block_pos.up();
```
4lve, line 99: Fix variable naming

pumpkin-data/build/block.rs
unschlagbar: it has nothing todo with solid, look at the java code

unschlagbar: is_full_cube just checks if it is "normal" block, a grindstone for example shouldnt be influencing the door placement even when it is solid

pumpkin-data/build/block.rs
```diff
@@ -1098,6 +1098,12 @@ pub(crate) fn build() -> TokenStream {
...
+                !self.collision_shapes.is_empty() && self.collision_shapes[0] == 0
+            }
+        }
```
Snowiiii, line 1105: Please use the extractor to get the Vanilla `is_full_cube` value

unschlagbar: Vanilla does it the same way, full blocks have always collsion_shape[0] wich goes from 0,0,0 to 1,1,1.
i also think that the fn sits in the wrong place tho, how about making the BlockState struct in a seperate file and import it to the genereated file?

unschlagbar: Why does the bool properties use Boolean instead of bool. it makes changing them so unclean because of the to_bool()

4lve: > Vanilla does it the same way, full blocks have always collsion_shape[0] wich goes from 0,0,0 to 1,1,1. i also think that the fn sits in the wrong place tho, how about making the BlockState struct in a seperate file and import it to the generated file?

Putting it in the extractor gives in compile time cache.

Boolean is an enum due to how state ids are calculated

unschlagbar: i changed the calculation, easy fix ;)

4lve: Cool, then it can be a bool

Snowiiii:
> (know issue some blocks like leaves also connect because there marked as solid wich is wrong)

we extract the vanilla value, how this can be wrong ?

4lve:
> > (know issue some blocks like leaves also connect because there marked as solid wich is wrong)
> 
> we extract the vanilla value, how this can be wrong ?

```java
public static boolean cannotConnect(BlockState state) {
  return state.getBlock() instanceof LeavesBlock
     || state.isOf(Blocks.BARRIER)
     || state.isOf(Blocks.CARVED_PUMPKIN)
     || state.isOf(Blocks.JACK_O_LANTERN)
     || state.isOf(Blocks.MELON)
     || state.isOf(Blocks.PUMPKIN)
     || state.isIn(BlockTags.SHULKER_BOXES);
}
```

Cause vanilla checks if it's leaves

Snowiiii:
why people want to change everything in one PR

unschlagbar:
merges are to slow, dont want to have muliple workspaces

unschlagbar:
This rp should be ready

Snowiiii:
> merges are to slow, don't want to have muliple pumpkin workspaces

What do you expect now, That i will just merge this then ?. You also don't clone multiple times its pretty easy to just create and switch branches

unschlagbar:
we can't just get rid of BlockStateRef, because you have to assign the right id to the blockstate and I can't just get it from the index
image

unschlagbar:
The output files would explode, there must be another solution

4lve:
> The output files would explode, there must be another solution

Ye, but it's worth it, you could possibly make the line size smaller with a macro but that complicates how it looks

unschlagbar:
Anyway, I'm done with it for now

pumpkin-world/src/chunk/palette.rs
```diff
@@ -166,7 +164,10 @@ impl<V: Hash + Eq + Copy + Default, const DIM: usize> PalettedContainer<V, DIM>
             }
 
             // TODO: Can we do this all with an `array::from_fn` or something?
-            let mut cube = Box::new([[[V::default(); DIM]; DIM]; DIM]);
+            // The uninit is totally safe since we don't read it before we assign a proper value
```
kralverde, line 167: This is wrong; we expect to have the default value in error cases

kralverde (changes requested):

kralverde:
I'd prefer if this PR was split into separate PRs:
* pumpkin data changes
* re-naming/refactoring
* the actual door implementation

unschlagbar:
> I'd prefer if this PR was split into separate PRs:
> 
> * pumpkin data changes
> * re-naming/refactoring
> * the actual door implementation

Your right i have dont way to much here but i dont want to do these splitting as i already spend way to much time. The next pr would be definitely smaller

pumpkin/src/item/items/hoe.rs
```diff
@@ -90,8 +95,20 @@ impl PumpkinItem for HoeItem {
...
                 let entity = world.create_entity(location, EntityType::ITEM);
                 // TODO: Merge stacks together
-                let item_entity =
-                    Arc::new(ItemEntity::new(entity, Block::HANGING_ROOTS.item_id, 1).await);
+                let item_entity = Arc::new(
+                    ItemEntity::new(
+                        entity,
+                        Block::HANGING_ROOTS.item_id,
+                        1,
+                        Vector3::new(
```
kralverde, line 103: This code is duplicated above, can this be condensed?

kralverde (changes requested):

PR #714, opened 2025-04-06
implement item durability

thiago-buarqque:
## Description

- What was changed?
  - It is possible now to increment the damage of a given item after use_item, use_item_on, on_attack and normal_use events or whatever other event that an item might handle.
  - We're now [saving](https://github.com/Pumpkin-MC/Pumpkin/pull/714/files#diff-e9446902ce129f03dc215f96671f01948d5662ee2923ca8e2092578ca85f7638R142) and [reading](https://github.com/Pumpkin-MC/Pumpkin/pull/714/files#diff-e9446902ce129f03dc215f96671f01948d5662ee2923ca8e2092578ca85f7638R165) item's damage for a given player.

- Why were these changes necessary?
It is part of the vanilla game.
- What is the impact of this change?
This is the first step to read and write item components (data) when saving/reading a player's data in a world. Also this change introduces [a way of sending structured components](https://github.com/Pumpkin-MC/Pumpkin/pull/714/files#diff-06dc7126c393ec5adad4a7769e2d5f7ab1475fb99502cd754675cb30ace807f7R36) for a given `Slot`.

- Are there any known issues or limitations?
This PR does not account for the unbreakable enchantment since it's not yet implemented.

## Demo
Turn sound on :loud_sound: to hear the breaking sound :D

https://github.com/user-attachments/assets/3b636303-0772-431d-8e7f-5fbf7652e453

### DIFF
--- a/pumpkin/src/item/items/hoe.rs
+++ b/pumpkin/src/item/items/hoe.rs
@@ -8,6 +8,7 @@ use pumpkin_data::block::Block;
 use pumpkin_data::entity::EntityType;
 use pumpkin_data::item::Item;
 use pumpkin_data::tag::Tagable;
+use pumpkin_util::GameMode;
 use pumpkin_util::math::position::BlockPos;
 use pumpkin_world::block::BlockDirection;
 use std::sync::Arc;
@@ -31,6 +32,19 @@ impl ItemMetadata for HoeItem {
 
 #[async_trait]
 impl PumpkinItem for HoeItem {
+    async fn on_after_dig(
+        &self,
+        _item: &Item,
+        player: &Player,
+        _location: BlockPos,
+        _block: &Block,
+        _server: &Server,
+    ) {
+        if player.gamemode.load() != GameMode::Creative {
+            self.increment_damage(player).await;
+        }
+    }
+
     async fn use_on_block(
         &self,
         _item: &Item,
@@ -94,6 +108,10 @@ impl PumpkinItem for HoeItem {
                     Arc::new(ItemEntity::new(entity, Block::HANGING_ROOTS.item_id, 1).await);
                 world.spawn_entity(item_entity.clone()).await;
                 item_entity.send_meta_packet().await;
+            }
+
+            if player.gamemode.load() != GameMode::Creative {
+                self.increment_damage(player).await;
             }
         }
     }
--- a/pumpkin/src/item/pumpkin_item.rs
+++ b/pumpkin/src/item/pumpkin_item.rs
@@ -1,3 +1,4 @@
+use crate::entity::Entity;
 use crate::entity::player::Player;
 use crate::server::Server;
 use async_trait::async_trait;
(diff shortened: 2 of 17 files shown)


pumpkin/src/item/items/hoe.rs
```diff
@@ -31,6 +32,19 @@ impl ItemMetadata for HoeItem {
 
 #[async_trait]
 impl PumpkinItem for HoeItem {
+    async fn on_after_dig(
```
Snowiiii, line 35: is this vanilla ?

pumpkin/src/item/pumpkin_item.rs
```diff
@@ -13,6 +14,17 @@ pub trait ItemMetadata {
 #[async_trait]
 pub trait PumpkinItem: Send + Sync {
     async fn normal_use(&self, _block: &Item, _player: &Player) {}
+    async fn on_attack(&self, _item: &Item, _victim: &Entity) {}
```
Snowiiii, line 17: is this vanilla?

Snowiiii:
I also wonder if the slot components are properly saved, so if you rejoin or the server restarts its saved in the player data

thiago-buarqque:
@Snowiiii 
> I also wonder if the slot components are properly saved, so if you rejoin or the server restarts its saved in the player data
Yes, it is. It's mentioned in the description

> We're now [saving](https://github.com/Pumpkin-MC/Pumpkin/pull/714/files#diff-e9446902ce129f03dc215f96671f01948d5662ee2923ca8e2092578ca85f7638R142) and [reading](https://github.com/Pumpkin-MC/Pumpkin/pull/714/files#diff-e9446902ce129f03dc215f96671f01948d5662ee2923ca8e2092578ca85f7638R165) item's damage for a given player.

Not very clearly thought, my bad.

thiago-buarqque:
> i also think instead of using duplicated logic for all the tools you can just use a trait like Mojang does (abstract class)
I don't have access to Mojang code, how can I get access to it?

I did not create a trait for that because not all items is damaged after the same events. For instance

All tools are damaged after breaking a block or using it in a block (if it's possible)

Swords break when hitting an entity.

Elytra breaks when in the chest slot and flying

others...

thiago-buarqque:
@Snowiiii I like the explicit way of saying "hey, after this event happen (normal use, use on block etc) damage the item

tomasalias:
@thiago-buarqque What happened? Are you still working on implementing this?

thiago-buarqque:
@tomasalias Hello, man. Sorry, I forgot to mention here that I stopped working on this. I've got some personal issues, and I am not really coding outside work

pumpkin/src/item/pumpkin_item.rs
thiago-buarqque, line 17: This is `postHit` in vanilla

pumpkin/src/item/items/hoe.rs
thiago-buarqque, line 35: This is `postMine` in vanilla

PR #725, opened 2025-04-11
Plugin improvements

suprohub:
## Description
This pr tries optimize and improve plugin system. Now plugin loader dont have in memory unloaded plugins. Its just a memory waste. Like we can just load it from disk.

Now plugin manager have plugin loader. Anyone can make own plugin what introduces own loader like lua loader or wasm loader. By default we have native loader.

## Testing
`cargo run`, see what its works

Create plugin by tutorial and fix errors
`cargo build` & `./pumpkin`, see what its works.


kralverde:
Thank you!

Issue #734, opened 2025-04-12
Water flows up to two blocks from source


NielsMooren:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

MacOS

### Server Software Version/Commit

0.1.0-dev+1.21.5 (8246e968)

### What happened?

When water flows from a source block, after at least two blocks it stops and does not flow further.

### To Reproduce

Find a water source block, remove at least 3 or more blocks next to the source block, watch how it doesn't flow further.

### Expected behavior

Water should flow farther than 2 blocks.

PR #735, opened 2025-04-12
fix(fluid): fix flowing mechanics

NielsMooren:
## Description
Fixes https://github.com/Pumpkin-MC/Pumpkin/issues/734

Water now flows for the right amount of time.

Official rules: https://minecraft.fandom.com/wiki/Water

## Testing
I have tested holes, the right length, drops, larger drops, and spreads.

Sorry for the compression, the source video was 300mb.

https://github.com/user-attachments/assets/ce2a11ab-8775-4210-b174-b62106f6f8a4

### DIFF
--- a/pumpkin/src/block/blocks/fluids/flowing_fluid.rs
+++ b/pumpkin/src/block/blocks/fluids/flowing_fluid.rs
@@ -1,4 +1,3 @@
-use std::collections::HashMap;
 use std::sync::Arc;
 
 use async_trait::async_trait;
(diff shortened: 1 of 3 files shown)


teknostom (changes requested):
I'm sorry, but this PR currently does not have any merit. You removed a bunch of code that is made to match vanilla behaviour. Please try to understand how the code works before removing stuff that you think is not needed. There is documentation on how to get the source code for Minecraft, and that is what we are working towards matching.

PR #742, opened 2025-04-15
Fix position desync between clients

### LINKED ISSUE #363: Teleport de-sync

### Operating System

Manjaro Linux

f29cfbc

### What happened?

Player position gets de-synced between clients when teleporting more than 8 blocks. This results in the player falling through the ground on other players POV.

### To Reproduce

/teleport ~ ~10 ~
You should then be in the ground for other players when falling to the ground.

### Expected behavior

Teleport 10 blocks up and then fall down stopping at the ground for all players.

Alvsch:
## Description
This PR makes sure the player position isn't desynced between players when moving too fast (>8 blocks/tick).
Also fixes `/tp` desync (Fixes #363)  

## Testing
Two minecraft players looking at each other
- `/tp ~ ~10 ~` and make sure no weird stuff happens
- Move faster than 8 blocks and make sure no weird stuff happens (hacked clients are good for this)


pumpkin-protocol/src/client/play/synchronize_vehicle_position.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use crate::{
+    ClientPacket, PositionFlag, VarInt,
+    ser::{NetworkWriteExt, WritingError},
+};
+
+// https://minecraft.wiki/w/Java_Edition_protocol#Synchronize_Vehicle_Position
+// Badly documented and confusing packet imo
+#[packet(PLAY_TELEPORT_ENTITY)]
+pub struct CSynchronizeVehiclePosition<'a> {
```
Snowiiii, line 15: Why is this called `CSynchronizeVehiclePosition` ? also we already have a packet called `CTeleportEntity`

Alvsch: check out the linked docs, idk everything is very confusing

Alvsch: they renamed the packet resources in 1.21 or smth

Snowiiii:
Looks good. Nice to have that finally fixed. Thank you @Alvsch

zly2006:
Hi, I disagree with this PR. I think the packets naming should follow mojang's names, so relatives should be in CTeleportEntity. Accordingly, another packet (the new CTeleportEntity after this PR) should be named CEntityPositionSync.

In fact, the original CTeleportEntity is the packet sent when calling player.teleport in vanilla, CSynchronizeVehiclePosition totally makes no sense.

PR #744, opened 2025-04-17
Add sound for others when placing blocks

Alvsch:
## Description
Send WorldEvent packet which plays sound and particles.
The one who places the block predicts the sound clientside

## Testing
Place blocks next to another player and listen for sound on both clients

### DIFF
--- a/pumpkin/src/net/packet/play.rs
+++ b/pumpkin/src/net/packet/play.rs
@@ -35,8 +35,7 @@ use pumpkin_inventory::player::{
 };
 use pumpkin_macros::send_cancellable;
 use pumpkin_protocol::client::play::{
-    CBlockUpdate, COpenSignEditor, CPlayerInfoUpdate, CPlayerPosition, CSetContainerSlot,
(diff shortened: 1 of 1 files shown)


pumpkin/src/net/packet/play.rs
```diff
@@ -1735,6 +1735,14 @@ impl Player {
                 .player_placed(world, &block, new_state, &final_block_pos, face, self)
                 .await;
 
+            // Block place particles + sound
+            world
+                .broadcast_packet_except(
+                    &[self.gameprofile.id],
+                    &CWorldEvent::new(2001, location, i32::from(new_state), false),
+                )
+                .await;
```
Snowiiii, line 1744: I don't think this is vanilla, and this is also for breaking. i also wrote an extractor for world events so instead of 2001 you could just use `WorldEvent::BlockBroken as i32`. We need to verify how vanilla handles the placing sounds and particles before adding that

Alvsch: You're right, this is for breaking.

PR #747, opened 2025-04-17
Fix rubberbanding after death

### LINKED ISSUE #728: Movement desync(?) when dying

### Operating System

Arch Linux 6.14.1-arch1-1

Pumpkin 0.1.0-dev+1.21.5 (c1387751)

### What happened?

It seems like dying while moving causes constant rubberbanding, leading to stuttering while on the respawn screen.

### To Reproduce

Execute `/kill` while moving quickly, or die from fall damage in survival. (This happens in any direction, not just falling)

https://github.com/user-attachments/assets/7e74c136-d53f-4324-8c88-92e350a9e9ee

### Expected behavior

The normal vanilla death animation plays and the player/camera stops moving.

Alvsch:
## Description
Make so the server doesn't send death packets every tick after a living entity is dead, even if the entity were to tick after it's supposed removal.
Fixes #728 

## Testing
Slam myself into the ground in survival mode, could not replicate mentioned issue.

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -256,7 +256,9 @@ impl EntityBase for LivingEntity {
             let time = self
                 .death_time
                 .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
-            if time >= 20 {
+
+            // Only remove entity and send remove packets once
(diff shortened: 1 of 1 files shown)


kralverde:
Thank you!

PR #754, opened 2025-04-20
Fix chunk loading and light data for vanilla imported save

zly2006:
## Description

Fix chunk serializer and network for light only chunks.
Added a basic light engine that only stores data for further calculating light.

## Testing

I tested this using a private map of MC 1.21.5.

### DIFF
--- a/pumpkin-protocol/src/client/play/chunk_data.rs
+++ b/pumpkin-protocol/src/client/play/chunk_data.rs
@@ -48,8 +48,8 @@ impl ClientPacket for CChunkData<'_> {
         let mut sky_light_buf = Vec::new();
         let mut block_light_buf = Vec::new();
         // mark extra chunk data as empty, finish it when we have a light engine.
-        let mut sky_light_empty_mask = 1 + (1 << self.0.section.sections.len() + 2);
-        let mut block_light_empty_mask = 1 + (1 << self.0.section.sections.len() + 2);
+        let mut sky_light_empty_mask = 1 + (1 << (self.0.section.sections.len() + 2));
+        let mut block_light_empty_mask = 1 + (1 << (self.0.section.sections.len() + 2));
         let mut sky_light_mask = 0;
         let mut block_light_mask = 0;
 
--- a/pumpkin-protocol/src/client/play/chunk_data.rs
+++ b/pumpkin-protocol/src/client/play/chunk_data.rs
@@ -60,7 +60,7 @@ impl ClientPacket for CChunkData<'_> {
             let light_data_size: VarInt = LightContainer::<16>::array_size().try_into().unwrap();
 
             if let LightContainer::Full(data) =
-                &*self.0.light_engine.sky_light[light_index].lock().await
+                &*self.0.light_engine.sky_light[light_index].read().await
             {
                 let mut buf = Vec::new();
                 buf.write_var_int(&light_data_size)?;
@@ -73,7 +73,7 @@ impl ClientPacket for CChunkData<'_> {
 
             // Write block light
             if let LightContainer::Full(data) =
-                &*self.0.light_engine.block_light[light_index].lock().await
+                &*self.0.light_engine.block_light[light_index].read().await
             {
                 let mut buf = Vec::new();
                 buf.write_var_int(&light_data_size)?;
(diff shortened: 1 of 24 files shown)


jobpaardekooper:
This does not properly fix the problem although it likely helps prevent a crash or chunk load error. The approach to fix this should be a little different. This problem is also related to current situation where light levels are not correctly being displayed for worlds that were generated by the vanilla generator and moved to pumpkin.

zly2006:
> This does not properly fix the problem although it likely helps prevent a crash or chunk load error. The approach to fix this should be a little different. This problem is also related to current situation where light levels are not correctly being displayed for worlds that were generated by the vanilla generator and moved to pumpkin.

I have fixed this issue. Its a network bug, the light data send is actually the data of the subchunk above it.

pumpkin-world/src/chunk/format/mod.rs
```diff
@@ -149,6 +222,82 @@ pub struct ChunkSectionBlockStates {
...
 }
 
+#[derive(Debug, Clone)]
+pub enum LightContainer<const DIM: usize> {
+    Empty(u8),
+    Full(Box<[u8]>),
+}
+
+impl LightContainer<16> {
+    pub const fn array_size() -> usize {
```
kralverde, line 232: You can create a const value in the impl and set it to DIM

kralverde (changes requested):
Why was it necessary to make `ClientPacket` an async trait?

zly2006:
> Why was it necessary to make `ClientPacket` an async trait?

I added a lock to chunk light engine for future use, and the chunk data packet encodes a Arc chunk, so I still have to wait for locks.

https://github.com/Pumpkin-MC/Pumpkin/blob/269ae1630efa09e5b9d38ee9e4df17e2e98ed102/pumpkin-protocol/src/client/play/chunk_data.rs#L146

Do you have any better.ideas?

zly2006:
If you don't like it, I can replace it with a atomic cell

kralverde:
It encodes a reference to a chunk not an arc chunk, im not sure I like having this be an async trait, ill take a deeper look later

zly2006:
> And then why are you locking something in a chunk that already is wrapped in a lock? the light engine should just be owned by the chunk

In vanilla the light calculating is multithreaded, I think maybe we also need doing that asynchronously, so I added locks to every chunk section light container.

kralverde:
Adding locks will actually slow it down, just make it owned by the chunk

kralverde:
Can you also remove making it an async trait

pumpkin-protocol/src/client/play/chunk_data.rs
```diff
@@ -165,29 +130,60 @@ impl ClientPacket for CChunkData<'_> {
...
-            write.write_slice(&sky_buf)?;
-        }
+        {
+            let mut sky_light_buf = Vec::new();
+            let mut block_light_buf = Vec::new();
+            let mut sky_light_empty_mask = 0;
+            let mut block_light_empty_mask = 0;
+            let mut sky_light_mask = 0;
+            let mut block_light_mask = 0;
+            for light_index in 0..self.0.light_engine.sections {
```
kralverde, line 140: Can this be a map-collect?

pumpkin-world/src/chunk/format/mod.rs
```diff
@@ -149,6 +220,80 @@ pub struct ChunkSectionBlockStates {
     pub(crate) palette: Vec<PaletteBlockEntry>,
 }
 
+#[derive(Debug, Clone)]
+pub enum LightContainer {
+    Empty(u8),
+    Full(Box<[u8]>),
```
kralverde, line 226: You can probably use the abstract cube from the palette with this

pumpkin-protocol/src/client/play/chunk_data.rs
zly2006: Im going to remove the `buf`s to avoid double-copy

kralverde:
I'm not sure why you made those changes, I'm talking about using x.iter().map().collect() instead of iterating and pushing

zly2006:
> I'm not sure why you made those changes, I'm talking about using x.iter().map().collect() instead of iterating and pushing

In the chunk data packet, originally, we are using some buffers of `Vec<u8>`, in the format of a byte array and its length at the beginning. We are copying it twice, first from light data to buf, then from the buf to the final `write`. so I reduced the copies, now it will only copy once. Therefore I added two for loops to copy only once.

kralverde:
Got it, thank you :)

PR #756, opened 2025-04-22
Fix dead lock

zly2006:
## Description

Fix `world.players` dead lock on stress test, by releasing read locks as early as possible.

Most players read locks has been replaced with `world.current_players()`, which returns a copy of current players in a `Vec`. I have not yet documented to warn developers do not use players directly in read only operations.

This PR also fixed a typo in `/pumpkin`

### DIFF
--- a/pumpkin/src/net/mod.rs
+++ b/pumpkin/src/net/mod.rs
@@ -146,6 +146,7 @@ pub struct Client {
     pub closed: Arc<AtomicBool>,
     /// The client's IP address.
     pub address: Mutex<SocketAddr>,
+    pub added_to_server_listing: Mutex<bool>,
     /// The packet encoder for outgoing packets.
     network_writer: Arc<Mutex<NetworkEncoder<BufWriter<OwnedWriteHalf>>>>,
     /// The packet decoder for incoming packets.
@@ -184,6 +185,7 @@ impl Client {
             tasks: TaskTracker::new(),
             outgoing_packet_queue_send: send,
             outgoing_packet_queue_recv: Some(recv),
+            added_to_server_listing: Mutex::new(false),
         }
     }
 
--- a/pumpkin/src/net/mod.rs
+++ b/pumpkin/src/net/mod.rs
@@ -146,6 +146,7 @@ pub struct Client {
     pub closed: Arc<AtomicBool>,
     /// The client's IP address.
     pub address: Mutex<SocketAddr>,
+    /// Indicates if the client is added to the server listing.
     pub added_to_server_listing: AtomicBool,
     /// The packet encoder for outgoing packets.
     network_writer: Arc<Mutex<NetworkEncoder<BufWriter<OwnedWriteHalf>>>>,
--- a/pumpkin/src/net/mod.rs
+++ b/pumpkin/src/net/mod.rs
@@ -146,7 +146,7 @@ pub struct Client {
     pub closed: Arc<AtomicBool>,
     /// The client's IP address.
     pub address: Mutex<SocketAddr>,
-    pub added_to_server_listing: Mutex<bool>,
+    pub added_to_server_listing: AtomicBool,
     /// The packet encoder for outgoing packets.
     network_writer: Arc<Mutex<NetworkEncoder<BufWriter<OwnedWriteHalf>>>>,
     /// The packet decoder for incoming packets.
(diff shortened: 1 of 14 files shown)


pumpkin/src/net/mod.rs
```diff
@@ -146,6 +146,7 @@ pub struct Client {
     pub closed: Arc<AtomicBool>,
     /// The client's IP address.
     pub address: Mutex<SocketAddr>,
+    pub added_to_server_listing: Mutex<bool>,
```
neeleshpoli, line 149: It might be better to use an `AtomicBool` here

zly2006: Thanks! fixed

zly2006:
After my test, the Option in on_player_collision is unnecessary, we can broadcast packets in a read lock, so I removed it.

pumpkin/src/world/mod.rs
```diff
@@ -1139,11 +1138,13 @@ impl World {
     ///
     /// * `uuid`: The unique UUID of the player to add.
     /// * `player`: An `Arc<Player>` reference to the player object.
-    pub async fn add_player(&self, uuid: uuid::Uuid, player: Arc<Player>) {
-        {
-            let mut current_players = self.players.write().await;
-            current_players.insert(uuid, player.clone())
-        };
+    pub async fn add_player(&self, uuid: uuid::Uuid, player: Arc<Player>) -> Result<(), String> {
+        if let Ok(mut players) = self.players.try_write() {
```
kralverde, line 1142: This should just be a normal write lock

kralverde (changes requested):
I also think we do not need to use `current_players_clone` simply take the read lock of current players

pumpkin/src/server/connection_cache.rs
```diff
@@ -90,20 +90,34 @@ impl CachedStatus {
...
         self.status_response_json = serde_json::to_string(&status_response)
             .expect("Failed to parse status response into JSON");
     }
 
-    pub fn remove_player(&mut self) {
+    pub fn remove_player(&mut self, player: &Player) {
         let status_response = &mut self.status_response;
         if let Some(players) = &mut status_response.players {
-            players.online -= 1;
+            if player
```
kralverde, line 110: This should be a compare swap; right now this is still a race condition

kralverde (changes requested):

pumpkin/src/server/connection_cache.rs
```diff
@@ -90,20 +90,31 @@ impl CachedStatus {
...
 
     // TODO: Player samples
-    pub fn add_player(&mut self) {
+    pub fn add_player(&mut self, player: &Player) {
         let status_response = &mut self.status_response;
         if let Some(players) = &mut status_response.players {
+            player
+                .client
+                .added_to_server_listing
+                .store(false, Ordering::Relaxed);
```
kralverde: shouldnt this be true?

pumpkin/src/world/mod.rs
```diff
@@ -680,24 +685,28 @@ impl World {
                     entity.velocity.load(),
                 ))
                 .await;
+            let config = existing_player.config.read().await;
+            let mut buf = Vec::new();
+            for meta in [
+                Metadata::new(17, MetaDataType::Byte, config.skin_parts),
```
kralverde, line 691: can we remove these magic numbers

kralverde (changes requested):

pumpkin/src/world/mod.rs
zly2006: 17 is DATA_PLAYER_MODE_CUSTOMISATION
18 is DATA_PLAYER_MAIN_HAND
Currently there is no such constant numbers in pumpkin.

These numbers are copied from `send_client_information`. Do you think I should add the constants?

kralverde: yes

zly2006: Idk if it is possible to grab field names from minecraft, I just did a part of it (I think this part is enough for current development), I think we should use a automated tool to define constants.

zly2006:
Im planning to use a python script to analysis decompiled minecraft code, should I also open source the script, like, for future updates? if so, where should I open a pr in the current procedure of grabbing minecraft data?

zly2006:
Added all entity data names.

PR #757, opened 2025-04-22
Fix vanilla generated chunk lighting, load extra top and bottom light sections, and fix build limit check

jobpaardekooper:
## Description

These changes correctly load light data from worlds that were generated by the vanilla game and were moved to this server implementation. The main changes that fixes this is using the correct index for the light data masks since the client expects to receive light data for one section above and another section below the world. These extra "light only" sections also get properly loaded and sent to the client.

The block data and biome data fields have been marked optional for chunk sections since these "light only" chunks don't include those.

Additionally, there is a small fix for when the build limit is checked. Before, the wrong height value was used so the build height limit message was never shown to the player. Could have also been a separate fix but since I was moving around and creating some constants I included it here. 

## Testing

Tested by loading a vanilla generated world with a chunk that has the extra top and bottom light only sections.

### DIFF
--- a/pumpkin/src/net/packet/play.rs
+++ b/pumpkin/src/net/packet/play.rs
@@ -36,7 +36,8 @@ use pumpkin_inventory::player::{
 use pumpkin_macros::send_cancellable;
 use pumpkin_protocol::client::play::{
     CBlockUpdate, COpenSignEditor, CPlayerInfoUpdate, CPlayerPosition, CSetContainerSlot,
-    CSetHeldItem, CSystemChatMessage, EquipmentSlot, InitChat, PlayerAction,
+    CSetHeldItem, CSystemChatMessage, EquipmentSlot, InitChat, PlayerAction, WORLD_HIGHEST_Y,
+    WORLD_LOWEST_Y,
 };
 use pumpkin_protocol::codec::item_stack_seralizer::ItemStackSerializer;
 use pumpkin_protocol::codec::var_int::VarInt;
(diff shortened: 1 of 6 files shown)


kralverde:
How does this differ from https://github.com/Pumpkin-MC/Pumpkin/pull/754?

jobpaardekooper:
@kralverde It's basically a different approach to solve the same problem. We briefly talked on discord and I think #754 is the better approach. This PR here does include a small fix though w.r.t. the build height limit message. But it would probably be better to fix that in a different PR.

pumpkin/src/net/packet/play.rs
```diff
@@ -1650,16 +1648,16 @@ impl Player {
...
         }
 
         // Check the world's max build height
-        if location.0.y + face.to_offset().y >= i32::from(Self::WORLD_MAX_Y) {
+        if location.0.y + face.to_offset().y >= WORLD_HIGHEST_Y {
             self.send_system_message_raw(
                 &TextComponent::translate(
                     "build.tooHigh",
-                    vec![TextComponent::text((Self::WORLD_MAX_Y - 1).to_string())],
+                    vec![TextComponent::text((WORLD_HIGHEST_Y - 1).to_string())],
```
jobpaardekooper, line 1660: This is the build height fix

Snowiiii:
So @jobpaardekooper, https://github.com/Pumpkin-MC/Pumpkin/pull/754 was merged. Is this still needed ?

jobpaardekooper:
No longer needed

PR #759, opened 2025-04-23
Finalise the /effect command

hthug06:
## Description
Finalise the /effect command
Add the duration, amplifier and hide particule
Add everything about the /effect clear

## Testing
Give yourself potions effects, clear them

## Other
While doing tests, i see that potion particles don't work for some reason. Also some visual effect are not working like the glowing effect.


kralverde:
Thanks, I think the particle issue can be addressed later

Issue #752, opened 2025-04-19
Deduct life from jumping into water at high altitude


nachonekoawa:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows11

### Server Software Version/Commit

0.1.0-dev+1.21.5 (fbf18cc1)

### What happened?

Watch this video.

https://github.com/user-attachments/assets/fca5819a-0f97-4500-995b-21c32ed88d9a

### To Reproduce

Jumping into water from a high altitude instantly triggers this lifesteal bug

### Expected behavior

There should be no deduction of life points

azomDev:
I can't test this week to make sure the PR fixed this. Has this been fix or do we need to keep the issue opened?

azomDev:
This issue has been fixed by #761 so it can be closed

PR #761, opened 2025-04-24
Fix fall damage when in water

TheDarkSword:
## Description
This pull request fix the issue #752.
As the title explain the fall damage is ignored when in water

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -186,11 +186,10 @@ impl LivingEntity {
     pub async fn is_in_water(&self) -> bool {
         let world = self.entity.world.read().await;
         let block_pos = self.entity.block_pos.load();
-        if let Ok(block) = world.get_block(&block_pos).await {
-            block.eq(&Block::WATER)
-        } else {
-            false
-        }
+        world
+            .get_block(&block_pos)
+            .await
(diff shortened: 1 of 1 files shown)


4lve:
I'm pretty sure this isn't equivalent to vanilla, also the server would crash if the player went above or below world limit

pumpkin/src/entity/living.rs
```diff
@@ -182,6 +183,20 @@ impl LivingEntity {
...
+    pub async fn is_in_water(&self) -> bool {
+        let world = self.entity.world.read().await;
+        let pos = self.entity.pos.load();
+        // Use the floor function to get the block position correctly (e.g. 1.5 -> 1, -1.5 -> -2)
+        let block_pos = BlockPos::new(
+            pos.x.floor() as i32,
+            pos.y.floor() as i32,
+            pos.z.floor() as i32,
+        );
+        let block = world.get_block(&block_pos).await.unwrap();
```
kralverde, line 196: This will crash if the player goes outside of world bounds (too high or below bedrock)

kralverde (changes requested):

kralverde:
Thank you! If any vanilla-matching issues occur in the future, we can just fix them then

PR #763, opened 2025-04-24
Fixing repeater power on placement

### LINKED ISSUE #762: Redstone repeater not turning on when placed

### Operating System

Fedora 42

0.1.0-dev+1.21.5 (abb4e541)

### What happened?

When placing a repeater in a place where it should be powered, it does not power on until it gets a block update (for example when whatever block is placed on the side of the repeater).

### To Reproduce

Explained above

### Expected behavior

It should power on directly.

azomDev:
## Description
(edited after I realized a part of what I was doing was probably out of the scope of this PR)

This pull request introduces the following change:

1. **Power Check for Repeater Placement**: Added code to check if a repeater should be powered on placement.

This fixes #762. For example if you try having a redstone torch and then place a repeater, it will be placed in a powered state.

## Testing
I tested manually, not sure if I need to do more

### DIFF
--- a/pumpkin/src/block/blocks/redstone/redstone_wire.rs
+++ b/pumpkin/src/block/blocks/redstone/redstone_wire.rs
@@ -103,20 +103,6 @@ impl PumpkinBlock for RedstoneWireBlock {
             }
         }
 
-        // Schedules a block tick for the connected block if the side is connected and wire is powered.
-        if new_side.is_connected() && wire.power != Integer0To15::L0 {
-            let block_pos = block_pos.offset(direction.to_offset());
-            let block = world.get_block(&block_pos).await.unwrap();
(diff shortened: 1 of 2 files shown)


azomDev:
image

Found a case where my code makes the repeater flickers fast (the unpowered repeater in the image) when placing that repeater in that setup

4lve:
A repeater can be powered by more ways than just redstone dust, I suggest checking decompiled source

azomDev:
@4lve yeah that's why there's some changes in `repeater.rs`. There, on placement, I'm checking whether it is powered at all, regardless of the source. (taking inspiration from the `on_neighbor_update` function where it updates the power)

kralverde:
Thank you!

PR #767, opened 2025-04-24
fixed rotate_ccw

azomDev:
## Description
The rotate_ccw had a small error. This fixes an issue with repeater locking. I don't know if I should, but I made this in a separate PR since it's probably affecting more than just that.

## Testing
I don't think this needs testing, but the game runs and repeater locking works in normal cases now


Snowiiii:
Nice catch. Thank you @azomDev

PR #774, opened 2025-04-26
Fixed hunger depletion and hunger health regen

azomDev:
## Description
The previous implementation was not accurate to vanilla. Some data like difficulty was not being used and some logic branches were missing. 

## Testing
It's the same as in vanilla, but I manually tested and it feels more like vanilla when testing with normal and peaceful difficulties.


azomDev:
Actually merge #775 before so I can use it

PR #775, opened 2025-04-26
Custom difficulty serialization and can_place_at with block face

Guibi1:
## Description

This PR is very small and adds ZERO functionality.
However, it made the `LevelData::difficulty` field use the actual enum instead of a u8.
It also adds the `face` prop to the `PumpkinBlock::can_place_at`, which will be useful for weird blocks like the bell (see [this message in the discord](https://discord.com/channels/1268592337445978193/1268611318617866261/1365485927010406471)).

## Testing


kralverde (changes requested):
The functions should take ownership of the block direction enum, not a reference to it. Copying enums is very cheap

Guibi1:
After review, i've found that `BasicConfiguration` has a `default_difficulty`, which was using the strignified version of the Difficulty enum (aka "Normal")
With my PR, that changes to the number representation. Is this an acceptable change?

kralverde:
No, add a conversion from config to enum

pumpkin/src/world/mod.rs
```diff
@@ -1595,16 +1595,18 @@ impl World {
     pub async fn update_neighbors(
         self: &Arc<Self>,
         block_pos: &BlockPos,
-        except: Option<&BlockDirection>,
+        except: Option<BlockDirection>,
     ) {
         let source_block = self.get_block(block_pos).await.unwrap();
         for direction in BlockDirection::update_order() {
-            if Some(&direction) == except {
```
kralverde, line 1602: Why was this changed?

kralverde (changes requested):
Can you rename c like enum to serialize enum as integer or something?

pumpkin/src/world/mod.rs
Guibi1: Stumbled upon it while reading and made it more readable. I can revert if needed.

PR #777, opened 2025-04-27
Fix last_pos of new LivingEntity

akashbeh:
## Description
Fix the initial last_pos of a LivingEntity.
In LivingEntity::new(), set last_pos to entity.pos instead of 0,0,0.
This is important because otherwise, when first sending position of a new mob, assuming you use the difference _pos - last_pos_ to send position, the entity will appear to jump somewhere else.

## Testing
Passed cargo check, cargo fmt --check, and cargo test.


kralverde:
Thank you :)

PR #780, opened 2025-04-27
Fix effect command crashing the server because of invalid bounds

Dewyer:
## Description

This commit fixes the server crash on invalid arguments for the effect command, hower it does not provide translation support for the error message and also doesn't change a lot of other things, the primary goal here is really just to fix this one crash.

Previously you could just crash the Pumpkin server by providing an invalid value for a bounded argument when using the `effect` command. Error handling for `give`, and `particle` was also inconsistent.

In the handler for `particle` the server just ignored the invalid argument and used defaults instead which is not the same behaviour that the java server exhibits.

Give just didn't provide the proper error message, also note that the current upper bound for the command is `i32::MAX` which is not the same as in the java server, this PR doesn't address that either however.

## Testing

Tested the commands manually after running the server, unit tests were not implemented as the command executor doesn't have any automatic test coverage at the moment, and setting that up would be out of scope for this PR.

### DIFF
--- a/pumpkin/src/command/commands/effect.rs
+++ b/pumpkin/src/command/commands/effect.rs
@@ -62,8 +62,7 @@ impl CommandExecutor for GiveExecutor {
                     .name("seconds")
                     .min(1)
                     .max(1_000_000)
-                    .find_arg_default_name(args)?
-                    .unwrap()
+                    .find_arg_default_name(args)??
                     * 20
             }
             Time::Infinite => -1,
@@ -75,8 +74,7 @@ impl CommandExecutor for GiveExecutor {
                 .name("amplifier")
                 .min(0)
                 .max(255)
-                .find_arg_default_name(args)?
-                .unwrap(),
+                .find_arg_default_name(args)??,
         };
 
         let mut hide_particles = self.2;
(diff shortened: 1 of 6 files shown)


Snowiiii:
One thing i would have done for a while now is to use translations for arguments, Can you may use translations here?

Dewyer:
sure thing, ill look into how to do it

pumpkin/src/command/commands/effect.rs
```diff
@@ -75,8 +74,7 @@ impl CommandExecutor for GiveExecutor {
                 .name("amplifier")
                 .min(0)
                 .max(255)
-                .find_arg_default_name(args)?
-                .unwrap(),
+                .find_arg_default_name(args)??,
```
kralverde, line 77: Can you use `flatten` here?

kralverde (changes requested):

pumpkin/src/command/commands/effect.rs
Dewyer: this is actually a ? .map_err(into::into)? situation that ? handles

kralverde:
Thank you :)

PR #782, opened 2025-04-27
Read commands from stdin

zly2006:
## Description

This PR allows reading commands from stdin. Stdin is very important since many server hosting softwares redirect stdin and executing console commands that way.

Without this PR, any commands in non-tty mode will be simply ignored.

## Needs improvement

When the user press Ctrl+C, the signal is correctly captured, and the server saves and stops. However, since `stdin().read_line` is blocking, it waits for a newline if there is not a newline. Users may have to press an extra enter to fully close the server. Server hosting softs that send a `stop` command via stdin will not be effected.

## Testing

### DIFF
--- a/pumpkin/src/lib.rs
+++ b/pumpkin/src/lib.rs
@@ -11,7 +11,7 @@ use pumpkin_config::{BASIC_CONFIG, advanced_config};
 use pumpkin_macros::send_cancellable;
 use pumpkin_util::text::TextComponent;
 use rustyline_async::{Readline, ReadlineEvent};
-use std::io::stdin;
+use std::io::{IsTerminal, stdin};
 use std::str::FromStr;
 use std::sync::atomic::AtomicBool;
 use std::{
@@ -135,7 +135,7 @@ pub static LOGGER_IMPL: LazyLock<Option<(ReadlineLogWrapper, LevelFilter)>> = La
             .and_then(Result::ok)
             .unwrap_or(LevelFilter::Info);
 
-        if advanced_config().commands.use_tty {
+        if advanced_config().commands.use_tty && stdin().is_terminal() {
             match Readline::new("$ ".to_owned()) {
                 Ok((rl, stdout)) => {
                     let logger = simplelog::WriteLogger::new(level, config.build(), stdout);
@@ -392,6 +392,10 @@ fn setup_stdin_console(server: Arc<Server>, handle: Handle) {
                 break;
             };
             let server_clone = server.clone();
+            if line.len() == 0 || line.as_bytes()[line.len() - 1] != b'\n' {
+                log::warn!("Console command was not terminated with a newline");
+                continue;
+            }
             handle.spawn(async move {
                 let command = &line[..line.len() - 1]; // Remove the newline
                 send_cancellable! {{
--- a/pumpkin/src/lib.rs
+++ b/pumpkin/src/lib.rs
@@ -150,11 +150,6 @@ pub static LOGGER_IMPL: LazyLock<Option<(ReadlineLogWrapper, LevelFilter)>> = La
                 }
             }
         } else {
-            if advanced_config().commands.use_tty && !stdin().is_terminal() {
-                log::warn!(
-                    "The input is not a TTY; falling back to simple logger and ignoring `use_tty` setting"
-                );
-            }
             let logger = simplelog::SimpleLogger::new(level, config.build());
             Some((ReadlineLogWrapper::new(logger, None), level))
         }
@@ -213,6 +208,11 @@ impl PumpkinServer {
                 if let Some(rl) = wrapper.take_readline() {
                     setup_console(rl, server.clone());
                 } else {
+                    if advanced_config().commands.use_tty {
+                        log::warn!(
+                            "The input is not a TTY; falling back to simple logger and ignoring `use_tty` setting"
+                        );
+                    }
                     setup_stdin_console(server.clone()).await;
                 }
             }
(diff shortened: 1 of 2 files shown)


zly2006:
still ned to fix a panic bug

pumpkin/src/lib.rs
```diff
@@ -133,7 +135,7 @@ pub static LOGGER_IMPL: LazyLock<Option<(ReadlineLogWrapper, LevelFilter)>> = La
             .and_then(Result::ok)
             .unwrap_or(LevelFilter::Info);
 
-        if advanced_config().commands.use_console {
+        if advanced_config().commands.use_tty && stdin().is_terminal() {
```
kralverde, line 138: Why check for `is_terminal` here?

pumpkin/src/lib.rs
```diff
@@ -378,6 +384,40 @@ impl PumpkinServer {
...
+                }
+            } else {
+                break;
+            };
+            let server_clone = server.clone();
+            if line.is_empty() || line.as_bytes()[line.len() - 1] != b'\n' {
+                log::warn!("Console command was not terminated with a newline");
+                continue;
+            }
+            handle.spawn(async move {
```
kralverde, line 404: Create a dedicated tokio task for this and use a mpsc channel to communicate between them

kralverde (changes requested):

pumpkin/src/lib.rs
zly2006: In the case of `echo stop | ./pumpkin`, the program attempts to read as a terminal arbitrarily and panic

kralverde: Can you at least add a warning since this changes expected behavior based on the config

pumpkin/src/lib.rs
```diff
@@ -378,6 +383,46 @@ impl PumpkinServer {
     }
 }
 
+async fn setup_stdin_console(server: Arc<Server>) {
+    let (tx, mut rx) = tokio::sync::mpsc::channel(10);
```
kralverde, line 387: You should only need to handle one at a time

pumpkin/src/lib.rs
```diff
@@ -378,6 +383,46 @@ impl PumpkinServer {
...
+            let mut line = String::new();
+            if let Ok(size) = stdin().read_line(&mut line) {
+                // if no bytes were read, we may have hit EOF
+                if size == 0 {
+                    break;
+                }
+            } else {
+                break;
+            };
+            if line.is_empty() || line.as_bytes()[line.len() - 1] != b'\n' {
```
kralverde, line 400: This behavior should be handled by readline

pumpkin/src/lib.rs
```diff
@@ -378,6 +383,46 @@ impl PumpkinServer {
...
+                log::warn!("Console command was not terminated with a newline");
+                continue;
+            }
+            rt.block_on(tx.send(line[..line.len() - 1].to_string()))
```
kralverde, line 404: You should use trim here

kralverde (changes requested):

pumpkin/src/lib.rs
zly2006: readline did not handle it in the case `echo list | ./pumpkin`, I suppose the server should keep running if stdin closed.

kralverde: I see what you mean

zly2006: and if you mean the '\n' check, the returned value will not always ends with it (maybe EOF), and it will panic when trying to remove the trailing '\n'. After replacing it with `trim`, I think I can remove this check. Im gonna do some tests on it.

pumpkin/src/lib.rs
```diff
@@ -149,6 +150,11 @@ pub static LOGGER_IMPL: LazyLock<Option<(ReadlineLogWrapper, LevelFilter)>> = La
                 }
             }
         } else {
+            if advanced_config().commands.use_tty && !stdin().is_terminal() {
```
kralverde, line 153: Just check if we expected to use tty here

kralverde (changes requested):

pumpkin/src/lib.rs
zly2006: sorry wdym?

kralverde: Don't check if it's not a terminal again, if we are in this branch and we expect to use the tty, log the warning

zly2006: Removing the `stdin().is_terminal()` will cause a panic in non-tty mode? is that OK?

kralverde: I'm talking about specifically for the log message

zly2006: Ah ok

pumpkin/src/lib.rs
```diff
@@ -133,7 +134,21 @@ pub static LOGGER_IMPL: LazyLock<Option<(ReadlineLogWrapper, LevelFilter)>> = La
             .and_then(Result::ok)
             .unwrap_or(LevelFilter::Info);
 
-        if advanced_config().commands.use_console {
+        if advanced_config().commands.use_tty {
+            if !stdin().is_terminal() {
```
kralverde, line 138: You can just defer this to when we actually have a logger

kralverde (changes requested):

kralverde (changes requested):

pumpkin/src/lib.rs
```diff
@@ -202,9 +203,18 @@ impl PumpkinServer {
...
 
-        if let Some((wrapper, _)) = &*LOGGER_IMPL {
-            if let Some(rl) = wrapper.take_readline() {
-                setup_console(rl, server.clone());
+        if advanced_config().commands.use_console {
+            if let Some((wrapper, _)) = &*LOGGER_IMPL {
+                if let Some(rl) = wrapper.take_readline() {
+                    setup_console(rl, server.clone());
+                } else {
+                    if advanced_config().commands.use_tty {
```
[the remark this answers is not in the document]
zly2006: logger already initialized here

kralverde:
Thank you :)

PR #784, opened 2025-04-29
Fix NaN rotation in player move packets

### LINKED ISSUE #781: Proper sanitize client position

No

Client position should be properly sanitized, f64.clamp isn't enough

Currently server can be easily crashed via massively sending client position at unloaded chunks, 
Or worse, since clamp doesn't eliminate NaNs.

A PoC: https://github.com/CCBlueX/LiquidBounce/blob/9eae73066a01409ff294da077fb721878e673464/src/main/kotlin/net/ccbluex/liquidbounce/features/module/modules/exploit/servercrasher/exploits/InvalidPositionExploit.kt

Related:
https://github.com/Pumpkin-MC/Pumpkin/blob/c07882369c14e3394ba230c356cd72a216270e6a/pumpkin/src/net/packet/play.rs#L208

This allows crashing the server within a packet.

zly2006:
## Description

Fix #781 

`is_infinite` does not check for NaN, so use `!is_finite` instead.

The original report is wrong: We have checked for NaN position before clamping, but there is still a NaN rotation.

## Testing

Have not tested yet.

### DIFF
--- a/pumpkin-protocol/src/server/play/player_position_rotation.rs
+++ b/pumpkin-protocol/src/server/play/player_position_rotation.rs
@@ -8,5 +8,6 @@ pub struct SPlayerPositionRotation {
     pub position: Vector3<f64>,
     pub yaw: f32,
     pub pitch: f32,
-    pub ground: bool,
+    /// bit 0: on_ground, bit 1: in_wall
+    pub ground: u8,
 }
--- a/pumpkin-protocol/src/server/play/player_position_rotation.rs
+++ b/pumpkin-protocol/src/server/play/player_position_rotation.rs
@@ -2,12 +2,15 @@ use pumpkin_data::packet::serverbound::PLAY_MOVE_PLAYER_POS_ROT;
 use pumpkin_macros::packet;
 use pumpkin_util::math::vector3::Vector3;
 
+pub static BIT_ON_GROUND: u8 = 0x01;
+pub static BIT_IN_WALL: u8 = 0x02;
+
 #[derive(serde::Deserialize)]
 #[packet(PLAY_MOVE_PLAYER_POS_ROT)]
 pub struct SPlayerPositionRotation {
     pub position: Vector3<f64>,
     pub yaw: f32,
     pub pitch: f32,
-    /// bit 0: on_ground, bit 1: in_wall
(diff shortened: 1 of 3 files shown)


pumpkin/src/net/packet/play.rs
```diff
@@ -352,8 +352,8 @@ impl Player {
         if position.x.is_nan()
```
kralverde, line 352: I think all of these should be checked to be not finite

pumpkin/src/net/packet/play.rs
```diff
@@ -384,14 +384,14 @@ impl Player {
 
                 let height_difference = pos.y - last_pos.y;
                 if entity.on_ground.load(std::sync::atomic::Ordering::Relaxed)
-                    && !packet.ground
+                    && (packet.ground & 1) != 0
```
kralverde, line 387: Create constant values to compare; don't use magic numbers

kralverde (changes requested):

pumpkin/src/net/packet/play.rs
```diff
@@ -286,15 +286,15 @@ impl Player {
                 self.living_entity.set_pos(pos);
 
                 let height_difference = pos.y - last_pos.y;
-                if entity.on_ground.load(Ordering::Relaxed) && !packet.ground && height_difference > 0.0 {
+                if entity.on_ground.load(Ordering::Relaxed) && !packet.collision & BIT_ON_GROUND != 0 && height_difference > 0.0 {
```
kralverde, line 289: Double negation

pumpkin-protocol/src/server/play/player_position_rotation.rs
```diff
@@ -2,11 +2,15 @@ use pumpkin_data::packet::serverbound::PLAY_MOVE_PLAYER_POS_ROT;
 use pumpkin_macros::packet;
 use pumpkin_util::math::vector3::Vector3;
 
+pub static BIT_ON_GROUND: u8 = 0x01;
```
kralverde, line 5: call these flags

kralverde (changes requested):

kralverde:
Thank you :)

PR #802, opened 2025-05-09
Skip hunger in Creative/Spectator

5jd:
## Description
A small fix that disables hunger mechanics and starvation damage when the player is in Creative or Spectator mode.
This might not be vanilla accurate.

## Testing


azomDev:
Is there a bug currently that makes creative or spectator that makes you take damage from that? When I fixed hunger, I checked the vanilla function and there was no check for this. At least not in that function

Issue #788, opened 2025-05-01
Effects are cleared on rejoin


Ancient77:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Debian 13

### Server Software Version/Commit

e8a50aa

### What happened?

Effects are cleared (even if time is still left) when you leave and rejoin.

### To Reproduce

- join
- /effect give @a minecraft:absorption 10
- leave
- rejoin
- see that the effect is gone

### Expected behavior

The vanilla behavior is that you keep the effect

PR #807, opened 2025-05-11
Save/Load entity effects and send to player on join

TheDarkSword:
## Description

This pull request enables saving and loading of entity effects and delivers them to players when they join.
Fixes issue #788 

## Testing


pumpkin/src/entity/living.rs
```diff
@@ -354,13 +354,94 @@ impl NBTStorage for LivingEntity {
...
+                            ambient,
+                            show_particles,
+                            show_icon,
+                            blend: true, // Copied from effect give command
+                        };
+                        active_effects.insert(effect.r#type, effect);
+                    }
+                }
+            }
+        }
```
Snowiiii, line 444: I recommend to add NBTStorage to the effect struct and put the nbt reading and writing code in there instead cluttering living entity

TheDarkSword: Done

pumpkin/src/entity/living.rs
```diff
@@ -398,45 +383,9 @@ impl NBTStorage for LivingEntity {
...
-                            duration,
-                            amplifier: amplifier as u8,
-                            ambient,
-                            show_particles,
-                            show_icon,
-                            blend: true, // Copied from effect give command
-                        };
+                        let mut effect = Effect::default();
+                        effect.read_nbt(&mut effect_nbt.clone()).await;
+                        effect.blend = true; // TODO: change, is taken from effect give command
```
Snowiiii, line 388: I think this does not make sense ?, why you are using the default here ?

pumpkin/src/entity/effect/mod.rs
```diff
@@ -10,3 +14,57 @@ pub struct Effect {
...
+        Self {
+            r#type: EffectType::Absorption,
+            duration: 0,
+            amplifier: 0,
+            ambient: false,
+            show_particles: true,
+            show_icon: true,
+            blend: false,
+        }
+    }
```
Snowiiii, line 29: I also don't like the default here it does not really make sense

pumpkin/src/entity/living.rs
TheDarkSword: Ok, the other way is to add to NBTStorage a method to read data from NBT and return T (where T (generico is the struct, in our example is Effect). Or maybe there is a better way?

Snowiiii:
Looks good now. Thank you @TheDarkSword :+1:

PR #811, opened 2025-05-12
Waterlogging and bucket improvements

teknostom:
## Description

Added waterlogging and inventory management for buckets.

## Testing


Snowiiii:
Hey @teknostom, please do heavy testing until opening a PR as non draft, I'm really tired of people making PR's which do not work when just simply tested

https://github.com/user-attachments/assets/3ab21fa5-56df-44d6-b35f-2d2de08ad85f

teknostom:
Sorry, worked perfectly for me, will look into it

teknostom:
@Snowiiii Added #815 to track that issue. I have retested the waterlogging and seems to work for all blocks. The only issue I know of is that bounding boxes aren't implemented, so waterloggable blocks act weird sometimes. Do you want me to do that in this PR? The building blocks should all be there, I just need to add the check.

If anyone testing this has the same issue, just swap to a different item bar location, and back. Then the bucket should work.

teknostom:
Ok, found an issue. The lava bucket also waterlogged, will fix.

Snowiiii:
Seems like you have commits here from master

teknostom:
Damn, must have done something wrong when merging from master. Will fix later. Should be ready next week though.

teknostom:
This should be ready now, remember to avoid the inv bug. It also currently uses collision boxes instead of bounding boxes, so fences and walls will be weird until that is available. It should work fine for now though.

pumpkin/src/item/items/bucket.rs
```diff
@@ -67,28 +95,110 @@ impl PumpkinItem for EmptyBucketItem {
...
+                    BlockFlags::NOTIFY_NEIGHBORS,
+                )
+                .await;
+        } else {
+            let block = world
+                .get_block(&block_pos.offset(direction.to_offset()))
+                .await;
+            let state_id = world
+                .get_block_state_id(&block_pos.offset(direction.to_offset()))
+                .await;
```
Snowiiii, line 152: you should use `world.get_block_and_block_state`

teknostom: Fixed

Snowiiii:
Looks good now. Thank you @teknostom :D

Issue #186, opened 2024-10-25
Teleport exploit and Infinite placing glitch with trapdoors


Z-fentom:
- [ ] I've searched existing issues and couldn't find a duplicate.

**Describe the bug**
Bug 1: walking and placing a block on the side of grass causes a teleport maximum of 3 blocks forward direction depends blocks behind the grass.
Bug 2: Placing a trapdoor of any kind and right clicking it will cause it to place blocks randomly in next to it self and will cause the player to place trapdoors with a empty hand

**To Reproduce**
Bug 1: place a grass block, start walking forwards and place a block on the side.
Bug 2: place a trapdoor and right click it then swap to a empty hand

**Expected behavior**
Bug 1: grass getting replaced by the chosen block instead of getting teleported also happens probably due to blocks not detecting player collision
Bug 2: trapdoor being placed and opened upon right click

**Environment (please complete the following information):**

- Operating System: Arch Linux BTW
- Server Software Version/Commit: [e.g., 1.0.0/39b4cb3]

**Additional context**

TonimatasDEV:
@Snowiiii That should be fixed by #812

RoosterBooster007:
This seems to have been fixed a while ago. I'm closing this issue. Great work, y'all!

PR #812, opened 2025-05-13
Trapdoors

hthug06:
## Description
Add trapdoor

## Testing
Place a trapdoor, play with it (right click, redstone)

## Other
Waterlogging is a little bit glitched like you can place trapdoor in the water but can't place water on a trapdoor

### DIFF
--- a/pumpkin/src/block/blocks/trapdoor.rs
+++ b/pumpkin/src/block/blocks/trapdoor.rs
@@ -1,0 +1,111 @@
+use crate::block::BlockIsReplacing;
+use crate::block::blocks::redstone::block_receives_redstone_power;
+use crate::block::pumpkin_block::{BlockMetadata, PumpkinBlock};
+use crate::block::registry::BlockActionResult;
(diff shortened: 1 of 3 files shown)


pumpkin/src/block/blocks/trapdoor.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut trapdoor_props = TrapDoorProperties::default(block);
+        trapdoor_props.waterlogged = replacing.water_source();
+
+        let powered = block_receives_redstone_power(world, block_pos).await;
+        let direction = player
+            .living_entity
+            .entity
+            .get_horizontal_facing()
+            .rotate_clockwise()
+            .rotate_clockwise();
```
Snowiiii, line 126: can't you just use opposite?

hthug06:
Also would fix #186

Snowiiii:
Everything works and the good looks good, Awesome. Thank you @hthug06 :smile:

PR #813, opened 2025-05-15
Added TNT (check)

TheDarkSword:
## Description

This pull request is for checking if what I'm doing is ok or not (entity-data branch) @kralverde

### DIFF
--- a/pumpkin/src/entity/mod.rs
+++ b/pumpkin/src/entity/mod.rs
@@ -39,6 +39,7 @@ use std::{
         },
     },
 };
+use enum_dispatch::enum_dispatch;
 use tokio::sync::RwLock;
 use r#type::entity_base_from_type;
 use uuid::Uuid;
@@ -61,15 +62,16 @@ mod combat;
 
 pub type EntityId = i32;
 
+#[enum_dispatch]
 #[async_trait]
 pub trait WorldEntityExt: EntityBase {
     /// Gets Called every tick
-    async fn tick(&mut self, world: &World);
+    async fn tick(&mut self, server: &Server, world: &World);
 
     /// Returns if damage was successful or not
     async fn damage(&self, amount: f32, damage_type: DamageType, world: &World) -> bool;
 
-    async fn init_data_tracker(&self, world: &World);
+    async fn init_data_tracker(&mut self, world: &World);
 
     /// Called when a player collides with a entity
     async fn on_player_collision(&self, player: Arc<Player>, world: &World);
@@ -92,7 +94,7 @@ pub trait WorldEntityExt: EntityBase {
             .await;
     }
 
-    async fn global_tick(&mut self, world: &World) {
+    async fn global_tick(&mut self, server: &Server, world: &World) {
         // Movement
         let velocity = self.get_velocity();
         let pos = self.pos();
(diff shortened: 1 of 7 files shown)


Snowiiii:
naming it TNTEntity or similar would be much nicer than just Tnt, i also wonder what check do you mean here ?, you should provide a more clear PR title and description

TheDarkSword:
> naming it TNTEntity or similar would be much nicer than just Tnt, i also wonder what check do you mean here ?, you should provide a more clear PR title and description

I opened this PR because kralverde will be away for a while and had asked on Discord if someone could help out. I volunteered, and he briefly explained what needed to be done. I opened this PR mainly to confirm if I'm going in the right direction, so I don’t end up doing unnecessary work.

The names like Tnt were based on what kralverde used, as he had already renamed quite a few things (like NonPlayerCommon, LivingCommon, etc.).

Thanks again, and sorry once more — I’ll make sure to be more careful with naming and PR descriptions next time.

pumpkin/src/entity/mod.rs
```diff
@@ -61,15 +62,16 @@ mod combat;
...
 pub trait WorldEntityExt: EntityBase {
     /// Gets Called every tick
-    async fn tick(&mut self, world: &World);
+    async fn tick(&mut self, server: &Server, world: &World);
 
     /// Returns if damage was successful or not
     async fn damage(&self, amount: f32, damage_type: DamageType, world: &World) -> bool;
 
-    async fn init_data_tracker(&self, world: &World);
+    async fn init_data_tracker(&mut self, world: &World);
```
Snowiiii, line 74: Why you made this &mut ?

TheDarkSword: Because in EntityBase trait there is the function "fn set_velocity(&mut self, velocity: Vector3<f64>);" that I call in the init_data_tracker for TNT (as the actual implementation in the master branch does)

Snowiiii: The trait `EntityBase` does not have a velocity function, Entity does and it does of course not take in &mut self, Otherwise it would make no sense to use Mutexes and Atomics everywhere 
https://github.com/Pumpkin-MC/Pumpkin/blob/60da1edb3c61dc9c96c669575e42313e5244dacd/pumpkin/src/entity/mod.rs#L209C18-L209C30

TheDarkSword: Kralverde said that in this implementation are not needed Mutexes and Atomics inside Entities because the lock is done at entity level (when getting it) then are not longer needed.
Anyway the EntityBase trait is made by kralverde, I just used the already implemented method

TheDarkSword: In my opinion this is the best way to manage entities, for example, Minestom does the same, the lock is at entity level, then the fields are not atomic, because is not necessary

Snowiiii: Minestom is written in Java, Without locking inside Entity you can't access it multiple times at the same time which is really useful to boost performance and have multiple operations at once

TheDarkSword: The logic of lock at entity level is done by kralverde

TheDarkSword: This is his exact message in discord:
"It is locked in level. Should be queried with get entity mut/iter entities mut
Should not use internal locks/mutexs anymore"

PR #814, opened 2025-05-16
Fix called `Result::unwrap()` on an `Err` value: BlockOutOfWorldBounds

vyPal:
## Description
Fixes the `called 'Result::unwrap()' on an 'Err' value: BlockOutOfWorldBounds` error by replacing the `.unwrap()` call with a if statement:
```rs
if let Ok((block, state)) = world.get_block_and_block_state(&pos).await {
    world
        .block_registry
        .on_entity_collision(block, &world, entity, pos, state)
        .await;
}
```

## Testing


Snowiiii:
Thanks

PR #819, opened 2025-05-17
add console summon command support

qwixck:
## Description
added support of summon command for console
## Testing


pumpkin/src/command/commands/summon.rs
```diff
@@ -28,23 +27,46 @@ impl CommandExecutor for Executor {
...
+                sender
+                    .send_message(TextComponent::translate(
+                        "commands.summon.success",
+                        [TextComponent::text(format!("{entity:?}"))],
+                    ))
+                    .await;
+            }
+            CommandSender::Rcon(_) => {
+                // TODO: make this work in rcon
+            }
```
qwixck, line 69: maybe remove this? does the commands sent from rcon are executing as `CommandSender::Console`? I'm just very new with this project and I know not that much about rcon. If no, any suggestions about naming a var in `CommandSender::Rcon`? It's a vec of a strings and I have no idea what that vec is about.

Snowiiii, line 69: Instead having RCON not supported, you could just use an OR. Like this
`CommandSender::Console | CommandSender::Rcon(_)`

qwixck: > Instead having RCON not supported, you could just use an OR. Like this `CommandSender::Console | CommandSender::Rcon(_)`

done

Snowiiii:
Looks good. Thank you @qwixck :D

PR #824, opened 2025-05-21
Entity physics

akashbeh:
## Description
Add entity physics: gravity, friction, block collisions on movement. This is in entity/mod.rs.
Also fixed knockback.

Note, it's based off old code so 
1. I used some old methods with GetBlockError in world/mod.rs, they're all marked as "deprecated"
2. Physics is incomplete. no_clip (being in a block) and being in a fluid will have to be redone with the new Entity::check_block_collisions. E.g. if in water: entity.in_water.store(true), change velocity. The old code is currently commented out. 

Another issue: liquid flow pushing is not done.

## Testing
cargo run
Entities land on the ground correctly. 
Though sometimes they go into walls when knocked back.

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -95,6 +95,22 @@ impl LivingEntity {
         self.last_pos.store(self.entity.pos.load());
         self.entity.set_pos(position);
     }
+    
+    pub async fn send_pos(&self) {
+        let old_pos = self.last_pos.load();
+        let pos = self.entity.pos.load();
+        self.last_pos.store(pos);
+        
+        self.entity.send_pos(pos, old_pos).await;
+    }
+    
+    pub async fn send_pos_rot(&self) {
+        let old_pos = self.last_pos.load();
+        let pos = self.entity.pos.load();
+        self.last_pos.store(pos);
+        
+        self.entity.send_pos_rot(pos, old_pos).await;
+    }
 
     pub async fn heal(&self, additional_health: f32) {
         assert!(additional_health > 0.0);
@@ -211,7 +227,7 @@ impl LivingEntity {
     ) {
         if ground {
             let fall_distance = self.fall_distance.swap(0.0);
-            if fall_distance <= 0.0 || dont_damage || self.is_in_water().await {
+            if fall_distance <= 0.0 || dont_damage || self.entity.in_water.load(Relaxed) {
                 return;
             }
 
@@ -259,6 +275,7 @@ impl LivingEntity {
             .await;
     }
 
+/*
     async fn tick_move(&self, entity: &dyn EntityBase, server: &Server) {
         let velo = self.entity.velocity.load();
         let pos = self.entity.pos.load();
@@ -271,6 +288,7 @@ impl LivingEntity {
             .store(velo.multiply(multiplier, 1.0, multiplier));
         Entity::check_block_collision(entity, server).await;
     }
+*/
 
     async fn tick_effects(&self) {
         let mut effects_to_remove = Vec::new();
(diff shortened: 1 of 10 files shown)


Snowiiii:
> I used some old methods with GetBlockError in world/mod.rs, they're all marked as "deprecated"

Please just remove them

Q2297045667:
Is there any progress?

Snowiiii:
When you think this is ready ?

akashbeh:
Very soon. Just testing items rn and then the merge can go ahead

akashbeh:
Testing:
LivingEntity works
ItemEntity works

Only player's velocity, not pos, has been changed, so it should work, but it has not been tested in multiplayer

akashbeh:
Collision physics were messed up in the merge. Currently trying to fix

Snowiiii:
> Collision physics were messed up in the merge. Currently trying to fix

So its still not ready?

akashbeh:
> So its still not ready?

No but very close

pumpkin-data/build/block.rs
```diff
@@ -326,6 +326,10 @@ impl ToTokens for BlockPropertyStruct {
                     }
                     block_props
                 }
+
+                fn name(&self) -> &'static str {
+                    "#name"
+                }
```
Snowiiii, line 332: why ?, the block struct already has its name pub

pumpkin/src/entity/living.rs
```diff
@@ -342,6 +332,360 @@ impl EntityBase for LivingEntity {
...
+        let world = self.entity.world.read().await;
+
+        let (block, state) = world.get_block_and_block_state(&pos).await;
+        let name = block.properties(state.id).map(|props| props.name());
+        if let Some(name) = name {
+            if name == "LadderLikeProperties"
+                || name == "ScaffoldingLikeProperties"
+                || name == "CaveVinesLikeProperties"
+                || name == "CaveVinesPlantLikeProperties"
+            {
```
Snowiiii, line 569: This looks ugly, from my unsterstanding vanilla just checks if the block is contained in a block tag
```java
if (blockState.isIn(BlockTags.CLIMBABLE)) {
``

File: pumpkin/src/world/mod.rs [hunk 3]
```diff
@@ -190,9 +169,26 @@ impl World {
             dimension_type,
             weather: Mutex::new(Weather::new()),
             block_registry,
-            sea_level: generation_settings.sea_level,
+            sea_level: 0,
```
Snowiiii: why this was changed ?

Snowiiii:
https://github.com/user-attachments/assets/b272df48-190b-4354-a86b-059c020779c3

pumpkin-data/build/block.rs
akashbeh: This is the name for the BlockProperties which I use below because we don't have block tags

pumpkin/src/entity/living.rs
akashbeh: Yes this is just a workaround since we don't have that yet

pumpkin/src/world/mod.rs
akashbeh: Well the sea level is set later on using get_generation_settings (rather than duplicating code). It's because get_generation_settings is called on World. It would ideally be called on DimensionType, but it's impossible because pumpkin data is built first.

I can fix this by changing the generation settings to an associated method.

pumpkin/src/entity/living.rs
Snowiiii: we have block tags

akashbeh:
> 2025-06-30.10-16-21.mp4

This has been fixed by making velocity dirtier.

The bouncing also doesn't happen in vanilla. Not sure why because the code is the same. Switched some Ordering::Relaxed to SeqCst in case it was a backwards condition, but there is still bouncing. Otherwise functional.

Snowiiii:
@akashbeh 
Can you please fix the bouncing and also implement Physics for TNT Entities before we merge this

akashbeh:
> @akashbeh Can you please fix the bouncing and also implement Physics for TNT Entities before we merge this

Ok. Bouncing is a simple fix (just strange because the code appears non-vanilla) but TNT entities will take some time

Snowiiii:
> > @akashbeh Can you please fix the bouncing and also implement Physics for TNT Entities before we merge this
> 
> Ok. Bouncing is a simple fix (just strange because the code appears non-vanilla) but TNT entities will take some time

How long do you think you will need ?

Snowiiii:
@akashbeh i'm fine with having TNT not included in this PR, but can you at least fix all conflicts then

PR #825, opened 2025-05-21
Fixed overflow for 32 bits

ioterw:
## Description
There was an usize overflow error in compile time, due it is 32 bits for 32 bit arch, so I made it 64 bits, then casted to usize

## Testing
I tested nothing


unschlagbar:
LGTM

Snowiiii:
Looks good. Thank you @ioterw :D

PR #826, opened 2025-05-21
Removed fall damage when landing on powdered snow

Nicolai-Holmgaard:
## Description
When testing i found a bug where you take fall damage when landing on powdered snow, which shouldn't happen.
Added quick check to remove fall damage when landing on the block.

## Testing
Tested it in game, also tested how fall damage works after landing on the block in the air. Everything is good except the view is reacting like the powdered snow wasnt there, but you take the correct fall damage.

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -203,6 +203,13 @@ impl LivingEntity {
         world.get_block(&block_pos).await == Block::WATER
     }
 
+    // Check if the entity is in powdered snow
+    pub async fn is_in_powder_snow(&self) -> bool {
+        let world = self.entity.world.read().await;
+        let block_pos = self.entity.block_pos.load();
+        world.get_block(&block_pos).await == Block::POWDER_SNOW
+    }
+
     pub async fn update_fall_distance(
         &self,
         height_difference: f64,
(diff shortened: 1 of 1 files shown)


Snowiiii:
Hey @Nicolai-Holmgaard Thanks for your PR, but i don't think `is_in_powder_snow` is a good method to use and i also don't like the already existing `is_in_water` method, How does vanilla check this?

Nicolai-Holmgaard:
Hi @Snowiiii I haven't really found out how they do the powder snow falling yet, but with water, vanilla code is this:
``` java
if (!this.isInWater() && fall_distance < 0.0) {
            this.fallDistance -= fall_distance ;
}
``` 
So vanilla uses an `is_in_water` method.

And for the blocks, they have a function that uses a method on the block the entity falls on, to check the fall damage.
So if the entity falls on 
Powder snow, it wouldn't take fall damage, 
Haybales, it would be 20% fall damage,
Normal block, it would be 100%
Stalagmites, there it would be 200%

So maybe that would be the way to do fall damage on blocks, and then have water as its own like in vanilla?

Snowiiii:
I mean this is not optimal but works. So thanks @Nicolai-Holmgaard :+1:  :t-rex:

PR #829, opened 2025-05-22
fixed assert for valid y coordinate

ioterw:
## Description
Breaking lowest bedrock block triggers this assert. This does not make sence. Lets not trigger it. Breaking lowest block should be fine

## Testing
It seems to work fine


Snowiiii:
Oh this must be an upsi. Good catch. Thank you @ioterw

PR #830, opened 2025-05-22
Arc context for plugins

vyPal:
## Description
Until now, plugins only received a `&Context` in the `on_load()` and `on_unload()` methods. This PR changes the `&Context` to a `Arc<Context>` so that plugins can store the context to use later. This will allow plugins to interact with the server at any time, not only when the server requests interaction.

## Testing

### DIFF
--- a/pumpkin/src/plugin/api/context.rs
+++ b/pumpkin/src/plugin/api/context.rs
@@ -20,9 +20,9 @@ use super::{Event, EventPriority, PluginMetadata};
 /// - `server`: A reference to the server on which the plugin operates.
 /// - `handlers`: A map of event handlers, protected by a read-write lock for safe access across threads.
 pub struct Context {
-    metadata: PluginMetadata<'static>,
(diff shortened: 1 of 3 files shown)


Snowiiii:
I'm not sure why the plugin should not be just able to clone Server which is already an Arc?

vyPal:
The server can be cloned, but the other fields of the context (`PluginMetadata` and `HandlerMap`) are private by default and therefore can't be cloned, and they can currently only be accessed by the `on_load` and `on_unload` methods. I was originally only going to make them public, so that the plugin could clone them, but that would mean the plugin would have to implement the helper methods (like `register_event`) so I thought making the whole context an Arc would be easier, as the plugin wouldn't have to individually clone each field.

That said, maybe I could remove the `Arc` from the server and handler map, so that we don't have nesting arcs

vyPal:
I'm going to close this PR and open a new one which will incorporate these changes + some other visibility changes to make necessary methods available and accessible to the plugin API

PR #831, opened 2025-05-23
Implement string-based permission system

vyPal:
## Description
This PR implements a string-based, namespaced permission system (not only) for commands.
Permissions are prefixed with a namespace, that can either be:
- `minecraft:` for vanilla builtin commands
- `pumpkin:` for builtin commands that have been added on top of vanilla
- or when used with plugins, the plugin identifier is used (plugins may only register permissions within their namespace, there is a workaround, but it is only intended for advanced use cases)

The value of each permission is a list of identifiers separated by the `.` character. There can be as many identifiers as necessary.

When registering a permission, the default value must be provided. This can be one of:
- `PermissionDefault::Allow` - which will allow the permission by default
- `PermissionDefault::Deny` - which will deny the permission by default
- `PermissionDefault::Op(PermissionLvl)` - will allow the permission for any player with a minimum PermissionLvl

Permissions can also have children. Children are a list of other permissions, which the player will be granted if they are granted the base permission. Example: if the permission `minecraft:command.teleport` has `minecraft:command.teleport.self = true` and `minecraft:command.teleport.other = true`, any player that is granted the `minecraft:command.teleport` permission will also be granted both of the child permissions.

Similarly to how other mc server software works, the permission system will only store the permissions (and any changes to them) in memory while the server is running. This means that plugins will be in charge of storing permissions in some non-volatile location.

## Testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2003,6 +2003,8 @@ dependencies = [
  "serde",
  "serde_json",
  "syn",
+ "tokio",
+ "uuid",
 ]
 
 [[package]]
--- a/pumpkin-util/Cargo.toml
+++ b/pumpkin-util/Cargo.toml
@@ -19,3 +19,6 @@ quote = "1.0"
 syn = "2.0"
 proc-macro2 = "1.0"
 enum_dispatch = "0.3.13"
+
+uuid.workspace = true
+tokio.workspace = true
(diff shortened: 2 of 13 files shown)


Snowiiii:
Looks good. Thank you @vyPal :D, Good work :+1:

PR #833, opened 2025-05-24
fix: config's log level and env fields are ignored

qwixck:
## Description
for some reason the config's log level and env fields are ignored, so it's always using `RUST_LOG` environment variable

## Testing
tested, everything works fine


qwixck:
small note:
why in `level` fields I didn't used `LevelFilter`? Because `Serialize` and `Deserialized` aren't implement for `LevelFilter`. If you try, you'll get an error because you can't implement foreign traits for foreign structs. So I created a custom `LoggingLevel` and just convert it into original `LevelFilter`.

Snowiiii:
i would prefer if it would be parsed directly from env instead using the config

PR #838, opened 2025-05-31
Implement End

teknostom:
## Description

Added end portal frames directional placing.
Added placing eyes of ender into portal frames.
Added creation of end portal when criteria are fulfilled

Also redid buckets to before the merge fault. (#811)

## Testing

Manual testing


Snowiiii:
Hey @teknostom, Please resolve conflicts

teknostom:
Also added block boundaries (outline shapes) for better collision and raytracing.

Snowiiii:
Please only update the necessary json assets and not all. Also if you change the extractor you need to make a PR for that on our Extractor repo.
Also what is the difference in outline shape between the current collision box shape ?

unschlagbar:
It makes no sense to add outlines and filling it with the same values as collision shapes

teknostom:
Added this commit to Extractor, forgot it last night: https://github.com/Pumpkin-MC/Extractor/pull/17

The bounding boxes are basically the same as collision shapes, but for some blocks, there are mismatches between the collision hitbox, and the shown hitbox. Mojang wouldn't have added them if they were not needed.

teknostom:
Will open again from new branch

PR #839, opened 2025-05-31
usize crashes my server all the time

ioterw:
## Description
usize crashes all around for me, when I play super cool 32 bit pumpkin.
Can we not use usize in some cool complicated multiplications and other math operations where big numbers could possibly come?
It could seem fun, but it crashes indeed. So I removed some usize from project

## Testing
Tested a bit


Snowiiii:
Looks good. Thank you @ioterw :kangaroo:

Snowiiii:
Just fix the conflict please

ioterw:
done

PR #841, opened 2025-06-03
Small changes to plugin API

vyPal:
## Description
This PR contains several changes to the plugin API and it's surrounding mechanisms.
### Changing method/module visibility

The first commit of this PR changes the `pub(crate)` visibility specifier on several builtin methods and modules to only `pub` to make them accessible from the plugin API. The main changes are exposing blocks, all argument types, and some fields and methods on the `CommandDispatcher`. The reason for exposing the `CommandDispatcher` methods is so that plugins can register and unregister commands and any moment during the life-cycle, not only during loading.

### Allow plugins to register custom plugin loaders

This change adds a couple helper methods to allow plugins to register their own custom plugin loaders with the central `PLUGIN_MANAGER`, which was not possible until now, as there are two separate PluginManagers, one used by the server itself (registered in the `main.rs` file) which is not accessible from plugins, and another registered in the `lib.rs` file, which is accessible from plugins, but isn't used (it is only declared there due to rust-analyzer complaining about it's absence).

This change also slightly modifies the behavior when registering a new loader: when a new loader is added, the PluginManager will go through a list of plugin files which it failed to load (for whatever reason) and test if the file can be loaded with the newly registered loader. If yes, the manager will attempt to load the plugin file with the new manager.

### Unloading all plugins when the server is shutting down

This change simply adds a new method for unloading all still loaded plugins, and calls it during the server shutdown process, after finishing all player tasks, but before the worlds are saved, so that plugins can make some last-minute changes to the worlds.

## Testing


Snowiiii:
@vyPal please fix conflicts

Snowiiii:
Looks good. Thank you @vyPal :D :rabbit2:

PR #842, opened 2025-06-03
removed scary unwrap

ioterw:
## Description
Server crashes in case of blocks being placed in various ways outside acceptable y bounds.
So I removed scary .unwrap()


Snowiiii:
Looks good. Thank you @ioterw

PR #845, opened 2025-06-04
packet type in "Failed to send packet" error

ioterw:
## Description
It feels cooler to have some error description about packet like this:
`Failed to send packet pumpkin_protocol::client::play::chunk_data::CChunkData to client`
Than like this
`Failed to send packet to client`


unschlagbar:
whats the point

Snowiiii:
I wonder if this is really necessary and how `std::any::type_name::<P>()` works, because i assume this will may add some debug information at runtime so get the names which would add more bloat

ioterw:
I used it to fix nether portals, but ended to use via proxy instead. I think, it is kinda useful, but maybe really not as useful as other approaches. I think, I will close.

PR #846, opened 2025-06-05
nether fix indeed

ioterw:
## Description
nether and end are 256 blocks tall. client will not get the server if, server sends 8 sections instead of 16...

## Testing
a bit


ioterw:
bad fix

Issue #801, opened 2025-05-08
Invalid Spawn Y location


Snowiiii:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

---

### Server Software Version/Commit

latest

### What happened?

Currently 200 is used for the default Spawn Y until its saved in the player data

### To Reproduce

Join a World the first time and see your in the sky

### Expected behavior

Use the Motion height map to get Spawn Y

PR #852, opened 2025-06-06
fix: spawn player on top block instead of 200 y (#801)

DocteurPing:
## Description

This PR was made to resolve [this issue](https://github.com/Pumpkin-MC/Pumpkin/issues/801).
When a player first spawns in a world, their y coordinate is set to 200, which typically results in their death on the first spawn. To address this issue, I modified the player’s spawning mechanism to ensure they spawn on the top block of the x and z coordinates. However, this modification, coupled with the current codebase, caused the player to spawn on the top block, but since the chunk generation occurred after the player spawned, the player would glitch out of the map. Consequently, I implemented a change to generate the chunk and send it to the client before the player spawns. 

## Testing

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -562,6 +562,11 @@ impl World {
                 true,
             ))
             .await;
+
+        // Spawn in initial chunks
+        // This is made before the player teleport so that the player doesn't glitch out when spawning
+        chunker::player_join(&player).await;
+
         // Permissions, i.e. the commands a player may use.
         player.send_permission_lvl_update().await;
         {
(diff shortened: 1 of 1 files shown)


qwixck:
that does fix the main issue, but, it would be way better if the `world.level_info.spawn_y` would be a correct coordinate based of heightmap. The player will spawn correctly, but, every other command and etc will still use the default 200 Y level for spawning an entity (if coordinates are not specified)

qwixck:
> that does fix the main issue, but, it would be way better if the `world.level_info.spawn_y` would be a correct coordinate based of heightmap. The player will spawn correctly, but, every other command and etc will still use the default 200 Y level for spawning an entity (if coordinates are not specified)

of course it's only my opinion, you can ignore it

pumpkin/src/world/mod.rs
```diff
@@ -562,6 +562,11 @@ impl World {
                 true,
             ))
             .await;
+
+        // Spawn in initial chunks
+        // This is made before the player teleport so that the player doesn't glitch out when spawning
+        chunker::player_join(&player).await;
+
         // Permissions, i.e. the commands a player may use.
         player.send_permission_lvl_update().await;
```
Snowiiii, line 567: this should be send before sending chunks

DocteurPing: fixed

Snowiiii:
Looks good. Thank you @DocteurPing 
Next we should use the best chunk for spawning, so you don't spawn in an ocean

PR #855, opened 2025-06-06
Support swap item action

sprinox:
## Description

Support swap item action
Also fix the saving of equipment slots

## Testing

Test in minecraft
动画

### DIFF
--- a/pumpkin-inventory/src/player/player_inventory.rs
+++ b/pumpkin-inventory/src/player/player_inventory.rs
@@ -45,7 +45,10 @@ impl PlayerInventory {
     }
 
     pub async fn swap_item(&self) -> (ItemStack, ItemStack) {
-        let slot = self.equipment_slots.get(&40).unwrap();
+        let slot = self
+            .equipment_slots
+            .get(&PlayerInventory::OFF_HAND_SLOT)
+            .unwrap();
         let mut equipment = self.entity_equipment.lock().await;
         let binding = self.held_item();
         let mut main_hand_item = binding.lock().await;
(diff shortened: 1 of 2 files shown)


OmarAfet:
The item on the off-hand will disappear after reconnecting to the server, just tested it.

sprinox:
> The item on the off-hand will disappear after reconnecting to the server, just tested it.

I’m aware of this issue, but I noticed that even if I manually move an item to the off-hand and re-enter the game, it still disappears. So this is probably unrelated to the code in question

sprinox:
now it is impossible to lead to data race

Snowiiii:
You don't send any packets to other clients so they don't see an item in the offhand

sprinox:
> You don't send any packets to other clients so they don't see an item in the offhand

it seems that dropping a stack of items and picking item from block without slot changing dont send update packets.

sprinox:
> You don't send any packets to other clients so they don't see an item in the offhand

i have solved the problem. can you check it out?

Snowiiii:
> > You don't send any packets to other clients so they don't see an item in the offhand
> 
> i have solved the problem. can you check it out?

Okay this seems fixed now. Now please fix that the item in the offhand is not saved and loaded when rejoining a world, Also please test and fix any bugs, When your sure that everything works and is fine unmark this as draft

sprinox:
> Okay this seems fixed now. Now please fix that the item in the offhand is not saved and loaded when rejoining a world, Also please test and fix any bugs, When your sure that everything works and is fine unmark this as draft

i have fixed the saving of equipment slots

Snowiiii:
Nice. Looks good now, Thank you @spr-equinox :D :panda_face:

PR #857, opened 2025-06-07
Prevent command block placement in survival mode

BatuhanAksoyy:
## Description
Players should not be able to place command blocks in survival mode.

## Testing
test


unschlagbar:
maybe you can also implement a check when a survival player interacts with the cmd block. cmd blocks are not implemented yet but it doesnt matter

BatuhanAksoyy:
Why would it be necessary since we cannot open them in survival mode already?

Snowiiii:
Looks good, Thank you @BatuhanAksoyy

PR #858, opened 2025-06-07
Fix permissions not being registered from plugins

vyPal:
## Description
Fixes the issue where plugins are unable to register permissions by passing a clone of the permission manager to the plugin's context.

The original issue is caused by the fact that both the `main.rs` and `lib.rs` files define a `PERMISSION_MANAGER`, and plugins use the instance created by `lib.rs` whereas the internal server uses the instance created from `main.rs`.

## Testing


Snowiiii:
Thanks @vyPal

PR #859, opened 2025-06-07
Implement Swimming (touching water check, submerged in water check, drowning)

lwalton101:
## Description
This PR implements an Air value for all entities as well as logic for detecting if they are submerged in water or touching water.
It also sets the player into the swimming pose when they start swimming.
The swimming eye height for the player is added, but I have moved the eye_position function to the living entity class as it is relevant for all living entities.
Drowning also works.

## Testing

There are 2 problems I am yet to fix
When standing in a door underwater you still drown, due to Pumpkin still thinking there is water there.
Air is not reset to 300 when the player dies


Ancient77:
> Air is not reset to 300 when the player dies

There should be a method called "kill" or something where you can restore this

akashbeh:
I added an in_water AtomicBool in Entity physics so I am working on that and water physics for entities in general. A lot of this code seems only relevant to players. E.g. not sure what happens when you set an item's pose to swimming.

Snowiiii:
@lwalton101 do you still want to work on this or should be move to https://github.com/Pumpkin-MC/Pumpkin/pull/986 ?

Issue #766, opened 2025-04-24
Redstone issues


azomDev:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Fedora 42

### Server Software Version/Commit

0.1.0-dev+1.21.5 (abb4e541)

### What happened?

There is a few issues with redstone since it is a complicated part. This issue is to put all of them into one place so it's easier to see the current problems with redstone. Many of these issues can end up being linked to each other. I have already found a few and I didn't want to open one issue for each bug, so I'm doing this. I'll put the ones I find as comments to this issue, and anyone can add to it.

I'm doing this since I think it will make it easier to track these, but if it's not a good way to handle that feel free to let me know, I'm new to the project :)

### To Reproduce

See comments

### Expected behavior

See comments

azomDev:
![Image](https://github.com/user-attachments/assets/22b206da-91d2-46ca-8410-fa5b09e33396)

In some configurations, repeaters are not getting locked. The redstone lamp is used to indicate North.

azomDev:
![Image](https://github.com/user-attachments/assets/c31625ef-1fc9-4b3d-beb9-0e01635d2db5)

When the target block is the last block to be placed in this setup (redirecting the redstone), the light does not power on.

azomDev:
![Image](https://github.com/user-attachments/assets/21e939fc-ad92-4f33-8013-700f880ffa54)
When the repeater is the last block to be placed in this setup (redirecting the redstone), the light does not power on.

azomDev:
Comparator does not work at all. (no `comparator.rs` in redstone directory)

azomDev:
![Image](https://github.com/user-attachments/assets/f3e19fa8-db69-46c1-9e69-2e9d472048bc)

Really cursed, but this setup in vanilla (which the screenshot was taken in) flickers a lot for a few seconds until it powers completely down, creating a little cloud of smoke (see second image). While in pumpkin it never stops.

![Image](https://github.com/user-attachments/assets/ef3c9a22-8f83-4a85-97f8-975505cd2139)

azomDev:
![Image](https://github.com/user-attachments/assets/2945185d-c0d2-472b-bc1f-3fadb1664836)

The lamp should power on when an entity goes on the string (not sure exactly if it's on entry or something close to that)

azomDev:
[...]

PR #861, opened 2025-06-07
Fix issue with Level::is_block_tick_scheduled, related to problems with redstone.

Piet06:
## Description

Level::is_block_tick_scheduled did not detect block-ticks that will be executed later in the current tick,
leading to incorrect behavior of redstone torches and repeaters, like redstone torches
reacting to 2 gametick pulses, and repeaters in some cases flashing when being turned off. 

is_block_tick_scheduled should detect block-ticks that will be executed later in the current tick,
but not detect block-ticks that were already processed.

To implement this, is_block_tick_scheduled needs to know which block-ticks were already executed
by the World::tick_scheduled_block_ticks function in the current tick, therefore we store the block-ticks,
that still need to be processed in a field in the Level called "remaining_block_ticks_this_tick", where both
methods can access it.

This is to kind of related to #766, however the redstone problems fixed with this PR are not mentioned there explicitly.

The same could also be done with fluid-ticks, however is_fluid_tick_scheduled is currently not used
anywhere, and i don't know of any case where this matters with fluids.

## Testing

Passed all tests.

### DIFF
--- a/pumpkin-world/src/chunk/mod.rs
+++ b/pumpkin-world/src/chunk/mod.rs
@@ -101,7 +101,7 @@ impl From<i32> for TickPriority {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Copy)]
 pub struct ScheduledTick {
     pub block_pos: BlockPos,
     pub delay: u16,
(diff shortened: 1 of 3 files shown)


Snowiiii:
Looks good. Thank you @Piet06 :+1:  :polar_bear:

PR #862, opened 2025-06-07
Implementation of QuickCraft, PickupAll and Throw in inventory

azomDev:
## Description

Quickcraft is when holding left or right click and dragging an item in the inventory. PickupAll is when double clicking an object to collect most of them on the cursor.


azomDev:
for QuickCraft left click, I'm unsure how I can keep reference to the slots/itemstack I'm using. Because when you hold drag with left click with a block on the cursor, with every new slot that it goes on, it might change all of the others from the same drag

Snowiiii:
This was resolved in https://github.com/Pumpkin-MC/Pumpkin/pull/940. Still Thank you @azomDev

PR #864, opened 2025-06-08
Added fire_charge item handler and burning util for exclude DRA. 

destroydevs:
## Description
Added fire_charge handler in pumpkin/src/item/items/fire_charge.rs
Added burning util in pumpkin/src/world/burning.rs which contains burn_block function for exclude DRA (don't repeat yourself).
Added some TODO's for future.

Removed unused imports.

> This PR adds to Pumpkin server a handler for fire_charge.

## Testing
While testing I build the project on Windows and starts server, fire_charge is works, but item not removing in survival. 
It need to do in future.


Snowiiii:
i don't like that you just moved this method into world, where it does not even is implemented in World.
Also flint and steel use different function logic (sound for example) that's also the reason they are in separate functions in vanilla

destroydevs:
> i don't like that you just moved this method into world, where it does not even is implemented in World. Also flint and steel use different function logic (sound for example) that's also the reason they are in separate functions in vanilla

Okay, I'll fix this soon

destroydevs:
@Snowiiii would you have time to review the PR? I'm keen to get started on some other features, so a review would be much appreciated

PR #866, opened 2025-06-08
Tick block entities

Piet06:
## Description

Block entities should be ticked at the end of every tick.

There was an old comment about this causing a deadlock when done at the beginning of the tick,
when testing this change i did not encounter any deadlocks.

Currently piston are the only block entities that tick and they work a lot better with this change, but still have some issues.

I also removed two annoying dbg! statements from the piston implementation, that messed up the console log.

## Testing

Passed all tests

### DIFF
--- a/pumpkin/src/block/blocks/piston/mod.rs
+++ b/pumpkin/src/block/blocks/piston/mod.rs
@@ -50,13 +50,7 @@ impl<'a> PistonHandler<'a> {
         self.moved_blocks.clear();
         self.broken_blocks.clear();
         let (block, block_state) = self.world.get_block_and_block_state(&self.pos_to).await;
-        dbg!(PistonBlock::is_movable(
-            &block,
-            &block_state,
-            self.motion_direction,
-            false,
-            self.piston_direction,
-        ));
+
         if !PistonBlock::is_movable(
             &block,
             &block_state,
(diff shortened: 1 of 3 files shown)


Snowiiii:
this still dead locks and causes chunks to load very slow, tbh the current tick_blocks function is really not optimal and there will be more changes needed

Piet06:
I think the deadlock happens because the chunk loading has a mutable reference into the chunk dashmap,
when we are creating the iterator in the tick_block_entities funtion, see [this](https://docs.rs/dashmap/6.1.0/dashmap/struct.DashMap.html#method.iter).

To do this properly we should store a list of block entity that should be ticked in the level.
That would give us the correct block entity update order, allow us to not have to tick 
block entities that don't have a tick function and also eliminate the deadlock.

Snowiiii:
> I think the deadlock happens because the chunk loading has a mutable reference into the chunk dashmap, when we are creating the iterator in the tick_block_entities funtion, see [this](https://docs.rs/dashmap/6.1.0/dashmap/struct.DashMap.html#method.iter).
> 
> To do this properly we should store a list of block entity that should be ticked in the level. That would give us the correct block entity update order, allow us to not have to tick block entities that don't have a tick function and also eliminate the deadlock.

We should also consider to somehow implement random ticks for blocks, Storing them separately would seem like a waste of memory

Issue #815, opened 2025-05-16
When logging in, selected slot is not sent.


teknostom:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows/wsl

### Server Software Version/Commit

latest

### What happened?

When joining, the client assumes the first item slot. If the stored slot is not the first, then there is a discrepancy, and when you try to use it, it will not work. Example: https://github.com/Pumpkin-MC/Pumpkin/pull/811#issuecomment-2884483884

### To Reproduce

Join the server, put an item in the first slot, select a slot other than the first.

When you rejoin, the visually selected and held item should be from the first slot, but that item will not work. If you had an item in the selected when you left, that item will be used instead.

### Expected behavior

The server should send the selected slot on join.

PR #867, opened 2025-06-08
send SetSelectedSlot on player join

Ancient77:
## Description
Fixes: #815:
Send the selected slot from NBT to the client upon joining.

This has to be done since the inventory sync_handler doesn't keep the selected slot in sync.
(At least, that's how I understand it. Maybe @4lve can confirm?)


Snowiiii:
Looks good @Ancient77, but can you solve the conflict

Issue #507, opened 2025-01-27
Erroneous Fall damage


Pookachu:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

linux-6.13-arch1-1

### Server Software Version/Commit

Pumpkin 0.1.0-dev (ea0cc20)

### What happened?

I'm on the test server `pumpkin.kralverde.dev` and while falling along the edge of a cliff I took fall damage without hitting the ground.

### To Reproduce

Join `pumpkin.kralverde.dev`, fall along the edge of the cliffs from the spawn platform.

### Expected behavior

To only take fall damage when you hit the ground.

ploxxxy:
Looks like the issue has worsened after dbbf745da27cc19ec519030cbbd0eccc3951c03f. Now you can die while falling and you retain some kind of falling/momentum data after respawning, hence me dying in the video for the second time

https://github.com/user-attachments/assets/7c929a66-ef7d-4a3f-a25f-5192e4f24412

DavyCosta701:
If you fall on water you take damage too

RobIux:
Issue seems to relate to the is_grounded check,

4lve:
The server needs to calculate the ground check instead of the client

DataM0del:
> The server needs to calculate the ground check instead of the client

I don't think that's related to this issue.

4lve:
> > The server needs to calculate the ground check instead of the client
> 
> I don't think that's related to this issue.

When I checked the code quickly before the clients onGround is used for when damaging the player. So it's most likely the player sending invalid onGround, anyways we need to calculate it ourselves instead of trusting the client

DataM0del:
> > > The server needs to calculate the ground check instead of the client
> > 
> > 
> > I don't think that's related to this issue.
> 
> When I checked the code quickly before the clients onGround is used for when damaging the player. So it's most likely the player sending invalid onGround, anyways we need to calculate it ourselves instead of trusting the client

Yes, but I want my No Fall moment (one of the 5 trillion No Fall modes Liquid Bounce has will probably work) 😡
I'll probably do it anyway 💀. (also, it's **client's** onGround, not **clients** onGround, spelling gripe moment)

Snowiiii:
I can't reproduce this anymore, does this bug still exist ?

PR #882, opened 2025-06-09
Fixed weird behavior of fall damage when player is in water #507

firelop:
## Description
The fall distance was not reset when the player fell into water or powder snow, causing the player to take damage when returning to the ground. 

The bug wasn't really in #507 but is similar.

## Testing
In-Game testing

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -238,14 +238,15 @@ impl LivingEntity {
                     .await;
             }
         } else if height_difference < 0.0 {
-            let mut new_fall_distance = 0f32;
+            let new_fall_distance = if !self.is_in_water().await && !self.is_in_powder_snow().await
+            {
+                let distance = self.fall_distance.load();
+                distance - (height_difference as f32)
+            } else {
+                0f32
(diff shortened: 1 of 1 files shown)


Ancient77:
Should this only run if the height difference is negativ?
I assume it should be reset every time you are in water

firelop:
> Should this only run if the height difference is negativ?
> I assume it should be reset every time you are in water

I think it's useless checks.

akashbeh:
This is going to be fixed in my Entity physics PR

Snowiiii:
It feels like https://github.com/Pumpkin-MC/Pumpkin/issues/507 is fixed on master ?

firelop:
I wasn't able to reproduce #507 but while trying to do so, I discovered a strange behavior with fall damage when the player is in water or powder snow: while the player was in these blocks, LivingEntity fall_distance was still updated which caused the player to take damage when returning on a solid block. Sorry if I wasn't clear in my original post.

Snowiiii:
Sorry for the late reply. Looks good to me. Thank you @firelop :D 🦥

Issue #659, opened 2025-03-24
E0283 while compiling


ZackZhao233:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11 Pro 24H2

### Server Software Version/Commit

failed to compile

### What happened?

while executing cargo run --release, an error occurs

error[E0283]: type annotations needed
    --> pumpkin\src\net\packet\play.rs:1467:69
     |
1467 |         if location.0.y + face.to_offset().y < Self::WORLD_LOWEST_Y.into() {
     |                                              -                      ^^^^
     |                                              |
     |                                              type must be known at this point
     |
     = note: multiple `impl`s satisfying `i32: PartialOrd<_>` found in the following crates: `core`, `deranged`:
             - impl PartialOrd for i32;
             - impl<MIN, MAX> PartialOrd<deranged::RangedI32<MIN, MAX>> for i32
               where the constant `MIN` has type `i32`, the constant `MAX` has type `i32`;
help: try using a fully qualified path to specify the expected types
     |
1467 |         if location.0.y + face.to_offset().y < <i8 as Into<T>>::into(Self::WORLD_LOWEST_Y) {
     |                                                ++++++++++++++++++++++                    ~

error[E0283]: type annotations needed
    --> pumpkin\src\net\packet\play.rs:1472:67
     |
1472 |         if location.0.y + face.to_offset().y >= Self::WORLD_MAX_Y.into() {
     |                                              --                   ^^^^
     |                                              |
     |                                              type must be known at this point
     |
     = note: multiple `impl`s satisfying `i32: PartialOrd<_>` found in the following crates: `core`, `deranged`:
             - impl PartialOrd for i32;
             - impl<MIN, MAX> PartialOrd<deranged::RangedI32<MIN, MAX>> for i32
               where the constant `MIN` has type `i32`, the constant `MAX` has type `i32`;
help: try using a fully qualified path to specify the expected types
     |
1472 |         if location.0.y + face.to_offset().y >= <u16 as Into<T>>::into(Self::WORLD_MAX_Y) {
     |                                                 +++++++++++++++++++++++                 ~

For more information about this error, try `rustc --explain E0283`.
error: could not compile `pumpkin` (lib) due to 2 previous errors

### To Reproduce

git clone https://github.com/Pumpkin-MC/Pumpkin.git
cd Pumpkin
cargo run --release

### Expected behavior

compile successfully

[...]

PR #883, opened 2025-06-10
max_distance tweak

ioterw:
## Description
For some reason, if I add one of libraries, I am testing out right now into pumpkin, I get E0283 error in only one specific file across the whole project. To be honest, I am not sure why it is happening, and only in one place, but I made a fix that changes nothing during normal Pumpkin build/usage process. Also it is somehow similar to: https://github.com/Pumpkin-MC/Pumpkin/issues/659
Also removed underscore from _min_volume, as it is actually tested to be positive.

## Testing
- checked that types are correct, otherwise nothing to test

### DIFF
--- a/pumpkin/src/command/commands/playsound.rs
+++ b/pumpkin/src/command/commands/playsound.rs
@@ -103,7 +103,7 @@ impl CommandExecutor for Executor {
         };
 
         // Get optional minimum volume (currently unused in implementation)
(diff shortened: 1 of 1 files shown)


ioterw:
```
error: unused import: `super::entity_status::*`
  --> pumpkin-data/src/lib.rs:74:13
   |
74 |     pub use super::entity_status::*;
```
Strange (not from my commit)

Snowiiii:
Thank you @ioterw :sheep:

PR #888, opened 2025-06-11
Fix: Range limit for teleport command.

Adamex2204:
Fix: Enforce /teleport range limits to match Minecraft (-30M to +30M)

### DIFF
--- a/pumpkin/src/command/commands/teleport.rs
+++ b/pumpkin/src/command/commands/teleport.rs
@@ -49,6 +49,14 @@ fn yaw_pitch_facing_position(
     (yaw_degrees as f32, pitch_degrees as f32)
 }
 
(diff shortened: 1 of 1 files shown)


pumpkin/src/command/commands/teleport.rs
```diff
@@ -49,6 +49,14 @@ fn yaw_pitch_facing_position(
     (yaw_degrees as f32, pitch_degrees as f32)
 }
 
+fn dest_in_range(dest: &Vector3<f64>) -> bool {
+    let range = -30_000_000.0..=30_000_000.0;
```
teknostom, line 53: I think this is creating a range on each function call? I think it would be better to have this as constant(s).

PR #889, opened 2025-06-11
Fix: Added range limit for teleport command.

Adamex2204:
Fix: Enforce /teleport range limits to match Minecraft default (-30M to +30M)

### DIFF
--- a/pumpkin/src/command/commands/teleport.rs
+++ b/pumpkin/src/command/commands/teleport.rs
@@ -1,7 +1,6 @@
 use async_trait::async_trait;
 use pumpkin_util::math::vector3::Vector3;
 use pumpkin_util::text::TextComponent;
-use std::ops::RangeInclusive;
 
 use crate::command::CommandError;
 use crate::command::args::ConsumedArgs;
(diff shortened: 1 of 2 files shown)


pumpkin/src/command/commands/teleport.rs
```diff
@@ -49,6 +49,12 @@ fn yaw_pitch_facing_position(
     (yaw_degrees as f32, pitch_degrees as f32)
 }
 
+fn dest_in_range(dest: &Vector3<f64>) -> bool {
+    let range = -30_000_000.0..=30_000_000.0;
```
teknostom: I think this is creating a range on each function call? I think it would be better to have this as constant(s).

Snowiiii:
remove the underscore in `_sender`, this is only used for unused variables

pumpkin/src/command/commands/teleport.rs
```diff
@@ -49,13 +53,17 @@ fn yaw_pitch_facing_position(
     (yaw_degrees as f32, pitch_degrees as f32)
 }
 
+fn dest_in_range(dest: &Vector3<f64>) -> bool {
+    RANGE.contains(&dest.x) && RANGE.contains(&dest.y) && RANGE.contains(&dest.z)
+}
```
Snowiiii: I think instead of this it would be better to create a `is_valid` function in `World` because we will need that in other Commands as well, Also we should check for -20000000...20000000 for Y like vanilla and not 30000000

Snowiiii:
Please fix CI

Snowiiii:
Feel free to reopen

PR #886, opened 2025-06-11
Implemented NamespacedKey

tn-lorenz:
## Description
Implemented `NamespacedKey` and a macro `ns_key!` that automatically fetches the `CARGO_PKG_NAME` for ease of use 

## Testing

Please follow our [Coding Guidelines](https://github.com/Pumpkin-MC/Pumpkin/blob/master/CONTRIBUTING.md#coding-guidelines)

PR #891, opened 2025-06-12
Add: NamespacedKey, PersistentDataContainer

tn-lorenz:
## Description
This pull request introduces support for `NamespacedKey` and `PersistentDataContainer`.

The `NamespacedKey` struct is implemented alongside a convenient macro, `ns_key!`, which automatically uses the `CARGO_PKG_NAME` as the namespace to simplify key creation.

The `PersistentDataContainer` system works differently from the typical Java implementation (e.g., in Bukkit and its derivatives). In this Rust version, `PersistentDataHolder` is not a trait that needs to be implemented by other structs. Instead, it is a generic wrapper struct that can hold a reference to any arbitrary type while associating it with a `PersistentDataContainer`.

This design enables developers to persistently attach data to any type without modifying its definition or implementing boilerplate interfaces. It also makes serialization and deserialization more flexible and manageable.

## Testing
Unit tests have been added to verify functionality.


tn-lorenz:
This needs to be merged after #886, but I think I fucked it up, anyway

PR #893, opened 2025-06-13
Plugin API / player interact event

topi-banana:
## Description

Added player interact event to the Plugin API events.

## Testing

### DIFF
--- a/pumpkin-protocol/src/server/play/interact.rs
+++ b/pumpkin-protocol/src/server/play/interact.rs
@@ -52,7 +52,7 @@ impl ServerPacket for SInteract {
     }
 }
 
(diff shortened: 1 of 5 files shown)


tn-lorenz:
I think it would be better if the file is called `player_interact.rs`, so it matches the others and also maybe let's rename the `InteractAction` Enum to `Action`, so it matches bukkit

topi-banana:
I thought so too, but
`pumpkin/src/net/packet/play.rs` already imports an enum named `Action`, so I named it `InteractAction`.
This enum is only used as an internal value for events, so I don't think the difference in name is an issue.

Snowiiii:
Looks good. Thank you @topi-banana :+1: :wolf:

PR #894, opened 2025-06-13
fix: Do not generate fire block when use the flint and steel on tnt block

### LINKED ISSUE #875: TNT and fire

### Operating System

Windows 11 专业工作站版 Insider Preview

27871.1000

### What happened?

https://github.com/user-attachments/assets/ee8364a6-878e-4176-9be0-8991f612a6d1

### To Reproduce

First, place the fire and then use TNT to click on the fire.

### Expected behavior

Vanilla

XYZboom:
## Description

Do not generate fire block when use the flint and steel on tnt block.
Fixed #875 

## Testing

Tested manually.

### DIFF
--- a/pumpkin/src/item/items/flint_and_steel.rs
+++ b/pumpkin/src/item/items/flint_and_steel.rs
@@ -26,9 +26,13 @@ impl PumpkinItem for FlintAndSteelItem {
         player: &Player,
         location: BlockPos,
         face: BlockDirection,
-        _block: &Block,
+        block: &Block,
         _server: &Server,
     ) {
+        if block.eq(&Block::TNT) {
+            // if use on tnt block, do not generate a new fire block
+            return;
+        }
         // TODO: check CampfireBlock, CandleBlock and CandleCakeBlock
         let world = player.world().await;
         let pos = location.offset(face.to_offset());


pumpkin/src/item/items/flint_and_steel.rs
```diff
@@ -29,6 +29,10 @@ impl PumpkinItem for FlintAndSteelItem {
         _block: &Block,
         _server: &Server,
     ) {
+        if _block.eq(&Block::TNT) {
```
kralverde: Don't have an underscore prefix here

kralverde (changes requested):

Snowiiii:
Hardcoding TNT feels like a bad approche tbh, would like to know how vanilla does it

XYZboom:
@Snowiiii Actually, we have already implemented the vanilla TNT logic, but the fire block seem to be generated by world updates rather than flint.
image

XYZboom:
Sorry, I made a mistake. The screenshot shows the logic of blocks such as candles

XYZboom:
For vanilla, they seem unwilling to determine the type of block and instead use the state to judge

XYZboom:
Vanilla call use_on_block after use_with_item but we do not. That's why an extra fire block generated. I will create a new issue and close this pr since this pr can not fix the problem.
image

PR #897, opened 2025-06-14
fix fence gate's facing when open

spr-equinox:
## Description

the gate's facing should be changed to opposite direction when player open it on the opposite side
fix fence gate's facing when open

## Testing


Snowiiii:
Nice job, I also noticed that we currently ignore redstone power in fence gates. Thank you @spr-equinox :D :crocodile:

PR #899, opened 2025-06-14
fix lever can place at

topi-banana:
## Description

image

## Testing

### DIFF
--- a/pumpkin/src/block/blocks/redstone/lever.rs
+++ b/pumpkin/src/block/blocks/redstone/lever.rs
@@ -10,7 +10,10 @@ use pumpkin_data::{
 use pumpkin_macros::pumpkin_block;
 use pumpkin_protocol::server::play::SUseItemOn;
 use pumpkin_util::math::position::BlockPos;
-use pumpkin_world::{BlockStateId, world::BlockFlags};
+use pumpkin_world::{
+    BlockStateId,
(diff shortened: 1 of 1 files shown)


pumpkin/src/block/blocks/redstone/lever.rs
```diff
@@ -67,6 +70,26 @@ impl PumpkinBlock for LeverBlock {
...
+        face: BlockDirection,
+        _use_item_on: Option<&SUseItemOn>,
+    ) -> bool {
+        if let Some(world) = world {
+            let place_block_pos = block_pos.offset(face.to_offset());
+            let place_block_state = world.get_block_state(&place_block_pos).await;
+            place_block_state.is_side_solid(face)
+        } else {
+            false
+        }
```
Snowiiii, line 90: instead of world you should use block_accessor, Would be the same thing but supports world generation and would make the code cleaner

Snowiiii:
Nice work, I would prefer if you would implement all "WallMountedBlock's" from Vanilla, so they all have the same logic

topi-banana:
https://github.com/Pumpkin-MC/Pumpkin/pull/908

Snowiiii:
Looks like you also implemented lever in #908. Can this be closed?

topi-banana:
yes.

Issue #865, opened 2025-06-08
Server crash when throw items one by one continuously


sprinox:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11

### Server Software Version/Commit

0.1.0-dev+1.21.5 (1e8687a6)

### What happened?

```
thread 'tokio-runtime-worker' panicked at pumpkin-inventory\src\slot.rs:57:14:
Timed out while trying to acquire lock: Elapsed(())
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library\std\src\panicking.rs:695
   1: core::panicking::panic_fmt
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library\core\src\panicking.rs:75
   2: core::result::unwrap_failed
             at /rustc/05f9846f893b09a1be1fc8560e33fc3c815cfecb/library\core\src\result.rs:1704
   3: enum2$<core::result::Result<tokio::sync::mutex::MutexGuard<pumpkin_world::item::ItemStack>,tokio::time::error::Elapsed> >::expect
             at C:\Users\user\.rustup\toolchains\stable-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\result.rs:1061
   4: pumpkin_inventory::slot::Slot::get_cloned_stack::async_block$0<pumpkin_inventory::slot::NormalSlot>
             at .\pumpkin-inventory\src\slot.rs:55
   5: pumpkin_inventory::screen_handler::ScreenHandler::send_content_updates::async_block$0<pumpkin_inventory::player::player_screen_handler::PlayerScreenHandler>
             at .\pumpkin-inventory\src\screen_handler.rs:316
   6: pumpkin::entity::player::impl$1::tick::async_fn$0
             at .\pumpkin\src\entity\player.rs:598
   7: pumpkin::world::impl$1::tick::async_fn$0
             at .\pumpkin\src\world\mod.rs:413
   8: pumpkin::server::impl$0::tick::async_fn$0
             at .\pumpkin\src\server\mod.rs:583
   9: pumpkin::server::ticker::impl$0::run::async_fn$0
             at .\pumpkin\src\server\ticker.rs:30
  10: pumpkin::impl$2::new::async_fn$0::async_block$1
             at .\pumpkin\src\lib.rs:267
[...]

PR #902, opened 2025-06-15
Fix deadlock when throwing items (#865)

spr-equinox:
## Description

Fix deadlock when throwing items #865
`self.current_screen_handler.lock().await;` should do first otherwise cause deadlock in tick() (this thread lock stack and wait screen_handler. that thread lock screen_handler and wait stack)

In addition, i notice `screen_handler.get_slot_index` and `screen_handler.set_received_stack` don't use any async fn in them but are marked as async. Should i remove async mark of the two fn?

## Testing

### DIFF
--- a/pumpkin-inventory/src/screen_handler.rs
+++ b/pumpkin-inventory/src/screen_handler.rs
@@ -165,7 +165,7 @@ pub trait ScreenHandler: Send + Sync {
         }
     }
 
-    async fn set_received_hash(&mut self, slot: usize, hash: OptionalItemStackHash) {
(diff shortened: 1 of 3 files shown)


Snowiiii:
> In addition, i notice screen_handler.get_slot_index and screen_handler.set_received_stack don't use any async fn in them but are marked as async. Should i remove async mark of the two fn?

Yes

sprinox:
done

Snowiiii:
LGTM. Thank you @spr-equinox :+1: :hedgehog:

Issue #876, opened 2025-06-09
I can't see the items in the hotbar.


Q2297045667:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11 专业工作站版 Insider Preview

### Server Software Version/Commit

27871.1000

### What happened?

I can't see the items in the hotbar.

![Image](https://github.com/user-attachments/assets/dcc82873-1d6d-4964-ae5b-18f417276aae)

### To Reproduce

Sudden death without warning (even in creative mode)

![Image](https://github.com/user-attachments/assets/f90fe985-889c-4f36-8a3f-d8df1d932f95)

### Expected behavior

I can see the items.

PR #914, opened 2025-06-18
fix: Server and client inventories out of sync after respawn_player

AyaSanae:
## Description
After the player respawns, items that were originally in the backpack are not visible in the client's hotbar. However, they can still be placed by right-clicking.
This bug is mentioned in the issue: https://github.com/Pumpkin-MC/Pumpkin/issues/876


Snowiiii:
Looks good. Thank you @AyaSanae :+1: :goat:

PR #915, opened 2025-06-18
add fluid mixing

teknostom:
## Description

Added impls for all fluid mixings:

- basalt
- cobble
- stone
- obsidian

## Testing

Manual

### DIFF
--- a/pumpkin-data/src/block_direction.rs
+++ b/pumpkin-data/src/block_direction.rs
@@ -131,6 +131,16 @@ impl BlockDirection {
         ]
     }
 
+    pub fn flow_directions() -> [BlockDirection; 5] {
+        [
+            BlockDirection::Down,
+            BlockDirection::North,
+            BlockDirection::South,
+            BlockDirection::West,
+            BlockDirection::East,
+        ]
+    }
+
     pub fn is_horizontal(&self) -> bool {
         matches!(
             self,
--- a/pumpkin/src/block/fluid/lava.rs
+++ b/pumpkin/src/block/fluid/lava.rs
@@ -1,17 +1,69 @@
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use pumpkin_data::fluid::Fluid;
+use pumpkin_data::{
+    Block, BlockDirection,
+    fluid::{Falling, Fluid, FluidProperties, Level},
+};
 use pumpkin_macros::pumpkin_block;
 use pumpkin_util::math::position::BlockPos;
-use pumpkin_world::BlockStateId;
+use pumpkin_world::{BlockStateId, world::BlockFlags};
 
 use crate::{block::pumpkin_fluid::PumpkinFluid, entity::EntityBase, world::World};
 
 use super::flowing::FlowingFluid;
+type FlowingFluidProperties = pumpkin_data::fluid::FlowingWaterLikeFluidProperties;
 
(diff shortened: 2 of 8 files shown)


pumpkin-data/src/block_direction.rs
```diff
@@ -131,6 +131,16 @@ impl BlockDirection {
...
 
+    pub fn flow_directions() -> [BlockDirection; 5] {
+        [
+            BlockDirection::Down,
+            BlockDirection::North,
+            BlockDirection::South,
+            BlockDirection::West,
+            BlockDirection::East,
+        ]
+    }
```
Snowiiii, line 142: Why not just using `all()` ?

pumpkin/src/block/fluid/lava.rs
```diff
@@ -1,34 +1,88 @@
...
+        let below_is_soul_soil = world
+            .get_block(&block_pos.offset(BlockDirection::Down.to_offset()))
+            .await
+            == Block::SOUL_SOIL;
+        let is_still = world.get_block_state_id(block_pos).await == Block::LAVA.default_state_id;
+
+        for dir in BlockDirection::flow_directions() {
+            let neighbor_pos = block_pos.offset(dir.opposite().to_offset());
+            if world.get_block(&neighbor_pos).await.default_state_id
+                == Block::WATER.default_state_id
```
Snowiiii: why not just `world.get_block(&neighbor_pos).await == Block::WATER` ?

pumpkin-data/src/block_direction.rs
teknostom: From java source:

   public static final ImmutableList<Direction> FLOW_DIRECTIONS = ImmutableList.of(
      Direction.DOWN, Direction.SOUTH, Direction.NORTH, Direction.EAST, Direction.WEST
   );

If we have Up in the directions, then we get the wrong logic.

pumpkin/src/block/fluid/lava.rs
teknostom: Done

Snowiiii:
you also forgot to add 
```java
private void playExtinguishEvent(WorldAccess world, BlockPos pos) {
        world.syncWorldEvent(WorldEvents.LAVA_EXTINGUISHED, pos, 0);
    }
```
That's the reason the Sound and Particles are missing, look at 
`world.sync_world_event`

Snowiiii:
Also this seems like a bug

https://github.com/user-attachments/assets/a6201ab3-0698-4816-8909-e993a7df8524

teknostom:
Fixed world event, and placing liquids in liquids, and now water in the nether plays the correct sound. But we don't have a global random in world yet, so can't do full impl. Added a TODO for that.

teknostom:
Pretty sure CI is failing because of pumpkin_data issues

Snowiiii:
> Fixed world event, and placing liquids in liquids, and now water in the nether plays the correct sound. But we don't have a global random in world yet, so can't do full impl. Added a TODO for that.

We usually only use "Mojang's" Random in the Chunk generation, since this is the only place where we care about being 100% Vanilla, In other places you can use the `rand` crate, `rand::rng().random()`

Snowiiii:
Good Work @teknostom!. Thank you :D :parrot:

PR #908, opened 2025-06-15
implement wall mounting blocks

topi-banana:
## Description

- new block GrindstoneBlock
- implement WallMountingBlock

## Testing

Please follow our [Coding Guidelines](https://github.com/Pumpkin-MC/Pumpkin/blob/master/CONTRIBUTING.md#coding-guidelines)

PR #920, opened 2025-06-20
Drops loot when blocks are changed by get_state_for_neighbor_update

topi-banana:
## Description

https://github.com/Pumpkin-MC/Pumpkin/pull/908#issuecomment-2980293019

## Testing


pumpkin/src/world/mod.rs
```diff
@@ -1766,6 +1766,7 @@ impl World {
 
         if new_state_id != block_state.id {
             self.set_block_state(block_pos, new_state_id, flags).await;
+            block::drop_loot(self, &block, block_pos, true, block_state.id).await;
         }
     }
```
Snowiiii: That logic is a bad wrong, here is the vanilla code
```java
if (newState.isAir()) {
                if (!world.isClient()) {
                    world.breakBlock(pos, (flags & SKIP_DROPS) == 0, null, maxUpdateDepth);
                }
            } else {
                world.setBlockState(pos, newState, flags & ~SKIP_DROPS, maxUpdateDepth);
            }
   ```
   So its only dropped when the new state is air, and `SKIP_DROPS` is not given

Comment (topi-banana) [reply pumpkin/src/world/mod.rs]: In reality, the SKIP_DROPS flag is ignored, and only GameRule checks are performed.
(For example, in creative mode, destroying a block does not result in an item drop, but item drops caused by neighbor_update.)

File: pumpkin/src/world/mod.rs [hunk 2]
```diff
@@ -1786,7 +1786,12 @@ impl World {
             .await;
 
         if new_state_id != block_state.id {
-            self.set_block_state(block_pos, new_state_id, flags).await;
+            let flags = flags & !BlockFlags::SKIP_DROPS;
+            if Block::from_state_id(new_state_id) == Some(Block::AIR) {
```
Snowiiii: Instead of checking for block air. You should do it like Vanilla and get a block state and then check `is_air()` since there are multiple air block types

Snowiiii:
Looks good now. Thank you @topi-banana :+1: :lion:

PR #922, opened 2025-06-20
Implement Game Mode Swapping

WorriedArrow:
## Description
This allows the player to use F3+F4 to switch gamemodes, using the change_game_mode packet.

## Testing
Fully tested in-game in 1.21.6.


pumpkin/src/entity/player.rs
```diff
@@ -2055,6 +2055,10 @@ impl Player {
                 self.handle_confirm_teleport(SConfirmTeleport::read(payload)?)
                     .await;
             }
+            SChangeGameMode::PACKET_ID => {
+                self.set_gamemode(SChangeGameMode::read(payload)?.game_mode)
+                    .await;
```
Snowiiii: This is really insecure since any Client can now just send the Gamemode packet and change his gamemode (including Hacked Clients for example), We should instead do it like Vanilla and check if the player has at least the permission level 2

WorriedArrow: Completely slipped my mind lol! Should be fixed now.

Snowiiii:
I would also prefer if you would create a new function in play.rs like for every other packet instead handling it directly

Snowiiii:
you need to run `cargo fmt` to apply formatting

WorriedArrow:
ah whoops, missed that, running now

Snowiiii:
Looks good. Thank you @WorriedArrow :D :monkey:

PR #923, opened 2025-06-21
Implement BlockSetType And Improve Door

sprinox:
## Description

Fix a panic caused by place a door next to a different type of door

Port `BlockSetType` from java(except `SoundType`)
Change the way to registering door blocks so that we dont need to query block id each time to find out the type of the door
Add `fn play_sound_expect`. Based on the Java code, we do not send a packet to the player entity passed into `playSound`.
Now we can play the sound of door opening and closing correctly.

Todo:

The door toggle handling code differs somewhat from the Java implementation. In Java, it seems that the update of the other half of the door is handled in `neighborChanged`.

Additionally, I suspect that the server may crash when interacting with just one half of the door, as the code doesn't appear to account for this special case.

After adding BlockSetType, trapdoor.rs can be updated too.

------------

There are some things I'm a bit confused about.
Why use a `HashMap<Vec<String>, Arc<dyn PumpkinBlock>>` to store blocks and fluids, instead of a `HashMap<String, Arc<dyn PumpkinBlock>>` where the strings in the `Vec` are individually inserted into the map?
In `get_pumpkin_block`, the current approach uses:

```rust
self.blocks.iter().find_map(|(ids, pumpkin_block)| {
    ids.contains(&format!("minecraft:{}", block.name))
        .then_some(pumpkin_block)
})
```

which seems to be linear in complexity.

## Testing

tested manually.

### DIFF
--- a/pumpkin/src/block/blocks/block_set_type.rs
+++ b/pumpkin/src/block/blocks/block_set_type.rs
@@ -9,26 +9,28 @@ pub enum PressurePlateSensitivity {
 }
 
 #[derive(Copy, Clone)]
+#[allow(dead_code)]
 pub struct BlockSetType {
-    pub name: &'static str,
+    // pub name: &'static str, unused
     pub can_open_by_hand: bool,
     pub can_open_by_wind_charge: bool,
     pub can_button_be_activated_by_arrows: bool,
     pub pressure_plate_sensitivity: PressurePlateSensitivity,
     // TODO pub soundType: SoundType,
-    pub door_close: Sound, //java SoundEvent
-    pub door_open: Sound, //java SoundEvent
-    pub trapdoor_close: Sound, //java SoundEvent
-    pub trapdoor_open: Sound, //java SoundEvent
+    pub door_close: Sound,               //java SoundEvent
+    pub door_open: Sound,                //java SoundEvent
+    pub trapdoor_close: Sound,           //java SoundEvent
+    pub trapdoor_open: Sound,            //java SoundEvent
     pub pressure_plate_click_off: Sound, //java SoundEvent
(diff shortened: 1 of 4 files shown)


pumpkin/src/block/blocks/block_set_type.rs
```diff
@@ -0,0 +1,320 @@
...
+        // soundType: SoundType.METAL,
+        door_close: Sound::BlockIronDoorClose,
+        door_open: Sound::BlockIronDoorOpen,
+        trapdoor_close: Sound::BlockIronTrapdoorClose,
+        trapdoor_open: Sound::BlockIronTrapdoorOpen,
+        pressure_plate_click_off: Sound::BlockMetalPressurePlateClickOff,
+        pressure_plate_click_on: Sound::BlockMetalPressurePlateClickOn,
+        button_click_off: Sound::BlockStoneButtonClickOff,
+        button_click_on: Sound::BlockStoneButtonClickOn,
+    };
```
Snowiiii, line 81: Did you really implemented them all by Hand ?, Why not just an extractor, this would make it also safe for the future when this gets updated

Snowiiii:
I'm also not sure why you implemented `BlockSetType` when it currently does nothing ?

sprinox:
> I'm also not sure why you implemented `BlockSetType` when it currently does nothing ?

I want to properly handle the sound effects for various materials of doors, fence gates, trapdoors, etc.
I noticed that the Java code uses BlockSetType to determine them, so I started working on implementing it.
But later I realized that doors actually only use two variables: door_close and door_open.
So I'm wondering — should I just store the sound-related variables directly in the DoorBlock instead?

PR #925, opened 2025-06-21
Change BlockRegistry to use single String instead of Vec<String>

sprinox:
## Description

In `get_pumpkin_block`, the current approach uses:

```rust
self.blocks.iter().find_map(|(ids, pumpkin_block)| {
    ids.contains(&format!("minecraft:{}", block.name))
        .then_some(pumpkin_block)
})
```

which seems to be linear in complexity.

`HashMap<String, Arc<dyn PumpkinBlock>>` instead of `HashMap<Vec<String>, Arc<dyn PumpkinBlock>>` may improve performance

## Testing

tested in game and it seems everything works correctly

### DIFF
--- a/pumpkin/src/block/registry.rs
+++ b/pumpkin/src/block/registry.rs
@@ -26,8 +26,8 @@ pub enum BlockActionResult {
 
 #[derive(Default)]
 pub struct BlockRegistry {
-    blocks: HashMap<Vec<String>, Arc<dyn PumpkinBlock>>,
-    fluids: HashMap<Vec<String>, Arc<dyn PumpkinFluid>>,
+    blocks: HashMap<String, Arc<dyn PumpkinBlock>>,
+    fluids: HashMap<String, Arc<dyn PumpkinFluid>>,
 }
 
 #[async_trait]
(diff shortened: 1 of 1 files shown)


Snowiiii:
Now instead of storing multiple names in one row you just push them multiple times into the Vec, I don't think this makes it any more efficient, I rather think this is actually more inefficient since it now has to push it multiple times and create more allocations then needed

CuzImClicks:
```
#![feature(test)]
extern crate test;

use pumpkin::block::default_registry;
use test::{bench, Bencher};

//#[bench]
//fn bench_get_pumpkin_block_old(b: &mut Bencher) {
//    let registry = default_registry();
//    b.iter(|| {
//        registry.blocks.iter().find_map(|(ids, pumpkin_block)| {
//            ids.contains(&format!("minecraft:{}", "dirt"))
//                .then_some(pumpkin_block)
//        })
//    });
//}

#[bench]
fn bench_get_pumpkin_block_new(b: &mut Bencher) {
    let registry = default_registry();
    b.iter(|| {
        registry.blocks.get(&format!("minecraft:{}", "dirt"))
    });
}
```

```
running 1 test
test bench_get_pumpkin_block_new ... bench:          59.31 ns/iter (+/- 3.04)
```

```
running 1 test
test bench_get_pumpkin_block_old ... bench:       3,879.57 ns/iter (+/- 688.45)
```

fun :D 
`Finished `bench` profile [optimized + debuginfo] target(s) in 9m 41s`
`Finished `bench` profile [optimized + debuginfo] target(s) in 11m 15s`

Snowiiii:
Oh okay, didn't expect to see such an performance change, but yeah makes sense since it don't need to loop trough the blocks. Thank you @spr-equinox :+1: :fish:

Snowiiii:
The only thing i would prefer if you would choose a better branch and PR name

PR #929, opened 2025-06-22
Add Gamerule

topi-banana:
## Description

- Add GameRules in LevelData
- https://github.com/Pumpkin-MC/Extractor/pull/20

## Testing


topi-banana:
it would be better to use `HashMap<GameRule, GameRuleValue>` for GameRuleRegistry?

Snowiiii:
> it would be better to use `HashMap<GameRule, GameRuleValue>` for GameRuleRegistry?

I think yes, it would be better

topi-banana:
But it is not type safe.
for example, can put a boolean into randomTick.

Snowiiii:
Or actually i think its fine, if there are any problems with this in the future we can change this. Thank you @topi-banana :deer:

PR #932, opened 2025-06-23
Add `/tick` command

OmarAfet:
## Description

An implementation of the `/tick` command, used Minecraft source code with Mojang mappings along the way and `/tick` documentation here: https://minecraft.wiki/w/Commands/tick

## Testing

Manual testing


Snowiiii:
Looks good, Good Job @OmarAfet, Thanks :100: :rooster:

PR #935, opened 2025-06-24
Add random block ticks and crop impls

teknostom:
## Description

Added support for random block ticks and fixed cacti.

Also implemented crop growing (carrot, beetroot, potatoes, torch_flower, wheat).

## Testing

Manual


teknostom:
I need to checkout if this can be done better, I am getting 40ms ticks on dev build compared to 0.5 ms without the changes.

teknostom:
Got it down to 5/3ms on dev/release respectivley.

pumpkin/src/world/mod.rs
```diff
@@ -565,6 +565,29 @@ impl World {
...
+            }
+        }
+        let third = Instant::now();
+        log::info!(
+            "Gathering took {}ms, processing {} random ticks took {}ms, total: {}ms",
+            second.duration_since(first).as_millis(),
+            random_ticks_len,
+            third.duration_since(second).as_millis(),
+            third.duration_since(first).as_millis()
+        );
```
Snowiiii: seems like you forgot to remove that ?

teknostom: Yes, removed

teknostom:
I don't want to merge random ticks with scheduled ticks for the following reasons. If someone wants to do it, it's fine to steal the PR. But I will only do merge conflict resolutions from now on.

Scheduled ticks never write lock until in <pumpkin_block>::on_scheduled_tick, and the chunks are RwLock, so they can have multiple readers.
And by merging this, we are not only moving away from vanilla logic, but we are also worsening performance for basically no gain. To do it, we would need to register what chunk coord each scheduled tick is in, then random tick every chunk that has scheduled ticks, then see what chunks wern't random ticked, and random tick those. And for seemingly no gain.
To have the write locks from both used in the same lock, we would also have to send in the locked chunks to the on_scheduled_tick function.

teknostom:
The random ticks are now down to sub millisecond levels per tick. Can't do much else without perf, which does not work on wsl.
image

Snowiiii:
@teknostom I'm sorry for the conflicts

teknostom:
> @teknostom I'm sorry for the conflicts

No problem, it improves the performance by a lot, so no complaints here. Will fix them tonight.

teknostom:
This still seems to be causing deadlocks, so I will make it draft.

teknostom:
Ticks are now at 2mspt, on release. Seems to be without deadlocks now

Snowiiii:
Looks good. Thank you @teknostom :D 🐞

PR #939, opened 2025-06-27
Arc<Mutex<ItemStack>> for the Item argument of PumpkinBlock::use_with_item

topi-banana:
## Description

uses items on blocks may consume items, so it is better to provide ItemStack as a mutable.

## Testing


4lve:
Nice, if you have time to could you make a struct for the args instead of it being super many of them, would make it easier to update in the future

topi-banana:
ok, I think it would be better to open another pull request

pumpkin/src/block/blocks/tnt.rs
```diff
@@ -60,11 +62,13 @@ impl PumpkinBlock for TNTBlock {
...
+        item_stack: &Arc<Mutex<ItemStack>>,
         _server: &Server,
         _world: &Arc<World>,
     ) -> BlockActionResult {
-        if *item != Item::FLINT_AND_STEEL || *item == Item::FIRE_CHARGE {
+        if item_stack.lock().await.item != &Item::FLINT_AND_STEEL
+            || item_stack.lock().await.item == &Item::FIRE_CHARGE
+        {
             return BlockActionResult::Continue;
         }
```
TheDarkSword, line 72: This code may cause a deadlock due to multiple .lock().await calls on the same Mutex.
Each call to lock().await acquires the mutex asynchronously, but the first lock isn't explicitly released before the second call. Since the MutexGuard is not stored in a variable and the .await happens inline, it's not guaranteed that the first lock is dropped before the second one is attempted. (It's not guaranteed becase you don't know how the compiler will optimize the code, and if it decide to drop the lock or not).

Maybe is better to acquire the lock once, store the guard in a variable, and extract the item before doing comparisons.

PR #951, opened 2025-06-28
Fix Resetting of Block States

CuzImClicks:
## Description

For some reason we were discarding the rest of the block state values when getting the new block state... which meant that for example the campfire rotation is reset.

## Testing

Ingame


Snowiiii:
Looks good, But please use more descriptive titles and description next time, because its not all block states and only about ignition, Thanks @CuzImClicks :1234: :cat2:

PR #869, opened 2025-06-08
Add crafting support

Guibi1:
## Description

Adds full crafting support.

## Testing

Via the player's 2x2 crafting grid. Code should work the same way in a crafting table.

Please follow our [Coding Guidelines](https://github.com/Pumpkin-MC/Pumpkin/blob/master/CONTRIBUTING.md#coding-guidelines)

PR #952, opened 2025-06-28
Feat: crafting system

zly2006:
## Description

Crafting table & inventory crafting, based on #869 

Thanks to @Guibi1 

## Testing

I am sure it works fine without dedlocks. However, many names in inv system are misleading and missing documents. I tried to add some documents but I cannot promise it works fine in all cases. 

This PR is ready to merge, but may need to test for edge cases.


Snowiiii:
Wow, Awesome. Thanks @Guibi1 and @zly2006 :100: :frog:

PR #959, opened 2025-06-29
improve fill command

spr-equinox:
## Description

improve fill command
fix the issue that filling blocks won't update neighors
add strict mode

reference: <https://minecraft.wiki/w/Commands/fill>

support replace with filter (without nbt support)

TODO: 
translate the command error message, which need to change `CommandError::GeneralCommandIssue(String)` to `CommandError::GeneralCommandIssue(TextComponent)`
check isInWorldBounds and throw argument.pos.outofbounds

**the check of block tag is extremely slow**

## Testing


pumpkin/src/command/args/block.rs
```diff
@@ -71,3 +70,91 @@ impl<'a> FindArg<'a> for BlockArgumentConsumer {
...
+    ) -> Result<Option<Vec<CommandSuggestion>>, CommandError> {
+        Ok(None)
+    }
+}
+
+impl DefaultNameArgConsumer for BlockPredicateArgumentConsumer {
+    fn default_name(&self) -> &'static str {
+        "filter"
+    }
+}
```
Snowiiii, line 130: Is this used multiple times in other Commands?, Because we usally put our args struct into the args folder

sprinox: /clone /execute use it too

<https://minecraft.wiki/w/Argument_types#block_predicate>

sprinox:
i have apply some translations and fix the build error

Snowiiii:
Looks good. Thank you @spr-equinox 🐈

Issue #892, opened 2025-06-13
Blocking CPU bound work regression


kralverde:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Debian

### Server Software Version/Commit

fd426273

### What happened?

In the last few weeks, a regression was introduced to due removing the rayon task group for cpu bound tasks. CPU bound tasks will now clog the tokio work pool and prevent async operations from taking place (users will time out, commands cannot run, packets not handled, etc.) This is especially noticeable on machines with fewer cores.

A fix should re-introduce a kind of work pool for CPU bound tasks (probably rayon again) and maybe a generic abstraction for communicating work from CPU bound thread pool land to async task land and vice versa (originally tokio mpsc channels were used which worked fine).

### To Reproduce

Load lots of (new) chunks and try to send chat messages for example.

### Expected behavior

Packets should be handled while chunks load.

TheDarkSword:
Is that hard to reimplement?

kralverde:
Relatively, no

Snowiiii:
Should be fixed now

PR #966, opened 2025-06-30
Reimplement Rayon for chunk generation

TheDarkSword:
## Description
Use a thread‑local Tokio runtime to drive the async fn generate_chunk inside Rayon worker threads. Each Rayon thread now initializes its own single‑threaded Tokio executor once, then calls block_on to run the generate_chunk future. This preserves Rayon’s efficient work‑stealing for CPU‑bound chunk generation while ensuring all internal .await calls execute correctly on a real async runtime—without rebuilding or contending over a global Tokio executor.

Issue: #892

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ edition = "2024"
 
 
 [profile.dev]
-opt-level = 1
+opt-level = 0
 
 [profile.release]
 lto = true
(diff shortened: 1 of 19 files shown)


tn-lorenz:
I think Alvsch was already working on a state-machine based way to do this, but I might be tripping

PR #970, opened 2025-06-30
refactor: cleanup Player::handle_player_action

Alvsch:
## Description
- Previously when the player break a block they can't reach, the server didn't send a sync to fix the client prediction.
- Split `Player::handle_player_action` into multiple functions and remove code duplication

## Testing

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -11,6 +11,7 @@ use crossbeam::atomic::AtomicCell;
 use log::warn;
 use pumpkin_world::chunk::{ChunkData, ChunkEntityData};
 use pumpkin_world::inventory::Inventory;
+use pumpkin_world::world::BlockFlags;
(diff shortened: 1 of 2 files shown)


pumpkin/src/entity/player.rs
```diff
@@ -1172,6 +1174,144 @@ impl Player {
...
+            return;
+        }
+
+        match status {
+            Status::StartedDigging => self.handle_started_digging(location, server).await,
+            Status::CancelledDigging => self.handle_cancelled_digging(location).await,
+            Status::FinishedDigging => self.handle_finished_digging(location, server).await,
+            _ => unreachable!(),
+        }
+    }
```
Snowiiii: I don't like this function, especially the `unreachable`, Instead you could just call all the functions directly

Snowiiii:
Feel free to reopen

PR #971, opened 2025-06-30
PluginAPI / Added useful helper functions for `Player`

tn-lorenz:
## Description
This pr aims to add useful helper functions to the `Player` struct inside `entity`, which mimic the ones familiar from bukkit and its derivatives.

## Testing
None :O)

Edit: jk, I have these exact helpers in my plugin implemented for `Arc<Player>` and they all work fine

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -8,6 +8,7 @@ use std::time::{Duration, Instant};
 
 use async_trait::async_trait;
 use crossbeam::atomic::AtomicCell;
+use futures::future::join_all;
 use log::warn;
 use pumpkin_world::chunk::{ChunkData, ChunkEntityData};
 use pumpkin_world::inventory::Inventory;
@@ -18,7 +19,7 @@ use uuid::Uuid;
 use pumpkin_config::{BASIC_CONFIG, advanced_config};
 use pumpkin_data::damage::DamageType;
 use pumpkin_data::entity::{EffectType, EntityPose, EntityStatus, EntityType};
-use pumpkin_data::item::Operation;
+use pumpkin_data::item::{Item, Operation};
 use pumpkin_data::particle::Particle;
 use pumpkin_data::sound::{Sound, SoundCategory};
 use pumpkin_data::tag::Tagable;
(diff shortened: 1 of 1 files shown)


Snowiiii:
tbh this feels a bit pointless, these are just wrappers ?

tn-lorenz:
I think (especially since Pumpkin is focussed around pertaining to the java minecraft server) it's in the interest of the project to have good, and by that I mean: easy to use, familiar, API design. So, I don't see a problem with having wrapper functions, if that means no slowdown + plugin developers who are coming from spigot/paper can easily understand it. This is especially important because we want Pumpkin to be adopted as easily and quickly as possible, in order to have it grow in popularity.
Consider this: We are implementing features with 1:1 logic from an existing backend AND there are A LOT of people on the dc who are already yearning for scripting support (because it is easier), so...I frankly don't get why we shouldn't merge this o.O

But yeah I kinda need to fix the todo's 

Edit: Also, plugins will get bloated if devs have to import many different things, just to acces stuff, that manage player fields anyway and especially with the dependency tree of pumpkin being kinda messy, it can be tough for newcomers (like me) to find what they need.

unschlagbar:
thats a good point

CuzImClicks:
In all honesty, this feels more like a documentation/tutorial problem...

Snowiiii:
> I think (especially since Pumpkin is focussed around pertaining to the java minecraft server) it's in the interest of the project to have good, and by that I mean: easy to use, familiar, API design. So, I don't see a problem with having wrapper functions, if that means no slowdown + plugin developers who are coming from spigot/paper can easily understand it. This is especially important because we want Pumpkin to be adopted as easily and quickly as possible, in order to have it grow in popularity. Consider this: We are implementing features with 1:1 logic from an existing backend AND there are A LOT of people on the dc who are already yearning for scripting support (because it is easier), so...I frankly don't get why we shouldn't merge this o.O
> 
> But yeah I kinda need to fix the todo's
> 
> Edit: Also, plugins will get bloated if devs have to import many different things, just to acces stuff, that manage player fields anyway and especially with the dependency tree of pumpkin being kinda messy, it can be tough for newcomers (like me) to find what they need.

Bukkit has so many functions because the player is a interface, I do not want to clutter everything with functions and want to make use of direct variables has, I know that you are coming from Java and it is coming to have methods everything called "get_..." or "set_...",  They can make sense if there is actual logic involved, In this case these are just Wrapper functions which can be easily implemented by  Developers.
Pumpkin was never build with the Idea of being an 1:1 Bukkit replacement and i dislike many things about Bukkit's API

I also want to say that things will change before Pumpkin v1.0 release and making such "helper/wrapper" functions should be only considered in the end phase

PR #977, opened 2025-07-01
support dropper

sprinox:
## Description

dropper now can drop item and be placed with correct facing , but can't add item to container(todo)

## Testing

### DIFF
--- a/pumpkin-world/src/block/entities/mod.rs
+++ b/pumpkin-world/src/block/entities/mod.rs
@@ -13,8 +13,8 @@ use pumpkin_util::math::position::BlockPos;
 use sign::SignBlockEntity;
 
 use crate::{
-    block::entities::chiseled_bookshelf::ChiseledBookshelfBlockEntity, dropper::DropperBlockEntity,
-    inventory::Inventory, world::SimpleWorld,
+    block::entities::chiseled_bookshelf::ChiseledBookshelfBlockEntity,
+    block::entities::dropper::DropperBlockEntity, inventory::Inventory, world::SimpleWorld,
 };
 
 pub mod barrel;
--- a/pumpkin/src/block/blocks/redstone/dropper.rs
+++ b/pumpkin/src/block/blocks/redstone/dropper.rs
@@ -1,7 +1,7 @@
 use crate::block::blocks::redstone::block_receives_redstone_power;
 use crate::block::pumpkin_block::{
     NormalUseArgs, OnNeighborUpdateArgs, OnPlaceArgs, OnScheduledTickArgs, OnStateReplacedArgs,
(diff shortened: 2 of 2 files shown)


topi-banana:
Does not work with weak power

sprinox:
the arg of is_emitting_redstone_power in block_receives_redstone_power looks incorrect. i have fixed it.

image

pumpkin-world/src/block/entities/dropper.rs
```diff
@@ -0,0 +1,132 @@
...
+        // this.unpackLootTable(null);
+        let mut ret = None;
+        for i in 0..self.items.len() {
+            let item = self.items[i].lock().await;
+            if !item.is_empty() && rng().random_range(0..=i) == 0 {
+                ret = Some(item);
+            }
+        }
+        ret
+    }
```
topi-banana: Sometimes it returns None even though the slot is not empty.

```rust
pub async fn get_random_slot(&self) -> Option<Arc<Mutex<ItemStack>>> {
    let mut non_empty_items = Vec::new();
    for item in &self.items {
        if !item.lock().await.is_empty() {
            non_empty_items.push(Arc::clone(item));
        }
    }

    non_empty_items.into_iter().choose(&mut rand::rng())
}
```

sprinox: solved

pumpkin-world/src/block/entities/dropper.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    async fn get_stack(&self, slot: usize) -> Arc<Mutex<ItemStack>> {
+        self.items[slot].clone()
+    }
+
+    async fn remove_stack(&self, slot: usize) -> ItemStack {
+        let mut removed = ItemStack::EMPTY;
+        let mut guard = self.items[slot].lock().await;
+        std::mem::swap(&mut removed, &mut *guard);
```
Snowiiii, line 112: this looks way too complicated, why not just 
`*self.items[slot].lock().await = ItemStack::EMPTY`

sprinox: i copied the code from barrel

it need to return the removed  stack removed

Snowiiii: > i copied the code from barrel
> 
> it need to return the removed stack removed

Oh, Than a much better solution would be to use `remove(slot)`, this will give you also the removed value, I would also appreciate if you would change the other implementation than as well

sprinox: sorry i didn't find `remove(slot)`, and it seems all remove_stack implement in that way

Snowiiii:
LGTM. Thank you @spr-equinox :100: 🐳

PR #998, opened 2025-07-02
Add anvil block rotation

tom-devv:
## Description
Add's correct rotation to anvil blocks when placed

## Testing

Manual testing, try all 3 types of anvil in all 4 directions.


teknostom:
You should be able to use:

```rust
impl BlockMetadata for AnvilBlock{
    fn namespace(&self) -> &'static str {
        "minecraft"
    }

    fn ids(&self) -> &'static [&'static str] {
        get_tag_values(RegistryKey::Block, "minecraft:anvil").unwrap()
    }
}
```
Instead of these:
```rust
#[pumpkin_block("minecraft:anvil")]
#[pumpkin_block("minecraft:chipped_anvil")]
#[pumpkin_block("minecraft:damaged_anvil")]
```

That way you don't have to extract the logic to a function, you just use put the function in the general impl.

CuzImClicks:
LGTM! 👍

Snowiiii:
Looks good. Thank you @tom-devv :D 🦒

PR #1001, opened 2025-07-03
fix: Can't load standard library when using VSCode on NixOS

HttpRafa:
## Description
When using NixOS and VS Code with the `flake.nix` file, the Rust Analyzer extension cannot locate the Rust standard library. This pull request updates `flake.nix` so that it always installs the standard library, which resolves the issue. In addition, I have updated the Nix flake to the latest version.

## Testing
Before this change, I encountered the error:
`ERROR: can't load standard library; try installing rust-src sysroot_path=/nix/store/f293vck2j3zf8rb8phc5zb08lsvihsff-rust-default-1.87.0`

After applying the change, the error no longer appears.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -19,11 +19,11 @@
         "nixpkgs-lib": "nixpkgs-lib"
       },
       "locked": {
-        "lastModified": 1749398372,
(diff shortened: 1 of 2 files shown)


Snowiiii:
Oh okay. Thank you @HttpRafa 🦅

PR #1007, opened 2025-07-04
Fix wall mechanics 

tom-devv:
## Description
This aims to fix wall mechanics as some were not working. From my testing this has fixed all issues that I could create and walls should now behave like vanilla again. 

This also fixes fence gates being attatched to walls properly when placed adjacent

## Testing

Here are some vanilla mechanics: [video](https://img.million.beer/u/dP5Cln.mp4) this PR implements that and each case (besides pistons as they do not work) has been tested


Snowiiii:
Looks good. Thank you @tom-devv :D 🐡

Issue #996, opened 2025-07-02
none armor items in armor slots


Commandcracker:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.7 (2e643cde)

### What happened?

You can place anything in the armor slots.
![Image](https://github.com/user-attachments/assets/240a3943-1145-4672-a4cb-b7937660ec5c)

### To Reproduce

Place something that is not armor in ther armor slot.

### Expected behavior

not being able to do it.

Commandcracker:
only works in survival

Lesiuk:
https://github.com/Pumpkin-MC/Pumpkin/pull/1008

PR #1008, opened 2025-07-04
fixes "none armor items in armor slots #996"

Lesiuk:
## Testing

Manual


pumpkin-inventory/src/slot.rs
```diff
@@ -271,11 +272,16 @@ impl Slot for ArmorSlot {
...
-    async fn can_insert(&self, _stack: &ItemStack) -> bool {
-        // TODO: return this.entity.canEquip(stack, this.equipmentSlot);
-        true
+    async fn can_insert(&self, stack: &ItemStack) -> bool {
+        match self.equipment_slot {
+            EquipmentSlot::Head(_) => stack.is_helmet(),
+            EquipmentSlot::Chest(_) => stack.is_chestplate() || stack.item == &Item::ELYTRA,
+            EquipmentSlot::Legs(_) => stack.is_leggings(),
+            EquipmentSlot::Feet(_) => stack.is_boots(),
+            EquipmentSlot::MainHand(_) => true,
```
TonimatasDEV: I think `EquipmentSlot::MainHand(_) => true` is unnecessary because you have the `_ => true` in the next one, you can simply delete it.

pumpkin-inventory/src/slot.rs
```diff
@@ -271,11 +272,15 @@ impl Slot for ArmorSlot {
         self.set_stack_no_callbacks(stack).await;
     }
 
-    async fn can_insert(&self, _stack: &ItemStack) -> bool {
-        // TODO: return this.entity.canEquip(stack, this.equipmentSlot);
-        true
+    async fn can_insert(&self, stack: &ItemStack) -> bool {
+        match self.equipment_slot {
+            EquipmentSlot::Head(_) => stack.is_helmet(),
```
Commandcracker: You'r missing the most important item the carved PUMPKIN!

Commandcracker (changes requested):

pumpkin-inventory/src/slot.rs
Commandcracker: Skulls also missing (minecraft:skulls)

Commandcracker:
It would also be nice if you could add that you equip an item on Shift + Left/Right Click.
(As a former pvp player I really need this :D)

Commandcracker (changes requested):
When you shift click equipment in vanilla minecraft then you are not able to equip stacked skulls. it will equip one and move the other ones in your hotbar. Everything else looks good

pumpkin-inventory/src/player/player_screen_handler.rs
```diff
@@ -75,6 +75,42 @@ impl PlayerScreenHandler {
...
+            let armor_slot = self.get_slot(armor_slot_index).await;
+            if !armor_slot.has_stack().await && armor_slot.can_insert(&one_item).await {
+                armor_slot.set_stack(one_item).await;
+                return true;
+            }
+        }
+        false
+    }
+
+    async fn handle_inventory_move(&mut self, slot_index: i32, slot_stack: &mut ItemStack) -> bool {
```
zly2006, line 96: please no. we are trying to keep the same code structure as vanilla. I suppose you should only modify `can_insert` function. If there are any reasons you have to make this change please tell me.

zly2006 (changes requested):

pumpkin-inventory/src/player/player_screen_handler.rs
Lesiuk: What do you mean by as vanilla? I cannot split code into separate methods? can_insert will not handle quick_move.

zly2006: > What do you mean by as vanilla?

Currently we are copying the code structure, functions from vanilla. Making this change may be confusing. In my opinion refactoring should be done after the inventory system is complete.

> I cannot split code into separate methods? can_insert will not handle quick_move.

You can add check for `can_insert` in `insert_item`, I suppose it is incomplete so the check is missing. I suggest to check vanilla code.

Snowiiii:
Looks good. Thank you @Lesiuk 🦏

PR #1009, opened 2025-07-05
Skip unloadable plugins and log loaded plugins

Commandcracker:
## Description

- When loading all Plugins make it so that when try_load_plugin fails that other plugins will still be loaded.
- Log loaded plugins

## Testing
grafik


Snowiiii:
please be more specific with the PR title, also i'm not sure if we want to handle this there

Snowiiii:
Looks good. Thank you @Commandcracker :+1: 🐸

PR #1011, opened 2025-07-05
Implements some AI Goals for Zombie

TheDarkSword:
## Description
This PR implement some AI goals for Zombie


TheDarkSword:
There are 3/4 goals that are not implement yet, I will do them in another PR.
This PR will refactor the MobEntity with new Trait (because every mob can override some methods) and implemented some goals for Zombie

TheDarkSword:
I will implement Path and collisions in another PR

Snowiiii:
For some reason Zombies don't do anything now
image

TheDarkSword:
Is missing goal priority

Snowiiii:
now the Zombie sometimes turns but its still buggy, he also does not move which he is on master
image

TheDarkSword:
I will check why he don’t move, but the cause of “buggy” rotation is because (I think) that is not yet implemented the dirty flag that send the new location and velocity of an Entity, then the LookAroundGoal and LookEntityGoal sends both the rotation packet and then the zombie is buggy. I will remove the packet send in LookAroundGoal (in vanilla there is nothing about sending packet), then it will work correctly when dirty is implemented

TheDarkSword:
Now moving works, but it don't turn around anymore, I think because is not yet implemented the position (physics) of entities, then at numeric level (yaw, pitch, ecc...) is all ok, but visually no

Snowiiii:
LGTM. Thank you @TheDarkSword :D 🐩

PR #1012, opened 2025-07-05
Feature: `CombatProfile` support

tn-lorenz:
## Description
This pr aims to add the ability to enable classic pvp mechanics and alter their values (for custom knockback) in the `features.toml`.
I will also add the ability to manipulate these mechanics from the plugin API, but there's lots of stuff I still need to add.

## Testing
Will be tested in game.


ghost:
Respectfully, I think this is a really bad idea - implementing and maintaining vanilla minecraft is already a large, difficult task. This feature opens the door to adding more non-vanilla features that will take even more time to create and keep up to date, or otherwise bloat Pumpkin and impact performance. At most, I think there should be support for modifying these values with a plugin, but not more.

Edit:  As an example, features have already broken/bugged when moving to newer versions, meaning the project clearly doesn't have the manpower to test & maintain just vanilla by itself.

tn-lorenz:
@Miasmusa Respectfully, this pr is not really complicated, so maintaining the changes shouldn't be hard. The classic combat mechanics in essence only change how knockback is calculated, which is easily done, as you can see by the code I already pushed. The only thing that might apply to your logic, is that making a setting that automatically makes swords blockable might be kinda weird to implement, but again, it is not hard to maintain.

This is also a feature many people want.

ghost:
> This is also a feature many people want.

I think this is the crux of the issue. There are many features many people want, but in my opinion these will only suffocate the project. You mention that the PR is simple, which I agree with, but that doesn't gurantee other features will be - or that the build-up of these features as a whole will still be easy to track and maintain. 

Is there a reason this can't/shouldn't be implemented in a plugin?

tn-lorenz:
@Miasmusa I don't think this pr increases the chance of making the project break down, because all I did is making a system that calls another `impl` of a `fn` depending on what `CombatProfile` is loaded. These `fn`s **exactly** replicate the behaviour that was already there (BTW: without removing the `knockback` fn from `entity > mod.rs`), so depending on _what_ changes, it can cleanly be implemented in the original `knockback` fn, which is now a kind of wrapper, or, in case the knockback logic itself changes (which I highly doubt will happen in the foreseable future), directly inside the call in `combat.rs`.

There is also a graceful fallback to default values in case somebody accidentally uses the `features.toml` incorrectly (granted, I need to make the error handling there better, but as this pr is still a draft, obviously it will get better before I mark it as ready for review).

Edit: I forgot to reply:
> Is there a reason this can't/shouldn't be implemented in a plugin?

Well yes, in the same way implementing support for multiple protocols is hacky, slow and more error-prone when doing it in a plugin, doing this in a plugin is hacky, slower and more error-prone.
There is a reason why traditionally spigot has been forked to give the ability to have custom knockback, just look at:

- Pandaspigot
- FlamePaper
- TacoSpigot

deslunes:
Hello. If I may add something. I understand @Miasmusa 's point, that implementing non-vanilla behaviors would lead to other people asking for non vanilla features. But I think we should also consider that:
- This PR aims to bring old vanilla behavior, so it's not really "implementing a non-vanilla behavior"
- I think that it might not be a good idea to let the PluginAPI access this kind of things, what happens if multiple plugins customize the knockback logic ? It might lead to other, bigger, issues.
- If we take example from Paper, the only plugin that does something similar is OldCombatMechanics, which uses a bunch of events to track the player's inventory. It's not easier to provide this inside of pumpkin, it's also more efficient.
- People might end up forking Pumpkin in order to implement CombatProfiles, because it may not fit in the scope of Pumpkin. Which might lead to a trend of forking pumpkin. I personally think it would be better if we could all build nice things around one and single project: Pumpkin; instead of everyone forking it and tunning it to fit their use case. That happened for Paper since CraftBukkit > Spigot > Paper > Pufferfish > Purpur > LeafMC. It also happens for Minestom, since it's a framework with no plugin system, they can't share around a big project. 

My point is, not doing it might cause bigger trouble than doing it. Plus, as @tn-lorenz said, it's a feature that defaults to the vanilla behavior, that's async, simple logic, and will be easy to maintain/ask little to no maintenance.

Snowiiii:
This is also already open for way too long, And using something like
```
/// The type of combat mechanics that are used by default. Options: "Legacy" (MC 1.7.10), "Classic" (MC 1.8), "Modern" (Current)
pub combat_type: String,
```
Is really bad practice, I think i already once said in the Discord that we should leave custom features to Plugins, The only thing i plan would be to expose all the "magic" Knockback numbers in the config and also allow Plugins to replace the Knockback with a custom implementation

PR #1015, opened 2025-07-05
fix: /tp command server crash

GreenedDev:
Set bounds for /tp command arguments

## Description
if you executed command /tp 1000000000000000000000000000000 100 10000
a few times, server would crash.
this bound check fixes the issue.
## Testing
to replicate the crash execute command
/tp 1000000000000000000000000000000 100 10000 a few times, you may need to change coordinates on each command.


Snowiiii:
Hey @GreenedDev, Thank you for your PR!. 

Can you please create a function in World, Like in 
https://github.com/Pumpkin-MC/Pumpkin/pull/889

I also would prefer if you would use just a return instead having an else,
Also please use translations

Snowiiii:
Instead of return `Ok` which is not true since the Command did fail, you could return the error and also eliminate the need to send it to the sender manually
For example
` return Err(CommandFailed(Box::new(TextComponent::translations(...)));`

Snowiiii:
Looks good. Thank you @GreenedDev :+1: 🦀

PR #1024, opened 2025-07-07
fix: the fields of the struct CookingRecipe should be pub.

AyaSanae:
## Description
I forgot to add the pub when writing the CookingRecipe struct.


Snowiiii:
I would prefer if you would make larger PR's where you use the things you modify instead of just modifying them an leave them as it is

AyaSanae:
> I would prefer if you would make larger PR's where you use the things you modify instead of just modifying them an leave them as it is

I found this issue while writing the tick() for the furnace. I think I'll include it in the furnace's commit

PR #1025, opened 2025-07-07
feat: Implement hand swing animation for beds

HttpRafa:
## Description
This PR implements the swing_hand function that sends the required packets to the players. So that everyone can see the animation

### PR branch
https://github.com/user-attachments/assets/1150d762-d59f-48ae-945d-338189c6a2a5

### Master branch
https://github.com/user-attachments/assets/563f32ce-2d04-4488-ba34-afb926ee1775


Snowiiii:
Isn't this already implemented ?, The player sends the swing packet to the server and we just braodcast it

HttpRafa:
> Isn't this already implemented ?, The player sends the swing packet to the server and we just braodcast it

I looked in the vanilla code and they call the swing hand method on the player when the use_item return success

HttpRafa:
> > Isn't this already implemented ?, The player sends the swing packet to the server and we just braodcast it
> 
> I looked in the vanilla code and they call the swing hand method on the player when the use_item return success

And in my testing if i rightclick a bed in 1.21.7 without this PR the hand doesnt swing.

Here is the vanilla code that i implemented into pumpkin:

image

HttpRafa:
> Isn't this already implemented ?, The player sends the swing packet to the server and we just braodcast it

In vanilla there is a InteractionResult called SUCCESS_SERVER if this is return the server should sent a Animate packet. The best example for such a block is the bed.

HttpRafa:
> > Isn't this already implemented ?, The player sends the swing packet to the server and we just braodcast it
> 
> In vanilla there is a InteractionResult called SUCCESS_SERVER if this is return the server should sent a Animate packet. The best example for such a block is the bed.

For the other ones you are correct we just need to broadcast it.

Snowiiii:
Looks good. Thank you @HttpRafa 🐆

PR #1026, opened 2025-07-07
Plugin Commit Verification on Load

CuzImClicks:
## Description

Adds a warning to the loading process of plugins. It checks the git commit hash locally against the git commit hash that was saved at compile time into the metadata of the plugin. Also adds a warning for loading plugins that werent compiled with release on windows.

## Testing

Done but this takes AGES due to compile times.


Snowiiii:
I really don't think we need this especially when this adds a new crate and also uses our old git crate

CuzImClicks:
🥀

PR #1034, opened 2025-07-09
Feat/log4rs

zly2006:
## Description

Use log4rs for logging.

Improvements:
- log4rs
- reset terminal state before panicking, so panic message is more pretty now.

## Testing

- custom log config
- panic


Snowiiii:
I don't think we need this

PR #1035, opened 2025-07-09
Change BlockRegistry and ItemRegistry  to use u16

spr-equinox:
## Description

optimize BlockRegistry and several small change

## Testing

before
```
get_pumpkin_block       time:   [52.939 ns 53.738 ns 54.587 ns]
                        change: [+569.28% +587.62% +602.61%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 14 outliers among 100 measurements (14.00%)
  2 (2.00%) high mild
  12 (12.00%) high severe
```
after
```
     Running benches\bench_block.rs (target\release\deps\bench_block-c1fa4b06525777e2.exe)
Gnuplot not found, using plotters backend
get_pumpkin_block       time:   [7.5270 ns 7.5767 ns 7.6414 ns]
                        change: [−4.8425% −2.5811% −0.0623%] (p = 0.05 < 0.05)
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  5 (5.00%) high severe
```

i also tried other optimizations, but failed.
apply simd to aquifer_sampler.rs, but it runs a bit slow than before
<https://gist.github.com/spr-equinox/e11f6f49bd810caa70f6c2009d09ac8e>
use binary search/phf_set in is_tagged_with, but it also runs slower. it seems tags are too small to use hashmap
the only way to optimize it may be getting rid of str

### DIFF
--- a/pumpkin-data/build/fluid.rs
+++ b/pumpkin-data/build/fluid.rs
@@ -594,6 +594,7 @@ pub(crate) fn build() -> TokenStream {
     let properties = property_enums.values().map(|prop| prop.to_token_stream());
 
     quote! {
+        use std::hash::{Hash, Hasher};
         use crate::tag::{Tagable, RegistryKey};
         use pumpkin_util::resource_location::{FromResourceLocation, ResourceLocation, ToResourceLocation};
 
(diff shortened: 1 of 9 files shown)


CuzImClicks:
great idea :D 👍

topi-banana:
I think it would be better to implement `Hash` for `Block` and use `HashMap<Block, T>` instead.
```
impl Hash for Block {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.id.hash(state);
    }
}
```

sprinox:
> I think it would be better to implement `Hash` for `Block` and use `HashMap<Block, T>` instead.
> 
> ```
> impl Hash for Block {
>     fn hash<H: Hasher>(&self, state: &mut H) {
>         self.id.hash(state);
>     }
> }
> ```

i agree with you. i added the hash fn
but use `HashMap<Block, T>` need to copy Block into HashMap.
I think both codes are almost equally fast, but `HashMap<Block, T>` use more memory.

topi-banana:
You can use `&'static Block` as the key instead.

Snowiiii:
Looks good. Thank you @spr-equinox :1st_place_medal: 🐑

PR #1037, opened 2025-07-09
Optimization `ChunkNoiseFunctionComponent`

wHoIsDReAmer:
This PR significantly optimizes the performance of world generation by introducing a buffer pooling strategy for noise cache structures.

It addresses a bottleneck caused by frequent memory allocations during the creation of `ChunkNoiseRouter`.

### Problem

During world generation, `ChunkNoiseFunctionComponent` are created for every chunk. These components allocated `f64` arrays on the heap for caching, leading to overhead from repeated allocation and deallocation cycles.

### Benchmark Results

A new benchmark was added to measure the performance of `ChunkNoiseRouter::generate`. The results demonstrate a substantial improvement:

| Benchmark                                 | Before Optimization (µs) | After Optimization (µs) | Change      |
| ----------------------------------------- | ------------------------ | ----------------------- | ----------- |
| `noise_router_creation`                   | **150.30 µs**            | **38.958 µs**           | **285%** |

This represents an approximate **3.85x speedup** in the creation of the noise router, a critical part of the chunk generation process. The improvement is statistically significant (p < 0.05).

### DIFF
--- a/pumpkin-world/Cargo.toml
+++ b/pumpkin-world/Cargo.toml
@@ -70,3 +70,7 @@ harness = false
 [[bench]]
 name = "chunk_gen"
 harness = false
+
+[[bench]]
+name = "noise_router"
+harness = false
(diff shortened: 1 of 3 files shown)


Snowiiii:
Looks good. Thank you @wHoIsDReAmer :+1: 🦤

PR #1038, opened 2025-07-09
add tab_list packet

topi-banana:
## Description

Add packet `tab_list`

## Testing

```rust
player.client.enqueue_packet(&CTabList::new(TextComponent::text("Hello"), TextComponent::text("World"))).await;
```

image


Snowiiii:
I don't like "blank" PR's with no addition or logic, this feels unnecessary

topi-banana:
I think this packet is not used in vanilla.
but very useful for plugins etc.

Snowiiii:
> I think this packet is not used in vanilla. but very useful for plugins etc.

You are aware that when having a plugin API there will be raw Packets needed to do such stuff, for example setting a title would be done in a seperate function, this has a bunch of benefits, For example when the Packet changes that the plugin does not break or we can add wrappers for bedrock edition

topi-banana:
This packet does not exist in Bedrock Edition.

Snowiiii:
> This packet does not exist in Bedrock Edition.

That one more reason we want wrappers, or abstraction. We don't want to accidentally send this packet to a bedrock player which would cause him to disconnect. It is also just much more convenient to have these functions instead relying on packets

pumpkin-protocol/src/java/client/play/tab_list.rs
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    #[must_use]
+    pub fn with_footer(self, footer: TextComponent) -> Self {
+        Self {
+            header: self.header,
+            footer,
+        }
+    }
```
Snowiiii, line 33: I don't think this is needed

Snowiiii:
Also in the future there should be now added a packet wrapper for bedrock. Feel free to reopen

PR #1045, opened 2025-07-10
Implement drop player inventory on death

awalki:
## Description
Implement drop player inventory on death

## Testing
I have noticed only one bug that gamerule keepinventory changes only after server restart


sprinox:
Did you handle dropping the equipment and setting the velocity for the item entity?

pumpkin/src/entity/player.rs
```diff
@@ -1406,6 +1408,59 @@ impl Player {
...
+            if !item_stack.is_empty() {
+                let entity = Entity::new(
+                    Uuid::new_v4(),
+                    self.world().await,
+                    item_pos,
+                    EntityType::ITEM,
+                    false,
+                );
+
+                let mut rng = Xoroshiro::from_seed(get_seed());
```
CuzImClicks: i dont know if this is possible, but could you not create this rng only once outside of the loop?

navlas08: done

pumpkin/src/entity/living.rs
```diff
@@ -276,7 +278,19 @@ impl LivingEntity {
...
+        if let Some(player) = self
+            .entity
+            .world
+            .read()
+            .await
+            .get_player_by_id(entity_id)
+            .await
+        {
+            player.drop_all().await;
+        }
```
Snowiiii, line 293: This should not be handled here, instead we should make use of the player struct

Snowiiii:
Feel free to reopen

PR #1047, opened 2025-07-10
support offhand

sprinox:
## Description

we need to add an itemstack arg to normal_use and use_on_block, so that we can decrease the item in it.
should we make item args packed like blocks?

## Testing

### DIFF
--- a/pumpkin/src/net/java/play.rs
+++ b/pumpkin/src/net/java/play.rs
@@ -1566,12 +1566,11 @@ impl JavaClientPlatform {
         }
 
         let inventory = player.inventory();
-        let hand = Hand::try_from(use_item.hand.0);
-        if hand.is_err() {
+        let Ok(hand) = Hand::try_from(use_item.hand.0) else {
             self.kick(TextComponent::text("InvalidHand")).await;
             return;
-        }
-        let item_in_hand = if hand.unwrap_or(Hand::Left) == Hand::Left {
+        };
+        let item_in_hand = if hand == Hand::Left {
             inventory.held_item()
         } else {
             inventory.off_hand_item().await


pumpkin/src/net/java/play.rs
```diff
@@ -1560,7 +1566,11 @@ impl JavaClientPlatform {
         }
 
         let inventory = player.inventory();
-        let binding = inventory.held_item();
+        let binding = if use_item.hand == VarInt::from(0) {
+            inventory.held_item()
+        } else {
+            inventory.off_hand_item().await
+        };
```
Snowiiii: `binding` is a bad name and this is also confusing, use Hand::from instead

Snowiiii:
Instead of having unwrap_or when trying to get the hand I think it would be better to just kick the player, We actually did this somewhere iirc

pumpkin/src/net/java/play.rs
```diff
@@ -1566,7 +1566,12 @@ impl JavaClientPlatform {
         }
 
         let inventory = player.inventory();
-        let item_in_hand = if Hand::try_from(use_item.hand.0).unwrap_or(Hand::Left) == Hand::Left {
+        let hand = Hand::try_from(use_item.hand.0);
+        if hand.is_err() {
+            self.kick(TextComponent::text("InvalidHand")).await;
+            return;
+        }
+        let item_in_hand = if hand.unwrap_or(Hand::Left) == Hand::Left {
```
Snowiiii:
```rust
let Ok(hand) = Hand::from(...) else {
            // kick
};

hand == Hand::....
```

Snowiiii:
Looks good. Thank you @spr-equinox :D 🪶

PR #1054, opened 2025-07-11
Support Spawn

sprinox:
## Description

Added support for mob spawning, along with a bunch of TODOs and a generous amount of potential bugs. Performance took a major hit (missing heightmap, naive implementation for collecting surrounding chunks).

image

add the spawn to all mob category.
fix incorrect entity chunk pos
make all entitytype static ref
extract more data from java
port natural spawn almost completely

todo：
~~use height map (#1083)~~
track chunk around player
~~entity spawn slower than it should be~~ fixed
~~get chunk biome~~ partly
check collision
some spawn check should be made for each entity individually
~~get world min y~~
probably multithread

## Testing

### DIFF
--- a/pumpkin/src/world/natural_spawner.rs
+++ b/pumpkin/src/world/natural_spawner.rs
@@ -411,7 +411,7 @@ pub async fn spawn_category_for_position(
                 inc += 1;
                 continue;
             }
-            let entity = from_type(entity_type, new_pos_center, world, Uuid::new_v4());
+            let entity = from_type(entity_type, new_pos_center, world, Uuid::new_v4()).await;
             entity
                 .get_entity()
                 .set_rotation(rng().random::<f32>() * 360., 0.);
--- a/pumpkin/src/entity/ai/goal/active_target_goal.rs
+++ b/pumpkin/src/entity/ai/goal/active_target_goal.rs
@@ -19,14 +19,14 @@ pub struct ActiveTargetGoal {
     track_target_goal: TrackTargetGoal,
     target: Mutex<Option<Arc<dyn EntityBase>>>,
     reciprocal_chance: i32,
-    target_type: EntityType,
+    target_type: &'static EntityType,
     target_predicate: TargetPredicate,
 }
 
 impl ActiveTargetGoal {
     pub async fn new<F, Fut>(
         mob: &MobEntity,
-        target_type: EntityType,
+        target_type: &'static EntityType,
         reciprocal_chance: i32,
         check_visibility: bool,
         check_can_navigate: bool,
@@ -55,7 +55,7 @@ impl ActiveTargetGoal {
     #[must_use]
     pub async fn with_default(
         mob: &MobEntity,
-        target_type: EntityType,
+        target_type: &'static EntityType,
         check_visibility: bool,
     ) -> Self {
         let track_target_goal = TrackTargetGoal::with_default(check_visibility);
@@ -74,7 +74,7 @@ impl ActiveTargetGoal {
     async fn find_closest_target(&self, mob: &MobEntity) {
         let mut target = self.target.lock().await;
         let world = mob.living_entity.entity.world.read().await;
-        if self.target_type == EntityType::PLAYER {
+        if self.target_type == &EntityType::PLAYER {
             *target = world
                 .get_closest_player(
                     mob.living_entity.entity.pos.load(),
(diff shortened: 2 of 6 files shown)


Snowiiii:
use a better PR description

sprinox:
relate to Pumpkin-MC/Extractor#26

Snowiiii:
This makes chunk gen way too slow, It almost feels like a dead lock, Before merging this the performance needs to be fixed

sprinox:
> This makes chunk gen way too slow, It almost feels like a dead lock, Before merging this the performance needs to be fixed

it won't effect chunk gen, but make tick much slower. we don't have height map, so i have to get the highest block in a extremely slow way

Snowiiii:
> > This makes chunk gen way too slow, It almost feels like a dead lock, Before merging this the performance needs to be fixed
> 
> it won't effect chunk gen, but make tick much slower. we don't have height map, so i have to get the highest block in a extremely slow way

This was just merged

Snowiiii:
Chunk gen still seems really slow

sprinox:
> Chunk gen still seems really slow

I am writing codes using heigh map. There will be a commit later

sprinox:
mspt is much better now

~~but it seems chunk gen still slow~~
`pumpkin::world: Ticking world took 50.6394ms, loaded chunks: 241, chunk tick took 7.5391ms`
in master, tps will drop down when generation tasks are heavy.
i think slow generation speed it is a normal behavior.

Snowiiii:
For some reason there are random mobs spawned sometimes, Currently the light level is always at max since we don't calculate light so i think you want to check if the mobs can be actually spawned at a certain light level
image

pumpkin/src/world/natural_spawner.rs
```diff
@@ -0,0 +1,566 @@
...
+}
+
+pub async fn is_valid_spawn_position_for_type(
+    world: &Arc<World>,
+    block_pos: &BlockPos,
+    category: &'static MobCategory,
+    entity_type: &'static EntityType,
+    distance: f64,
+) -> bool {
+    // TODO !SpawnPlacements.checkSpawnRules(entityType, level, EntitySpawnReason.NATURAL, pos, level.random)
```
sprinox, line 507: i left todo

sprinox:
> For some reason there are random mobs spawned sometimes, Currently the light level is always at max since we don't calculate light so i think you want to check if the mobs can be actually spawned at a certain light level

light is checked in checkSpawnRules and isValidPositionForMob which i leave them todo

Snowiiii:
> > For some reason there are random mobs spawned sometimes, Currently the light level is always at max since we don't calculate light so i think you want to check if the mobs can be actually spawned at a certain light level
> 
> light is checked in checkSpawnRules and isValidPositionForMob which i leave them todo

Can you maybe implement that before this gets merged ?

sprinox:
image

it is too complicated and it should be implemented in path finding

pumpkin/src/world/natural_spawner.rs
```diff
@@ -0,0 +1,566 @@
...
+    fn can_spawn(
+        &mut self,
+        entity_type: &'static EntityType,
+        pos: &BlockPos,
+        _world: &Arc<World>,
+    ) -> bool {
+        self.last_checked_pos = *pos;
+        self.last_checked_type = entity_type;
+        // TODO get biome
+        let biome = &Biome::PLAINS;
```
Snowiiii: At least implement getting the biome here

sprinox: i implement get_rough_biome instead, getBiome need complicated caculaton and LinearCongruentialGenerator

Snowiiii:
For some reason all entities despawn when rejoining the world

spr-equinox:
> For some reason all entities despawn when rejoining the world

i think it is because height map isn't save to file and i only create it once when chunk gen not chunk load.

sprinox:
> For some reason all entities despawn when rejoining the world

i accidently found a dropper block panic. i fixed it in this pr.

i don't know why i need to spawn a entity in this way so that it won't be invisiable after rejoin.

image

image

but i noticed that item entity will be invisiable after rejoin too.
it can't see them but i can pick it up

sprinox:
@Snowiiii could this be merged?

Snowiiii:
> @Snowiiii could this be merged?

Is the despawning bug fixed ?

sprinox:
despawn fixed

Snowiiii:
Okay. LGTM then. Thank you @spr-equinox 🦙

PR #1057, opened 2025-07-12
Improve tags

sprinox:
TODO replace get_tag_values is_tagged_with get_tag_values

## Description

`is_tagged_with` impove

master
```
     Running benches\bench_tag.rs (target\release\deps\bench_tag-43ea3f1e7dba8419.exe)
is_tagged_with          time:   [228.23 ns 232.88 ns 237.53 ns]
                        change: [+132.06% +137.24% +142.85%] (p = 0.00 < 0.05)
                        Performance has regressed.

get_tag_values          time:   [104.31 ns 105.67 ns 107.10 ns]
                        change: [+16.844% +18.699% +20.677%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

     Running benches\chunk.rs (target\release\deps\chunk-373d9b9036131ec0.exe)
Benchmarking overworld biome: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.8s, enable flat sampling, or reduce sample count to 60.
overworld biome         time:   [1.1219 ms 1.1525 ms 1.1821 ms]
                        change: [+24.556% +27.944% +31.776%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild

overworld noise         time:   [15.773 ms 16.076 ms 16.382 ms]
                        change: [+22.347% +24.699% +27.154%] (p = 0.00 < 0.05)
                        Performance has regressed.

overworld surface       time:   [21.742 ms 22.001 ms 22.267 ms]
                        change: [+5.7484% +7.0275% +8.3882%] (p = 0.00 < 0.05)
                        Performance has regressed.

     Running benches\chunk_gen.rs (target\release\deps\chunk_gen-2de1a7336234841d.exe)
Benchmarking chunk generation seed 0: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 180.0s. You may wish to increase target time to 216.2s or enable flat sampling.
chunk generation seed 0 time:   [3.9264 s 3.9346 s 3.9501 s]
                        change: [−0.2786% +0.8497% +1.6652%] (p = 0.13 > 0.05)
                        No change in performance detected.
```

v1
```
     Running benches\bench_tag.rs (target\release\deps\bench_tag-43ea3f1e7dba8419.exe)
[...]

PR #1059, opened 2025-07-13
Refactor ItemStack to Support Vanilla-Like Data Components

spr-equinox:
## Description

rely on #1057

The goal of this PR is to implement the vanilla-like `DataComponentType` system. The current `ItemStack` and `ItemComponents` design does not support the powerful and flexible item component system found in vanilla Minecraft (for example, you can even change the max stack size using commands like `/give @s minecraft:ender_pearl[minecraft:max_stack_size=99] 99`).

With future features like enchantments and better maintainability in mind, I’ve come up with a potential approach (see the code for details). Developers can implement specific `DataComponentType` in corresponding `XxxxImpl` structs. ~~Then, in the `Item` struct, a static reference like `&'static [&'static DataComponent]` is used as the default property list.~~ The `Item` remains globally constant, and `ItemStack` can override those defaults using a `Vec<DataComponent>`.

However, I’m not familiar with the current system for packet (de)serialization in this project, so I hope someone else can help implement the networking part that supports sending `ItemStack` with `DataComponent`.

This is just one possible implementation I came up with — I’m very open to better ideas if anyone has a cleaner or more efficient approach.

~~i don't want to use lazy_static, but have to~~ use Cow

## Testing


4lve:
Hey, nice initiative. I made a concept script a little while ago that might be useful

```rs
use std::any::{Any, TypeId};
use std::collections::HashMap;
use std::hash::Hash;

pub trait ComponentType: Hash + Eq + 'static {
    type Data: Clone + Hash + Eq + 'static;
}

pub struct ComponentMap(HashMap<TypeId, Box<dyn Any>>);

impl ComponentMap {
    pub fn new() -> Self {
        Self(HashMap::new())
    }

    pub fn insert<C: ComponentType>(&mut self, data: C::Data) {
        self.0.insert(TypeId::of::<C>(), Box::new(data));
    }

    pub fn get<C: ComponentType>(&self) -> Option<&C::Data> {
        self.0
            .get(&TypeId::of::<C>())
            .and_then(|boxed_data| boxed_data.downcast_ref::<C::Data>())
    }

    pub fn get_mut<C: ComponentType>(&mut self) -> Option<&mut C::Data> {
        self.0
            .get_mut(&TypeId::of::<C>())
            .and_then(|boxed_data| boxed_data.downcast_mut::<C::Data>())
    }
}

#[derive(Debug, Hash, Eq, PartialEq, Clone, Copy)]
struct MaxStackSize {}

impl ComponentType for MaxStackSize {
    type Data = u8;
}

#[derive(Debug, Hash, Eq, PartialEq, Clone, Copy)]
struct ItemLevel;

impl ComponentType for ItemLevel {
    type Data = u8;
}

#[derive(Debug, Hash, Eq, PartialEq, Clone, Copy)]
struct Durability;

impl ComponentType for Durability {
    type Data = u16;
}

fn main() {
    let mut component_map = ComponentMap::new();

    component_map.insert::<MaxStackSize>(64u8);

    component_map.insert::<ItemLevel>(10u8);

    component_map.insert::<Durability>(100u16);

    component_map.insert::<MaxStackSize>(60u8);

    // Retrieve MaxStackSize
    if let Some(data) = component_map.get::<MaxStackSize>() {
        println!("MaxStackSize data: {}", data);
    }

    println!("{:?}", component_map.0);
}
```

sprinox:
Actually ~~I try to get rid of dyn call, because it may be slow and need to save ptr of fn~~ still use dyn, and hashmap may not perform better than array when data size is really small. We also can't use &'static in component data. It needs to contain item names, custom model data and so on.

unschlagbar:
You should make the json more like this: 
{
    "minecraft:custom_data": 0,
    "minecraft:max_stack_size": 1,
    "minecraft:max_damage": 2,
    "minecraft:damage": 3,
    ...
},

sprinox:
> You should make the json more like this: { "minecraft:custom_data": 0, "minecraft:max_stack_size": 1, "minecraft:max_damage": 2, "minecraft:damage": 3, ... },

yes, i used to want to extract more information from java. i am changing it

unschlagbar:
I think that `for i in &x` should be prefered over `for i in x.iter()` its just cleaner

sprinox:
> I think that `for i in &x` should be prefered over `for i in x.iter()` its just cleaner

Cow will compile error in that way

4lve:
We should look into using a registry instead of extracting item components. This is due to each item component having a unique Serializer and the possibility for plugins to register new ones in the future

pumpkin/src/block/blocks/candle_cakes.rs
```diff
@@ -18,26 +20,27 @@ use crate::{
...
-    (&Item::LIGHT_GRAY_CANDLE, &Block::LIGHT_GRAY_CANDLE_CAKE),
-    (&Item::CYAN_CANDLE, &Block::CYAN_CANDLE_CAKE),
-    (&Item::PURPLE_CANDLE, &Block::PURPLE_CANDLE_CAKE),
-    (&Item::BLUE_CANDLE, &Block::BLUE_CANDLE_CAKE),
-    (&Item::BROWN_CANDLE, &Block::BROWN_CANDLE_CAKE),
-    (&Item::GREEN_CANDLE, &Block::GREEN_CANDLE_CAKE),
-    (&Item::RED_CANDLE, &Block::RED_CANDLE_CAKE),
-    (&Item::BLACK_CANDLE, &Block::BLACK_CANDLE_CAKE),
-];
+lazy_static! {
```
CuzImClicks: what is the point of this? the old version still compiles and doesn't need a dependency.

sprinox: Oh, I forgot to change back it

sprinox:
> We should look into using a registry instead of extracting item components. This is due to each item component having a unique Serializer and the possibility for plugins to register new ones in the future

item components need to send to client, so we can't add a new one without a modded client. as far as i know, spigot api don't support to add a new component. <https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/inventory/meta/components/package-summary.html>

unschlagbar:
> > We should look into using a registry instead of extracting item components. This is due to each item component having a unique Serializer and the possibility for plugins to register new ones in the future
> 
> item components need to send to client, so we can't add a new one without a modded client. as far as i know, spigot api don't support to add a new component. https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/inventory/meta/components/package-summary.html

We want to make pumpkin compatible with forge client mods, that means that when you want to port a mod to pumpkin you only need to remake the server side.

pumpkin/Cargo.toml
```diff
@@ -84,6 +84,7 @@ dhat = { version = "0.3.3", optional = true }
 
 flate2 = "1.1.2"
 console-subscriber = { version = "0.4.1", optional = true }
+lazy_static = "1.5.0"
```
Snowiiii: Why you added `lazy_static`, it is outdated and you can just use the std LazyLock

sprinox: forget to remove

Snowiiii:
LGTM. Thank you @spr-equinox 🫎

PR #1060, opened 2025-07-14
PluginAPI / Add `NamespacedKey` and `PersistentDataContainer`

tn-lorenz:
## Description
This PR aims to add the covetted `PersitentDataContainer` feature to the Plugin API, but there are still some things I need to consider:
- [ ] Possibly convert `PersistentDataHolder` from a derive macro to an attribute macro for more flexibility
- [ ] Better way to "add" values to existing `NamespacedKey`, instead of overriding every time (automatic listing?)
- [ ] Add error handling when plugin dev tries to register an already existing key
- [ ] Implement for more types

## Testing
TODO

### DIFF
--- a/pumpkin-macros/src/lib.rs
+++ b/pumpkin-macros/src/lib.rs
@@ -3,7 +3,7 @@ use proc_macro::TokenStream;
 use proc_macro_error2::{abort, abort_call_site, proc_macro_error};
 use quote::quote;
 use syn::spanned::Spanned;
-use syn::{self};
+use syn::{self, DeriveInput};
 use syn::{
     Block, Expr, Field, Fields, ItemStruct, Stmt,
     parse::{Nothing, Parser},
@@ -351,4 +351,21 @@ pub fn block_property(input: TokenStream, item: TokenStream) -> TokenStream {
     };
 
     code.into()
+}
+
+// Assumes the struct has a `container: PersistentDataContainer` field
+#[proc_macro_derive(PersistentDataHolder)]
+pub fn derive_persistent(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = input.ident;
+
+    let expanded = quote! {
+        impl HasPersistentContainer for #name {
+            fn persistent_container(&self) -> &PersistentDataContainer {
+                &self.container
+            }
+        }
+    };
+
+    TokenStream::from(expanded)
 }
--- a/pumpkin/src/plugin/api/persistence/mod.rs
+++ b/pumpkin/src/plugin/api/persistence/mod.rs
@@ -1,0 +1,162 @@
+use dashmap::DashMap;
+use serde::{Deserialize, Serialize};
+
+/// The `NamespacedKey` struct
+#[allow(dead_code)]
+#[derive(Eq, Hash, PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub struct NamespacedKey {
+    namespace: String,
+    key: String,
+}
+
+#[derive(Debug)]
+pub enum NamespacedKeyError {
+    NonAsciiNamespace,
+    NonAsciiKey,
+}
+
+/// The `NamespacedKey` constructor
+///
+/// `new()` must only be called via the `ns_key!` macro.
+///
+/// # Parameters
+/// - `namespace`: namespace of the key, must be equal to the `CARGO_PKG_NAME`
+/// - `key`: The key as a String
+///
+/// # Returns
+/// - Self
+impl NamespacedKey {
+    #[allow(dead_code)]
+    pub(crate) fn new(namespace: &str, key: &str) -> Result<Self, NamespacedKeyError> {
+        if !namespace.is_ascii() {
+            log::error!("Invalid namespace: '{namespace}' is not pure ASCII.");
+            return Err(NamespacedKeyError::NonAsciiNamespace);
+        }
+        if !key.is_ascii() {
+            log::error!("Invalid key: '{key}' is not pure ASCII.");
+            return Err(NamespacedKeyError::NonAsciiKey);
+        }
+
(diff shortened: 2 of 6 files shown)


pumpkin-macros/src/lib.rs
```diff
@@ -352,3 +352,62 @@ pub fn block_property(input: TokenStream, item: TokenStream) -> TokenStream {
...
+                }
+            }
+            None
+        })
+    } else {
+        None
+    };
+
+    let field = match field {
+        Some(f) => f,
```
MatrixFrog: there's a somewhat new feature you can use that's a little shorter than this match statement

https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html

pumpkin/src/plugin/api/persistence/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    };
+}
+
+/// The `PersistentDataContainer` struct
+///
+/// This struct contains `NamespacedKey`s and associates them with `PersistentValue`s using a `DashMap` for maximum concurrency.
+#[allow(dead_code)]
+#[derive(Default, Debug)]
+pub struct PersistentDataContainer {
```
MatrixFrog: why make this a struct rather than just a type alias?

pumpkin-macros/src/lib.rs
```diff
@@ -352,3 +352,62 @@ pub fn block_property(input: TokenStream, item: TokenStream) -> TokenStream {
...
+#[proc_macro_derive(PersistentDataHolder, attributes(persistent_data))]
+pub fn derive_persistent(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = input.ident;
+
+    let field = if let syn::Data::Struct(data) = &input.data {
+        data.fields.iter().find_map(|f| {
+            for attr in &f.attrs {
+                let path = attr.path();
+                if path.is_ident("persistent_data") {
```
MatrixFrog: would it be considered an error to put this annotation on a field which is something other than PersistentDataContainer? if so do you want to check the type here and report that error?

pumpkin/src/plugin/api/persistence/mod.rs
tn-lorenz: I was thinking that we don't need to expose every method on DashMap (idk if stuff like `shards()` or `shrink_to_fit()` are necessary here) and also otherwise I probably couldn't use the `get_as` fn

P.S.: I'm trying to mimic this feature of an existing java minecraft plugin API: https://docs.papermc.io/paper/dev/pdc/#:~:text=The%20Persistent%20Data%20Container%20(PDC,World

pumpkin-macros/src/lib.rs
tn-lorenz: Yes I forgot that, thank you :) 👍🏼

pumpkin/src/entity/mod.rs
```diff
@@ -76,15 +81,20 @@ pub trait EntityBase: Send + Sync {
...
     }
 
-    async fn write_nbt(&self, nbt: &mut pumpkin_nbt::compound::NbtCompound) {
+    async fn write_nbt(&self, nbt: &mut NbtCompound) {
+        let entity = self.get_entity();
+        let compound = from_pdc(&self.get_entity().container);
+
         if let Some(living) = self.get_living_entity() {
             living.write_nbt(nbt).await;
+            living.write_nbt(&mut compound.clone()).await;
```
ReCore-sys, line 90: Probably don't use clone here, make write_nbt take ownership so the caller has to clone, making it clearer when clones are happening

pumpkin/src/plugin/api/persistence/nbt.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn to_pdc(&self) -> PersistentDataContainer {
+        let container = PersistentDataContainer::new();
+
+        for (namespace, ns_tag) in &self.child_tags {
+            if let NbtTag::Compound(ns_compound) = ns_tag {
+                for (key, tag) in &ns_compound.child_tags {
+                    if let Ok(ns_key) = NamespacedKey::new(namespace, key) {
+                        let value = match tag {
+                            NbtTag::Byte(b) => PersistentDataType::Bool(*b != 0),
+                            NbtTag::Short(s) => PersistentDataType::I32(i32::from(*s)),
```
ReCore-sys, line 19: Why is a short 32 bits here? you already have a U16 type defined above

pumpkin/src/plugin/api/persistence/mod.rs
```diff
@@ -0,0 +1,185 @@
...
+///
+/// # Parameters
+/// - `$value`: The key string.
+///
+/// # Panics
+/// Panics if the key or namespace contains non-ASCII characters.
+#[macro_export]
+macro_rules! ns_key {
+    ($value:expr) => {
+        match ::pumpkin::plugin::persistence::NamespacedKey::new(env!("CARGO_PKG_NAME"), $value) {
```
ReCore-sys, line 71: If possible, get the macro to generate a literal instead of calling new()

unschlagbar:
The idea behind this is genius

Snowiiii:
@tn-lorenz Conflicts are getting more and more and i don't understand if this is done or not because it is not marked as draft but there are many things missing looking at the Checkboxes, The last commit was also already a while ago

tn-lorenz:
@Snowiiii Well as I pointed out in the important block at the beginning (and more than once on discord), it is currently not possible to go further and test, as NBT Data is never read in some cases. At the time, I verified this by adding debug statements, but only the one corresponding to saving got printed, never the one for reading. 
This was further emphasized by the fact that spawned mobs were not persistent between restarts.
Also, I'm heading for vacation now, so don't expect me finishing this in the next 7-9 days.

PR #1063, opened 2025-07-18
Bedrock player joining & random things

unschlagbar:
## Description
I am still going to work on this until alex is back.
This is just here for early review since this has some very big changes.

I added derive macros for bedrock packets, a better name suggestion is welcome.
Bedrock packets are now marked with B: BClientPacket, BServerPacket, The current issue is that my PacketWrite & Read traits return std::io::error while the serde ones return serde errors, im still looking for a solution.
In the future i want to replace all `put_T`, `get_T` with `T::write` and `T::read`, look into the bedrock packets that implement PacketWrite or PacketRead to see what i mean.

I added the bedrock blockstate nbt file because its seems like the way to caclulate the network block state id in bedrock is unknown and every bedrock server uses that file, its extracted from a bds mod. If anyone knows the way to calculate the right order, please let me know.

Changed bedrock_addr and java_addr with ports (binding to a addr other then the own one is impossible anyway) in inspiration from the bedrock dedicated server [Bedrock server download](https://www.minecraft.net/en-us/download/server/bedrock), also i think our release zip should have a similar structure, maybe we can build a very small rust programm that makes the structure: remove all source files, unnecessary assets and move the binary out of target, so we can make a release very fast

Alex please dont hate me because this:
Removed every local import of `std::sync::atomic::Ordering::Relaxed`
Replaced `where` with generics where ever possible for better readability

## Testing

Known issues:
I dont know how to get my pc into sending the renderdistance (it defaults to 16) my mobile does it tho
Bedrock players can experience a deadlock very fast
BlockStates are incorrect & Creative menu empty


pumpkin-data/build/block.rs
```diff
@@ -984,19 +984,21 @@ pub(crate) fn build() -> TokenStream {
                 }
             }
 
+            #[track_caller]
             #[doc = r" Get the properties of the block."]
             pub fn properties(&self, state_id: u16) -> Option<Box<dyn BlockProperties>> {
```
CuzImClicks, line 1031: Block.properties returns an Option
Block.from_properties panics

Fluid.properties panics
Fluid.from_properties panics

Why were these changed away from an Option?
Why is Block.properties different?

pumpkin/src/main.rs
```diff
@@ -122,9 +121,7 @@ async fn main() {
         // We need to abide by the panic rules here.
         std::process::exit(1);
     }));
-    log::info!(
-        "Starting Pumpkin {CARGO_PKG_VERSION} for Minecraft {CURRENT_MC_VERSION} (Protocol {CURRENT_MC_PROTOCOL})",
-    );
+    log::info!("Starting Pumpkin {CARGO_PKG_VERSION} Minecraft (Protocol {CURRENT_MC_PROTOCOL})",);
```
CuzImClicks, line 124: Why are we removing the mc version?

pumpkin/src/net/mod.rs
```diff
@@ -106,19 +106,31 @@ impl ClientPlatform {
...
     }
 
+    /// This function should only be used where you know that the client is bedrock!
+    #[inline]
     #[must_use]
-    pub fn closed(&self) -> bool {
-        match self {
-            Self::Java(java) => java.closed.load(Ordering::Relaxed),
-            Self::Bedrock(bedrock) => bedrock.closed.load(Ordering::Relaxed),
+    pub fn bedrock(&self) -> &Arc<BedrockClient> {
```
CuzImClicks, line 113: Why... match statements exist and they don't panic.

pumpkin/src/world/mod.rs
```diff
@@ -764,50 +766,84 @@ impl World {
...
-                        server_auth_sounds: false,
-                    },
-                    RakReliability::Unreliable,
-                )
-                .await;
-        } else {
-            panic!();
-        }
+
+        let client = player.client.bedrock();
```
CuzImClicks, line 787: This is just worse than if let

CuzImClicks (changes requested):

pumpkin/src/main.rs
unschlagbar: Its in the pkg version
Screenshot 2025-07-18 205115

pumpkin/src/net/mod.rs
unschlagbar: Because there are many cases where you know what the client is, its just a helper fn.
Alex complained about client match statements

pumpkin-data/build/block.rs
unschlagbar: It was unwraped every where so there is no point, this changes also somehow made the binary a bit smaller.
Not every block has properties so that must be handled but making like stone from props is an obvius error

unschlagbar: Hashmaps are very slow especialy in comparison to a vec with a average size of 3. Hashmaps start to make sense with 50+ entries

pumpkin/src/main.rs
unschlagbar: `Starting Pumpkin 0.1.0-dev+1.21.8 for Minecraft 1.21.8 (Protocol 772)` looks very stupid

CuzImClicks:
Hey can we try to get a bit more descriptive commit names xD just like a three word summary would go a long way :)

PS: you are doing great work!

pumpkin/src/net/bedrock/play.rs
```diff
@@ -0,0 +1,128 @@
...
+                    config.chat.format.clone(),
+                    gameprofile.name.clone(),
+                    message.clone(),
+                );
+
+                let entity = &player.living_entity.entity;
+                let world = &entity.world.read().await;
+                if BASIC_CONFIG.allow_chat_reports {
+                    //TODO Alex help, what is this?
+                    //world.broadcast_secure_player_chat(player, &message, decorated_message).await;
```
Snowiiii, line 112: When Mojang added Chat Reporting when moving to an more secure Chat system

Snowiiii:
Awesome work, Just wow. Big big thank you @unschlagbar :D 🐻

PR #1064, opened 2025-07-18
support Hopper, remove dropper todo, and support shulker box basically

spr-equinox:
## Description

I believe there is a very small chance that the current implementation could lead to a deadlock.
For example, if container A locks inventory A1 and tries to add an item to container B’s inventory B2, and at the same time, container B has already locked B2 and is trying to add an item to A1 — could this result in a circular lock?

If such a situation is indeed possible, I can't think of a good way to prevent it, unless we lock the entire container as a single unit.

## Testing


Snowiiii:
LGTM. Thank you @spr-equinox :+1: 🦏

PR #1067, opened 2025-07-21
Implment Heightmap

AyaSanae:
## Description
A HeightMap consists of a packed array of i64 type, where each 9 bits store a height value. Therefore, without considering concatenation, one i64 can store 7 height values.

The heightmaps.set() method adds world_min_y.abs() (which is 64 in the overworld)to the stored height value in the packed array to ensure it is positive.
Similarly, heightmaps.get() subtracts world_min_y.abs() from the value before returning it as the result.
For a given (x, z), if no valid y is found after iterating through all possible y values, the heightmap will be set to (x, min_y, z).

The log_heightmap() method can intuitively print the values in the heightmap for visualization.
like:
image
image

In the image, the short grass has a value of (2, 78, 0) in the WorldSurface heightmap. However, the MotionBlocking heightmap does not record the coordinates of the short grass because it does not block movement. Therefore, for (x:2, z:0), the MotionBlocking heightmap records (2, 77, 0)

## Testing
image

I directly modified get_top_block() and used the kill command for testing. It works fine in all dimensions.
In the Nether, the heightmap is the same as in vanilla, at 127.


Snowiiii:
Can you maybe fix this for Nether, and make sure it works fine for End

AyaSanae:
> Can you maybe fix this for Nether, and make sure it works fine for End

I've fixed it, and now it's working properly in all dimensions. Please check my updated description.

Snowiiii:
> > Can you maybe fix this for Nether, and make sure it works fine for End
> 
> I've fixed it, and now it's working properly in all dimensions. Please check my updated description.

Nice, now please only fix the conflicts

AyaSanae:
> > > Can you maybe fix this for Nether, and make sure it works fine for End
> > 
> > 
> > I've fixed it, and now it's working properly in all dimensions. Please check my updated description.
> 
> Nice, now please only fix the conflicts

Done
In the Nether, the heightmap is the same as in vanilla, at 127.

Snowiiii:
LGTM. Thank you @AyaSanae 🐝

Snowiiii:
@AyaSanae going from nether into overworld gives me
image

AyaSanae:
> @AyaSanae going from nether into overworld gives me image

I just realized. I think the issue is that I'm using version 1.21.7. I'll try migrating to 1.21.8.

Snowiiii:
> > @AyaSanae going from nether into overworld gives me image
> 
> I just realized. I think the issue is that I'm using version 1.21.7. I'll try migrating to 1.21.8.

there should be no chunk differences since it is only has hotfixes

AyaSanae:
> > > @AyaSanae going from nether into overworld gives me image
> > 
> > 
> > I just realized. I think the issue is that I'm using version 1.21.7. I'll try migrating to 1.21.8.
> 
> there should be no chunk differences since it is only has hotfixes

I just tried it, and this is indeed not a version-related issue.
I remember I successfully returned from the Nether to the Overworld. Let me troubleshoot the issue.🤔

PR #1069, opened 2025-07-22
PluginAPI / Add schedulers

tn-lorenz:
## Description
This PR aims to add the usefull makros `run_task_later!` and `run_task_timer!` to mimic the behaviour of bukkit runnables, which are frequently used in bukkit plugins.

### DIFF
--- a/pumpkin/src/plugin/api/task/mod.rs
+++ b/pumpkin/src/plugin/api/task/mod.rs
@@ -1,0 +1,77 @@
+use std::sync::Arc;
+use tokio::runtime::Runtime;
+use tokio::time::{Duration, sleep};
+use once_cell::sync::Lazy;
+
+pub static TOKIO_RUNTIME: Lazy<Runtime> =
+    Lazy::new(|| Runtime::new().expect("Failed to create global Tokio Runtime"));
+
+#[async_trait::async_trait]
+pub trait TaskHandler: Send + Sync {
+    async fn run(&self);
+    async fn cancel(&self);
+}
+
+pub fn start_loop<H>(delay: Duration, handler: Arc<H>)
+where
+    H: TaskHandler + 'static,
+{
+    TOKIO_RUNTIME.spawn(run_task_timer(delay, handler));
+}
+
+async fn run_task_timer<H>(delay: Duration, handler: Arc<H>)
+where
+    H: TaskHandler + 'static,
+{
+    loop {
+        sleep(delay).await;
+        handler.run().await;
+    }
+}
+
+async fn run_task_later<H>(delay: Duration, handler: Arc<H>)
+where
+    H: TaskHandler + 'static,
+{
+    sleep(delay).await;
+    handler.run().await;
+}
+
+#[macro_export]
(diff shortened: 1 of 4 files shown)


tn-lorenz:
Should I hook into the world ticker? Because currently the tasks will still tick normally, even if the world ticker is stopped/slowed down/sprinting.. o.O

connor33341:
> Should I hook into the world ticker? Because currently the tasks will still tick normally, even if the world ticker is stopped/slowed down/sprinting.. o.O

Ideally as that is how Bukkit Runable does it

HttpRafa:
I think there should also be an option to specify a Rust Duration, similar to how Paper does it using TimeUnit.MINUTE, etc. A lot of people work with the Paper API nowadays. https://jd.papermc.io/paper/1.21.7/io/papermc/paper/threadedregions/scheduler/AsyncScheduler.html

tn-lorenz:
@HttpRafa I'm aware of this and I'm planning on implementing that, but atm I can't even get `run_task_timer!` to work, so I'll probably need some help of a macro wizard before adding all the missing features 👍🏼

HttpRafa:
> @HttpRafa I'm aware of this and I'm planning on implementing that, but atm I can't even get `run_task_timer!` to work, so I'll probably need some help of a macro wizard before adding all the missing features 👍🏼

No problem. I just figured it was worth mentioning, since many people use this method to get predictable callbacks without relying on ticks.

pumpkin/src/plugin/api/task/mod.rs
```diff
@@ -0,0 +1,153 @@
...
+impl ScheduledHandle {
+    pub async fn cancel(&self) {
+        self.cancel_flag.store(true, Ordering::Relaxed);
+        self.handler.cancel().await;
+    }
+}
+
+#[derive(Clone)]
+pub struct RepeatingHandle {
+    pub(crate) cancel_flag: Arc<AtomicBool>,
```
unschlagbar, line 137: This struct should be removed! A single valued struct without any function other than set & get is useless verbosity

unschlagbar (commented):
Otherwise, looks usefull

pumpkin/src/plugin/api/task/mod.rs
tn-lorenz: Aight, will check it out 👍🏼

4lve:
Is there any reason we are using a macro for this instead of a function?

tn-lorenz:
@4lve Idk how to pass an arbitrary amount of arbitrary types and functions into a function (while also making it ergonomic to use), that's why :D

Edit: also, as you can see, there is stuff that needs to be generated for this to work

Snowiiii:
I think the macros make the code more complicated then it should be

tn-lorenz:
Yes, maybe this is not the right use-case, but - as the logic for durations and ticks will differ - I was thinking that this would allow for convenient use while having only one macro, instead of having multiple fns with similar names.

Snowiiii:
Feel free to reopen

PR #1071, opened 2025-07-23
Add macro for expanding tags to PumpkinBlock

teknostom:
## Description

Right now, many blocks are implementing the `tag -> blocks` logic every time, so i added a macro to do this automatically.

## Testing
 Manual

### DIFF
--- a/pumpkin/src/block/registry.rs
+++ b/pumpkin/src/block/registry.rs
@@ -80,6 +80,7 @@ impl BlockRegistry {
         let val = Arc::new(block);
         self.blocks.reserve(names.len());
         for i in names {
+            log::info!("{i}");
             self.blocks.insert(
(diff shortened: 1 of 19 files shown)


pumpkin/src/block/registry.rs
```diff
@@ -80,6 +80,7 @@ impl BlockRegistry {
         let val = Arc::new(block);
         self.blocks.reserve(names.len());
         for i in names {
+            log::info!("{i}");
```
HttpRafa: I think you forgot to remove this.

teknostom: Yeah, good catch. Will fix

Snowiiii:
Looks good. Thank you @teknostom :D 🐓

Issue #1077, opened 2025-07-27
Could not find en_us.json


AyaSanae:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

EndeavourOS Linux x86_64

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.8/8495a00

### What happened?

<img width="1899" height="768" alt="Image" src="https://github.com/user-attachments/assets/b5159877-9093-4e1a-80a1-d12d5e0b1e26" />

### To Reproduce

cargo build
cp target/debug/pumpkin ../pumpkin
cd .. 
./pumpkin

### Expected behavior

A binary program can run the game.

PR #1079, opened 2025-07-27
fix: could not find en_us.json

AyaSanae:
## Description
https://github.com/Pumpkin-MC/Pumpkin/issues/1077

When I placed the assets/en_us.json file from the source code into the assets directory relative to the pumpkin binary, it worked correctly.
So I used include_str!() to handle it.


Snowiiii:
Thank you @AyaSanae 🐟

PR #1080, opened 2025-07-27
Custom apis

vyPal:
## Description
This PR contains several substantial changes to the plugin API.

### Move `Context` behind an `Arc`
The plugin's `on_load` and `on_unload` methods traditionally received a reference to the Context object (`&Context`), I have replaced this with a `Arc<Context>`, which allows both the use of the same instance of the Context for both the loading an unloading of the plugin, as well as copying the Context object and using it across threads from within the plugin. 

Previously if a plugin wanted to use the Server, or PermissionManager, or any other field contained in the Context, the plugin either needed to use the field directly from the `on_load` method, or a method directly called from `on_load`. If the plugin wanted to use it in another thread, it would have to make clones of each of the field that it wanted to access. Now it is possible to just clone the Context itself, without having to clone each field individually. Also this allows using the Context's helper methods in other threads.

### Move `PluginManager` outside of `RwLock`
Since plugins will now require access to certain fields in the PluginManager while loading, I have decided to move the manager itself out of the RwLock, and instead put all fields that require mutation inside their own RwLocks. This means that plugins can access certain methods and fields of the PluginManager while they are being loaded, without causing a deadlock.

This also allows plugins to fire events, including their own custom events.

### Custom APIs
This change adds a `services` field to the PluginManager, where plugins can submit instances of their own structs, which can be accessed by other plugins. This allows for plugins to expose custom APIs to each other for easier interaction and sharing data between plugins. Provided below is a hypothetical example situation:
<details>
  <summary>Code sample</summary>
  
  Here are examples of two plugins, the first one being a API provider and the second a API consumer.
  #### API Provider
```rs
#[derive(Clone)]
pub struct EconomyApi {
    // internal state
}

impl EconomyApi {
    pub async fn get_balance(&self, player_uuid: &str) -> Result<f64, String> {
        // implementation
    }
    
    pub async fn withdraw(&self, player_uuid: &str, amount: f64) -> Result<(), String> {
        // implementation
    }
}

#[plugin_method]
async fn on_load(&mut self, server: &Context) -> Result<(), String> {
    let economy_api = EconomyApi::new();
    
    // Register API service
    server.register_api_service("economy", Box::new(economy_api)).await?;
    
    Ok(())
}
```

  #### API Consumer
```rs
#[plugin_method]
async fn on_load(&mut self, server: &Context) -> Result<(), String> {
    // Wait for economy plugin to be available
    let economy_api = server.get_api_service::<EconomyApi>("economy").await
        .ok_or("Economy plugin not found")?;
    
    // Store reference for later use
    self.economy = Some(economy_api);
    
    Ok(())
}

// In command executor
async fn execute(&self, sender: &mut CommandSender, server: &Server, args: &ConsumedArgs) -> Result<(), CommandError> {
    if let Some(economy) = &self.economy {
        let balance = economy.get_balance(&player_uuid).await?;
        // use balance...
    }
}
```
  
</details>

### Asynchronous plugin loading and dependency awaiting
This addition changes how plugins are loaded. When the plugin is first picked up by a loader (any loader returns `true` from `can_load()`), the plugin is added to a state map with a state of `PluginState::Loading`. After that, the `on_load` method is called from a separate task so that multiple plugins can load at once without the whole server having to wait for one plugin to fully load before another one starts loading. 

Once the plugin's `on_load` method has completed, the state changes to either `PluginState::Loaded` if the method completed successfully, or to `PluginState::Failed(String)` with the error returned.

The PluginManager also has several new methods to go with this new loading system and to allow plugins to wait for other plugins to fully load. For example, the new `wait_for_plugin` method allows a plugin to wait until another plugin is fully loaded:
```rs
#[plugin_method]
async fn on_load(&mut self, context: Arc<Context>) -> Result<(), String> {
    pumpkin::init_log!();

    log::info!("Waiting for PLua to load...");

    context.plugin_manager.wait_for_plugin("plua").await?; // Use the plugin's package name as specified in it's Cargo.toml
    
    log::info!("PLua loaded!");

    Ok(())
}
```

### DIFF
--- a/pumpkin-macros/src/lib.rs
+++ b/pumpkin-macros/src/lib.rs
@@ -100,8 +100,6 @@ pub fn send_cancellable(input: TokenStream) -> TokenStream {
             if let Some(cancelled_block) = cancelled_block {
                 quote! {
                     let event = crate::PLUGIN_MANAGER
-                        .read()
-                        .await
                         .fire(#event)
                         .await;
 
@@ -115,8 +113,6 @@ pub fn send_cancellable(input: TokenStream) -> TokenStream {
             } else {
                 quote! {
                     let event = crate::PLUGIN_MANAGER
-                        .read()
-                        .await
                         .fire(#event)
                         .await;
 
@@ -129,8 +125,6 @@ pub fn send_cancellable(input: TokenStream) -> TokenStream {
         } else if let Some(cancelled_block) = cancelled_block {
             quote! {
                 let event = crate::PLUGIN_MANAGER
-                    .read()
-                    .await
                     .fire(#event)
                     .await;
 
@@ -142,8 +136,6 @@ pub fn send_cancellable(input: TokenStream) -> TokenStream {
         } else {
             quote! {
                 let event = crate::PLUGIN_MANAGER
-                    .read()
-                    .await
                     .fire(#event)
                     .await;
             }
--- a/pumpkin/src/command/commands/plugin.rs
+++ b/pumpkin/src/command/commands/plugin.rs
@@ -36,8 +36,7 @@ impl CommandExecutor for ListExecutor {
         _server: &crate::server::Server,
         _args: &ConsumedArgs<'a>,
     ) -> Result<(), CommandError> {
-        let plugin_manager = PLUGIN_MANAGER.read().await;
-        let plugins = plugin_manager.active_plugins();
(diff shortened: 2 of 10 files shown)


Snowiiii:
Looks good. Thank you @vyPal 🕷️

PR #1081, opened 2025-07-27
Config set default values for values missing in file

FlorianLang06:
## Description

I implemented that the configs are merged with the default after loading. This means if a value is missing in the configuration it will be set with the default value. If a missing value is detected the config will be saved after the merge. Every value already set in the file will not be touched and will sill be the same.

Reason for the change:
- if a new config option will be added the user from now on doesn't need to delete the config/change it manually.
- you could set some of the values before the first start of the server

## Testing
If you want to test this feature just create the `configuration.toml` or `features.toml` and set one value. The other way would be to just delete one of the properties in a existing file.


jstindev:
bad idea

itzmxritz:
awesome

teknostom:
> bad idea

That is not an argument. I think this is a great idea.

jakubbbdev:
great idea

Snowiiii:
Seems like a good idea and i was thinking of that already for a long time, I would prefer if there would be some kind of console message like "config.toml changed..."

FlorianLang06:
> Seems like a good idea and i was thinking of that already for a long time, I would prefer if there would be some kind of console message like "config.toml changed..."

How should I do that? Because if I use the normal `log::info!()` its only working for `configration.toml` but not for `features.toml` because this is executed before the logger is initialized.

Snowiiii:
> > Seems like a good idea and i was thinking of that already for a long time, I would prefer if there would be some kind of console message like "config.toml changed..."
> 
> How should I do that? Because if I use the normal `log::info!()` its only working for `configration.toml` but not for `features.toml` because this is executed before the logger is initialized.

Oh, I mean i'm also fine with not including which file changed

FlorianLang06:
> > > Seems like a good idea and i was thinking of that already for a long time, I would prefer if there would be some kind of console message like "config.toml changed..."
> > 
> > 
> > How should I do that? Because if I use the normal `log::info!()` its only working for `configration.toml` but not for `features.toml` because this is executed before the logger is initialized.
> 
> Oh, I mean i'm also fine with not including which file changed

The problem is not that I can't show the name, the problem is that the normal log method only works in one of two cases. Because in the `init_log!()` macro is `LOGGER_IMPL` used and in there `advanced_config().logging.enabled` is used. This means in the case of loading the advanced_config the logger isn't initialized. 

I see 3 ways how we can solve that:
- only show info log for "configuration.toml"
- use `println!()`
- combine both and use `log::info!()` if possible and if not than `println!()` (than I need to know how I could find you which macro I should use)

Snowiiii:
> * println!()

Just use `println!()`

FlorianLang06:
image
@Snowiiii now you could check this PR again

Snowiiii:
Looks good now. Thank you @FlorianLang06 🐭

PR #1082, opened 2025-07-27
Fix Breaking the Block below Vegetation

CuzImClicks:
## Description

PlantBlockBase was added but not implemented for all types of foliage.

These two still have to be fixed but probably not in this pr:

- [ ] Lily Pad placement
- [ ] Sea Grass and Tall Sea Grass

## Testing

Yea

### DIFF
--- a/pumpkin/src/block/blocks/plant/short_plant.rs
+++ b/pumpkin/src/block/blocks/plant/short_plant.rs
@@ -1,8 +1,12 @@
 use async_trait::async_trait;
 use pumpkin_data::tag::Taggable;
 use pumpkin_data::{Block, tag};
+use pumpkin_world::BlockStateId;
 
-use crate::block::pumpkin_block::{BlockMetadata, CanPlaceAtArgs, PumpkinBlock};
+use crate::block::{
+    blocks::plant::PlantBlockBase,
+    pumpkin_block::{BlockMetadata, CanPlaceAtArgs, GetStateForNeighborUpdateArgs, PumpkinBlock},
+};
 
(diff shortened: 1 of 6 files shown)


pumpkin/src/block/blocks/plant/short_plant.rs
```diff
@@ -23,4 +27,19 @@ impl PumpkinBlock for ShortPlantBlock {
         block_below.is_tagged_with_by_tag(&tag::Block::MINECRAFT_DIRT)
             || block_below == &Block::FARMLAND
```
teknostom: This can be replaced by the trait right?

teknostom:
Other than my question above, LGTM.

Btw, what is wrong with the Lily Pad placement, and why can't that be fixed in this PR?

Also, we should probably rename `pos` in PlantBlockBase to `floor_pos`, to make it easier to tell that it is the floor that is being checked.

pumpkin/src/block/blocks/plant/short_plant.rs
CuzImClicks: yep good catch

CuzImClicks:
> Btw, what is wrong with the Lily Pad placement, and why can't that be fixed in this PR?

Well it literally doesn't work... none of the placement logic works when you place on water. I logged on all of the lily pad functions and they only get called when you try to place on a block below the water, not on just water. I tried to fix it but was just too tired yesterday to figure it out.

Snowiiii:
LGTM. Thank you @CuzImClicks :+1:  🐮

PR #1083, opened 2025-07-28
Implement heightmap

AyaSanae:
## Description
A HeightMap consists of a packed array of i64 type, where each 9 bits store a height value. Therefore, without considering concatenation, one i64 can store 7 height values.

The heightmaps.set() method adds world_min_y.abs() (which is 64 in the overworld)to the stored height value in the packed array to ensure it is positive.
Similarly, heightmaps.get() subtracts world_min_y.abs() from the value before returning it as the result.
For a given (x, z), if no valid y is found after iterating through all possible y values, the heightmap will be set to (x, min_y, z).

The log_heightmap() method can intuitively print the values in the heightmap for visualization.
like:
image
image

As shown in the figure:  
- **pale_oak_leaves** coordinates: `(2, 118, 1)`  
  - HeightMap for `(x,z) = (2,1)`:  
    - `WorldSurface`: 118  
    - `MotionBlocking`: 118  
    - `MotionBlockingNoLeaves`: 114  

- **cherry_sapling** coordinates: `(2, 118, 0)`  
  - HeightMap for `(x,z) = (2,0)`:  
    - `WorldSurface`: 118  
    - `MotionBlocking`: 117  
    - `MotionBlockingNoLeaves`: 117  

And in the End:
image
image

In the Nether, the heightmap is the same as in vanilla, at 127.

TODO:
    Figure out what caused the network issue.

## Testing
image

I directly modified get_top_block() and used the kill command for testing. It works fine in all dimensions.


AyaSanae:
For some reason, after executing rm -rf world && ./pumpkin, returning from the Nether or the End to the overworld still causes network issues—or the overworld appears 'empty' with chunks failing to load.

AyaSanae:
> For some reason, after executing rm -rf world && ./pumpkin, returning from the Nether or the End to the overworld still causes network issues—or the overworld appears 'empty' with chunks failing to load.

I don't think this issue was caused by this PR.
because: https://github.com/Pumpkin-MC/Pumpkin/issues/1084

Snowiiii:
> > For some reason, after executing rm -rf world && ./pumpkin, returning from the Nether or the End to the overworld still causes network issues—or the overworld appears 'empty' with chunks failing to load.
> 
> I don't think this issue was caused by this PR. because: #1084

Oh okay. Looks like this also happens on master. Thank you then again

Issue #1086, opened 2025-07-28
Server deadlock


spr-equinox:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

.

### Server Software Version/Commit

master

### What happened?

server is stuck

### To Reproduce

quit and rejoin for many times

https://github.com/user-attachments/assets/1f35c511-3cb2-4713-8794-d84403b8b777

### Expected behavior

run normally

sprinox:
i dump the server when it is stuck at stopping
```
2025-07-30 09:31:50 [ERROR] TASK 0:
2025-07-30 09:31:50 [ERROR] ╼ <tokio_util::task::task_tracker::TrackedFuture<F> as core::future::future::Future>::poll at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-util-0.7.15/src/task/task_tracker.rs:671:9
  └╼ pumpkin_world::level::Level::receive_chunks::{{closure}} at /workspaces/Pumpkin/pumpkin-world/src/level.rs:544:13
     └╼ <core::future::poll_fn::PollFn<F> as core::future::future::Future>::poll at /usr/local/rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/poll_fn.rs:151:9
        ├╼ pumpkin_world::level::Level::receive_chunks::{{closure}}::{{closure}} at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.47.0/src/macros/select.rs:708:49
        │  └╼ pumpkin_world::level::Level::fetch_chunks::{{closure}} at /workspaces/Pumpkin/pumpkin-world/src/level.rs:902:14
        │     └╼ <core::pin::Pin<P> as core::future::future::Future>::poll at /usr/local/rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/future/future.rs:124:9
        │        └╼ <pumpkin_world::chunk::io::file_manager::ChunkFileManager<S> as pumpkin_world::chunk::io::FileIO>::fetch_chunks::{{closure}} at /workspaces/Pumpkin/pumpkin-world/src/chunk/io/file_manager.rs:253:45
[...]

PR #1091, opened 2025-07-30
Solve deadlock

spr-equinox:
## Description

Solve #1086

According to dump log, i think server is stuck at notify.notified().await.

notify.notified() will never awake after notify_waiters has been called.
So this may happen
```
T = thread

T1 let mut locks = self_clone.chunk_generation_locks.lock().await;
T1 let Some(existing) = locks.get(&pos)  pumpkin-world/src/level.rs:800
T1 Some(existing.clone())
T1 // chunk_generation_locks unlock

T2 let mut locks = self_clone.chunk_generation_locks.lock().await;
T2 if let Some(notify) = locks.remove(&pos) {
T2 notify.notify_waiters();

T1 if let Some(notify) = notify {
T1 notify.notified().await; // forever waiting here. nobody can send notify
```

I know we can solve by using condition variable, but tokio doesn't have that.
Using `watch` may be a bad solution and I can't find out a right place to drop `Receiver<bool>`, but i can't find any other way.

## Testing


CuzImClicks:
```diff
diff --git a/pumpkin-world/src/level.rs b/pumpkin-world/src/level.rs
index 8156423f..531b6bd6 100644
--- a/pumpkin-world/src/level.rs
+++ b/pumpkin-world/src/level.rs
@@ -860,7 +860,7 @@ impl Level {
                             handle.spawn(async move {
                                 let mut locks = self_clone.chunk_generation_locks.lock().await;
                                 if let Some(notify) = locks.remove(&pos) {
-                                    notify.notify_waiters();
+                                    notify.notify_one();
                                 }
                             });
                         }
```

i cannot test this but notify_one() stores a permit so the next task future task that comes along can immediately consume that permit

sprinox:
there may be multiple waiters. The reason we need to use notify is that there may be more than one thread require for a same chunk.

Snowiiii:
For some reason it looks like there is still a dead lock, chunks load very slow and when rejoin they don't even load

sprinox:
> For some reason it looks like there is still a dead lock, chunks load very slow and when rejoin they don't even load

is cpu usage very high? Like almost all the cores are 100% usage.

The deadlock I encountered caused the program to use almost no CPU. After joining the world, no chunks would load at all—I just kept falling into the void, and the server couldn’t be stopped.

Snowiiii:
> The deadlock I encountered caused the program to use almost no CPU. After joining the world, no chunks would load at all—I just kept falling into the void, and the server couldn’t be stopped.

I will get empty chunks and fall into the void

sprinox:
> > The deadlock I encountered caused the program to use almost no CPU. After joining the world, no chunks would load at all—I just kept falling into the void, and the server couldn’t be stopped.
> 
> I will get empty chunks and fall into the void

When chunk loading is slow or stalls entirely, and the client falls into the void while the CPU usage remains high, I believe it's not a deadlock. It's more likely a typical case where the server is overwhelmed with chunk loading tasks.

`2025-07-30 14:01:54 [DEBUG] (4) pumpkin::world: Chunks queued after 1574340ms`

I'm running the server in a GitHub Codespace, and I flew a long distance in-game. Although the server was barely responsive and the client disconnected shortly after falling into the void, the CPU usage remained at 100%, and chunks were still being generated—just extremely slowly.

Snowiiii:
LGTM. Thank you @spr-equinox 🐆

PR #1095, opened 2025-07-31
Fix damage & death msg & portal delay

zly2006:
## Description

Fix damage sound & death message display.

## Testing

Fall into void.


Snowiiii:
Looks good. Thank you @zly2006 🐄

PR #1098, opened 2025-08-01
Refactor entity argument

zly2006:
## Description

Entity command arguments are using `Arc<Player>` instead of `Arc<dyn EntityBase>`. So in this PR:

- [x] Support `@e` `@n`
- [x] Fix entity argument usage to support all entities
- [x] Move NBT serde fron EntityBase to its super trait NBTStorage, for consistent behavior like block entity etc.
- [x] /kill
- [x] /tp
- [x] /clear (refactor args)
- [x] entity type filter & limit & sort (`@e[type=xxx, !type=xxx]`)

Other entity selector(filter) will not be implemented int his PR.

## Testing

```
/summon zombie
/kill @e
/kill @e[type=zombie]
```

### DIFF
--- a/pumpkin/src/entity/mod.rs
+++ b/pumpkin/src/entity/mod.rs
@@ -181,7 +181,12 @@ pub trait EntityBase: Send + Sync + NBTStorage {
 
     /// Kills the Entity.
     async fn kill(&self) {
-        self.damage(f32::MAX, DamageType::GENERIC_KILL).await;
+        if let Some(living) = self.get_living_entity() {
+            living.damage(f32::MAX, DamageType::GENERIC_KILL).await;
+        } else {
+            // TODO this should be removed once all entities are implemented
+            self.get_entity().remove().await;
+        }
     }
 
     /// Returns itself as the nbt storage for saving and loading data.


Snowiiii:
When executin `/kill @e[type=minecraft:zombie]` it kills me as well

zly2006:
> When executin `/kill @e[type=minecraft:zombie]` it kills me as well

Try removing `minecraft:`?

Snowiiii:
> > When executin `/kill @e[type=minecraft:zombie]` it kills me as well
> 
> Try removing `minecraft:`?

Build is now failing

zly2006:
> > Try removing `minecraft:`?
> 
> Build is now failing

The latest commit should be fine. `minecraft:` may cause issues because parsing entity type from name does not strip this prefix, and yes there is no check for invalid entity types, I will fix it.

zly2006:
Done

fucksophie:
Should strip any prefix. Plugins/mods could add new prefixes.

zly2006:
> Should strip any prefix. Plugins/mods could add new prefixes.

Its absolutely OK for now. We dont have an API for adding entities and all entity types are generated when compiling.

zly2006:
@Snowiiii Could you please test and review this PR? All bugs are fixed.

Snowiiii:
For some reason when spawning an Creeper via an Spawn Egg, it does not kill the creeper via the Command has this to do with the Creeper not being implemented via LivingEntity and just being a Entity
image

zly2006:
> has this to do with the Creeper not being implemented via LivingEntity and just being a Entity 

Its possible, I will debug.

zly2006:
@Snowiiii We even have not implemented creeper entity, so currently it is an `Entity`, this is not a bug.

You can use an egg to spawn zombie and it can be killed.

Snowiiii:
> so currently it is an `Entity`

My Question was if currently only Living Entities are Killed and Entities not ?, Because a Creeper if not a Living Entity but an `Entity`, Also does this affect other non Living Entities like Item's or Paintings ?

zly2006:
> My Question was if currently only Living Entities are Killed and Entities not ?

Only living entities.

> Also does this affect other non Living Entities like Item's or Paintings ?

Yes.

Snowiiii:
> > My Question was if currently only Living Entities are Killed and Entities not ?
> 
> Only living entities.
> 
> > Also does this affect other non Living Entities like Item's or Paintings ?
> 
> Yes.

> Yes.

I guess this should be fixed ?

zly2006:
> I guess this should be fixed ?

So for normal entities, they are removed arbitrarily, and for livings, they are killed and drops. Im gonna do this.

Snowiiii:
Looks good now. Thank you @zly2006 :+1: 🐺

Issue #1103, opened 2025-08-02
bug


FJKiXfaR:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

win server 2019

### Server Software Version/Commit

a4180e0512f8eacd656f8f7657cde4757778896e

### What happened?

PS C:\Users\Administrator> cd D:\minecraft\Pumpkin\target\release
PS D:\minecraft\Pumpkin\target\release> D:\minecraft\Pumpkin\target\release\pumpkin.exe
2025-08-02 22:36:55 [INFO] (1) Starting Pumpkin 0.1.0-dev+1.21.8 Minecraft (Protocol 772)
2025-08-02 22:36:55 [WARN] Pumpkin is currently under heavy development!
2025-08-02 22:36:55 [INFO] (1) Report issues on https://github.com/Pumpkin-MC/Pumpkin/issues
2025-08-02 22:36:55 [INFO] (1) Join our Discord for community support: https://discord.com/invite/wT8XjrjKkf
2025-08-02 22:36:55 [WARN] Failed to get level_info, using default instead: Info not found!
2025-08-02 22:36:55 [WARN] Failed to load favicon from 'icon.png': not found; using default.
2025-08-02 22:36:55 [INFO] (1) Loading Overworld: -4208046055586752277
2025-08-02 22:36:55 [INFO] (1) Loading Nether: -4208046055586752277
2025-08-02 22:36:55 [INFO] (1) Loading End: -4208046055586752277

thread 'main' panicked at D:\minecraft\Pumpkin\pumpkin\src\lib.rs:206:10:
Failed to start `TcpListener`: Os { code: 10013, kind: PermissionDenied, message: "以一种访问权限不允许的方式做了一个访 问套接字的尝试。" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

### To Reproduce

build and run

### Expected behavior

can start

PR #1104, opened 2025-08-02
Improve error message on  binding failure

FJKiXfaR:
## Description
When I was running it, I found that the port was occupied, so I want to add a prompt.[issues/1103](https://github.com/Pumpkin-MC/Pumpkin/issues/1103)
## Testing

### DIFF
--- a/pumpkin/src/lib.rs
+++ b/pumpkin/src/lib.rs
@@ -207,7 +207,7 @@ impl PumpkinServer {
             Ok(l) => l,
             Err(e) => {
                 panic!(
-                    "faild to bind TCP port: {}, error: {}.",
+                    "failed to bind TCP port: {}, error: {}.",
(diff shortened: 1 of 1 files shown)


unschlagbar:
its better if you catch the error when binding the socket, just replace expect with a match

Snowiiii:
This does not check the port only, Also please use a better PR title

FJKiXfaR:
> This does not check the port only, Also please use a better PR title

Improved the title.

Snowiiii:
And again saying "failed to bind TCP/UDP port:" is wrong, It is not only about the port but also about the Address

FJKiXfaR:
> And again saying "failed to bind TCP/UDP port:" is wrong, It is not only about the port but also about the Address

Is there anything else that needs to be changed?

PR #1107, opened 2025-08-03
feat: Drop player inventory & mob experience

zly2006:
## Description

as title.

**Why so many `dyn_self`s?**

In the case like this:

`Player::damage -> LivingEntity::set_health -> LivingEntity::on_death -> Player::drop_loot`, without dyn_self, the vtable pointer is lost and may never invoke `Player::drop_loot`.

## Testing


Snowiiii:
@zly2006 Please do more Testing before submitting PR's,
All Mobs (e.g. Zombies) don't drop Loot anymore when Killed

TheDarkSword:
Instead of dyn_self you can use self.get_player and if it return Some execute player method (or what you want)

zly2006:
> Instead of dyn_self you can use self.get_player and if it return Some execute player method (or what you want)

Hi I don't think this would work. When calling `LivingEntity::damage` in `Player::damage`, the data pointer is already pointed to the living entity and the vtable pointer is lost, so `get_player` will return None even it is actually a player.

zly2006:
> @zly2006 Please do more Testing before submitting PR's, All Mobs (e.g. Zombies) don't drop Loot anymore when Killed

@Snowiiii  Sorry,  have fixed them and added mob & player experience drop, please review them.

I think maybe we need a integration test instead of testing them manually?

Snowiiii:
> > @zly2006 Please do more Testing before submitting PR's, All Mobs (e.g. Zombies) don't drop Loot anymore when Killed
> 
> @Snowiiii Sorry, have fixed them and added mob & player experience drop, please review them.
> 
> I think maybe we need a integration test instead of testing them manually?

Maybe we could add some test which Spawns in an Mob or Animal Entity and then check if there is an Item dropped when killed

pumpkin/src/entity/mob/mod.rs
```diff
@@ -57,6 +60,55 @@ impl MobEntity {
...
+            10
+        } else if entity_type.id == EntityType::PIGLIN_BRUTE.id
+            || entity_type.id == EntityType::RAVAGER.id
+        {
+            20
+        } else {
+            0
+        }
+    }
 }
```
Snowiiii, line 112: This looks ugly, is this `protected int getExperienceToDrop(ServerWorld world)` in Java ?, When yes you defenitly want to implement that as in the EntityBase trait

zly2006: Other entities are not implemented currently, how to implement them or only implement zombies?

Snowiiii: > Other entities are not implemented currently, how to implement them or only implement zombies?

Only implement it for Zombie for now, this function is wrong anyway since there is often more logic involved

DaniD3v:
Why was this closed?

zly2006:
> Why was this closed?

I dont have enough time to further work on this. You can feel free to open a new PR.

PR #1108, opened 2025-08-04
Add carved pumpkin and jack o lantern roation

tom-devv:
## Description

Adds rotation logic to carved pumpkin and jack o lantern. They are sorta bundled like this in minecraft code too as they share most (all?) logic.

## Testing

Just place it in every direction


Snowiiii:
Looks good. Thank you @tom-devv 🦄

PR #1115, opened 2025-08-08
Rework chunk generation system

4lve:
## Description
Instead of spawning new tasks every time we want to generate chunks we instead have a set amount of threads that are ready to pick up chunk generation work at any time.

## Testing


HttpRafa:
What does this do? Does it speed up the generation?

Q2297045667:
> What does this do? Does it speed up the genMy guess is that it will limit the total number of threads for block generation calls, and I think adding a maximum number of threads to the configuration file is a good idea.

Not everyone has the same number of CPU cores.

Q2297045667:
> What does this do? Does it speed up the generation?

This is probably used to limit the number of threads for block generation calls. However, I still suggest adding a maximum thread limit in the configuration file.
Because not all users have the same number of cores.

Snowiiii:
@4lve can you please provide a benchmark comparison

sprinox:
try to add a priority system. https://github.com/4lve/Pumpkin/pull/2

Snowiiii:
Looks good. Thank you @4lve 🐁

PR #1119, opened 2025-08-12
build: Make file generation deterministic & skip unchanged files

Ancient77:
## Description
I noticed that running `cargo check`, `cargo build`, and `cargo clippy` back-to-back would always rewrite the generated files. This slowed down incremental builds because crates that depended on those files had to be recompiled every time.

This was happening for three main reasons:
- Changing tools doesn't honor `cargo:rerun-if-changed`
-  We used HashMaps, which don't guarantee a consistent order.
-  We always rewrote the output file, even if the content was identical.

This PR fixes this by switching all the HashMaps in the build functions to BTreeMaps to ensure the output is always in the same order. I also added a lint to prevent anyone from accidentally using a HashMap here again.

Additionally, the script now checks if the file content has actually changed before writing. If it's the same, we skip the write.

I also did some minor refactoring around the `rustfmt` formatting logic.

## Testing

I've tested this on Linux and it fixes the issue for me.

### DIFF
--- a/pumpkin-data/Clippy.toml
+++ b/pumpkin-data/Clippy.toml
@@ -1,0 +1,3 @@
+disallowed-types = [
+    { path = "std::collections::HashMap", reason = "If you need a map, use BTreeMap for deterministic ordering. Otherwise it forces a rebuild after every change.", replacement = "std::collections::BTreeMap" }
+]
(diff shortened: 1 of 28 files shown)


Snowiiii:
Looks good. Thank you @Ancient77 :D 🐯

PR #1121, opened 2025-08-12
fix world y offset & bedrock update

unschlagbar:
## Description
Updated Bedrock to 1.21.100.
Fixed the World having a 64 blocks y offset.
Chunks now have the same section count as before if there get unloaded and loaded again.
Added logic to send muliple bedrock packets in a frameSet (kinda wip still).
Bedrock pos packets now only get proceed if the client loading finished.

## Testing
The bedrock clients stops rendering chunks, if you are a bit fast (need help, can't figure out why, spend so much hours debuging this)


CuzImClicks:
Banger commit messages as always🔥

Snowiiii:
Could you include the movement fix in this PR please

Snowiiii:
when trying to join with bedrock
```
thread 'tokio-runtime-worker' panicked at /home/alex/Documents/Development/Rust/Pumpkin/pumpkin-protocol/src/bedrock/client/level_chunk.rs:30:9:
                        assertion `left == right` failed
                                                          left: 26
                                                                   right: 24
                                                                            note: run with `RUST_BACKTRACE=1` environment 
 ```

Snowiiii:
Looks good. Thank you @unschlagbar 🦂

PR #1124, opened 2025-08-13
Fix redstone wire with non-solid blocks

BramOtte:
## Description
This pull requests aims to fix issues regarding redstone wire and non-solid blocks.
Specifically it makes it so blocks like glass and slabs don't block redstone wire from going diagonal and only allow wire signals to move up but not down them.
Additionally it corrects some missuses of the is_solid property.

This is achieved by adding the is_solid_block property to blocks.json, the required changes to the extractor can be found here: https://github.com/Pumpkin-MC/Extractor/pull/32
I also moved has_random ticks to stateFlags, this isn't really needed but it makes the blocks.json file a bit smaller.

## Testing
All existing unit tests seem to pass, I have not yet added any of my own automated tests.
I did test manually in game and it seems to work, see examples.
Additionally I briefly looked into the java source code to make sure the behaviors match up.

## Examples
Redstone should only move up slabs and glass
slabs

Redstone should be able to go through slabs and glass diagonally
slabs2

Counter which was previously non-functional now works
counter

### DIFF
--- a/pumpkin-data/src/block_state.rs
+++ b/pumpkin-data/src/block_state.rs
@@ -123,6 +123,19 @@ impl BlockState {
 
         Some(shapes)
     }
+
+    /// Returns whether a block prevents a wire from connect diagonally.
+    pub fn blocks_wire(&self) -> bool {
+        // TODO: check minecraft java code to see if this is correct
+        self.is_solid() && self.opacity == 15
+    }
+
+    /// Returns whether a block if below a wire prevents its signal going down.
+    /// For java this is basically just the inverse of blocks_wire,
+    /// but on bedrock redstone should be able to go down solid sides of blocks including glass
+    pub fn blocks_wire_down(&self) -> bool {
+        !self.blocks_wire()
+    }
 }
 
 #[derive(Clone, Debug)]
--- a/pumpkin-data/src/block_state.rs
+++ b/pumpkin-data/src/block_state.rs
@@ -6,7 +6,7 @@ use crate::{Block, BlockDirection, CollisionShape};
 #[derive(Debug)]
 pub struct BlockState {
     pub id: u16,
-    pub state_flags: u8,
+    pub state_flags: u16,
     pub side_flags: u8,
     pub instrument: Instrument,
     pub luminance: u8,
@@ -14,7 +14,6 @@ pub struct BlockState {
     pub hardness: f32,
     pub collision_shapes: &'static [u16],
     pub outline_shapes: &'static [u16],
-    pub has_random_tick: bool,
     /// u8::MAX is used as None
     pub opacity: u8,
     /// u16::MAX is used as None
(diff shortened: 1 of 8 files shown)


pumpkin-data/src/block_state.rs
```diff
@@ -6,15 +6,14 @@ use crate::{Block, BlockDirection, CollisionShape};
 #[derive(Debug)]
 pub struct BlockState {
     pub id: u16,
-    pub state_flags: u8,
+    pub state_flags: u16,
```
Snowiiii, line 9: I don't like this, lets use a u8 please

BramOtte: Should I make a state_flags2 then? alternatively I can also make it 2 separate booleans or have stateflags be a [u8; 2].

CuzImClicks:
If only all of our pr descriptions were this detailed😁

pumpkin-data/src/block_state.rs
Snowiiii: I'm talking about the higher memory usage, using `[u8; 2]` would change nothing. Currently the better option would be properly to just add two booleans. 
@unschlagbar What do you think ?

BramOtte: If its for memory usage, I mean either way its probably not going to matter too much.
but adding a state_flags2 would be most efficient. it would require 2 bytes for state_flags and the other 2 flags to be represented and these 2 bytes would also be able to be placed anywhere in the struct with an align of 1.

having the state_flags be u8 and 2 booleans would take up 3 bytes but everything would only have an align of 1. Main advantage of this approach is however that it requires the least changes to the code currently on master.

having the state_flags u16 would take up 2 bytes but have an align of 2 potentially increasing the size of the struct.

There is also the question of how to represent it in the json, just having the json be a 1 to 1 representation would be least confusing (and probably what i'd go for), but putting all the flags into 1 number could make potentially make it smaller (less keys to serialize)

pumpkin-data/src/block_state.rs
```diff
@@ -123,6 +134,18 @@ impl BlockState {
...
+    pub fn blocks_wire(&self) -> bool {
+        self.is_solid_block()
+    }
+
+    /// Returns whether a block if below a wire prevents its signal going down.
+    /// For java this is basically just the inverse of blocks_wire,
+    /// but on bedrock redstone should be able to go down solid sides of blocks including glass
+    pub fn blocks_wire_down(&self) -> bool {
+        !self.is_solid_block()
+    }
```
Snowiiii: Are these methods exist in Vanilla too ?, I'm not sure if we really want to have them in block state, we could also just the raw `is_solid_block` function

BramOtte: No vanilla does not have these methods, I made these methods before I looked at the vanilla code, and left them in since bedrock does things slightly differently and if would be easier to know what part of the code to change this way.
But considering supporting bedrock mechanics isn't really a goal right now I can remove them.

pumpkin-data/src/block_state.rs
```diff
@@ -6,7 +6,8 @@ use crate::{Block, BlockDirection, CollisionShape};
 #[derive(Debug)]
 pub struct BlockState {
     pub id: u16,
-    pub state_flags: u16,
+    pub state_flags: u8,
+    pub state_flags2: u8,
```
Snowiiii: Okay fine, lets use u16 and only one state_flags

pumpkin-data/src/block_state.rs
Snowiiii: It would be better to have them removed

Snowiiii:
Looks good. Thank you @BramOtte :D 🦢

PR #1125, opened 2025-08-13
fix: Make all client packet members public.

utfunderscore:
## Description
This pull request makes all parameters of client packets public and accessible to any projects that may use pumpkin for the client protocol. BotMark makes use of the protocol but does not currently need access to any of these packets, but it should open the opportunity for people to use it that way.


Snowiiii:
Looks good. Thank you @utfunderscore :D 🦜

PR #1126, opened 2025-08-14
Fix broken nix flake, add rust-analyzer to devshell and move to fenix from rust-overlay

CrabbyDisk:
## Description
Like the title says, this fixes the flake that previously didn't break, adds rust-analyzer to the nix devshell and moves to fenix (a rust toolchain provider) from rust-overlay.

## Testing
This doesn't affect any of the code and I have successfully built the nix package which should be reproducible.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,5 +1,26 @@
 {
   "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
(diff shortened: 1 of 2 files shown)


Snowiiii:
LGTM. Thank you @CrabbyDisk 🦏

PR #1127, opened 2025-08-14
Add block light

teknostom:
## Description

This PR adds block light to the server :)

I have tested a bunch, but might still have some issues. But from my testing it works really well.

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -34,7 +34,6 @@ use crate::{
 };
 use crate::{block::BlockEvent, entity::item::ItemEntity};
 use border::Worldborder;
-use bytes::BufMut;
 use crossbeam::queue::SegQueue;
 use explosion::Explosion;
 use pumpkin_config::BasicConfiguration;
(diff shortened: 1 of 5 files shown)


pumpkin/src/world/mod.rs
```diff
@@ -2380,6 +2386,180 @@ impl World {
         .await;
     }
 
+    pub fn queue_decrease(self: &Arc<Self>, pos: BlockPos, level: u8) {
```
CuzImClicks: What a great day to start documenting functions! (Also should probably have a more descriptive name to go along with it)

CuzImClicks (commented):
Great work :)

teknostom:
The vertical skylight now works fully at initialization. I won't be able to work FOSS for a couple weeks/months so if anyone wants to take over the PR, they are welcome to.

teknostom:
Current problems:

- No horizontal sky light.
- Light updates are slow af. (Maybe look at starlight)
- Block light needs to be initialized for each section, to check if there are current blocks with luminance.

teknostom:
I commented out sky light, because right now there is no way to propagate through chunks. So, we can jsut add this later. So i think we can merge this in now.

Snowiiii:
@teknostom, Is this still active ?

PR #1128, opened 2025-08-15
fix: Remove liftimes from CStatusResponse and CPlayerPosition

utfunderscore:
## Description
I have removed lifetimes from CPlayerPosition and CStatusResponse. Input on what should happen to the other client packets with lifetimes is needed.


utfunderscore:
Some are still some remaining client packets with lifetimes, but a deeper look is needed. For now, i have updated the low hanging fruit where the data did not need to be a lifetime to begin with.

Snowiiii:
I wonder if we could use Cow for this

utfunderscore:
> I wonder if we could use Cow for this

The current changes I've made here  have no effect on memory as the memory is created, and is only ever used in the packet then gets dropped. 
Cow also needs a lifetime still which defeats the purpose of these changes.

Snowiiii:
Looks good. Thank you @utfunderscore 🐨

PR #1143, opened 2025-08-20
add enchant command

sprinox:
## Description

add enchant command
should work as same as vanilla but only work on player

## Testing

test in game

### DIFF
--- a/pumpkin/src/command/commands/enchant.rs
+++ b/pumpkin/src/command/commands/enchant.rs
@@ -9,7 +9,6 @@ use crate::command::tree::CommandTree;
 use crate::command::tree::builder::argument_default_name;
 use crate::command::{CommandError, CommandExecutor, CommandSender};
 use pumpkin_data::data_component_impl::EnchantmentsImpl;
-use pumpkin_util::text::color::{Color, NamedColor};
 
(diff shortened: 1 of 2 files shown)


pumpkin/src/command/commands/enchant.rs
```diff
@@ -0,0 +1,174 @@
...
+            let lock = player.inventory.held_item();
+            let mut item = lock.lock().await;
+            if item.is_empty() {
+                if only_one {
+                    let msg = TextComponent::translate(
+                        "commands.enchant.failed.itemless",
+                        [targets[0].get_display_name().await],
+                    );
+                    sender.send_message(msg).await;
+                    return Ok(());
```
Snowiiii: Instead of this, you should return this
```rust
return Err(CommandError::CommandFailed(Box::new(
                TextComponent::translate("argument.pos.outofbounds", []),
            )));
```
this also now makes the error message red like in Vanilla

sprinox: I think so too, but I've found that many other commands do this as well.
sender.send_message also will cause difficulty to implement command block
will commit tomorrow

Snowiiii:
Looks good. Thank you @spr-equinox :D 🦃

PR #1147, opened 2025-08-23
use HungerManager::eat for cakes

CuzImClicks:
## Description

As requested

## Testing

Shouldnt be controversial


Snowiiii:
still wrong, looks at vanilla

CuzImClicks:
But like the wiki says these values and ingame it matches vanilla

Snowiiii:
> But like the wiki says these values and ingame it matches vanilla

I trust code more than the wiki, I think it also mentions that you should use it as a modifer and not the value itself

PR #1151, opened 2025-08-26
Optimized redundant logic and added vanilla Simplified Chinese

FandMC:
This is my first time submitting code, and I'm not good at describing it. I hope you can be patient with me

## Description

As requested

## Testing


Q2297045667:
由原来的加载全部文件改成了加载单个文件。
哈基米，你这家伙。

Snowiiii:
I'm not sure if this is the right place for this. Vanilla translations are quiet large and would heavy impact the executable size. The current translations are custom translations only

Q2297045667:
> I'm not sure if this is the right place for this. Vanilla translations are quiet large and would heavy impact the executable size. The current translations are custom translations only

Maybe we shouldn't compile it into an executable file. I think the external file should be loaded when the program starts.

Snowiiii:
@FandMC, Sorry but i think for now we should only stick with en_us, maybe we can make a selector in the future

PR #1172, opened 2025-09-14
feat: implement mangrove root block & barrier block

Aditeya:
## Description

I have implemented the mangrove root block and the barrier block, both of which have very similar implementations as they use the `MangroveRootsLikeProperties`.

### Barrier Block

---

Please do let me know if there is anything more to do.


Snowiiii:
LGTM. Thank you @Aditeya 🦏

PR #1182, opened 2025-10-05
fix(auth): Implement Secure Bedrock JWT Chain Validation

harihar-nautiyal:
## Description:

This pull request completely overhauls the Bedrock Edition login handler to be secure, robust, and correct. The previous implementation was a placeholder that was vulnerable to server crashes and, most critically, did not perform any cryptographic verification of the player's identity, allowing anyone to impersonate any player.

This PR fixes these issues by implementing the full `x5u` certificate chain validation flow as required by the protocol.

#### **Key Changes:**

*   **Implements Full Cryptographic Verification:**
    *   The login handler now performs a step-by-step cryptographic verification of the three-token JWT chain sent by the client.
    *   It uses the hardcoded Mojang ECDSA public key as the anchor of trust to validate the integrity of the entire chain, preventing player impersonation attacks.
    *   The signature of each token is manually verified using the `p384` and `spki` crates to ensure correctness, mirroring the logic of trusted libraries like Cloudburst's protocol implementation.

*   **Correct and Robust Payload Parsing:**
    *   Replaced all panicking code (e.g., `.unwrap()`, direct indexing) with proper `Result`-based error handling. The server will no longer crash on malformed login packets.
    *   Correctly handles the double-encoded JSON payload where the certificate chain is embedded as a string within a larger JSON object.

*   **Comprehensive Error Handling:**
    *   Introduced a detailed `AuthError` enum to provide specific, actionable error messages for every step of the validation process (e.g., `SignatureMismatch`, `BrokenChainLink`, `UntrustedMojangKey`). This will make future debugging much easier.

#### **How This Was Tested:**

The invalid token chain was debugged and validated using independent, external tools, including `openssl` and a series of Python scripts using the `PyJWT` library. These tests definitively proved:
1.  The `x5u` chain-of-trust was the correct authentication flow.
2.  The hardcoded Mojang public key was the correct anchor of trust.
3.  A final Python script that cleaned the tokens before verification was able to **successfully validate the entire chain**, confirming the logic implemented in this PR is correct.

This fix makes our Bedrock authentication secure and reliable.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "55248b47b0caf0546f7988906588779981c43bb1bc9d0c44087278f80cdb44ba"
 
 [[package]]
+name = "bedrock-jwt"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94de89773692f545fd487d926c5c293cabcee6226a8f47dd8bcf6e53067caebc"
+dependencies = [
+ "base64 0.22.1",
+ "ecdsa",
+ "p384",
+ "serde",
+ "serde_json",
+ "thiserror",
+]
+
+[[package]]
 name = "bitflags"
 version = "2.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -476,6 +490,12 @@ dependencies = [
  "tracing-core",
  "tracing-subscriber",
 ]
+
+[[package]]
+name = "const-oid"
+version = "0.9.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
 
 [[package]]
 name = "const-oid"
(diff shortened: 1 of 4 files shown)


unschlagbar:
Overall this looks good, but i don't like that we are using an extra crate you made for this. we have enough space in `pumpkin_utils`, i have nothing against a public crate but using it in pumpkin makes the the security parts a bit more decentralized and harder to spot insecurities and optimization opportunities, if you understand what i mean, but clearly its wwaayy better than what we have now. i sure you used some help from an ai for the description ;)

unschlagbar:
already found a piece in you code that could be optimized:
```
    let mut seq = Vec::new();
    seq.push(0x02);
    seq.push(r_enc.len() as u8);
    seq.extend_from_slice(&r_enc);
    seq.push(0x02);
    seq.push(s_enc.len() as u8);
    seq.extend_from_slice(&s_enc);

    let mut der = Vec::new();
    der.push(0x30);
    der.push(seq.len() as u8);
    der.extend_from_slice(&seq);
```
you can easily calculate the required capacity to save multiple allocations

harihar-nautiyal:
sure I will implement chain validation in pumpkin_utils and optimize it.

unschlagbar:
LGTM

Snowiiii:
Looks good. Thank you @harihar-nautiyal :D 🦌

Issue #1183, opened 2025-10-05
Copper bars not working


JunkyDeveloper:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Ubuntu 25.04

### Server Software Version/Commit

58583da4

### What happened?

Copper bars are not working

### To Reproduce

place copper bars

### Expected behavior

copper bars are not connecting

JunkyDeveloper:
I currently learning Rust and to learn a language you need to have a project to learn.
So i would like to fix this here, I say iron bars are working so that means i can copy it and bring it quickly working while try to understand the project

JunkyDeveloper:
Adding more and more copper blocks to it

PR #1185, opened 2025-10-06
Copper bars

JunkyDeveloper:
## Description
Added Copper Bars and functionality with refactoring of iron bars. Should be the behavior of iron bars! (Solves #1183
)

I removed `other_block == block` in the connected check because for me that doesn't make much sense. But i can add it back if someone knows why this was there in the first place because no changed behavior. (line 48 in bars.rs)

image

## Important
Waxing and wax removing doesn't work and oxidation also. If i checked correctly, no real copper was implemented, so should be done first and then add it here to it

## Testing


JunkyDeveloper:
On my machine the command `cargo fmt --check` runs through I have no idea why it fails here!
Also the results from the tests:
failures:
    generation::proto_chunk::test::test_no_blend_no_beard_badlands2
    generation::proto_chunk::test::test_no_blend_no_beard_frozen_ocean
    generation::proto_chunk::test::test_no_blend_no_beard_surface
    generation::proto_chunk::test::test_no_blend_no_beard_surface_badlands
    generation::proto_chunk::test::test_no_blend_no_beard_surface_badlands2
    generation::proto_chunk::test::test_no_blend_no_beard_surface_badlands3
    generation::proto_chunk::test::test_no_blend_no_beard_surface_biome_blend
    generation::proto_chunk::test::test_no_blend_no_beard_surface_frozen_ocean

test result: FAILED. 59 passed; 8 failed; 0 ignored; 0 measured; 0 filtered out; finished in 175.87s

I don't think it's on my here because I added only a block

PR #1188, opened 2025-10-06
create gh release on each push

jarjk:
# pr

## commits

- **ci: pre-release a nightly build on every push**
- **ci: upload only bin, with name `pumpkin-<arch>-<os>`**

## Description

creates a `Nightly Build` named GitHub Release on each push to `master`
also, the uploaded and now released artifacts are renamed by the template defined above, eg: `pumpkin-X64-Linux`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -68,9 +68,11 @@ jobs:
       - name: Export executable
         uses: actions/upload-artifact@v4
         with:
-          name: pumpkin-${{ matrix.os }}
+          name: pumpkin-${{ runner.arch }}-${{ runner.os }}
           compression-level: 9
-          path: target/${{ matrix.target }}/release/pumpkin*
+          path: |
+           target/release/pumpkin
+           target/release/pumpkin.exe
   clippy_release:
     name: Run lints in release mode
     runs-on: ubuntu-latest
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -84,3 +84,26 @@ jobs:
       - run: rustup component add clippy rustfmt --toolchain ${{ matrix.toolchain }}
       - uses: Swatinem/rust-cache@v2
       - run: cargo clippy --release --all-targets --all-features
+  draft_release:
+    permissions:
+      contents: write
+    needs: [build_release]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download prebuilt binary artifacts
+        uses: actions/download-artifact@v5
+        with:
+          merge-mulitple: true
+      - name: Generate release notes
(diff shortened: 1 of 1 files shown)


Snowiiii:
I don't like the idea of having this repo full of releases, i think we should may host them in a separate repo [like Servo does](https://github.com/servo/servo-nightly-builds)

jarjk:
indeed, luckily this action overwrites the `nightly` tag on each commit to `master`, this way only one release will be ever published, which is always built from the tip of the master branch 

is it fine that way?

jarjk:
also, what about building for more targets, such as macos-arm and linux-arm?
would be pretty obvious to implement, it's just

```diff
diff --git a/.github/workflows/rust.yml b/.github/workflows/rust.yml
index c72a2ee..d6e5759 100644
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,7 +41,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest, ubuntu-24.04-arm]
         toolchain:
           - stable
     steps:
```

jarjk:
> I don't like the idea of having this repo full of releases, i think we should may host them in a separate repo [like Servo does](https://github.com/servo/servo-nightly-builds)

@Snowiiii, also worth noting, that it's set as a pre-release, so it isn't something that would bother many

some examples of releasing this way:
- [yazi's nightly](https://github.com/sxyazi/yazi/releases/tag/nightly)
- [ghostty's tip](https://github.com/ghostty-org/ghostty/releases/tag/tip)

Snowiiii:
> also, what about building for more targets, such as macos-arm and linux-arm? would be pretty obvious to implement, it's just
> 
> ```diff
> diff --git a/.github/workflows/rust.yml b/.github/workflows/rust.yml
> index c72a2ee..d6e5759 100644
> --- a/.github/workflows/rust.yml
> +++ b/.github/workflows/rust.yml
> @@ -41,7 +41,7 @@ jobs:
>      runs-on: ${{ matrix.os }}
>      strategy:
>        matrix:
> -        os: [ubuntu-latest, windows-latest]
> +        os: [ubuntu-latest, windows-latest, macos-latest, ubuntu-24.04-arm]
>          toolchain:
>            - stable
>      steps:
> ```

Sure, why not

jarjk:
got it, check out <https://github.com/Pumpkin-MC/Pumpkin/actions/runs/19978663032>

Snowiiii:
Overall looks good. Great work, thank you @jarjk :D 🦜

PR #1196, opened 2025-10-13
Command data get block entity

JunkyDeveloper:
## Description
Added the /data get block <position> command is implemented only the offset from the position is incorrect because Position3DArgumentConsumer is wrong implemented, see for that the issue #1197


Snowiiii:
image

JunkyDeveloper:
Never checked that case

Snowiiii:
> Never checked that case

No surprise, no one testers their own PRs

Snowiiii:
So @JunkyDeveloper any plans to work further on this ?

JunkyDeveloper:
Not for the next two months

JunkyDeveloper:
Only wanted to say, I used that command but never used ~ because why should I want to get the properties of the nbt data of a block where I'm in. So if you use numbers it works.

Did you test all other commands if ~ works? didn't checked that and a code which I use from pumpkin converts the corts, so not my code which has this problem. So other commands should have the same problem or other commands replaced somehow ~

PR #1198, opened 2025-10-13
Replace some statics with consts

topi-banana:
## Description

These instances of `static` can be replaced with `const`, which is the appropriate choice for immutable, inlined constants.

## Testing


unschlagbar:
like this👍, the statics must be from someone that doesnt know rust to well

Snowiiii:
Looks good. Thank you @topi-banana :D

PR #1201, opened 2025-10-17
dynamic config

topi-banana:
## Description

Move **BasicConfiguration** and **AdvancedConfiguration** from **LazyLock** statics into the **Server** struct.

If we use static configurations,
When creating a Server programmatically, the current design prevents the use of different configurations per server instance because the configuration is static and shared.
This also applies to `#[test]` blocks within the code.

before:

```rust
static BASIC_CONFIG ...
static ADVANCED_CONFIG ...
````

after:

```rust
pub struct Server {
    basic_config: BasicConfiguration,
    advanced_config: AdvancedConfiguration,
}
```

## Testing


Snowiiii:
I'm not sure if this really has a benifit

topi-banana:
I believe moving the config from a static LazyLock into the Server struct does have real benefits, especially for testing:

- When tests need to override parts of the config, the current LazyLock approach reloads the file every time, so developers’ local config changes unintentionally affect the tests. By injecting a fixed config directly in the test (see the test changes in this PR), we ensure consistent, deterministic behavior.

- When running tests in parallel (--test-threads >1), modifying a global static config can cause race conditions and cross-test pollution. Having the config as a Server field makes each test completely isolated.

Snowiiii:
Overall looks good. Thank you @topi-banana :D 🦞

PR #1202, opened 2025-10-18
Organization

Blockaid0-0:
## Description
Added some simple organization like putting the Over world files into its own folder called DIM0 and a simple grammar mistake.
## Testing
Everything works the same


RealVaprant:
is that how Vanilla structures their worlds?

PR #1209, opened 2025-10-30
Update signature dependency to version 3.0.0-rc.2

txnnor:
## Description

RSA 0.10.0-rc.5 depends on signature 3.0.0-rc.2. I believe this is the root cause of plugin compilations failing. 

## Testing

I haven't tested, but I pulled it directly from RSA's 0.10.0-rc.5 update commit: https://github.com/RustCrypto/RSA/commit/472c98dd95c21cc123af77144bd2b65553ee8745


txnnor:
Closing PR because this alone does not fix the issue as "ecdsa" package still depends on signature 2.2.0. I also looked on crates.io and didn't see a version of ecdsa that goes with the signature version bump.

PR #1210, opened 2025-11-01
rewrite schedule thread

sprinox:
## Description

use a completely different way to schedule

it may improve performance and code readability 
i hope this can solve deadlock

## Testing


Snowiiii:
Can you please explain further what pros and cons this has ?

sprinox:
Pros:
Strictly tracks information for each chunk, with clear dependency relationships that make debugging easier.
No need to repeatedly check whether conditions are met and generation proceeds strictly by priority.
Changes are sent directly instead of being checked one by one.
Much higher readability; the old code was full of corner cases and had become difficult to maintain.

Cons:
None maybe

Snowiiii:
It looks like there is a memory leak and chunks get not deallocated

sprinox:
> It looks like there is a memory leak and chunks get not deallocated


I make chunk unload less frequently

From last pr message


> This PR does not touch entities or entity chunks. I’m not sure what mechanism is currently used for entities, but it causes them to keep being processed even when the player is far away. As a result, chunks are repeatedly requested, temporarily loaded, and then unloaded by the scheduling thread, which severely impacts performance. So, when testing performance, try freezing the game tick.

Snowiiii:
Overall looks good. Thank you @spr-equinox :D

PR #1214, opened 2025-11-13
fix: Fixing flake.

swagtop:
## Description

The current flake doesn't work, as the hash of the toolchain that `fenix` uses is out of date. Since Pumpkin is [developed on stable Rust][toolchain], there really isn't a reason to use `fenix` over the toolchain included in `nixpkgs` itself, which is always the stable version. 

[toolchain]: https://github.com/Pumpkin-MC/Pumpkin/blob/80d0482c787b13f45475f725f730c39752680846/rust-toolchain.toml

This could either be fixed by getting the toolchain with `fenix.packages.${system}.stable.toolchain`, as in using the stable channel through `fenix`, or removing `fenix` as a dependency and just getting it through `nixpkgs` for previously mentioned reasons.

This PR does the latter, and also scales back the role of Nix in the project, by moving the development shell stuff into the `flake.nix` itself, and removing the `nix/dev` directory, as there really shouldn't be any reasons to have two flakes in this project, and because it doesn't really seem like anyone is using whatever it is offering.

In short, merging this would fix the current broken flake, reduce the dependencies of it, and make it easier to update.

One could argue that the flake could be made even simpler, and have less dependencies, but some people might find using `naersk`, `flake-compat` and `flake-parts` useful, so I have not touched them.

## Testing

I have tested the flake with: `nix build`, `nix develop`, `nix flake check`, and used the development shell to do some developing. Everything looks good.


Snowiiii:
Looks good. Thank you @swagtop :D

PR #1231, opened 2025-12-08
fix:trait method signature mismatch for plugin_method

Sclock:
Background

Pumpkin’s Plugin trait requires plugin methods (e.g., on_load) to return:
```rust
Pin<Box<dyn Future<Output = Result<(), String>> + Send>>
```

However, the current implementation of the #[plugin_method] macro rewrites methods into an async fn, which compiles into:
```rust
impl Future<Output = Result<(), String>>
```

This causes a type mismatch when implementing the Plugin trait, resulting in compilation error E0053.

❗ Problem

When using:
```rust
#[plugin_method]
async fn on_load(...) -> Result<(), String> { ... }
```

the generated method has signature:
```rust
fn on_load(...) -> impl Future<...>
```
but the trait expects:
```rust
fn on_load(...) -> Pin<Box<dyn Future<...> + Send>>
```

This mismatch triggers:

error[E0053]: method `on_load` has an incompatible type for trait

✅ Solution

Modify the plugin_method macro to wrap the async block into a boxed future, producing the correct return type expected by the Plugin trait.

Key Changes

Replace the generated async fn with a regular fn returning:

```rust
Pin<Box<dyn Future<Output = Result<(), String>> + Send>>
```

Wrap the original async body using:

```rust
Box::pin(async move {
    crate::GLOBAL_RUNTIME.block_on(async move { ... })
})
```
Preserve the user-written method body while ensuring type compatibility.

📄 Summary of Changes

Updated the expansion of #[plugin_method] to generate a boxed future instead of impl Future.

Ensured all plugin methods rewritten by the macro now fully satisfy the Pumpkin Plugin trait signatures.

No changes required from plugin authors—existing macro usage continues to work.

🔍 Impact

Fixes compilation failure for any method tagged with #[plugin_method].

Makes macro behavior trait-compliant.

No breaking API changes.

### DIFF
--- a/pumpkin-api-macros/src/lib.rs
+++ b/pumpkin-api-macros/src/lib.rs
@@ -14,14 +14,16 @@ pub fn plugin_method(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let input_fn = parse_macro_input!(item as ItemFn);
     let fn_name = &input_fn.sig.ident;
     let fn_inputs = &input_fn.sig.inputs;
-    let fn_output = &input_fn.sig.output;
     let fn_body = &input_fn.block;
 
     let method = quote! {
-        #[allow(unused_mut)]
-        async fn #fn_name(#fn_inputs) #fn_output {
-            crate::GLOBAL_RUNTIME.block_on(async move {
-                #fn_body
+        fn #fn_name(#fn_inputs) -> std::pin::Pin<
+            Box<dyn std::future::Future<Output = Result<(), String>> + Send>
+        > {
+            Box::pin(async move {
+                crate::GLOBAL_RUNTIME.block_on(async move {
+                    #fn_body
+                })
             })
         }
     }


Sclock:
After testing, it can only solve the compilation problem, but it will get stuck during operation

2025-12-08 17:52:33 [INFO] (57) schedule thread start id: ThreadId(57) name: Schedule Thread
2025-12-08 17:52:33 [INFO] (1) Query protocol is enabled. Starting...
2025-12-08 17:52:33 [INFO] (6) Server query running on port 25565

Snowiiii:
> After testing, it can only solve the compilation problem, but it will get stuck during operation
> 
> 2025-12-08 17:52:33 [INFO] (57) schedule thread start id: ThreadId(57) name: Schedule Thread 2025-12-08 17:52:33 [INFO] (1) Query protocol is enabled. Starting... 2025-12-08 17:52:33 [INFO] (6) Server query running on port 25565

Any ideas why?, Would be great to have that fixed before merging

pumpkin-api-macros/src/lib.rs
```diff
@@ -14,14 +14,16 @@ pub fn plugin_method(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
-    let fn_output = &input_fn.sig.output;
     let fn_body = &input_fn.block;
 
     let method = quote! {
-        #[allow(unused_mut)]
-        async fn #fn_name(#fn_inputs) #fn_output {
-            crate::GLOBAL_RUNTIME.block_on(async move {
-                #fn_body
+        fn #fn_name(#fn_inputs) -> std::pin::Pin<
+            Box<dyn std::future::Future<Output = Result<(), String>> + Send>
```
vyPal, line 21: While it is true that all current plugin methods return a `Result<(), String>`, this may not be the case in the future. I'd personally prefer if the input function's return type was taken into account when generating the final function.

Sclock:
> > After testing, it can only solve the compilation problem, but it will get stuck during operation
> > 2025-12-08 17:52:33 [INFO] (57) schedule thread start id: ThreadId(57) name: Schedule Thread 2025-12-08 17:52:33 [INFO] (1) Query protocol is enabled. Starting... 2025-12-08 17:52:33 [INFO] (6) Server query running on port 25565
> 
> Any ideas why?, Would be great to have that fixed before merging

I'm sorry, my abilities are limited. Let's first address the issue with plugin_method

Sclock:
> > After testing, it can only solve the compilation problem, but it will get stuck during operation
> > 2025-12-08 17:52:33 [INFO] (57) schedule thread start id: ThreadId(57) name: Schedule Thread 2025-12-08 17:52:33 [INFO] (1) Query protocol is enabled. Starting... 2025-12-08 17:52:33 [INFO] (6) Server query running on port 25565
> 
> Any ideas why?, Would be great to have that fixed before merging

I have checked it and it seems that the problem is with on_load

Sclock:
The reason is the deadlock caused by `pumpkin::init_log!();`. Removing the log initialization in the plugin will prevent the failure to start

BjornTheProgrammer:
I've done my own investigation into this as well. Since dynamically included librarys have a different address space, the logger never is able to use the already defined logger in the pumpkin main process. Instead it must create its own first, but that is never done, so it just waits, this was exposed in https://github.com/Pumpkin-MC/Pumpkin/pull/1201.

The best way to fix this might be to remove the macro entirely, and to add the logger to the context instead. I'm working on that in my PR https://github.com/Pumpkin-MC/Pumpkin/pull/1241

Snowiiii:
fixed in https://github.com/Pumpkin-MC/Pumpkin/pull/1241

PR #1236, opened 2025-12-11
ci: build fully-static Linux binary with musl to fix glibc version errors

MiaM1ku:
### Problem  
The current nightly release (`pumpkin-X64-Linux`) is linked against **glibc 2.39**, so it fails on older distributions with:

```
GLIBC_2.39 not found
```

### Solution  
Add a new job `build_release_static` that compiles for `x86_64-unknown-linux-musl` and uploads the **fully-static** binary.  
The new artifact is named `pumpkin-x86_64-linux-musl` and is shipped together with the existing GNU build in the nightly release.

### Changes  
- Install `musl-tools` (provides `x86_64-linux-musl-gcc`)  
- Add `rustup target add x86_64-unknown-linux-musl`  
- Build with `cargo build --release --target x86_64-unknown-linux-musl`  
- Include the static binary in `draft_release`

### Verification  
- [x] CI passes  
- [x] Artifacts contain `pumpkin-x86_64-linux-musl`  
- [x] Downloaded binary shows `statically linked` via `file pumpkin-x86_64-linux-musl`

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -74,7 +74,7 @@ jobs:
           if [ "${{ runner.os }}" == "Windows" ]; then EXT=".exe"; fi
           cp "target/release/pumpkin$EXT" "dist/pumpkin-${{ runner.arch }}-${{ runner.os }}$EXT"
       - name: Export executable
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: pumpkin-${{ runner.arch }}-${{ runner.os }}
           compression-level: 9
@@ -127,9 +127,8 @@ jobs:
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     steps:
       - uses: actions/checkout@v6
-
-      - name: Download all binary artifacts
-        uses: actions/download-artifact@v6
+      - name: Download prebuilt binary artifacts
+        uses: actions/download-artifact@v7
         with:
           pattern: "**/pumpkin-*"
           path: dist/


.github/workflows/rust.yml
```diff
@@ -78,6 +80,32 @@ jobs:
           compression-level: 9
           path: dist/pumpkin-*
 
+  # ── 新增：完全静态链接的 musl 构建 ───────────────────────────────
```
illyrius666:
```suggestion
```

.github/workflows/rust.yml
```diff
@@ -91,20 +119,23 @@ jobs:
       - run: rustup component add clippy rustfmt --toolchain ${{ matrix.toolchain }}
       - uses: Swatinem/rust-cache@v2
       - run: cargo clippy --release --all-targets --all-features
+
   draft_release:
     permissions:
       contents: write
-    needs: [build_release]
+    needs: [build_release, build_release_static]   # ← 依赖静态构建
```
illyrius666:
```suggestion
    needs: [build_release, build_release_static]
```

HttpRafa:
We could also use cargo-zigbuild to build Pumpkin for an older glibc version. This might avoid some of the negative effects of using musl, such as increased binary size.

Snowiiii:
Hey, I'm sorry but i don' think this is necessary and the "userbase" for this is very small, glibc 2.39 released a year ago, enough time for the most distros to update their glibc package, Also if using an older distro the user still always has the option to install glibc manually

MiaM1ku:
> Hey, I'm sorry but i don' think this is necessary and the "userbase" for this is very small, glibc 2.39 released a year ago, enough time for the most distros to update their glibc package, Also if using an older distro the user still always has the option to install glibc manually

For some users, there are various reasons why they may still be on older systems. Upgrading glibc is not as straightforward as it might seem and can easily introduce system instability or breakages. Keeping a statically linked build provides a safer alternative and helps these users avoid the risks associated with manually upgrading glibc.

PR #1237, opened 2025-12-12
feat(plugin): add API version compatibility check for native plugins

Sclock:
Added a plugin ABI gate so outdated binary plugins can’t take the server down. Specifically:
﻿
Introduced PLUGIN_API_VERSION and documented how to bump it when changing the public plugin ABI (pumpkin/src/plugin/mod.rs:21).
Updated the plugin derive macro to export the API version symbol alongside the metadata so loaders can inspect it (pumpkin-api-macros/src/lib.rs:62-80).
Extended the native loader with an explicit version check and descriptive loader errors; incompatible or missing symbols now cause a clean load failure before any plugin code runs (pumpkin/src/plugin/loader/native.rs:5-87, pumpkin/src/plugin/loader/mod.rs:40-71).
Tests: cargo check

### DIFF
--- a/pumpkin-api-macros/src/lib.rs
+++ b/pumpkin-api-macros/src/lib.rs
@@ -66,6 +66,9 @@ pub fn plugin_impl(_attr: TokenStream, item: TokenStream) -> TokenStream {
             description: env!("CARGO_PKG_DESCRIPTION"),
         };
 
+        #[unsafe(no_mangle)]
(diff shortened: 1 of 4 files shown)


Snowiiii:
I'm not sure if this is the best solution, i think letting the plugin define a min target plugin version similar like in Android would be better since the developers can then support more versions

Snowiiii:
I mean for now its fine. Thank you @Sclock :D

PR #1241, opened 2025-12-16
Fix compilation of plugins and add examples

BjornTheProgrammer:
## Description

This adds an official example of a plugin using Rust. The purpose of this is twofold: to provide consistently up-to-date documentation, and to be a simple test to verify a plugin will actually compile without unintended breaking changes.

The motivation for doing this is that with the most recent commits, the current example code from the [website](https://docs.pumpkinmc.org/plugin-dev/plugin-template/basic-logic) does not work.

This doesn't seem to be completely intentional, and as a result, this would help ensure something like that doesn't happen again.

## Testing

The plugins were successfully compiled.


BjornTheProgrammer:
Note before merging, running the plugin on the server still causes a freeze to occur, and doesn't allow players to join

BjornTheProgrammer:
Fixed all of the previous issues. One note is that this changes the public api to no longer include the `init_log!` macro. Instead the user should call `server.init_log()` in the `on_load` funciton, like follows.

```rust
#[plugin_method]
async fn on_load(&mut self, server: Arc<Context>) -> Result<(), String> {
    server.init_log();

    log::info!("Hello, Pumpkin!");

    Ok(())
}
```

The reason for this is that it will now actually share the logging with the main crate now.

examples/hello-world-plugin/src/lib.rs
vyPal, line 1: I believe I remember Alex being against having example plugins directly in the main repo. Could you instead make an additional PR for https://github.com/PumpkinPlugins/Hello-Pumpkin and probably also a PR for https://github.com/Pumpkin-MC/Pumpkin-Website to update the plugin tutorial on the website?

pumpkin-api-macros/src/lib.rs
```diff
@@ -2,11 +2,13 @@ use proc_macro::TokenStream;
...
 use proc_macro2::Ident;
 use quote::quote;
+use std::collections::HashMap;
 use std::sync::LazyLock;
 use std::sync::Mutex;
 use syn::{ImplItem, ItemFn, ItemImpl, ItemStruct, parse_macro_input, parse_quote};
 
-static PLUGIN_METHODS: LazyLock<Mutex<Vec<String>>> = LazyLock::new(|| Mutex::new(Vec::new()));
+static PLUGIN_METHODS: LazyLock<Mutex<HashMap<String, String>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
```
vyPal, line 11: I don't see a need for using a HashMap instead of a Vec, as the key (function name) is never used

pumpkin-api-macros/src/lib.rs
```diff
@@ -17,17 +19,25 @@ pub fn plugin_method(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
 
+    let output_type = match fn_output {
+        syn::ReturnType::Default => quote! { () },
+        syn::ReturnType::Type(_, ty) => quote! { #ty },
+    };
+
     let method = quote! {
         #[allow(unused_mut)]
-        async fn #fn_name(#fn_inputs) #fn_output {
-            crate::GLOBAL_RUNTIME.block_on(async move {
```
vyPal, line 23: This might cause issues with calling some internal methods from within plugins (I believe methods for playing sounds or teleporting were problematic), as they require being called from withing a tokio runtime, but even though the plugin method is called from a tokio runtime running on the server, this runtime isn't accessible to the plugin, as it is essentially running in it's own process

pumpkin-api-macros/src/lib.rs
```diff
@@ -42,7 +52,7 @@ pub fn plugin_impl(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let methods = PLUGIN_METHODS.lock().unwrap();
 
     let methods: Vec<proc_macro2::TokenStream> = methods
-        .iter()
+        .values()
```
vyPal, line 57: As mentioned above, since you only use the values, there is no need for a HashMap

Cargo.toml
```diff
@@ -11,6 +11,7 @@ members = [
     "pumpkin-world",
     "pumpkin/",
     "pumpkin-data",
+    "examples/*",
```
vyPal: Wouldn't include the example plugins as mentioned above

vyPal (changes requested):

examples/hello-world-plugin/src/lib.rs
BjornTheProgrammer: Will do

pumpkin-api-macros/src/lib.rs
BjornTheProgrammer: The reason for not using a vec is to ensure that there are not duplicate functions with the same name generated. I noticed this was an issue with incremental checks when using the expand macro recursively feature.

pumpkin-api-macros/src/lib.rs
BjornTheProgrammer: How do you recommend doing the equivilant to this with the new no async trait method?

BjornTheProgrammer:
I've made all the changes requested, except for the `HashSet` removal request, due to unnecessary duplications in macro generated code being created, as seen in the following screenshot.

image

Additionally I've submitted PRs to the docs (https://github.com/Pumpkin-MC/Pumpkin-Docs/pull/54) and hello-world plugin (https://github.com/PumpkinPlugins/Hello-Pumpkin/pull/3).

Snowiiii:
Overall looks good. Thank you @BjornTheProgrammer :D

PR #1242, opened 2025-12-18
BlockEntities move to 'pumpkin' from 'pumpkin-world'

topi-banana:
## Description

I said in the Discord(#pumpkin-dev) 👇
> It's impossible to implement Command Block without being able to use any of the structs from the pumpkin crate like Server, Entity, (etc)

Here are the key changes:

- `pumpkin-world/src/block/entities/*`
    -> `pumpkin/src/block/entities/~`

- `pumpkin-world/src/chunk/mod.rs`
    ```rust
    // old
    struct ChunkData {
      block_entities: HashMap<BlockPos, Arc<dyn BlockEntity>>,
    }
    // new
    struct ChunkData<T: BlockEntityCollection> {
      block_entities: T,
    }
    ```

While the code quality in this change isn't particularly high, it is necessary to address the issue.
I'm open to suggestions if there's a better approach or alternative solution.

## Testing

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,7 +41,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest, ubuntu-24.04-arm]
+        os: [ubuntu-latest,  ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-latest, macos-15-intel]
         toolchain:
           - stable
     steps:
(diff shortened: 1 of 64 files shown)


Snowiiii:
There has to be an other way

topi-banana:
I think so too, but...

topi-banana:
There is a issue with both the current code and this code.😢
BlockEntities are ticking per World, in the order they were loaded. but the current code is ticking them per chunk via HashMap iteration.
this needs to be changed properly, I'm going to close this PR.

PR #1243, opened 2025-12-24
Add advancements

vlado00001:
## Description

This new system automatically reads and organizes Minecraft advancement files- the code scans folders like "Story" or "Nether" and creates a live map of every goal in the game. This makes it easy for server owners to add custom advancements just by dropping a file into a folder, without ever touching the actual code.

To keep the server running smoothly, all this information is stored in a HashMap during startup. This means the server can instantly check if a player earned an achievement—like mining a diamond—without slowing down the game or causing lag. It also includes "safety brakes" that prevent the server from crashing if a file is broken or formatted incorrectly, simply skipping the bad file and keeping the game running.

By tracking "parent" links between goals, the system understands the proper order of progression, ensuring players can't skip ahead to end-game rewards.

## Testing

The test_load_advancements test performs the following checks:
-Directory Discovery: Confirms the system can find and enter the src/advancements/minecraft folders.
-Data Integrity: Validates that key advancements (like minecraft:story/root) are not only loaded but contain the correct titles and criteria.
-Error Resilience: Verifies that the AdvancementMap remains functional and populates fallback data even if a JSON file is missing expected fields.
-To run this specific test, use: cargo test advancements::tests::test_load_advancements


jackssrt:
The description of this PR reeks of AI. Most of the advancement JSON files are completely empty and the crate this PR adds is never used anywhere. This is not ready to be merged.

Q2297045667:
> The description of this PR reeks of AI. Most of the advancement JSON files are completely empty and the crate this PR adds is never used anywhere. This is not ready to be merged.

You're absolutely right.

Snowiiii:
@vlado00001 The CI is failing, do you still work on this?, when yes you should mark it as draft

PR #1247, opened 2025-12-25
chore: Remove Redundant Clippy Allows

CuzImClicks:
## Description

`#[expect(...lints)]` is better practice, because it warns you when the lints become redundant. I switched all allows to expects and removed all of the unnecessary ones. This is great for code hygiene :)

## Testing


Snowiiii:
I actually also prefer expect over allow, most of the allows are coming from before Rust 1.81. Thank you @CuzImClicks 🐑

Issue #1233, opened 2025-12-11
After blocking the water source, flowing water still appeared repeatedly.


RegulusSource:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

windows11

### Server Software Version/Commit

fba4d405

### What happened?

After blocking the water source, flowing water still appeared repeatedly.

### To Reproduce

Place the water source and block it with blocks.

### Expected behavior

After a moment, the running water stopped.

AyaSanae:
If you block the water source, the water will spread everywhere like vines. XD

PR #1249, opened 2025-12-26
fix: fluid system incorrectly treating air as flowing water

AyaSanae:
## Description
The issue: https://github.com/Pumpkin-MC/Pumpkin/issues/1233
I think the problem is: 
When water dries up (the block turns to air):
[world.get_block_state_id(block_pos)](https://github.com/Pumpkin-MC/Pumpkin/blob/42c903493a09a4c8d7f4eddc2644464e3871c736/pumpkin/src/block/fluid/flowing.rs#L345) returns the block state ID of air
the [from_state_id ](https://github.com/Pumpkin-MC/Pumpkin/blob/42c903493a09a4c8d7f4eddc2644464e3871c736/pumpkin-data/build/fluid.rs#L236)iterates through all fluid states for a matching state.block_state_id, but the air's state ID cannot match any fluid state.
So, returns the default state: [Self::from_index(0)](https://github.com/Pumpkin-MC/Pumpkin/blob/42c903493a09a4c8d7f4eddc2644464e3871c736/pumpkin-data/build/fluid.rs#L211)
The default state has Falling set to true, which causes the level to be set to 7 by the code following flow_to_sides.
So it caused the issue: "After blocking the water source, flowing water still appeared repeatedly."

 I simply checked whether the block_state_id obtained from the location is a fluid state ID.

## Testing
After blocking the water source or cutting off the water flow with blocks, it will not cause the water to repeatedly appear and disappear.

### DIFF
--- a/pumpkin/src/block/fluid/flowing.rs
+++ b/pumpkin/src/block/fluid/flowing.rs
@@ -344,6 +344,14 @@ pub trait FlowingFluid: Send + Sync {
         Box::pin(async move {
             let block_state_id = world.get_block_state_id(block_pos).await;
 
+            let is_fluid_state_id = fluid
+                .states
+                .iter()
+                .any(|state| state.block_state_id == block_state_id);
+            if !is_fluid_state_id {
+                return;
+            }
+
             let props = FlowingFluidProperties::from_state_id(block_state_id, fluid);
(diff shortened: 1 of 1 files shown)


Snowiiii:
Looks good. Thank you @AyaSanae :D

PR #1250, opened 2025-12-27
Reuse TickData allocations between cycles

NailLegProcessorDivide:
Reduce allocations that happen every tick for level tick data by storing the allocation across ticks


Snowiiii:
Looks good. Thank you @NailLegProcessorDivide :D

ruasidev:
> Looks good. Thank you @NailLegProcessorDivide :D

Pretty sure this caused a breaking change, no longer compiles ;w;

PR #1263, opened 2026-01-09
Fix a deadlock

sprinox:
## Description

It is really a hard work.

I think there is possibility that `self.queue.is_empty()` is true after `self.resort_work`. Then incorrectly goto `self.send_level.wait_and_get` and deadlock

## Testing


Snowiiii:
Looks good. Thank you @sprinox :D

Issue #1276, opened 2026-01-12
Headless pistons with beds


us3rn1me:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11

### Server Software Version/Commit

0.1.0-dev+1.21.11

### What happened?

You can make headless pistons using beds.

### To Reproduce

Place a bed, a piston facing the top or bottom half or the bed, and activate the piston, which breaks only one of the beds half, and then manually break the other half, and the piston becomes headless.

### Expected behavior

That should not happen

Poneaswaran:
Suggested Fix

Treat beds as a single block entity during piston interactions.

Ensure that when one half of a bed is destroyed or moved, the other half is also removed immediately.

Add validation to piston logic to prevent piston heads from being removed without removing the piston base.

Snowiiii:
closed in https://github.com/Pumpkin-MC/Pumpkin/commit/578bab259a089b2997b13619f5d318c2a2d46c09

PR #1285, opened 2026-01-13
fix(bed): remove paired bed half on state replacement issue #1276

Poneaswaran:
This PR implements on_state_replaced for BedBlock to ensure that beds are treated as atomic blocks during block replacement.

Key logic added in bed.rs:

When a bed block is replaced (e.g., by a piston):

The method checks if the block was moved (args.moved) and exits early if so

Determines the position of the other bed half using facing and part

Verifies that the neighboring block is the corresponding bed half

Removes the other half by setting it to AIR

This guarantees that no orphaned bed halves remain, preventing piston corruption.

Files Changed

pumpkin/src/block/blocks/bed.rs


Snowiiii:
Looks good. Thank you @Poneaswaran :D

PR #1289, opened 2026-01-13
fix: apply clippy lints to workspace

FabseGP:
## Description

Moved clippy lints to workspace Cargo.toml and made every crate reuse the lints

The changes are from running `cargo clippy --fix` to apply the default lints, which I looked through to ensure nothing was incorrectly changed. There were lots of other suggestions by clippy not applied though, later on I will look through and nitpick some of the suggestions

## Testing

I checked that the code was still compiling and that I could join a server and move around, the changes applied didn't change any core mechanic

### DIFF
--- a/pumpkin-nbt/src/deserializer.rs
+++ b/pumpkin-nbt/src/deserializer.rs
@@ -3,7 +3,7 @@ use std::io::{Seek, SeekFrom};
 
 use crate::{
     BYTE_ARRAY_ID, BYTE_ID, COMPOUND_ID, END_ID, Error, INT_ARRAY_ID, INT_ID, LIST_ID,
-    LONG_ARRAY_ID, LONG_ID, NbtTag, Write, get_nbt_string, io,
+    LONG_ARRAY_ID, LONG_ID, NbtTag, get_nbt_string, io,
 };
 use io::Read;
 use serde::de::{self, DeserializeSeed, IntoDeserializer, MapAccess, SeqAccess, Visitor};
--- a/pumpkin-util/src/jwt/verifier.rs
+++ b/pumpkin-util/src/jwt/verifier.rs
@@ -187,7 +187,7 @@ pub fn decode_header_get_x5u(header_b64: &str) -> Result<String, AuthError> {
     if let Some(x5u) = header_json.get("x5u")
         && let Some(s) = x5u.as_str()
     {
-        return Ok(s.to_string());
+        return Ok(s.to_owned());
     }
     Err(AuthError::MissingX5U)
 }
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -110,3 +111,41 @@ time = "0.3"
...
+panic = "warn"
+panic_in_result_fn = "warn"
+redundant_type_annotations = "warn"
+return_and_then = "warn"
+str_to_string = "warn"
+verbose_file_reads = "warn"
+use_debug = "warn"
+unwrap_used = "warn"
+unwrap_in_result = "warn"
+unused_trait_names = "warn"
```
CuzImClicks: i dont really like this one, it would really just be annoying while writing and the benefits seem to be only marginal

pumpkin/src/world/custom_bossbar.rs
```diff
@@ -30,7 +30,7 @@ pub struct CustomBossbar {
 impl CustomBossbar {
     #[deny(clippy::new_without_default)]
```
CuzImClicks: idk why this is here but should be removed

FabseGP: fixed

Cargo.toml
FabseGP: unused_trait_names I can live without (possibly use_debug as well, even though debug-formatting is not consistent, it doesn't harm anything), but warning against unwraps seems useful? 

Ideally no unwraps would be used and errors caught instead, smth like this (just a random example)

```rust
if let Err(err) = HTTP_CLIENT.post(url).bearer_auth(token).send().await {
	error!("Failed to report uptime: {:?}", &err);
}
```

Snowiiii:
@FabseGP do you still work on this ?, the last time i checked the CI was failing now the PR already got a bunch of conflicts

Cargo.toml
CuzImClicks: oh my comment was only meant for the unused_trait_names rule not the other ones!

FabseGP: oh my bad lol, I've removed it

FabseGP:
> @FabseGP do you still work on this ?, the last time i checked the CI was failing now the PR already got a bunch of conflicts

@Snowiiii ahh yeah, the merge conflicts are fixed now, the CI is failing now because of the clippy warnings from the added lints... it'll make this pr bigger fixing the warnings

pumpkin-protocol/Cargo.toml
CuzImClicks: while youre at it you can also remove these two unused features

CuzImClicks:
can you revert the `as _` that is now in every file?

pumpkin-protocol/Cargo.toml
FabseGP: fixed

FabseGP:
@CuzImClicks thats's fixed, but the only way to make the CI not fail is to resolve the clippy warnings manually, so is there any other lints that should be reverted? easier to do now than later lol

Snowiiii:
closed in https://github.com/Pumpkin-MC/Pumpkin/commit/fb13004b47a6c3f85158e29460f51fcadc871c45

PR #1292, opened 2026-01-13
feat: compile optimizations

FabseGP:
## Description

Prune unused features, strip release binary from debuginfo (and strip debug binary from line-tables data, only useful with an actual debugger like gdb) and compile for native CPU when not in CI. Added abort on panic as well, the unwraps I could find where the server panics seems worthy to abort on, like when failing to parse files

From a clean target on my machine (Ryzen 5 5600H), the compile time both in debug and release mode didn't change much (sometimes it took 15s less, other times only 3s less), but the binary size in debug mode was reduced from 321MB to 129MB and from 28MB to 25MB in release mode

I reckoned the CI-builds should be compiled for a generic CPU, which I added to the workflow. I also made the change to define all deps with "*" for the version with the reasoning that Cargo.lock would track a known, working state while it makes it easier to update all dependencies using cargo update

## Testing

Used [cargo-features-manager](https://github.com/ToBinio/cargo-features-manager) to prune unused features (it had some false positives, which I reverted). The binaries have been tested to run as expected (connect to server, move around etc.)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -7,8 +7,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTFLAGS: "-Dwarnings"
-  CARGO_BUILD_RUSTFLAGS: "-Dwarnings -C target-cpu=generic"
+  RUSTFLAGS: "-Dwarnings -Ctarget-cpu=generic"
 
 jobs:
   format:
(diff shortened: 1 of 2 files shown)


Snowiiii:
Hey, Sorry but i really don't like the changes in there, Defining every crate with the version * seems just pointless and can cause compiling to easly break when a crate has a new version with breaking system. And using `panic = "abort"` makes debugging a crash in release mode almost impossible

FabseGP:
I get your perspective regarding the crate versions, I still think having Cargo.lock is enough to keep a working state, but you're right it's easier to break this way

About `panic = "abort"`, that only becomes a problem when it is even possible for the code to crash, say from using unwrap, unchecked indexing etc., which you can eliminate by catching errors or use checked methods. You can still provide some logging when an error is catched for example

FabseGP:
Some of the changes were taken from here btw: https://nnethercote.github.io/perf-book/build-configuration.html

PR #1306, opened 2026-01-15
Binary tweaks

FabseGP:
## Description

Instead of defining the rust-version in each crate, define it in the workspace and apply it to all crates 

Strip debuginfo from the release binary (still keeps backtraces: https://doc.rust-lang.org/rustc/codegen-options/index.html#strip), while limits debug info in the dev-profile to line-tables-only (more info is only needed for smth like single step in gdb: https://doc.rust-lang.org/rustc/codegen-options/index.html#debuginfo), reducing the binary size in both profiles (and the incremental compile time in my experience)

## Testing

The server can still be compiled


Snowiiii:
@FabseGP please make sure that the CI passes before marking this as ready

FabseGP:
@Snowiiii sorry about that, all fixed now

Snowiiii:
I don't think there is any need reducing the debug information for debug builds

FabseGP:
Yeah it would reduce the binary size, but that's prob not so important in debug builds; fixed now

Snowiiii:
Looks good. Thank you @FabseGP :D 🦄

Snowiiii:
Also please choose a better title in the future

PR #1313, opened 2026-01-16
Use macros for i8/u8 reads & remove temp buffer

sshcrack:
## Description
This PR removes the manually implemented functions and uses the `get_number_be` macro instead.
Also it uses the `take` method to take exactly `bound +1` bytes to ensure that only a maximum of `bound` bytes can be read.
If the returned buffer is larger than `bound` (so it must be `bound +1`, an error is returned like in the original implementation).

## Testing
I've tested, that the macro behaves like the previous functions on the [rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=38bfac22810d8735f9b626fde82a6048).
Also I've tested the new read implementation on the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=63055238e0c606ce2e7a391075c6c0fb)


Snowiiii:
Looks good. Thank you @sshcrack :D

PR #1338, opened 2026-01-18
feat: Implemented Drowning

spectradevv:
## Description
Implemented oxygen depleting while underwater, causing damage when it runs out.

## Testing

### DIFF
--- a/pumpkin/src/entity/breath.rs
+++ b/pumpkin/src/entity/breath.rs
@@ -1,10 +1,10 @@
 use crate::entity::EntityBase;
 use crate::entity::player::Player;
-use pumpkin_data::Block;
 use pumpkin_data::damage::DamageType;
 use pumpkin_data::effect::StatusEffect;
-use pumpkin_data::fluid::Fluid;
 use pumpkin_data::meta_data_type::MetaDataType;
+use pumpkin_data::tag;
+use pumpkin_data::tag::Taggable;
 use pumpkin_data::tracked_data::TrackedData;
 use pumpkin_protocol::codec::var_int::VarInt;
 use pumpkin_protocol::java::client::play::Metadata;
@@ -97,7 +97,6 @@ impl BreathManager {
     async fn is_eye_in_water(&self, player: &Player) -> bool {
         let e = &player.living_entity.entity;
         let pos = e.pos.load();
-
         let eye_y = e.get_eye_y();
 
         let bp = BlockPos::new(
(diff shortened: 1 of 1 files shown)


FireAcidCow:
Just tried it, seems to work pretty well. I did notice two things. First, when you die in vanilla, your breath meter is fully reset. This doesn't seem to be the case in this branch. The other thing I noticed was that if you switch into creative mode in vanilla, your breath meter goes invisible but still starts to refill, so if you switch back fast enough it still has some air gone. In this branch, switching into creative mode and back to survival quickly fully refills your air meter. Overall, this seems like a good base.

spectradevv:
@FireAcidCow both issues fixed. thanks for the feedback!

pumpkin/src/entity/player.rs
```diff
@@ -1479,6 +1486,125 @@ impl Player {
...
+            let new_air = (prev + 4).min(300);
+            if new_air != prev {
+                self.air_supply.store(new_air, Ordering::Relaxed);
+                self.sync_air(new_air).await;
+            }
+            self.drowning_tick.store(0, Ordering::Relaxed);
+        }
+    }
+
+    async fn sync_air(&self, air: i32) {
```
CuzImClicks: for the player we use `async fn send_...` when sending packets instead of `sync_...`
also i think it would be better if we always send the value that is stored in the player's atomic values so desync becomes impossible.

```suggestion
    async fn send_air_supply(&self) {
```

pumpkin/src/entity/player.rs
```diff
@@ -1517,6 +1643,10 @@ impl Player {
             }
         }
 
+        // Reset air supply on death
+        self.air_supply.store(300, Ordering::Relaxed);
```
CuzImClicks: might also want to set drowning_tick to 0 just for completeness :)

CuzImClicks:
@Snowiiii do you think we should put the breath system into its own Manager like `HungerManager` to keep all of the code related to it in one file?

also @spectradevv you can also put into your pr description that you also implemented the gamerule and the water breathing effect

other than my comments... LGTM!

spectradevv:
@CuzImClicks sounds good i'll work on that tonight

pumpkin/src/entity/player.rs
```diff
@@ -506,6 +511,8 @@ impl Player {
             last_sent_health: AtomicI32::new(-1),
             last_sent_food: AtomicU8::new(0),
             last_food_saturation: AtomicBool::new(true),
+            air_supply: AtomicI32::new(300),
```
CuzImClicks: Shouldnt these be removed now that they are stored in the breath system?

pumpkin/src/entity/player.rs
```diff
@@ -1565,6 +1573,13 @@ impl Player {
             }
         }
 
+        // Reset air supply & drowning ticks on death
+        self.breath_manager.air_supply.store(300, Ordering::Relaxed);
```
CuzImClicks: Small suggestion maybe just add a reset function so you can just do reset() and then send_...()

CuzImClicks (commented):
Looking good!

CuzImClicks:
Just call it reset() xD but after that... LGTM!

pumpkin/src/entity/breath.rs
```diff
@@ -0,0 +1,156 @@
...
+            }
+            self.drowning_tick.store(0, Ordering::Relaxed);
+        }
+    }
+
+    async fn is_eye_in_water(&self, player: &Player) -> bool {
+        let e = &player.living_entity.entity;
+        let pos = e.pos.load();
+
+        let eye_y = pos.y + f64::from(e.entity_dimension.load().eye_height);
```
Snowiiii: should be converted to `e.get_eye_y()`

Snowiiii:
I also wonder if this allows to breath when inside of BUBBLE_COLUMNs?

spectradevv:
@Snowiiii i found out that it broke with water plants and fixed it. but i couldnt get bubble columns to work to test with it

pumpkin/src/entity/breath.rs
```diff
@@ -108,8 +109,19 @@ impl BreathManager {
...
+
+            let is_water_plant = block_here == &Block::SEAGRASS
+                || block_here == &Block::TALL_SEAGRASS
+                || block_here == &Block::KELP
+                || block_here == &Block::KELP_PLANT;
+
+            if !is_water_plant {
+                return false;
+            }
         }
```
Snowiiii: Are you sure vanilla does it the same ?, because just hardcoding them seems wrong

spectradevv: fixed it lol. my brain wasnt braining

Snowiiii:
Are you sure the implementation is the same as in Vanilla ?, maybe it is because of slow ticks but the damage time is way higher and it also sounds like you are getting double damage (maybe i'm wrong)

spectradevv:
@Snowiiii yea i think its just the slow ticks. not sure why the servers running so slow. its 150ms per tick

Snowiiii:
Looks good then. Thank you @spectradevv :D

Issue #1308, opened 2026-01-16
Using a bucket in survival crashes the server


FireAcidCow:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

0.1.0/ff1f3a0f

### What happened?

When you use a bucket in survival mode, the server crashes

### To Reproduce

1. Use a bucket in any fashion (full/empty, doesn't matter) while in survival mode.

### Expected behavior

Update the item in the player inventory to the correct state

musselmandev:
I was able to reproduce this crash, same os and server version

spectradevv:
Fixed in pr: https://github.com/Pumpkin-MC/Pumpkin/pull/1341

PR #1341, opened 2026-01-19
fix: server crash when placing liquids in survival

spectradevv:
## Description
Fixed bug when placing liquids in survival would cause the server to crash due to a lock issue
https://github.com/Pumpkin-MC/Pumpkin/issues/1308

## Testing


Snowiiii:
can't we just remove the   `drop(held);` above and use that ?

Snowiiii:
> can't we just remove the `drop(held);` above and use that ?

Found out that this actually deadlocks

Issue #1347, opened 2026-01-19
Waterlogged blocks can exploid


user123456000:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch linux

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.11 6fe57ff8..bcdde452

### What happened?

Waterlogged blocks are destroyed by explosions.

### To Reproduce

1. Place leaves  (or any other block that can be waterlogged).
2. Fill it with water.
3. Explode a TNT.

### Expected behavior

The block should not explode.

FMMazur:
@Snowiiii can be closed

PR #1352, opened 2026-01-20
fix: adjust explosion resistance calculation to account for fluid states

FMMazur:
## Description
Waterlogged blocks should not be destroyed in explosions

solves #1347

## Testing
Place water in watterloggable blocks then explode a TNT and check if it keeps in place

### DIFF
--- a/pumpkin/src/world/explosion.rs
+++ b/pumpkin/src/world/explosion.rs
@@ -58,7 +58,8 @@ impl Explosion {
                         // }
 
                         if !state.is_air() || !fluid_state.is_empty {
-                            let resistance = fluid_state.blast_resistance.max(block.blast_resistance);
+                            let resistance =
+                                fluid_state.blast_resistance.max(block.blast_resistance);
                             h -= resistance * 0.3;
                         }
 


pumpkin/src/world/explosion.rs
```diff
@@ -50,15 +50,20 @@ impl Explosion {
...
 
-                        if !state.is_air() {
-                            h -= (block.blast_resistance + 0.3) * 0.3;
-                        }
+                        let resistance = if state.is_air() && fluid_state.is_empty {
+                            0.0
+                        } else {
+                            fluid_state.blast_resistance.max(block.blast_resistance)
+                        };
+                        h -= (resistance + 0.3) * 0.3;
```
Snowiiii: h is not always subtracted in vanilla, here is the logic
```java
public Optional<Float> getBlastResistance(Explosion explosion, BlockView world, BlockPos pos, BlockState blockState, FluidState fluidState) {
		return blockState.isAir() && fluidState.isEmpty()
			? Optional.empty()
			: Optional.of(Math.max(blockState.getBlock().getBlastResistance(), fluidState.getBlastResistance()));
	}

Optional<Float> optional = this.behavior.getBlastResistance(this, this.world, blockPos, blockState, fluidState);
if (optional.isPresent()) {
    h -= (optional.get() + 0.3F) * 0.3F;
}
```

FMMazur:
@Snowiiii i think this solves the issue, but i assume it would be better to just skip with continue, instead of doing the calculation, i will test it

Snowiiii:
Looks good. Thank you @FMMazur :D

PR #1358, opened 2026-01-22
fix: placing blocks on composters

### LINKED ISSUE #942: Placing blocks on Composter's

### Operating System

Windows 11

Pumpkin 0.1.0-dev+1.21.6 (fe5ab68)

### What happened?

In vanilla you can place blocks on a composter without sneaking, but you cant do it on Pumpkin.

### To Reproduce

Try to place a block on a Composter (without sneaking).

### Expected behavior

The blocks to be palced.

RoosterBooster007:
## Description
This fix addresses the composter block placement bug, allowing players to now place blocks properly next to composters (without sneaking).

## Testing
You can now place blocks off of composters without the interaction and block being consumed (see attached image).
Screenshot 2026-01-22 153448

Fixes #942

### DIFF
--- a/pumpkin/src/block/blocks/composter.rs
+++ b/pumpkin/src/block/blocks/composter.rs
@@ -16,6 +16,7 @@ use pumpkin_data::{
     item::Item,
     world::WorldEvent,
 };
+use pumpkin_inventory::screen_handler::InventoryPlayer;
 use pumpkin_macros::pumpkin_block;
 use pumpkin_util::math::position::BlockPos;
 use pumpkin_world::{BlockStateId, item::ItemStack, tick::TickPriority, world::BlockFlags};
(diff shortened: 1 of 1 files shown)


pumpkin/src/block/blocks/composter.rs
```diff
@@ -50,24 +50,42 @@ impl BlockBehaviour for ComposterBlock {
...
+                    args.world
+                        .sync_world_event(WorldEvent::ComposterUsed, *args.position, 1)
+                        .await;
+                    true
+                } else {
+                    false
+                }
+            };
+
+            if handled {
```
Snowiiii: I think instead of defining handled it would be better to just early return `BlockActionResult::Consume` or ` BlockActionResult::Pass`

RoosterBooster007: Agreed, that's cleaner. I also made it so that items are now consumed if that's alright.

Snowiiii:

PR #1364, opened 2026-01-23
fix: implement broken behavior for piston block to break the head

### LINKED ISSUE #1271: Bugs with mining pistons

### Operating System

Windows 11

0.1.0-dev+1.21.11

### What happened?

When mining an extended piston, it only drops when you break the extended part. When you break the "base" part, it doesn't drop, and the extended part is left.

### To Reproduce

Mine an extended piston 2 times, 1st try the base part and then the extended part

### Expected behavior

Both parts should drop a piston and when breaking the base part the extended part should also break

FMMazur:
## Description
Adds the broken behaviour to the piston so it breaks the moving/head part

fixes #1271

## Testing
When you break a piston base then it break the movable piece.

### DIFF
--- a/pumpkin/src/block/blocks/piston/piston.rs
+++ b/pumpkin/src/block/blocks/piston/piston.rs
@@ -16,7 +16,7 @@ use pumpkin_world::{
 
 use crate::{
     block::{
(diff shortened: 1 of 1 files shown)


Snowiiii:
Looks good. Thank you @FMMazur :D

PR #1373, opened 2026-01-25
fix: initialize player abilities based on gamemode on first join

### LINKED ISSUE #1367: Unable to use properly when first entering Pumpkin server creative mode

### Operating System

Windows 11 24H2

0.1.0-dev+1.21.11/33cf5237

### What happened?

Upon first accessing the Pumpkin server, creative mode may not function correctly (the configuration file defaults to creative mode).
After manually switching to another mode and then switching back to Creative Mode, everything returned to normal.

### To Reproduce

The configuration file is set to Creative Mode by default, and this is the first time logging into a Pumpkin server (a newly created world save file, any seed).
The problem persists even after entering and exiting multiple times.

### Expected behavior

The blocks in the inventory are not displaying correctly. Players cannot fly.

Image
Image
Image
Image

Purdze:
## summary

- initialize player abilities based on gamemode during player creation
- call set_for_gamemode() in player::new() to set correct ability flags
- fixes creative mode not working properly on first join (flying disabled, blocks not displaying)

## changes

- pumpkin/src/entity/player.rs - initialize abilities based on gamemode before player struct creation

fixes #1367

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -448,6 +448,10 @@ impl Player {
             PlayerScreenHandler::new(&inventory, None, 0).await,
         ));
 
+        // Initialize abilities based on gamemode (like vanilla's GameMode.setAbilities())
+        let mut abilities = Abilities::default();
+        abilities.set_for_gamemode(gamemode);
+
         Self {
             living_entity,
             config: RwLock::new(config),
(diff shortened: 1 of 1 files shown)


Snowiiii:
Seems like i removed that in a recent commit, thanks @Purdze :D

Issue #881, opened 2025-06-09
add allow-end: true


Q2297045667:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

I would like to have a configuration option to control "the End" in the game.

### Describe the solution you'd like

I hope to add `allow-end: true` in the `configuration.toml` file, just like in `bukkit.yml`.

BigPebble123:
Maybe better two settings: first one for end generation and second for allowing players to enter the end?
For example `generate-end` and `allow-end-entering`?

Q2297045667:
> Maybe better two settings: first one for end generation and second for allowing players to enter the end? For example and ?`generate-end``allow-end-entering`

The second one seems to make little sense? I can’t imagine what kind of environment would require “allow-end-entering.”

eleanordoesntcode:
Maybe if you want to enable the End after a while on a server but not immediately - this is something I've seen servers doing.

Q2297045667:
> Maybe if you want to enable the End after a while on a server but not immediately - this is something I've seen servers doing.

No, sometimes people don't need it. My suggestion is, if the Nether can be closed, why can't the End?

Snowiiii:
Should be implemented now

PR #1374, opened 2026-01-25
fix: make entities respect the allow_nether setting and add an allow_end setting

awildergoose:
## Description
- Adds an allow_end setting to toggle end portals (#881)
- Fixes the allow_nether setting not being respected

## Testing
works 👍

### DIFF
--- a/pumpkin/src/entity/mod.rs
+++ b/pumpkin/src/entity/mod.rs
@@ -1353,22 +1353,20 @@ impl Entity {
             return;
         }
 
-        if portal_world.dimension == Dimension::THE_NETHER
+        if (portal_world.dimension == Dimension::THE_NETHER
             && !portal_world
                 .server
                 .upgrade()
                 .unwrap()
                 .basic_config
-                .allow_nether
-        {
-            return;
-        } else if portal_world.dimension == Dimension::THE_END
-            && !portal_world
-                .server
-                .upgrade()
-                .unwrap()
-                .basic_config
-                .allow_end
+                .allow_nether)
+            || (portal_world.dimension == Dimension::THE_END
+                && !portal_world
+                    .server
+                    .upgrade()
+                    .unwrap()
+                    .basic_config
+                    .allow_end)
         {
             return;
         }


Snowiiii:
This seems pretty inefficient since the World are still loaded, it would be better to not load the Worlds at all

Q2297045667:
> This seems pretty inefficient since the World are still loaded, it would be better to not load the Worlds at all

You're on the same wavelength as me.

awildergoose:
> This seems pretty inefficient since the World are still loaded, it would be better to not load the Worlds at all

I thought about doing this, but with that rises a new issue where if a player who was previously inside of the nether dimension tried rejoining after the server had disabled the nether, The server would then have to load the world to put the player in. Implementing the ability for the server to load the world at runtime would raise another issue, being that the `Server` struct would have to be mutated for the worlds `Vec` to be extended, and multiple functions would need to have their signatures changed to use `&mut Server` instead of `&Server`, in my opinion, I think the best solution is to keep all worlds loaded regardless of the config.

Q2297045667:
> > This seems pretty inefficient since the World are still loaded, it would be better to not load the Worlds at all
> 
> I thought about doing this, but with that rises a new issue where if a player who was previously inside of the nether dimension tried rejoining after the server had disabled the nether, The server would then have to load the world to put the player in. Implementing the ability for the server to load the world at runtime would raise another issue, being that the struct would have to be mutated for the worlds to be extended, and multiple functions would need to have their signatures changed to use instead of , in my opinion, I think the best solution is to keep all worlds loaded regardless of the config.`Server``Vec``&mut Server``&Server`

If that's the case, adding allow-endand allow_netherwould be meaningless—in certain situations ( Minigames or Lobby), there's no need to load any extra worlds at all.
Also, I think there’s absolutely no need to consider the issue of players rejoining, because this is mostly something that only needs to be changed the first time the server runs.

awildergoose:
> If that's the case, adding allow_end and allow_nether would be meaningless

Some players prefer to keep the end (or nether) disabled for having players be more interactive within a server, as after players go and beat the dragon, players tend to get less interested, I do this too with my friends whenever we make a server, just to keep the game fun for a little longer.

> in certain situations (Minigames or Lobby), there's no need to load any extra worlds at all.

You're right, I suggest possibly implementing that a different way for the reason above though.

> Also, I think there’s absolutely no need to consider the issue of players rejoining, because this is mostly something that only needs to be changed the first time the server runs.

As mentioned in my first point, it's possible for that to happen, plus, covering edge cases from early on is always good practice IMO.

pumpkin/src/entity/mod.rs
```diff
@@ -1352,6 +1352,29 @@ impl Entity {
...
+                .server
+                .upgrade()
+                .unwrap()
+                .basic_config
+                .allow_nether
+        {
+            return;
+        }
+
+        if portal_world.dimension == Dimension::THE_END
```
CuzImClicks: else if

Snowiiii:
I mean for now this seems fine later we can may directly check if a dimension/world is currently not loaded and fallback to an loaded world, but the logic works. Thank you @awildergoose :D

PR #1380, opened 2026-01-26
feat: add git commit hash to /version command

### LINKED ISSUE #1378: Add git commit to /version

The issue template suggests using the /version command, but the output hasn't been updated and it doesn't contain the current commit, so it's harder to tell if the issue was patched in a recent version or not

Add the current git commit hash to the /version command output, or put another way to uniquely identify which patch the server is running.

Purdze:
## summary
- adds git commit hash to the `/version` command output
- displays 7-character short hash in the version string (e.g., `0.1.0 (Commit: ca77306)`)
- shows full 40-character hash on hover for easy copying

closes #1378

### DIFF
--- a/pumpkin/build.rs
+++ b/pumpkin/build.rs
@@ -1,0 +1,30 @@
+use std::process::Command;
+
+fn main() {
+    // Get short hash (7 chars) for display
+    let short_output = Command::new("git")
+        .args(["rev-parse", "--short=7", "HEAD"])
+        .output();
+
+    let git_hash_short = match short_output {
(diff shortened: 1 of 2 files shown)


CuzImClicks:
I think this should also be logged on startup. This would provide even more info when people send logs. And we wouldnt constantly have to ask people to run /version.

Purdze:
> I think this should also be logged on startup. This would provide even more info when people send logs. And we wouldnt constantly have to ask people to run /version.

Will add this

Snowiiii:
Looks good. Thank you @Purdze :D

PR #1381, opened 2026-01-26
fix: check the nearest solid block to spawn players

DocteurPing:
## Description

This PR has been made as a follow up of [this one](https://github.com/Pumpkin-MC/Pumpkin/pull/852).
Instead of setting the player in (0,0), we will look for the nearest solid block that as 2 empty block above it. We need to check about the 2 empty block otherwise there is a possibility to spawn in a tree. I calculate the spawn position at the creation of the world and i save it in the config, so the server doesn't have to run the calculation each time a player is joining the world. 

## Testing

### DIFF
--- a/pumpkin/src/server/mod.rs
+++ b/pumpkin/src/server/mod.rs
@@ -316,12 +316,11 @@ impl Server {
                 let mut new_level_info = (**level_info).clone();
                 new_level_info.set_pos(safe_pos.x, safe_pos.y, safe_pos.z);
                 server.level_info.swap(Arc::new(new_level_info));
-                let test = server.level_info.load();
                 log::info!(
                     "Safe spawn location set to ({}, {}, {})",
-                    test.spawn_x,
-                    test.spawn_y,
-                    test.spawn_z
+                    safe_pos.x,
+                    safe_pos.y,
+                    safe_pos.z
                 );
             }
         }


pumpkin/src/server/mod.rs
```diff
@@ -290,6 +290,30 @@ impl Server {
         };
 
         log::info!("All worlds loaded successfully.");
+
+        // Find and set a safe spawn location if using default spawn
+        {
+            let level_info = server.level_info.read().await;
+            let is_default_spawn = level_info.spawn_x == 0 && level_info.spawn_z == 0;
```
CuzImClicks: `let is_default_spawn = { ... };`

To properly scope the guard

pumpkin/src/server/mod.rs
```diff
@@ -290,6 +290,30 @@ impl Server {
         };
 
         log::info!("All worlds loaded successfully.");
+
+        // Find and set a safe spawn location if using default spawn
+        {
```
CuzImClicks: This doesnt have to be in its own block... this just lets you get rid of indentation

pumpkin/src/server/mod.rs
```diff
@@ -290,6 +290,30 @@ impl Server {
...
+            let level_info = server.level_info.read().await;
+            let is_default_spawn = level_info.spawn_x == 0 && level_info.spawn_z == 0;
+            drop(level_info);
+
+            if is_default_spawn {
+                log::info!("Finding safe spawn location...");
```
CuzImClicks: Should probably be a debug not an info

pumpkin/src/server/mod.rs
```diff
@@ -290,6 +290,30 @@ impl Server {
...
+                let overworld = {
+                    let worlds = server.worlds.read().await;
+                    worlds.first().cloned()
```
CuzImClicks: The first isnt guaranteed to to also be the overworld... just check for dimension

pumpkin/src/world/mod.rs
```diff
@@ -1164,6 +1164,51 @@ impl World {
         self.dimension.min_y
     }
 
+    /// Finds a safe spawn location starting from (0, 0) and expanding outward
+    /// Returns (x, y, z) coordinates for a valid spawn location
+    pub async fn find_safe_spawn_location(&self) -> (i32, i32, i32) {
```
CuzImClicks: Return a BlockPos or a Vector3 instead

pumpkin/src/world/mod.rs
```diff
@@ -1164,6 +1164,51 @@ impl World {
...
+        const MAX_SEARCH_RADIUS: i32 = 100;
```
CuzImClicks, line 1156: Is this whole function the vanilla way of doing it?

pumpkin/src/world/mod.rs
```diff
@@ -1164,6 +1164,51 @@ impl World {
...
+
+        // Fallback to (0, 0) if no safe location found
+        (0, self.get_top_block(Vector2::new(0, 0)).await + 1, 0)
+    }
+
+    async fn check_spawn_safety(&self, x: i32, z: i32) -> Option<(i32, i32, i32)> {
+        let top_y = self.get_top_block(Vector2::new(x, z)).await;
+
+        // Check if there's solid ground and air above for spawning
+        if !(top_y < self.dimension.min_y || top_y >= self.dimension.height - 1) {
```
CuzImClicks: Use an early return instead

pumpkin/src/server/mod.rs
```diff
@@ -290,6 +290,36 @@ impl Server {
         };
 
         log::info!("All worlds loaded successfully.");
+
+        // Find and set a safe spawn location if using default spawn
+        let level_info = server.level_info.read().await;
+        let is_default_spawn = { level_info.spawn_x == 0 && level_info.spawn_z == 0 };
+        drop(level_info);
```
CuzImClicks:
```suggestion
       
        let is_default_spawn = { 

            let level_info = server.level_info.read().await;
            level_info.spawn_x == 0 && level_info.spawn_z == 0
};
```
Hey sorry there must have been a slight misunderstanding of what i meant... this is what i wanted to suggest

Snowiiii:
This seems very wrong and not vanilla behaviour, Vanilla uses biome noise to find the right biome chunk coordinate to spawn the player in, i already began implementing a FittestPositionFinder

DocteurPing:
ok got it i will close that PR then

PR #1386, opened 2026-01-26
feat: add /spawnpoint command

Purdze:
## summary
- implement `/spawnpoint` command to set player spawn points with optional position and angle
- add `force` flag to `RespawnPoint` to bypass bed/respawn anchor validation (matches vanilla behavior)
- support cross-dimension respawn (e.g., die in nether with overworld spawn point → respawn in overworld)
- fix respawn teleport by sending position packet after chunks load

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -28,6 +28,7 @@ use tokio::sync::{Mutex, RwLock};
 use tokio::task::JoinHandle;
 use uuid::Uuid;
 
+use pumpkin_data::block_properties::{BlockProperties, EnumVariants, HorizontalFacing};
 use pumpkin_data::damage::DamageType;
 use pumpkin_data::data_component_impl::{AttributeModifiersImpl, Operation};
 use pumpkin_data::data_component_impl::{EquipmentSlot, EquippableImpl, ToolImpl};
@@ -824,10 +825,10 @@ impl Player {
     /// Returns `Some(CalculatedRespawnPoint)` if a valid respawn point exists, `None` otherwise.
     ///
     /// # Behavior
-    /// - If `force` flag is set (via `/spawnpoint` command), returns the stored position directly
-    ///   without validating bed/anchor blocks.
-    /// - For beds (Overworld): validates the bed block still exists.
-    /// - For respawn anchors (Nether): validates the anchor block still exists.
+    /// - If `force` flag is set (via `/spawnpoint` command), validates the spawn position is safe
+    ///   (both the block and block above allow mob spawn).
+    /// - For beds: validates the bed block still exists and finds a valid spawn position around it.
+    /// - For respawn anchors (Nether): validates the anchor has charges and finds a valid spawn position.
     /// - Returns `None` if the spawn block is invalid/missing (caller should send
     ///   `NoRespawnBlockAvailable` game event and use world spawn).
     ///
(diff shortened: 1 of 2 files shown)


pumpkin/src/entity/player.rs
```diff
@@ -793,22 +793,65 @@ impl Player {
...
+            .send_packet_now(&CPlayerSpawnPosition::new(
+                block_pos,
+                yaw,
+                pitch,
+                dimension.minecraft_name.to_owned(),
+            ))
+            .await;
+    }
+
+    pub async fn get_respawn_point(&self) -> Option<(Vector3<f64>, f32, f32, Dimension)> {
```
CuzImClicks: While youre already changing stuff in this function, you might aswell rework it.

This function really needs it... first of all the return type is bad because its a tuple type, which means that important context such as variable names get lost. This is annoying when you later want to work with the function, for example when people develop plugins. 

The name also implies that it just returns the RespawnPoint, which it actually doesnt, because it **calculates** the respawn point based on the stored value. This function also shouldnt be sending a packet, because it could be called in places where its not actually wanted to send one.

There should also be a warning that the spawnpoint calculation can be overridden by the force flag in the RespawnPoint struct, which is a side effect that isnt directly visible when using the function. Maybe a mention in the function documentation would be nice :)

pumpkin/src/world/mod.rs
```diff
@@ -1875,21 +1876,106 @@ impl World {
 
         let data_kept = u8::from(alive);
 
-        // TODO: switch world in player entity to new world
+        let info = self.level_info.read().await;
```
CuzImClicks: This is holding a lock across an await point which we want to avoid. Obtain the lock, copy the info into local variables and release it again... you probably want something like

`let (spawn_x, spawn_y, spawn_z, spawn_yaw, spawn_pitch) = { ... };`

pumpkin/src/world/mod.rs
```diff
@@ -1875,21 +1876,106 @@ impl World {
...
+
+                (
+                    Vector3::new(
+                        f64::from(info.spawn_x) + 0.5,
+                        (top + 1).into(),
+                        f64::from(info.spawn_z) + 0.5,
+                    ),
+                    info.spawn_yaw,
+                    info.spawn_pitch,
+                    self.dimension,
```
CuzImClicks: This logic will have to be adjusted when #1381 gets merged

Purdze: Want me to factor that PR in based on it getting merged?

pumpkin/src/world/mod.rs
Purdze: Implemented, thank you!

pumpkin/src/world/mod.rs
CuzImClicks: I mean the calculation is wrong here anyways... if youre gonna figure out the y level you might aswell rerun the vanilla calculation... idk how much time the other pr is going to take, so i think you just add a fixme and make a comment on the other pr, mentioning that this would have to be changed

pumpkin/src/net/java/play.rs
CuzImClicks, line 1142: abilities update is sent twice

world/mod.rs L1986

pumpkin/src/world/mod.rs
```diff
@@ -1864,7 +1864,8 @@ impl World {
         }
     }
 
-    pub async fn respawn_player(&self, player: &Arc<Player>, alive: bool) {
+    #[allow(clippy::too_many_lines)]
+    pub async fn respawn_player(self: &Arc<Self>, player: &Arc<Player>, alive: bool) {
```
CuzImClicks, line 1864: this function also needs to handle what PlayerList::respawn in vanilla does.

- finding correct block to spawn on
- missing bed and respawn anchor

pumpkin/src/net/java/play.rs
Purdze: this has already been fixed, just not pushed

pumpkin/src/world/mod.rs
Purdze: this has already been fixed too, again just not pushed

Snowiiii:
Looks good overall. Thank you @Purdze :D

PR #1389, opened 2026-01-27
Prevent out-of-cache feature placement reads

antalervin19:
- **What was changed?** Guarded feature placement against reading block/height data outside the active generation cache. Specifically in the CountOnEveryLayer placement:
	- Updated `get_positions(...)` to skip positions when the relevant chunk is not present in the cache.
	- Updated `find_pos(...)` to early-return when the target or scanned positions move outside the cache bounds.
	- File: [pumpkin-world/src/generation/feature/placed_features.rs](pumpkin-world/src/generation/feature/placed_features.rs)

- **Why were these changes necessary?** Out-of-cache reads were being treated as air by `Cache::get_block_state`, which caused incorrect placement and surface checks along with excessive log spam (`"illegal get_block_state …"`). This could make chunks appear empty until a local block update forced a valid refresh.

- **What is the impact of this change?**
	- Prevents out-of-bounds reads from being interpreted as air.
	- Reduces noisy error logs during world generation.
	- Improves reliability of chunk population without requiring player interaction.
	- No behavioral change for in-bounds positions and only minimal overhead from the added checks.

- **Are there any known issues or limitations?** If a generation stage legitimately needs to sample outside the current cache window, these guards will skip those positions. In such cases, the stage’s `write_radius` should be increased or cache alignment verified. These checks address the symptoms but do not correct a mismatched cache center.

- **Include any relevant context, such as related issues or discussions.** The issue originates from bounds checks in `Cache::get_block_state()` (see [pumpkin-world/src/chunk_system.rs](pumpkin-world/src/chunk_system.rs)). Saw the errors during spinning up a new server and joining it:  
	`illegal get_block_state Vector3 { x: -313, y: 115, z: -317 } cache pos (-11, -11) size 3`, indicating placement positions outside the cache window.

- **No Clippy Warnings:** Pass ✅

- **Passing Unit Tests:** Pass ✅

- **FMT Formatting Check:** Pass ✅


pumpkin-world/src/generation/feature/placed_features.rs
```diff
@@ -344,12 +347,22 @@ impl CountOnEveryLayerPlacementModifier {
...
         let mut found_count = 0;
         let mut current_block_state = GenerationCache::get_block_state(chunk, &mutable_pos.0);
 
         for j in (chunk.bottom_y() as i32 + 1..=y).rev() {
             mutable_pos.0.y = j - 1;
+            // stop, scan is outside of cache bounds
+            if chunk
+                .get_chunk(mutable_pos.0.x >> 4, mutable_pos.0.z >> 4)
+                .is_none()
+            {
```
Purdze: checks not needed, this is checking the same chunk as the check at the start of the function. this will always pass if the first one passes too

antalervin19: True, true, ur right.

antalervin19: NGL if you want i can remove that one extra IF. takes nothing.

Purdze: yeah :D

pumpkin-world/src/generation/feature/placed_features.rs
```diff
@@ -305,15 +305,18 @@ impl CountOnEveryLayerPlacementModifier {
...
+        let mut positions = Vec::new();
+        let mut i = 0;
         let mut bl;
 
         loop {
             bl = false;
             for _j in 0..self.count.get(random) {
                 let x = random.next_bounded_i32(16) + pos.0.x;
                 let z = random.next_bounded_i32(16) + pos.0.z;
+                if chunk.get_chunk(x >> 4, z >> 4).is_none() {
```
Snowiiii, line 317: You now check for the chunk twice

Snowiiii:
Feel free to reopen, i also made the warns `log::debug

PR #1391, opened 2026-01-28
fix: armor stands now break instantly in creative mode

### LINKED ISSUE #1281: Armor stands are not breakable instantly

### Operating System

Windows 11

0.1.0-dev+1.21.11

### What happened?

You cannot break armor stands instantly in creative mode.

### To Reproduce

Try breaking an armor stand in creative mode and it is not instant and the armor stand item drops.

### Expected behavior

They should break instantly when the player is in creative mode.

Purdze:
## Summary
- pass `Player` instead of `Entity` as damage source so `get_player()` works
- creative mode: spawn break particles, play sound, remove without drops
- add `spawn_break_particles` method (TODO: use oak plank block particles, requires block state data in particle system)

Fixes #1281


Snowiiii:
Looks good. Thank you @Purdze :D

PR #1392, opened 2026-01-28
fix: optimized, working fluids

RoosterBooster007:
## Description
These changes fix the current fluid dynamics issues plaguing Pumpkin rn.

**Before:**
- Fluids ALWAYS ticked... even sources/oceans (leading to extreme lag, timeout kicks, and hangs)
- Fluids didn't flow/expand quite right
- Fluids used DFS and recursion for flow (vanilla), which is slower (memory inefficient)

**After:**
- Fluids do not always tick (they "sleep" after inactivity as expected)
- Fluids seem to flow/expand right (but may need adjustments down the line)
- Fluids use an iterative BFS algorithm (custom) for flow, which is estimated to be 3-5x faster and use less memory

Other fixes:
- Door tops now cause water updates underwater (instead of leaving an empty water pocket)
- Fluids hole-find for up to 5 blocks like vanilla MC (used to be 4)

## Testing
I did my best to test for issues... including the lava x leaf litter one 💀
I'm sure bugs will appear as they always do, but this is really just a better solution (compared to what we have atm)

I plan to add the gamerules for source creation in another PR...

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,17 +25,20 @@ and customizable experience. It prioritizes performance and player enjoyment whi
 - **Extensibility**: Provides a foundation for plugin development.
 
 > [!IMPORTANT]
-> Pumpkin is currently under heavy development
+> Pumpkin is currently under heavy development.
+>
+> [See what needs to be done before the 1.0.0 Release](https://github.com/Pumpkin-MC/Pumpkin/issues/449)
 
 ## Features
 
 - [x] Configuration (toml)
-- [x] Server Status/Ping
-- Networking
+- [Tracking: Protocol](https://github.com/Pumpkin-MC/Pumpkin/issues/1401)
+  - [x] Server Status/Ping
   - [x] Encryption
   - [x] Packet Compression
-- World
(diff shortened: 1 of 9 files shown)


pumpkin/src/block/fluid/physics.rs
```diff
@@ -0,0 +1,51 @@
...
+        || id == Block::LADDER.id
+        || id == Block::POINTED_DRIPSTONE.id
+        || id == Block::SCAFFOLDING.id
+    {
+        return false;
+    }
+
+    // Only replace air, explicitly replaceable blocks, or carpets
+    block_state.replaceable()
+        || id == Block::AIR.id
```
CuzImClicks: .is_air()

Snowiiii:
Did you make sure this matches vanilla?

RoosterBooster007:
Yes. Though some waterlogged blocks still behave strangely. I can fix this in a separate PR (as well as any issues that arise here).

Snowiiii:

PR #1398, opened 2026-01-28
Add /restart command with safe restart flow

xStarless-Skyx:
## Description

Adds a built‑in /restart command that cleanly restarts the server. The command sets a restart flag, triggers a normal shutdown, and on Unix/macOS replaces the process with a fresh instance so the console stays attached. Also registers the command and permission at OP level 4.

## Testing

cargo build --release
Ran ./target/release/pumpkin


xStarless-Skyx:
after further testing i found a bug ill fix and open a new pull request

PR #1399, opened 2026-01-29
Restart Command (take 2) 

xStarless-Skyx:
## Description

Adds a built‑in /restart command that cleanly restarts the server. The command sets a restart flag, triggers a normal shutdown, and on Unix/macOS replaces the process with a fresh instance so the console stays attached, on Windows it spawns a new process for the console barely noticeable. it registers the command and permission at OP level 4 and adds a  little bit of logging. Fairly useful for my case with my plugin development where im just rebuilding my plugin and replacing it and restarting the server :D 

## This also adds 3 new things to the server config file. 

1) Server restart delay 

2/3.   It lets you set your own message for server restart and server close. instead of showing the player server closed on a restart, it's set as default to server restarting when you do a restart and server closed when closing the server.

## Testing

cargo build --release
Ran ./target/release/pumpkin
Tested on both MacOs on an M1 Macbook and on Windows 11 with my pc.

### DIFF
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -9,6 +9,8 @@ use pumpkin_data::packet::CURRENT_MC_PROTOCOL;
 #[cfg(unix)]
 use std::os::unix::process::CommandExt;
 use std::process::Command;
+#[cfg(not(unix))]
+use std::process::Stdio;
 use std::{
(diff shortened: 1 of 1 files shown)


xStarless-Skyx:
this ones a total mess 💀

PR #1400, opened 2026-01-29
fix: eggs, snowballs, and babies

### LINKED ISSUE #567: Projectiles do not break

### Operating System

CachyOS

3f1fd5a

### What happened?

I threw a snowball, and when it hit a block, it did not break, i threw an egg too, same behavior. Also is limited to a diagonal on the x-z axis, and when you re-connect projectiles launch from where they we're launched again

### To Reproduce

- Get a throwable projectile, like an snowball
- Shoot it aiming to a block

### Expected behavior

The projectile must do it's thing (Spawning a chicken, do damage, tp, etc) and then break itself/disappear

### LINKED ISSUE #1332: When throwing eggs or snowballs as soon as they hit a block they are hanging in them and then despawn after some seconds

### Operating System

Windows

Pumpkin 0.1.0-dev+1.21.11

### What happened?

When throwing eggs or snowballs as soon as they hit a block they are hanging in them and then despawn after some seconds

and the new copper egg / blue egg you cant even throw them

### To Reproduce

throw an Egg / Snow ball

### Expected behavior

it should like show the vanilla egg / Snowball animation

RoosterBooster007:
## Description
- Thrown eggs (including blue and brown) and snowballs behave as expected now (they hit/clear)
- Eggs now have a chance at spawning (multiple) baby chickens 🍗
   - You can now set entity age before spawning (for babies)
- Snowballs damage blazes

## Testing
However, the chickens are all temperate/default as entity variants need to be implemented
Sounds and particles work (but could be tweaked later)
Also, velocity and drag values could be updated (projectiles seem fast atm)

Wind charges, epearls, arrows, and firework colors are next...

Fixed #567
Fixed #1332

### DIFF
--- a/pumpkin/src/entity/mod.rs
+++ b/pumpkin/src/entity/mod.rs
@@ -208,12 +208,6 @@ pub trait EntityBase: Send + Sync + NBTStorage {
         Box::pin(async {})
     }
 
-    /// Returns true if this entity is a projectile (egg, snowball, etc.).
-    /// Default is false.
-    fn is_projectile(&self) -> bool {
-        false
-    }
-
     fn get_entity(&self) -> &Entity;
     fn get_living_entity(&self) -> Option<&LivingEntity>;
 
--- a/pumpkin/src/entity/projectile/egg.rs
+++ b/pumpkin/src/entity/projectile/egg.rs
@@ -101,10 +101,6 @@ impl EntityBase for EggEntity {
         self
     }
 
-    fn is_projectile(&self) -> bool {
-        true
-    }
-
     fn on_hit(&self, hit: crate::entity::projectile::ProjectileHit) -> EntityBaseFuture<'_, ()> {
         Box::pin(async move {
             let world = self.get_entity().world.load();
--- a/pumpkin/src/entity/projectile/firework_rocket.rs
+++ b/pumpkin/src/entity/projectile/firework_rocket.rs
@@ -159,8 +159,4 @@ impl EntityBase for FireworkRocketEntity {
     fn as_nbt_storage(&self) -> &dyn crate::entity::NBTStorage {
         self
     }
-
-    fn is_projectile(&self) -> bool {
-        true
-    }
 }
--- a/pumpkin/src/entity/projectile/mod.rs
+++ b/pumpkin/src/entity/projectile/mod.rs
@@ -1,6 +1,7 @@
 use super::{Entity, EntityBase, NBTStorage, living::LivingEntity};
 use crate::server::Server;
 use pumpkin_data::BlockDirection;
+use pumpkin_data::entity::EntityType;
 use pumpkin_protocol::java::client::play::CEntityVelocity;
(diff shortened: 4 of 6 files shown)


pumpkin/src/entity/projectile/egg.rs
```diff
@@ -0,0 +1,176 @@
...
+use pumpkin_protocol::codec::item_stack_seralizer::ItemStackSerializer;
+use pumpkin_protocol::java::client::play::Metadata;
+use pumpkin_util::math::vector3::Vector3;
+use pumpkin_world::item::ItemStack;
+use tokio::sync::Mutex;
+use uuid::Uuid;
+
+pub struct EggEntity {
+    pub thrown: ThrownItemEntity,
+    pub item_stack: Mutex<ItemStack>,
```
CuzImClicks: Just for my understanding… why do we need to store the itemstack behind a mutex when we know that the itemstack (when initialized via new()) can and probably should only ever be a single egg?

pumpkin/src/entity/projectile/egg.rs
```diff
@@ -0,0 +1,176 @@
...
+            let r: u8 = rand::random(); // 0..=255
+            let to_spawn = if r == 0 {
+                4
+            } else if r < 32 {
+                1
+            } else {
+                0
+            };
+
+            // Spawn chickens in a separate task to prevent stack overflow
```
CuzImClicks, line 128: How does spawning 1 or 4 entities cause a stack overflow?

pumpkin/src/entity/projectile/egg.rs
RoosterBooster007: Nice catch, we don't. I'll update this to read or clone when I get back to my PC.

pumpkin/src/entity/projectile/egg.rs
RoosterBooster007: Something strange was happening when spawning multiple chickens in one tick, but now I think it's due to a projectile collision loop. I can try to fix this by using a simple ``has_hit`` boolean flag to stop the overflow, instead.

RoosterBooster007:
@Snowiiii, I finally fixed the clippy errors

pumpkin/src/entity/mod.rs
```diff
@@ -189,6 +204,16 @@ pub trait EntityBase: Send + Sync + NBTStorage {
         Box::pin(async {})
     }
 
+    fn on_hit(&self, _hit: crate::entity::projectile::ProjectileHit) -> EntityBaseFuture<'_, ()> {
+        Box::pin(async {})
+    }
+
+    /// Returns true if this entity is a projectile (egg, snowball, etc.).
+    /// Default is false.
+    fn is_projectile(&self) -> bool {
```
Snowiiii: This seems kinda unnecessary, i think instead we may should just check the type

RoosterBooster007:
@Snowiiii does this look okay now?

Snowiiii:
Overall looks fine. Thank you @RoosterBooster007 :D

PR #1408, opened 2026-01-29
fix: add out of bounds checking for some commands

Laptop59:
## Description

Previously, `/fill` and `/setblock` commands did not have out-of-bounds checking. This PR adds the checking, and also
- each mode in the `/fill` command has its own `Filler` which now removes the need for copy-pasting the same 3D loop. it has two methods, `execute_for_region` to do the 3D loop for each block in it and the other `execute_for_pos` for one block
- `Context` is passed in both methods, which stores information about the command parameters, the world, placed blocks and the `Vec` of `BlockPos` to update (from the previous implementation)
- moved edge checking to its own `const` inlined function in `Context`
- added checking in the `/fill` command if too many blocks were attempted to be filled (gamerule max_block_modifications)
- moved some error feedback into an `Err` result for red color
- added functions in `World` to return the minimum y and maximum y (differs for each `Dimension`), helper methods related to them and removed floating-point ones (replaced with `i32`)
- in the java protocol, removed hardcoded y limits of the world and replaced them with ones dependent on the `World`
- adds `strict` mode (1.21.5) for `/setblock`
- added a way to floor a Vector3 for the out-of-bounds checking change in the `/tp` command

## Testing
- commands can be run to check results

### DIFF
--- a/pumpkin-util/src/math/vector3.rs
+++ b/pumpkin-util/src/math/vector3.rs
@@ -338,6 +338,28 @@ impl<T: Math + Copy + Into<f64>> Vector3<T> {
 }
 
 impl<T: Math + Copy + Into<f64>> Vector3<T> {
+    pub fn floor_to_i32(&self) -> Vector3<i32> {
+        let x: f64 = self.x.into();
+        let y: f64 = self.y.into();
+        let z: f64 = self.z.into();
+        Vector3 {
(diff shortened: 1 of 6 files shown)


Snowiiii:
Looks good. Thank you @Laptop59 :D

PR #1410, opened 2026-01-29
fix: fluids now follow gamerules

RoosterBooster007:
## Description
Water and lava will now check their respective gamerules before converting to a source.

## Testing
I've tested it, and it works fine.


Snowiiii:

PR #1412, opened 2026-01-30
Feat /restart command (take 3) 

xStarless-Skyx:
## Description

Adds a built‑in /restart command that cleanly restarts the server. The command sets a restart flag, triggers a normal shutdown, and on Unix/macOS replaces the process with a fresh instance so the console stays attached, on Windows it spawns a new process for the console barely noticeable. it registers the command and permission at OP level 4 and adds a little bit of logging. 

Fairly useful for my case with my plugin development where im just rebuilding my plugin and replacing it and restarting the server :D

## This also adds 3 new things to the server config file.
- Server restart delay

- It lets you set your own message for server restart and server close. instead of showing the player server closed on a restart, it's set as default to server restarting when you do a restart and server closed when closing the server.

## Testing

cargo fmt --check
cargo clippy --all-targets --all-features
cargo clippy --release --all-targets --all-features

cargo build --release
Ran ./target/release/pumpkin

Tested on both MacOs on an M1 Macbook and on Windows 11 with my pc.

### Sorry for the mess. I've got things figured out now this sort of thing wont happen again

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -113,6 +113,12 @@ pub struct BasicConfiguration {
     pub white_list: bool,
     /// Whether to enforce the whitelist
     pub enforce_whitelist: bool,
+    /// Delay in milliseconds before restarting the server process.
+    pub restart_delay_ms: u64,
+    /// Message shown to players when the server stops.
+    pub server_stop_message: String,
+    /// Message shown to players when the server restarts.
+    pub server_restart_message: String,
 }
 
 impl Default for BasicConfiguration {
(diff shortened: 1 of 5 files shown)


Snowiiii:
Hi, /restart seems to not be a Vanilla command and seems to not provide enough value compared to the implementation effort and the increased security risk of this introduces

xStarless-Skyx:
> Hi, /restart seems to not be a Vanilla command and seems to not provide enough value compared to the implementation effort and the increased security risk of this introduces

It's not a vanilla command but I'm pretty sure paper and I think other server softwares have it, I don't mind that much anyways since I wrote something into my server wrapper to restart servers that don't have a restart command anyways lol. I just didn't really know what to work on heh

PR #1415, opened 2026-01-30
feat: make commands return i32 as result on success

Laptop59:
## Description
Every command in Java Edition results in an `int` as a result upon success (depending on the command itself). If the command fails, it always results in `0`. This PR provides a way for each command to propagate an `i32` upon command success.

- For example, `/random value 1..5` always gives an `int` between 1 and 5 (inclusive) which can be extracted (along with printing it to the command executor in the chat, but that is separate)
- This value can then be captured by commands like `/return run ...` and `/execute store result ...`. So a complete example of capturing the `int` and usage of it by another command is `/execute store result score @s random_rolls run random value 1..5`, which stores this random value in the executing entity's score entry in the `random_rolls` scoreboard.
- This PR tries to replicate the `i32` results that vanilla also provides upon command success.
- The hope for this PR is to make it easier to introduce the commands requiring this result (i.e. `/execute` and `/return`).

To achieve this in Pumpkin,
- `CommandResult<'a>` has been changed from `Pin<Box<dyn Future<Output = Result<(), CommandError>> + Send + 'a>>` to `Pin<Box<dyn Future<Output = Result<i32, CommandError>> + Send + 'a>>`.
- This does not affect the output of the command for the players running it.
- Each command will need to be adjusted to return an integer similar to what vanilla does in that command.

The docs will also need to be updated.

## Testing
all commands have been tested (`/help` does not work, but that was already the case before this PR)

### DIFF
--- a/pumpkin/src/block/blocks/signs.rs
+++ b/pumpkin/src/block/blocks/signs.rs
@@ -2,9 +2,10 @@ use std::sync::Arc;
 use std::sync::atomic::Ordering;
 
 use pumpkin_data::Block;
-use pumpkin_data::block_properties::BlockProperties;
+use pumpkin_data::BlockDirection;
 use pumpkin_data::block_properties::EnumVariants;
 use pumpkin_data::block_properties::Integer0To15;
+use pumpkin_data::tag::Taggable;
 use pumpkin_inventory::screen_handler::InventoryPlayer;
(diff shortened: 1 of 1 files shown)


Snowiiii:
Looks good. Thank you @Laptop59 :D

PR #1416, opened 2026-01-31
fix: make TargetSelector new function public

BjornTheProgrammer:
## Description

This just makes the new function in TargetSelector public, since it seems like it should be. I would also like to use it to get entities with PatchBukkit.

## Testing

Testing isn't really required for this since I'm just making the function public.


Snowiiii:
Looks good. Thank you @BjornTheProgrammer :D

Issue #1301, opened 2026-01-14
Assertion failed when pinging server status when behind mc-router


FabseGP:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Artix Linux

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.11

### What happened?

When the minecraft launcher tries to ping the status of the pumpkin-server behind [mc-router](https://github.com/itzg/mc-router), it panics with the following error:

```
[INFO] (1) Server is now running. Connect using port: Java Edition: 0.0.0.0:25599
$ thread 'tokio-runtime-worker' panicked at /home/fabse/.cache/paru/clone/pumpkin-git/src/pumpkin-git/pumpkin-protocol/src/ser/mod.rs:373:9:
     assertion failed: data.len() <= bound
```

Nothing in the logs from mc-router indicates what happened

### To Reproduce

Setup mc-router and link a domain to the pumpkin server. The following docker-compose can be used:

```
services:
  mc-router:
    image: itzg/mc-router
    restart: unless-stopped
    container_name: mc-router
    environment:
      MAPPING: |
        DOMAIN_NAME=SERVER_IP:PUMPKIN_PORT
    ports:
      - "25565:25565"
    networks:
      - mcrouter
networks:
  mcrouter:
    driver: bridge
```

Add the server with the chosen domain name to the launcher and let it ping the status

### Expected behavior

Pinging the status of the server should not make it crash

FabseGP:
It only happened when I set a path to a custom favicon and it was larger than 32x32 in dimensions

PR #1418, opened 2026-01-31
fix: incorrect favicon causes crashing during server ping

FabseGP:
## Description

The [server list ping interface](https://minecraft.wiki/w/Java_Edition_protocol/Server_List_Ping) specifies the following for the favicon:

> The favicon field is optional. If specified, it should be a [PNG](https://en.wikipedia.org/wiki/Portable_Network_Graphics) image that is [Base64](https://en.wikipedia.org/wiki/Base64) encoded (newlines (\n) are no longer supported since 1.13) and prepended with data:image/png;base64,. It should also be noted that the source image must be exactly 64x64 pixels, otherwise the Notchian client will not render the image.

But neither conditions is upheld currently. Change favicon_path to be `Option<String>` and ensure it's a PNG and 64x64; if not, fallback to the default favicon

## Testing

DIfferent sizes and formats of favicons have been tested, those not ending with .png or with a size of 64x64 fell back to the default favicon


FabseGP:
reopened because I cleaned up my branches

Snowiiii:
I don't like the way you fixed it tbh using manual byte reads and wrapping `favicon_path` in an Option which is not necessary since it will just not load the icon if there isn't one, 
There was a time where we checked the size of the icon via the png crate, but adding a new crate for just a small feature seems unjustified, I think a crash/panic is fine in this case when the user made a configuration error

FabseGP:
The reason for the Option was to avoid the warning that would show up with the default config (where use_favicon is set and the path is "icon.png"); the default config shipped should not show any warnings at all imo

And yeah, I was thinking about including either the png or image crate, but that's too much for a small feature. Still, I don't think a crash/panic is worth it too and even so, it's not clear for the user why a crash happened; I was debugging a crash earlier and thought it was because of the proxy I used, when it was the favicon that was the fault (#1301)

Snowiiii:
I mean i'm fine with manually checking the size using some Signature magic but it also looks like some things were just hardcoded here which are not necessary

pumpkin-config/src/lib.rs
```diff
@@ -201,8 +201,12 @@ pub trait LoadConfiguration {
             merged_config
         } else {
             let content = Self::default();
+            let mut file_content = toml::to_string(&content).unwrap();
+            if !file_content.contains("favicon_path") {
```
Snowiiii: It is not necessary to do anything here just because of one simple config option

FabseGP: True, I added it because otherwise the user would have to check the documentation on how to add a custom favicon (since it wouldn't be in the config anymore), but that's prob fine; fixed now

FabseGP:
The rest I don't think are hardcoded unnecessarily

Snowiiii:
Yeah looks definitely better now. Thanks @FabseGP :D

PR #1419, opened 2026-01-31
chore: detect unused dependencies in CI + rearrange CI-steps

FabseGP:
## Description

Use cargo-machete in CI to detect unused dependencies + rearrange CI-steps and have the steps depend on each other (no need to run clippy if rustfmt fails for example) 

## Testing

serde_json in pumpkin-world & phf in pumpkin-data had to be whitelisted


Snowiiii:
Looks good. Thank you @FabseGP :D 🦛

PR #1421, opened 2026-02-01
feat(data): add is_solid and is_air helper methods to Block

aShanki:
## Description

Adds convenience helper methods to `Block` for checking block properties based on the default state:
- `is_solid()` - Returns whether the block's default state is solid
- `is_air()` - Returns whether the block's default state is air

These are simple delegations to the existing `BlockState` methods, providing a more ergonomic API when working with `Block` directly.

## Testing

Trivial change - delegates to existing `BlockState::is_solid()` and `BlockState::is_air()` methods.


Snowiiii:
Looks fine. Thanks @aShanki

Issue #1315, opened 2026-01-16
Crawling under a block damages the player


FireAcidCow:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

0.1.0/6f18d165

### What happened?

When you get into crawl/swim/elytra fly mode and are underneath a block, you start taking damage as if you were suffocating.

### To Reproduce

1. put some blocks on the surface of the ocean
2. go into survival mode
3. swim underneath said blocks

### Expected behavior

Don't take damage because your hitbox isn't in a block

PR #1425, opened 2026-02-02
Fix suffocation damage when swimming or crawling under blocks #1315

Poneaswaran:
**Summary**

This PR fixes an issue where entities (especially players) would incorrectly take suffocation damage when swimming, crawling, or fall-flying beneath solid blocks, even though their hitbox was not intersecting any block.

**Problem**

When an entity entered certain poses (Swimming, FallFlying, crawl-equivalent states), the suffocation logic continued to assume a standing eye height. As a result:

The eye-level collision box was evaluated too high

Entities swimming under blocks (e.g., ocean surface slabs/blocks) were treated as being inside a block

Suffocation damage was applied incorrectly

This behavior was reproducible in survival mode by swimming under blocks placed at water level.

**Root Cause**

The suffocation check (tick_block_collisions) used a fixed or standing eye-height assumption instead of the pose-adjusted eye height, causing false-positive intersections between the eye-level AABB and block collision shapes.

**Fix**

Suffocation detection now uses get_eye_height() which correctly accounts for entity pose:

Swimming

FallFlying

Default standing pose

The eye-level bounding box is adjusted accordingly before intersection checks

Suffocation damage is only applied when the actual eye position intersects a solid block

Expected Behavior After Fix

No suffocation damage when:

Swimming under blocks

Crawling under blocks

Elytra/fall-flying beneath blocks

Suffocation damage still applies correctly when the eye position is genuinely inside a solid block

**Testing Performed**

Placed solid blocks at ocean surface level

Entered survival mode

Swam, crawled, and fall-flew underneath blocks

Verified:

No damage taken when not intersecting blocks

Damage still occurs when fully enclosed

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -90,12 +90,15 @@ use super::combat::{self, AttackType, player_attack_sound};
 use super::hunger::HungerManager;
 use super::item::ItemEntity;
 use super::living::LivingEntity;
-use super::{Entity, EntityBase, NBTStorage, NBTStorageInit};
+use super::{Entity, EntityBase, NBTStorage, NBTStorageInit, SWIMMING_EYE_HEIGHT};
 use pumpkin_data::potion::Effect;
 use pumpkin_world::chunk_system::ChunkLoading;
 
 const MAX_CACHED_SIGNATURES: u8 = 128; // Vanilla: 128
 const MAX_PREVIOUS_MESSAGES: u8 = 20; // Vanilla: 20
+
+/// Eye height for players in crouching pose
+const CROUCHING_EYE_HEIGHT: f64 = 1.27;
 
 pub const DATA_VERSION: i32 = 4671; // 1.21.11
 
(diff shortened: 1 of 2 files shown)


Snowiiii:
I also looks like there were changed made to the bed system?

pumpkin/src/block/blocks/bed.rs
BjornTheProgrammer, line 147: Why is there bed related code in this PR for suffocation damage? Consider removing it and creating a new PR with this change.

pumpkin/src/entity/player.rs
```diff
@@ -1479,25 +1068,28 @@ impl Player {
...
         self.living_entity.entity.pos.load()
     }
 
     pub fn eye_position(&self) -> Vector3<f64> {
-        let eye_height = f64::from(self.living_entity.entity.entity_dimension.load().eye_height);
+        let pose = self.living_entity.entity.pose.load();
+        let eye_height = match pose {
+            EntityPose::Crouching => CROUCHING_EYE_HEIGHT,
+            EntityPose::Swimming | EntityPose::FallFlying => f64::from(SWIMMING_EYE_HEIGHT),
+            _ => f64::from(self.living_entity.entity.standing_eye_height),
```
BjornTheProgrammer, line 1084: Is it possible for entities to have poses, and if so is there crouching, swimming, and fall flying speeds the same as the players? If they are different we might want to update `http://github.com/Pumpkin-MC/Extractor` to get those values, and feed it into `pumpkin-data/build/entity_type.rs`

BjornTheProgrammer (changes requested):
Great work! Just some questions before this can be merged.

Snowiiii:
Feel free to reopen

PR #1427, opened 2026-02-02
feat: add string reader from Brigadier and its errors

Laptop59:
## Description
The current command parser is not fundamentally accurate to what *Minecraft* uses. *Minecraft* uses a library called [*Brigadier*](https://github.com/Mojang/brigadier) to parse commands (made by Mojang). This can be demonstrated with an example:

### Example
```
/tellraw @s {"text":"hello","color":"green"}
/tellraw @s {text:"hello",color:"green"}
/tellraw @s {text:hello,color:green}
```
- The vanilla implementation responds to these commands correctly, and prints a green `hello` to the player.
- The problem with Pumpkin's implementation is that it only recognizes a string with a set of quotes (`""`). *Brigadier* actually also allows identifiers to be interpreted as strings, leading to the last two working properly. In short, it's not as simple as using a JSON parser.

### In general
- The errors currently thrown to the player don't have any context, (*i.e.* they don't show where the error actually occured). *Brigadier* can throw error messages like, for example, when running `/data get @s Health`:
> Incorrect argument for command  
data get <ins><span>@</span>s Health</ins>*<--[HERE]*

### Proposed Solution
- To get *Brigadier*-like parsing, we can start off with a struct called a **`StringReader`**. It does the same functions that the one in *Brigadier* does.
- It is not linked to any Pumpkin functionality (for now). Instead, this PR hopes to make a small step towards achieving *Brigadier*-like error messages from commands in *Minecraft*.
- `CommandSyntaxError`, which is an error homologous to the `CommandSyntaxException` from *Brigadier*, has also been added, which are errors instantiated from the type **`CommandErrorType<N>`** for **`N`** translation arguments. A table of relations are:

| **Struct**            | **Equivalent in Brigadier**    |
| --------------------- | ------------------------------ |
| `CommandErrorType<0>` | `SimpleCommandExceptionType`   |
| `CommandErrorType<1>` | `DynamicCommandExceptionType`  |
| `CommandErrorType<2>` | `Dynamic2CommandExceptionType` |
| `CommandErrorType<3>` | `Dynamic3CommandExceptionType` |
| `CommandErrorType<4>` | `Dynamic4CommandExceptionType` |
| `CommandErrorType<N>` | `DynamicNCommandExceptionType` |
> [!NOTE]  
> While implementing this type, I found out due to a Clippy error that a `TextComponent` has a size of `88` bytes.  
> After wrapping `content: TextContent` (which occupied `56` bytes) contained in `TextComponentBase` in a `Box`, `TextComponent`  now has a size of `40` bytes. I think this should be done as a lot of code relies on using them without references or `Box`es.
- Combining them with `Result`s will allow them to be propagated to the caller.

## Testing

Added test cases to the `StringReader` as well, and all of them pass.

### DIFF
--- a/assets/translations/vi_vn.json
+++ b/assets/translations/vi_vn.json
@@ -1,0 +1,12 @@
+{
+    "commands.pumpkin.version": "Pumpkin %s\n",
+    "commands.pumpkin.description": "Giúp bất cứ ai có thể host một server Minecraft nhanh và mượt.\n",
+    "commands.pumpkin.minecraft_version": "(Minecraft %1$s, Giao thức %2$s)\n",
+    "commands.pumpkin.github": "[Repositorys GitHub]",
+    "commands.pumpkin.website": "[Trang web]",
+    "commands.pumpkin.version.hover": "Click để sao chép phiên bản",
+    "commands.pumpkin.description.hover": "Click để sao chép mô tả",
+    "commands.pumpkin.minecraft_version.hover": "Click để sao chép phiên bản Minecraft",
+    "commands.pumpkin.github.hover": "Click để mở repository.",
+    "commands.pumpkin.website.hover": "Click để mở trang web."
+}
--- a/pumpkin-config/src/chat.rs
+++ b/pumpkin-config/src/chat.rs
@@ -1,9 +1,13 @@
 use serde::{Deserialize, Serialize};
 
+/// Configuration for in-game chat behaviour.
+///
+/// Controls chat formatting and display.
 #[derive(Deserialize, Serialize)]
 #[serde(default)]
 pub struct ChatConfig {
-    /// The custom chat format (doesn't work with secure chat)
+    /// The custom chat format.
+    /// `Note`: it does not apply when secure chat is enabled.
     pub format: String,
 }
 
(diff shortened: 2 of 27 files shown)


Snowiiii:
Hey, Why not link it directly to any Commands in this PR?, so for example tellraw would be fixed

Laptop59:
It shouldn't be linked to commands, but rather to argument type parsers. Commands just take their values. In Brigadier, commands don't do their own parsing.

That would require a fundamental change to the command system. Instead of consuming Args, argument types would need to take a StringReader (as seen in Brigadier's code).

That would require a lot of code changes. If you don't think that's too much for one PR, I could try continuing it. Let me know what you think!

Laptop59:
The text component in commands is actually SNBT, and not JSON. So for that we would actually need Ops and PackratParser as well (which takes a StringReader)

Snowiiii:
> It shouldn't be linked to commands, but rather to argument type parsers. Commands just take their values. In Brigadier, commands don't do their own parsing.
> 
> That would require a fundamental change to the command system. Instead of consuming Args, argument types would need to take a StringReader (as seen in Brigadier's code).
> 
> That would require a lot of code changes. If you don't think that's too much for one PR, I could try continuing it. Let me know what you think!

Ah okay i though it would be easier, than this is already good

Snowiiii:
LGTM. Thank you @Laptop59 :D

Issue #1429, opened 2026-02-03
Chests can be opened when block is placed on top of them


Frycek1:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11

### Server Software Version/Commit

0.1.0-dev+1.21.11 (Commit: 360e138)

### What happened?

As title said - it is possible to open chests despite having solid block on top of them. This bug affects chests, trapped chests in both single and double variants as well as ender chests.

### To Reproduce

* Place chest (single or double) / trapped chest (single or double) / ender chers
* Place solid block on top of chest
* Right click chest
* Chest is opened

### Expected behavior

It should not be possible to acces chest inventory when solid block is directly above it.

PR #1430, opened 2026-02-03
fix: block chest opening when solid block above

Frycek1:
Fixes: #1429

## Description
This PR fixes an issue where standard chests and ender chests could be opened despite having a solid block placed directly above them.

## Changes:
* Added a has_block_on_top  and is_chest_blocked helper functions to check for solid blocks at pos.up().
* Implemented the check in normal_use for both chest.rs and ender_chest.rs.
* Added logic to check the neighboring block if a double chest is interacted with.

## Testing
Tested locally:
Placed a single chest, double chest, and ender chest.
Placed a solid block (e.g., stone) above them.
Tried to interact (right-click) - the inventory GUI no longer opens and the lid animation does not trigger.
Placed a non-solid block (e.g., stairs) above them - chests open correctly.

### DIFF
--- a/pumpkin-data/src/blocks.rs
+++ b/pumpkin-data/src/blocks.rs
@@ -114,7 +114,7 @@ impl Block {
 
     /// Returns whether this block is solid (based on default state)
     pub fn is_solid(&self) -> bool {
-        self.default_state.is_solid_block()
+        self.default_state.is_solid()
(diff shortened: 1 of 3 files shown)


Snowiiii:
Looks good. Thank you @Frycek1 :D

PR #1434, opened 2026-02-03
feat: implement nether fortress structure generation

Purdze:
## summary
- implement all 14 nether fortress piece types
- includes blaze spawner placement, chest generation, and nether wart rooms

### DIFF
--- a/pumpkin-world/src/chunk_system/chunk_state.rs
+++ b/pumpkin-world/src/chunk_system/chunk_state.rs
@@ -137,6 +137,10 @@ impl StagedChunkEnum {
             // the Biome Step, this should be more efficient
             Self::Biomes => &[Self::Empty],
             Self::StructureStart => &[Self::Biomes],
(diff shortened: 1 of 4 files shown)


Snowiiii:
@Purdze when do you plan to finish this ?

RoosterBooster007:
The recent changes to `get_chunk` in `generation_cache.rs` seem to be causing a deadlock. If a neighbor chunk finishes generating quickly and becomes a `Level` chunk, the generator receives `None` and assumes the neighbor hasn't loaded yet. The scheduler retries the task infinitely, resulting in a permanent world-gen freeze. A "fully finished" chunk is being treated as a "missing" chunk. Right now, the world gen freezes, and I keep timing out.

When I first launched the server, with some of an existing world, I got the `set_block_state` out-of-chunk panic from earlier (`proto_chunk.rs`), too. This happened in the overworld.

PR #1437, opened 2026-02-04
feat: implement world carvers (caves and ravines)

aShanki:
1. Added cave and ravine carvers so the world generator can hollow out terrain to create natural cave systems and canyons
2. Created a carving mask to keep track of which blocks have been carved, preventing carvers from overwriting each other
3. Extended proto-chunks to support carving stages so caves can be carved before the chunk is fully generated
4. Hooked up the aquifer sampler so underwater caves properly fill with water instead of leaving air pockets below sea level
5. Wired up biome configuration so different biomes can have different cave/ravine generation settings

### DIFF
--- a/pumpkin-world/src/chunk_system/generation_cache.rs
+++ b/pumpkin-world/src/chunk_system/generation_cache.rs
@@ -203,11 +203,9 @@ impl GenerationCache for Cache {
         let dx = (pos.x >> 4) - self.x;
         let dz = (pos.z >> 4) - self.z;
         if !(dx < self.size && dz < self.size && dx >= 0 && dz >= 0) {
-            log::debug!(
+            debug!(
                 "illegal mark_pos_for_postprocessing {pos:?} cache pos ({}, {}) size {}",
-                self.x,
-                self.z,
-                self.size
+                self.x, self.z, self.size
             );
             return;
         }
@@ -215,7 +213,7 @@ impl GenerationCache for Cache {
         let block_pos = BlockPos::new(pos.x, pos.y, pos.z);
         match &mut self.chunks[(dx * self.size + dz) as usize] {
             Chunk::Level(_) => {
-                log::debug!("skipping mark_pos_for_postprocessing on Level chunk at {pos:?}");
+                debug!("skipping mark_pos_for_postprocessing on Level chunk at {pos:?}");
             }
             Chunk::Proto(data) => {
                 data.mark_pos_for_postprocessing(block_pos);
--- a/pumpkin-world/src/generation/biome.rs
+++ b/pumpkin-world/src/generation/biome.rs
@@ -143,9 +143,11 @@ impl<'a> NoiseBiomeSampler<'a> {
         biome_z: i32,
     ) -> &'static Biome {
         if dimension == Dimension::THE_END {
-            TheEndBiomeSupplier::biome(biome_x, biome_y, biome_z, &mut self.sampler, dimension)
+            TheEndBiomeSupplier.biome(biome_x, biome_y, biome_z, &mut self.sampler)
+        } else if dimension == Dimension::THE_NETHER {
+            MultiNoiseBiomeSupplier::NETHER.biome(biome_x, biome_y, biome_z, &mut self.sampler)
         } else {
-            MultiNoiseBiomeSupplier::biome(biome_x, biome_y, biome_z, &mut self.sampler, dimension)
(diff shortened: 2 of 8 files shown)


assets/carver.json
```diff
@@ -16,6 +16,7 @@
         "above_bottom": 8
       },
       "debug_settings": {
+        "debug_mode": false,
```
Snowiiii: it looks like you added this file manually, but it is auto generated and should not be edited

Snowiiii:
Also ideally we should add unit tests and compare it to vanilla data like we do for noise

assets/carver.json
aShanki: Oops, lemme remove it

aShanki: Oops, lemme remove it

aShanki:
@Snowiiii added unit tests and fixed a bug

Snowiiii:
Performance seems to be a lot slower and this also crashes after some seconds
```rust
thread 'Schedule' (27242) panicked at pumpkin-world/src/chunk_system.rs:2160:13:
nodes count error
```

aShanki:
I tracked this down to scheduler graph corruption during level downgrades: when targets dropped, pumpkin was removing some task nodes in a way that didn’t release dependency state, which left orphan nodes in the DAG. At the same time, some runnable nodes were never re-queued, and stale tasks could still be dispatched after their prerequisites were no longer valid, which is why you hit the `nodes count error` at shutdown (and a few panic paths). 

I fixed it by switching downgrade cleanup to dependency-aware removal, adding queue recovery so runnable nodes are re-enqueued when the queue goes empty, cancelling orphan nodes that are no longer needed by current target stage, validating dependencies/cache inputs before dispatch and dropping stale tasks safely, and making priority calculation robust when dependency-only positions are missing from `last_level` (fallback to `MAX_LEVEL`). 

After this, it should work cleanly, please lmk if you find any other bugs and lmk the repro steps

Snowiiii:
@aShanki many things have changed since the light system implementation, please merge master and resolve conflicts

aShanki:
Gocha, will resolve today

BjornTheProgrammer:
@aShanki, feel free to let me know when you manage to rebase. I'll give it look.

aShanki:
> Was the TrapezoidFloatProvider change intentional, or is this a simplification that needs to match vanilla?
> 
> The upgrading and old_noise_generation fields on ProtoChunk are always initialized to false and never set to true. Are these for future use?
> 
> Is the blending_data / OldGenerationBounds support complete, or is it for future use?

Thanks for the review!
1. Yes, this change was intentional. The old code did not match vanilla behavior so I changed it to use the same sampling approach as 1.21.11 vanilla
2. upgrading / old_noise_generation on ProtoChunk
- `upgrading`: right now this is basically a placeholder in this PR. It isn’t fully wired yet.
- `old_noise_generation`: this is read from `blending_data` when loading chunks, but we’re not fully using it yet.
3. As for blending_data / OldGenerationBounds, it is only partially implemented right now. In this PR it’s used for carver mask filtering near old-generation bounds. Full vanilla blending/retrogen support is still future work.

Purdze:
> > Was the TrapezoidFloatProvider change intentional, or is this a simplification that needs to match vanilla?
> > The upgrading and old_noise_generation fields on ProtoChunk are always initialized to false and never set to true. Are these for future use?
> > Is the blending_data / OldGenerationBounds support complete, or is it for future use?
> 
> Thanks for the review!
> 
> 1. Yes, this change was intentional. The old code did not match vanilla behavior so I changed it to use the same sampling approach as 1.21.11 vanilla
> 2. upgrading / old_noise_generation on ProtoChunk
> 
> * `upgrading`: right now this is basically a placeholder in this PR. It isn’t fully wired yet.
> * `old_noise_generation`: this is read from `blending_data` when loading chunks, but we’re not fully using it yet.
> 
> 3. As for blending_data / OldGenerationBounds, it is only partially implemented right now. In this PR it’s used for carver mask filtering near old-generation bounds. Full vanilla blending/retrogen support is still future work.

im going to be picky here, minor cleanup left.  can_reach is duplicated in cave.rs and ravine.rs, and the three carve_ellipsoid variants could be unified

Purdze:
What testing did you do?

Q2297045667:
resolve conflicts

Snowiiii:
Feel free to reopen

PR #1442, opened 2026-02-05
feat: Add `Lifecycle`, `DataResult`, `MapLike` and `DynamicOps` from DataFixerUpper

SomeYellowGuy:
## Description
[DataFixerUpper](https://github.com/Mojang/DataFixerUpper) is a library used in Minecraft for *serialization* and *datafixers*. A port of this will be necessary as upgrades can be done to game data (like worlds) easily using `DataFixer`s between versions.
Additionally, the `Codec` system in *DataFixerUpper* saves having to implement specific NBT serialization methods, instead specifying a specific `Codec`, which describes the way to serialize/deserialize an object into any required format. For example, many datapack objects (like advancements) require item stacks to be serialized in JSON, like in advancement icons.

- This PR majorly adds `DynamicOps`, which provides methods to convert a value into another value represented by the format of that ops. In DFU, this is implemented by classes like `NbtOps` (for NBT) and `JsonOps` (for JSON). A `Number` `enum` is also present to store a general number for `DynamicOps` conversion. For example, an `NbtOps` could convert an `i32` into an `NbtTag::Int`.
- A `DataResult` is a way to also describe a partial result (as `Error`), along with a complete result for a `Success` and no result (also as `Error`.
- A `Lifecycle` represents the current stage of something. A `Lifecycle` can be experimental, deprecated or stable.
- A `MapLike` is a map-view which has some map-like functions. For example, it can implement methods for an `NbtTag::Compound`. 

## Testing
Testing can only be done after more things in the library are implemented.

### DIFF
--- a/pumpkin-datafixerupper/src/serialization/dynamic_ops.rs
+++ b/pumpkin-datafixerupper/src/serialization/dynamic_ops.rs
@@ -151,13 +151,13 @@ pub trait DynamicOps {
         })
     }
 
-    /// Creates an `int` list that can be represented by this `DynamicOps` using a byte buffer.
+    /// Creates an `int` list that can be represented by this `DynamicOps` using an iterator.
     fn create_int_list(&self, int_iter: impl Iterator<Item = i32>) -> Self::Value {
         self.create_list(int_iter.map(|i| self.create_int(i)))
     }
 
-    /// Gets a `long` (`i32` in Rust) [`Iterator`] from a generic value represented by this `DynamicOps`.
-    /// This is the equivalent of DFU's `getIntStream()` function in `DynamicOps`.
+    /// Gets a `long` (`i64` in Rust) [`Iterator`] from a generic value represented by this `DynamicOps`.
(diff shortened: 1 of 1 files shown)


SomeYellowGuy:
Just to be clear, I made some changes to some things mentioned in this PR for a possible next PR involving some basic primitive codecs, which i haven't updated here (but in my next branch).

Snowiiii:
Hey, Tbh most of the seems like AI generated code providing no actual value nor really implementation, I also think to provide proper compatibility we already could use our own NBT crate and parse all needed fields in a safe and fallback friendly way

SomeYellowGuy:
The code is not connected to any of the other crates as it is meant to be used as a library. The main point of the serialization is to provide a format-agnostic way of working with data. Converting data to JSON and back will be heavily important for datapack definitions.
I was currently working on another branch which builds off of this one, including an implementation of `JsonOps` (which serves this purpose). I realize that I should really try to make an `NbtOps` struct as well, which uses the `pumpkin-nbt` crate. I will close this PR and create a new one with the new aim.

PR #1444, opened 2026-02-05
perf: chunk generation optimizations

Purdze:
## summary
- cache `SurfaceHeightEstimateSampler` height estimates between noise and surface stages, avoiding redundant recomputation of hundreds of height estimates per chunk
- lock sections mutex once in `upgrade_to_level_chunk` instead of ~500 times per chunk


Snowiiii:
@Purdze i changed the locks in the latest commit, also i think when having tracing implemented profiling would be way easier

PR #1447, opened 2026-02-06
fix: fall damage death loop (with incorrect respawn)

### LINKED ISSUE #1414: Player is being constantly killed when respawned

### Operating System

Mac OS Tahoe 26.2

0.1.0 (be322123cfa1290d8350a5707c3283a531e2c9a7)

### What happened?

When you fall from a high place close to the spawn you will be teleported into a area where you will die in a loop when respawning but sometimes it will teleport you back to the correct respawn point.

### To Reproduce

- generate a world with that seed (-8855202767317237656)
- fall from the block you spawned on
- you will die when respawn

### Expected behavior

Being teleported back to the spawn point

RoosterBooster007:
## Description
After dying of fall damage, you now respawn at the last set (correct) spawn point. You no longer continue to die in a loop.

The ``CPlayerSpawnPosition`` packet was being sent/delivered BEFORE the ``CRespawn`` packet, which was the main issue. I also added invulnerability ticks and set fall damage data to properly clear (to stop the death loop).

## Testing
Yes.

Fixes #1414

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -2718,7 +2718,7 @@ impl NBTStorage for Player {
             ) {
                 let dim = nbt
                     .get_string("SpawnDimension")
-                    .and_then(|s| Dimension::from_name(&s).copied())
+                    .and_then(|s| Dimension::from_name(s).copied())
                     .unwrap_or(self.world().dimension);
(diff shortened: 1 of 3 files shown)


Snowiiii:
Works and looks good. Thank you @RoosterBooster007 :D

PR #1448, opened 2026-02-06
fix: position args for command block summon

### LINKED ISSUE #1312: Summon command in command block always summons at command block

### Operating System

Windows 11

0.1.0-dev+1.21.11

### What happened?

When summoning mobs at a specific location via a command block, the mobs instead spawn at the location of the command block itself. The command by itself works, but not when put in a command block.

### To Reproduce

1. /summon a mob at a specific coordinate eg `/summon zombie 1 1 1`
2. Put that command in a command block
3. Power said command block

### Expected behavior

The mobs should be summoned at the coordinates specified in the command.

RoosterBooster007:
## Description
The position args now work properly for the summon command in command blocks.

## Testing
Yes.

Fixes #1312

### DIFF
--- a/pumpkin/src/command/commands/summon.rs
+++ b/pumpkin/src/command/commands/summon.rs
@@ -57,7 +57,8 @@ impl CommandExecutor for Executor {
                     (player.world(), pos)
                 }
                 CommandSender::CommandBlock(c, w) => {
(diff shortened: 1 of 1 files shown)


Snowiiii:

PR #1453, opened 2026-02-06
Fix block-break flickering

BraveDevelopment:
## Description

Move CAcknowledgeBlockChange sending out of immediate per-player send into the world flush path so acknowledgements are enqueued and sent during world flush. world/mod.rs now iterates players to swap and enqueue CAcknowledgeBlockChange for any pending sequence. Also call update_sequence(...) in Java play handlers to update the player's packet sequence when handling player actions.

## Testing

Tested. Flickering gone 🙂

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -587,7 +587,7 @@ impl World {
     pub async fn tick(self: &Arc<Self>, server: &Server) {
         let start = tokio::time::Instant::now();
 
-        // IMPORANT: send flush_block_updates first to prevent issues with CAcknowledgeBlockChange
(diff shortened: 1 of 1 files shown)


Snowiiii:
Hey, I'm pretty sure moving the `CAcknowledgeBlockChange` send logic into the world tick function is not necessary since directly after the player tick function is called i think the issue was fixed by just adding some missing `update_sequence` (already expected that)

Snowiiii:
Found out that calling `flush_block_updates` seems important before sending any `CAcknowledgeBlockChange` packets, otherwise the client will not like that

Snowiiii:
Thanks @BraveDevelopment :D

PR #1456, opened 2026-02-07
Basic pathfinder impl from vanilla

vyPal:
## Description
This PR serves as a basic implementation of the Pathfinder module from vanilla. Many things still need to be added/fixed, but basic pathfinding for mobs (currently only zombies) is implemented and functioning.

---

#### Fully implemented:
- [x] `BinaryHeap`
- [x] `Node` & `Target`
- [x] `NodeEvaluator` (represented as `struct BaseNodeEvaluator` and `trait NodeEvaluator`)
- [x] `PathComputationType`
- [x] `Path` & `PathType` & `PathTypeCache`
#### Partially implemented:
- [x] `PathFinder`
- [ ] `PathfindingContext`
- [ ] `WalkNodeEvaluator`
- [ ] `MobData` - even though it's not a part of the pathfinder, it is necessary for it's operation
#### Not implemented
- [ ] `AmphibiousNodeEvaluator`
- [ ] `FlyNodeEvaluator`
- [ ] `SwimNodeEvaluator`

---

### Known issues with implementation
- Zombie moves too fast - caused by wrong speed value for `NavigatorGoal`, in the future this should probably be inclued in `MobData`
- Pathfinding fails when two or more jumps are required to reach goal - idk why
- Zombie following range incorrect - pathfinding loop is set to a fixed number of iterations, but this should be calcualted based on `MobData`
- Zombie attacks through walls - visibility check (canSee) not implemented, needs world raycast

### DIFF
--- a/pumpkin/src/entity/ai/goal/active_target.rs
+++ b/pumpkin/src/entity/ai/goal/active_target.rs
@@ -12,7 +12,6 @@ use std::sync::Arc;
 
 const DEFAULT_RECIPROCAL_CHANCE: i32 = 10;
 
-#[expect(dead_code)]
 pub struct ActiveTargetGoal {
     track_target_goal: TrackTargetGoal,
     target: Option<Arc<dyn EntityBase>>,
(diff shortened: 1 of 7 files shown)


Snowiiii:
Thank you @vyPal, @Purdze (@Snowiiii :D)

PR #1460, opened 2026-02-07
feat: replace log with tracing

FabseGP:
## Description

Replace all instances of log with tracing

## Testing

WIP

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -2595,6 +2595,7 @@ impl World {
                 for player in current_players.iter() {
                     player.send_system_message(&event.join_message).await;
                 }
+                // TODO: Switch to structured logging, e.g. info!(player = %name, "connected")
                 info!("{}", event.join_message.to_pretty_console());
             }
         });


FabseGP:
Since the global tracing-subscriber doesn't spawn to any threads spawned ([post for context](https://www.reddit.com/r/rust/comments/y6s4lo/tokio_tracing_subscriber_and_dynamically_loaded/)), each plugin spawns their own tracing_subscriber and calls a log-function from the plugin context instead of log! directly. The output will look like this:

```rust
2026-02-08 17:51:49  INFO ThreadId(13) pumpkin_plugin: Hello! plugin="hello-pumpkin"
```

with "hello-plugin" being the name of the plugin

pumpkin/src/command/mod.rs
```diff
@@ -70,7 +70,8 @@ impl fmt::Display for CommandSender {
 impl CommandSender {
     pub async fn send_message(&self, text: TextComponent) {
         match self {
-            Self::Console => log::info!("{}", text.to_pretty_console()),
+            #[allow(clippy::print_stdout)]
+            Self::Console => println!("{}", text.to_pretty_console()),
```
BjornTheProgrammer, line 74: What is the reason as to using a println instead of a `tracing::info`?

pumpkin/src/plugin/api/context.rs
```diff
@@ -300,15 +303,42 @@ impl Context {
...
+                let fmt_layer = fmt_layer.without_time();
+                tracing_subscriber::registry()
+                    .with(*level)
+                    .with(fmt_layer)
+                    .init();
+            }
         }
     }
+
+    pub fn log(&self, message: impl std::fmt::Display) {
```
BjornTheProgrammer, line 336: Do we want to not use a macro instead? It's a little annoying to have to do server.log(format!(...)). Additionally maybe we shouldn't even have this since logging directly in the plugin using `tracing` or `log` seems to work just fine.

BjornTheProgrammer:
Forgot to mention it here, but I have run it on a simple plugin, and it seems to work perfectly in my testing. Great work!

pumpkin/src/command/mod.rs
FabseGP: Ahh that was before #1516 got merged (downgrading tracing-subscriber to a version which didn't escape ANSI-characters), but I'm not sure if I want to revert this

The argument for escaping ANSI-characters by the project was for security and for commands run by users in the console, they prob don't need the extra info from tracing

pumpkin/src/plugin/api/context.rs
FabseGP: True that will be tedious, I'll try and create a macro a bit later!

Regarding the last point, I think we should consolidate as many methods a plugin might use within Pumpkin itself, standardizing logs and whatever other functionality in the future (say log rate limiting or an api for getting the logs)

FabseGP:
I wonder too if we should consider adding a step in the CI to compile the Hello-Pumpkin plugin as a way to detect breakages with plugins (like with how it's broken now because CommandExecute now expects an integer on return)

Not in this pr, but a new one if useful?

BjornTheProgrammer:
I've actually tried to do the same thing in the past, but it wasn't merged. Alex wants to keep all plugins in the separate repo. Even if it is for testing, if I remember correctly.

FabseGP:
> I've actually tried to do the same thing in the past, but it wasn't merged. Alex wants to keep all plugins in the separate repo. Even if it is for testing, if I remember correctly.

Oh that can work still like in #1530, we can just point to the repository using the checkout action

EDIT: it'll fail now because the plugin is currently broken

pumpkin/src/plugin/api/context.rs
BjornTheProgrammer: Okay, I understand the rational, but I'll wait for Alex's input on this. It might be a week before we can get this merged!

BjornTheProgrammer: We've decided that this is actually fine.

FabseGP:
> if we're using tracking we should use something like (info!(player = %name, "conneted"))

That's a good idea, there are prob other places as well where structured logging would make sense! I'll look through later, for now the todo is added

PR #1461, opened 2026-02-08
fix: water flow not spreading correctly

### LINKED ISSUE #1459: Water sometimes does not flow correctly

### Operating System

macOS / Windows

0.1.0-dev+1.21.11 (Commit: 6ed8af2/release)

### What happened?

Water has problems with flowing sometimes.
It seems to have something to do with chunk borders but im not a 100% sure.
Image

### To Reproduce

I cant really say what the problem is or how to replicate.
Just place a water source on an even surface and check for missing water.

### Expected behavior

Water flowing correctly

Purdze:
## summary
- fix `get_max_flow_distance` returning 5 instead of 4, causing incorrect flow direction selection
- fix `flow_to_sides` computing fluid level from source position (`level - 1`) instead of using `getUpdatedState` for each target position, which considers all neighbors
- fix falling fluid side-spread using `(8 - drop_off).min(8)` instead of unconditionally using level 7
- update `get_spread` to return computed fluid states alongside directions, matching vanilla's `getSpread()` return structure

fixes #1459

### DIFF
--- a/pumpkin/src/block/fluid/flowing_trait.rs
+++ b/pumpkin/src/block/fluid/flowing_trait.rs
@@ -489,9 +489,8 @@ pub trait FlowingFluid: Send + Sync {
 
     /// Spreads fluid horizontally to adjacent positions using pathfinding.
     ///
-    /// Uses `get_spread` to find optimal flow directions (shortest distance to holes).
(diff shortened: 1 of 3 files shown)


Snowiiii:
LGTM. Thank you @Purdze :D

PR #1463, opened 2026-02-08
Implement trident melee and thrown states (1.21.11)

aShanki:
## Summary
Port trident behavior from 1.21.11 vanilla into Pumpkin, covering both melee use-state handling and thrown projectile behavior.

## Changes
- Add `release_use` to `ItemBehaviour` and dispatch support in the item registry.
- Route `ReleaseItemInUse` packet handling through item `release_use` and correctly sync/clear active hand state.
- Fix active-hand offhand-flag mapping and clearing in `LivingEntity`.
- Ensure active-item completion only auto-consumes consumable items.
- Mark trident as long-use (`72000`) in world item state.
- Add trident projectile registration and `ThrownTridentEntity` implementation.
- Implement trident release logic:
  - requires charge time before throw,
  - blocks throwing when the next durability hit would break,
  - plays throw sound, spawns thrown trident with vanilla-like throw power,
  - decrements held stack in survival and keeps inventory intact in creative.
- Apply thrown trident hit behavior:
  - trident damage type and base hit damage,
  - hit-ground/entity sounds,
  - survival drop of a trident item on impact.

## Notes
- This PR focuses on base trident throw/melee flow parity. Enchantment-specific behaviors beyond this scope are not included here.


pumpkin/src/entity/living.rs
```diff
@@ -172,7 +174,7 @@ impl LivingEntity {
             .store(stack.get_max_use_time(), Ordering::Relaxed);
         *self.item_in_use.lock().await = Some(stack);
         self.set_living_flag(Self::USING_ITEM_FLAG, true).await;
-        self.set_living_flag(Self::OFF_HAND_ACTIVE_FLAG, hand == Hand::Left)
+        self.set_living_flag(Self::OFF_HAND_ACTIVE_FLAG, hand == Hand::Right)
```
Snowiiii: I don't think this makes sense

Snowiiii:
I'm sorry but this is just spagetti code, you hard coded things like `held.item.id == Item::TRIDENT.id` instead of using abstractions, instead of changing something here a ground up rewrite/PR would be better in this case

PR #1464, opened 2026-02-08
fix: align player pose resolution with vanilla to prevent crawl in-wall damage

### LINKED ISSUE #1315: Crawling under a block damages the player

aShanki:
## Summary
- fix player pose selection to better mirror vanilla pose resolution rules
- update pose on Java movement/sneak/elytra transitions so hitbox metadata matches state changes
- run pose reconciliation each player tick to match vanilla tick behavior
- keep Bedrock unchanged in this PR as requested

## Why
Issue #1315 reports players taking `in_wall` damage while crawling/swimming under blocks. The root cause is stale/incorrect pose selection, which keeps a too-tall hitbox in constrained spaces.

## Parity notes
Compared against vanilla 1.21.11 `Player.updatePlayerPose()` behavior:
- spectator/passenger can keep desired pose
- crouching is suppressed while ability-flying
- pose is updated every tick

Closes #1315

### DIFF
--- a/pumpkin/src/server/connection_cache.rs
+++ b/pumpkin/src/server/connection_cache.rs
@@ -4,7 +4,7 @@ use core::error;
 use pumpkin_config::BasicConfiguration;
 use pumpkin_data::packet::CURRENT_MC_PROTOCOL;
 use pumpkin_protocol::{
-    Players, StatusResponse, Version,
+    Players, Sample, StatusResponse, Version,
     codec::var_int::VarInt,
     java::client::{config::CPluginMessage, status::CStatusResponse},
 };
@@ -13,8 +13,10 @@ use std::{
     fs::{self},
     path::Path,
 };
+use uuid::Uuid;
 
 const DEFAULT_ICON: &[u8] = include_bytes!("../../../assets/default_icon.png");
+const MAX_SAMPLE_PLAYERS: usize = 12;
 
 fn load_icon_from_file<P: AsRef<Path>>(path: P) -> Result<String, Box<dyn error::Error>> {
     let buf = fs::read(path)?;
@@ -41,6 +43,7 @@ pub struct CachedStatus {
     // We cache the json response here so we don't parse it every time someone makes a status request.
     // Keep in mind that we must parse this again when the StatusResponse changes, which usually happen when a player joins or leaves.
     status_response_json: String,
+    player_samples: Vec<(Uuid, String)>,
 }
 
 pub struct CachedBranding {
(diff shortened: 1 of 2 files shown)


pumpkin/src/net/java/play.rs
```diff
@@ -300,6 +300,7 @@ impl JavaClient {
                 let entity = &player.living_entity.entity;
                 let last_pos = entity.pos.load();
                 player.living_entity.entity.set_pos(pos);
+                player.update_player_pose().await;
```
BjornTheProgrammer: We should probably handle this in Bedrock as well. Are you able to confirm this is an issue in Bedrock, and if so would you be able to update the bedrock code as well? If not we can implement it in the future.

aShanki: I do have bedrock but I'd rather make that a seperate PR if possible

BjornTheProgrammer: Alright, I'll test this PR and see if Purdze is able to merge

BjornTheProgrammer (approved):

Purdze:
- Missing isSwimming() check — vanilla returns EntityPose.SWIMMING when the player is sprint-swimming in water; this never proactively enters the swimming pose
- Missing isUsingRiptide() / SPIN_ATTACK pose in the priority chain
- Sleeping check reads current_pose == Sleeping instead of a state flag — vanilla uses isSleeping() (checks sleepingPosition.isPresent()), not the current pose
- update_player_pose() is called from 5 packet handlers — vanilla only calls updatePose() once per tick from
PlayerEntity.tick(); the extra calls cause redundant async block lookups, potential race conditions, and double
metadata broadcasts
- Tick ordering is inverted — vanilla runs super.tick() (updates swimming/fluid state) before updatePose(), but this
PR calls update_player_pose() before living_entity.tick()
- The special-case crawl branch (current_pose == Swimming && !can_fit(Standing)) doesn't exist in vanilla — vanilla
handles crawling through its fallback chain (desired → crouching → swimming) which already covers this
- set_sneaking() unconditionally sets the pose to Crouching/Standing — vanilla's setSneaking() only sets the flag and
lets updatePose() handle pose resolution

If things cant be done yet, put a TODO

PR #1466, opened 2026-02-08
fix: avoid drop item lock inversion with screen handler

yunuservices:
## Problem
  A lock inversion could occur between:
  - `drop_held_item` path (held item stack lock -> screen handler lock)
  - player tick inventory sync path (screen handler lock -> slot/item stack lock)

  Under contention, this could block packet processing and present as:
  ## Fix
  - Refactored `drop_held_item` to avoid holding item stack and screen handler locks at the same time.
  - Captures drop result and updated stack in a scoped item lock.
  - Drops item after leaving that lock scope.

  - Behavior is preserved functionally (single item / full stack drop logic unchanged).
  - Change is targeted to the deadlock trigger path only.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -2129,27 +2129,34 @@ impl Player {
     }
 
     pub async fn drop_held_item(&self, drop_stack: bool) {
-        // should be locked first otherwise cause deadlock in tick() (this thread lock stack, that thread lock screen_handler)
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM. Thank you @yunuservices :D

yunuservices:
> LGTM. Thank you @yunuservices :D 

No worries 🥳

PR #1486, opened 2026-02-08
fix: collisions and other minor things

### LINKED ISSUE #1505: touching cactus doesnt deal damage

### Operating System

windows 11

0.1.0-dev+1.21.11

### What happened?

when touching a cactus it doesnt deal damage

### To Reproduce

touch a cactus

### Expected behavior

Deal damage

JustMarfix:
## Description

- fixed collision not working for some blocks (ex. cactus when standing nearby, berry bushes)
- added last movement delta field for entity
- non-living entities now can take damage, so they can be destroyed when colliding with cactus
- fixed fun bug when player could suffocate while being near a bamboo

upd: this also fixes campfire collision damage

fixes #1505

### DIFF
--- a/pumpkin/src/block/blocks/plant/crop/sweet_berry_bush.rs
+++ b/pumpkin/src/block/blocks/plant/crop/sweet_berry_bush.rs
@@ -9,7 +9,6 @@ use crate::{
     },
     world::World,
 };
-use pumpkin_data::world::WorldEvent;
 use pumpkin_data::{
     Block,
     block_properties::{BlockProperties, EnumVariants, Integer0To3, NetherWartLikeProperties},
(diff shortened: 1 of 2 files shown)


pumpkin/src/entity/mod.rs
```diff
@@ -318,6 +318,8 @@ pub struct Entity {
     pub pos: AtomicCell<Vector3<f64>>,
     /// The last known position of the entity.
     pub last_pos: AtomicCell<Vector3<f64>>,
+    /// The last movement vector
+    pub movement: AtomicCell<Vector3<f64>>,
```
Snowiiii, line 322: I don't think this is really needed, also movement is just the wrong name for this field

JustMarfix: This is the name from yarn, and used to deal damage in berry bushes (and I guess for a lot more things) only when player is moving. so idk, should I do it another way?

JustMarfix:
I removed that bone meal duct tape thing from this PR, will do it properly for all plants in another

Snowiiii:
please split all features/fixed bugs here into seperate PRs

JustMarfix:
Moved to #1522

PR #1488, opened 2026-02-09
fix: void damage

### LINKED ISSUE #1314: Falling into the void results in softlock

### Operating System

Arch Linux

0.1.0/1e691c71

### What happened?

When you fall into the void, whether it be in creative or survival, you fall until -130 ish, and then get locked in place by the server and take no damage. You also cannot fly with creative/spectator mode. The only ways to get out are /tp and /kill as far as I'm aware.

### To Reproduce

1. Get into the void
2. Start falling

### Expected behavior

Either match java edition and fall infinitely and start taking damage, or follow bedrock edition and fall onto a void floor at -104 and take damage there.

RoosterBooster007:
## Description
Before, players would get locked in the void due to an ``entity.remove()`` call on the player.
Now, players take damage in the void at/under y=-128 and die normally (in all game modes).

## Testing
Yes.

Fixes #1314

### DIFF
--- a/pumpkin/src/entity/mod.rs
+++ b/pumpkin/src/entity/mod.rs
@@ -1939,7 +1939,8 @@ impl Entity {
     }
 
     pub fn is_invulnerable_to(&self, damage_type: &DamageType) -> bool {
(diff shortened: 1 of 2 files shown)


Snowiiii:
Looks good. Thanks @RoosterBooster007 :D

PR #1490, opened 2026-02-09
fix: mob pathfinding bug

spectradevv:
## Description
Fixed mobs moving towards players who were in creative or spectator mode.

## Testing
behaves as expected.

### DIFF
--- a/pumpkin/src/entity/ai/goal/active_target.rs
+++ b/pumpkin/src/entity/ai/goal/active_target.rs
@@ -81,7 +81,10 @@ impl ActiveTargetGoal {
                     .target_predicate
                     .test(&world, Some(&mob.living_entity), living)
             {
-                if !potential_entity.get_player().is_some_and(|p| p.is_creative() || p.is_spectator()) {
+                if !potential_entity
+                    .get_player()
+                    .is_some_and(|p| p.is_creative() || p.is_spectator())
+                {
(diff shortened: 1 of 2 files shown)


Snowiiii:
I don't think this PR follows the Vanilla logic

spectradevv:
@Snowiiii could you elaborate?

illyrius666:
> @Snowiiii could you elaborate?

probably the way you `hide` the player from the hostile mobs eyes. ig this is what he means.

Snowiiii:
> @Snowiiii could you elaborate?

Can you prove that vanilla uses similar code in that places

Purdze:
> This actually looks largely correct. I might use a slightly different structure for represeting whether the enemy can attack the target or not, so that more than just the player can be added. It is true that if the difficulty is peaceful, the player is invulnerable, is a spectator, or is not alive, then the mob should not be able to target the player.
> 
> So instead of checking if creative, instead check if they have the invulnerable ability. Also add the other checks, and we'll get this merged.
> 
> Thanks for your great work!

to add, the fix should live in TargetPredicate::test() instead of being duplicated in each goal, that way any future targeting goal gets the check for free. Also, the return; in active_target.rs is outside the new if, so if the nearest candidate is a creative/spectator player the mob stops searching instead of checking other entities.

Issue #1492, opened 2026-02-09
Placing a block into an entity's hitbox


user123456000:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Linux

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.11 (Commit: b0ab500/release)

### What happened?

In vanilla Minecraft, a player can't place a block if an entity's hitbox prevents it (not always).

This doesn't apply to items or perhaps other things (I'm not sure).

### To Reproduce

1. /summon minecraft:zombie
2. F3 + B
3. Place a block in the entity's hitbox

### Expected behavior

You shouldn't be able to do this.

GreenedDev:
this appears to be fixed @Snowiiii

QuantumSegfault:
This should be fixed as of #1494. Could you please verify?

PR #1494, opened 2026-02-09
fix: check every hitbox onplace block, not just player

chocodev11:
#1492 

## Testing

### DIFF
--- a/pumpkin/src/net/java/play.rs
+++ b/pumpkin/src/net/java/play.rs
@@ -1993,7 +1993,7 @@ impl JavaClient {
     fn has_blocking_entity_in_box(world: &World, placed_box: &BoundingBox) -> bool {
         let players = world.players.load();
         if players.iter().any(|player| {
-            !player.is_spectator()
+            Self::entity_blocks_block_placement(player.as_ref())
                 && player
                     .get_entity()
                     .bounding_box


pumpkin/src/net/java/play.rs
```diff
@@ -2111,14 +2112,21 @@ impl JavaClient {
...
-        for player in world.get_nearby_players(location.0.to_f64(), 3.0) {
-            let player_box = player.living_entity.entity.bounding_box.load();
-            for shape in state.get_block_collision_shapes() {
-                if shape.at_pos(final_block_pos).intersects(&player_box) {
-                    return Ok(false);
-                }
+        for shape in state.get_block_collision_shapes() {
+            let placed_box = shape.at_pos(final_block_pos).to_bounding_box();
+
+            if !world.get_players_at_box(&placed_box).is_empty() {
```
BjornTheProgrammer: I might be wrong, but you can probably get rid of this check here since we now check for entities at box. Can you test that?

Purdze:
Looks good however one inconsistency, the entity path checks is_removed, no_clip, and is_spectator via entity_blocks_block_placement, but the player path only checks is_spectator.. a no-clip or removed player would still block placement, should run the same checks

PR #1497, opened 2026-02-09
 fix: reduce inventory lock contention and avoid stale empty slot snapshots

yunuservices:
This PR fixes inventory lock and stale empth slot snapshots.
  - Reduced lock scope around equipment maps so async work no longer runs while holding the main equipment lock.
  - `get_cloned_stack` no longer returns `ItemStack::EMPTY` on lock timeout.
  - On 5s timeout, it logs a warning and waits for the lock, preventing stale empty snapshots.

### DIFF
--- a/pumpkin-inventory/src/slot.rs
+++ b/pumpkin-inventory/src/slot.rs
@@ -72,22 +72,15 @@ pub trait Slot: Send + Sync {
         // Default implementation logic:
         Box::pin(async move {
             let stack = self.get_stack().await;
-            match timeout(Duration::from_secs(5), stack.lock()).await {
-                Ok(lock) => lock.clone(),
-                Err(_) => {
-                    log::warn!(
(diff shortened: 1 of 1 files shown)


yunuservices:
> the retry-on-timeout in get_cloned_stack doesnt really help, if the contention fixes work, it never triggers, and if it does, delaying the panic by 5s doesn't really help. mainhand isn't pre-populated in new() but the other slots are.., intentional? Also, the lock-per-slot pattern in the world code means concurrent modifications could give an inconsistent equipment snapshot (old helmet + new boots) when sending to clients on join

well, welp. I have to figure out this then rewrite it. Thanks for feedback!

yunuservices:
Batch queue reduction: keeping only #1666, #1654, #1653, #1647 for faster merge throughput. Will re-open/continue next batch after these land.

PR #1498, opened 2026-02-09
Fix ptero egg

vyPal:
## Description
This PR adds a bit more helpful info to the pterodactyl egg for pumpkin.

### List of changes:
- Updated RAM message to recommend increasing to 4GB as the build process sometimes failed with only 2GB
- Added recommendation to increase max cpu limit for faster compilation
- Added info to the Cargo build jobs env variable about minimum value and how to use as many jobs as possible
- Changed type of build jobs env var to string so that the value `default` can be input, and changed this to be default instead of an empty variable

## Recommendation
(as mentioned [on discord](https://discord.com/channels/1268592337445978193/1467166843746652180/1470455995842035928)) I'd highly recommend adding a second Dockerfile to the repo, specifically for Ptero/Pelican and build the image in action alongside the normal image. The reason for this is that even with an increased cpu limit during the build stage (200 instead of 100), the build still took 24 minutes on a average server. Since ptero also doesn't have any caching, this time is needed whenever a server is being deployed. In my opinion it's not really sustainable to build it on the machine, and since it's being built in docker, there are no real benefits to building on the target machine (the `target-cpu=native` setting is basically useless/would be the same building in actions)

## Testing

- [x] Works on self deployed Pterodactyl server
- [ ] Will test on Pelican as well (waiting for response)

### DIFF
--- a/egg-pumpkin.json
+++ b/egg-pumpkin.json
@@ -4,10 +4,10 @@
         "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2024-12-29T00:41:20+01:00",
+    "exported_at": "2026-02-10T08:07:42+00:00",
     "name": "Pumpkin",
     "author": "lilalexmed@proton.me",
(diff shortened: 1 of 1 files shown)


vyPal:
GitHub is experiencing degraded performance, I'd wait until it's fixed and rerun the actions: [https://www.githubstatus.com/incidents/smf24rvl67v9](https://www.githubstatus.com/incidents/smf24rvl67v9)

Snowiiii:
@vyPal maybe this should be regenerated ?, looking at `exported_at` it already seems pretty old not sure if there is maybe even a newer version already ?

vyPal:
Yeah I'm working on remaking it with a couple new features, will mark as draft for now

Snowiiii:
LGTM. Thank you @vyPal :D

PR #1500, opened 2026-02-09
fix: totem behavior

RoosterBooster007:
## Description
- Totems ignore /kill and void damage.
- Totems are properly removed from a player's inventory (instead of allowing infinite use).
- Totems now work for fall damage

## Testing
Yes.


Snowiiii:

PR #1508, opened 2026-02-10
feat(worldgen): implement blender biome supplier hook

yunuservices:
### What changed
  - Added a BlenderBiomeSupplier function type.
  - Updated BlenderImpl::get_biome_supplier to return a concrete supplier function.
      - Uses MultiNoiseBiomeSupplier for Overworld/Nether.
  - Updated ProtoChunk::populate_biomes to resolve biomes via Blender::NO_BLEND.get_biome_supplier().

  ### Why

  - Removes the remaining todo!() in Blender biome supplier logic.
  - Makes biome selection flow cleaner and extensible for future blending implementations.

  ### Notes
  - Current no-blend runtime behavior is preserved.

### DIFF
--- a/pumpkin-world/src/biome/end.rs
+++ b/pumpkin-world/src/biome/end.rs
@@ -1,4 +1,4 @@
-use pumpkin_data::{chunk::Biome, dimension::Dimension};
+use pumpkin_data::chunk::Biome;
 
 use crate::{
     biome::BiomeSupplier,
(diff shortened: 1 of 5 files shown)


BjornTheProgrammer (approved):

Purdze:
- get_biome_supplier should take a BiomeSupplier input and wrap it (decorator pattern), not return a standalone function, this blocks real blending implementation
- The dimension parameter doesn't exist in vanilla's BiomeSupplier - the correct BiomeSource is already selected at
the generator level
- Bare fn pointer can't capture state needed for actual blending

yunuservices:
> - get_biome_supplier should take a BiomeSupplier input and wrap it (decorator pattern), not return a standalone function, this blocks real blending implementation
> - The dimension parameter doesn't exist in vanilla's BiomeSupplier - the correct BiomeSource is already selected at
> the generator level
> - Bare fn pointer can't capture state needed for actual blending

Lemme fix them, thanks for feedback 😭

PR #1510, opened 2026-02-10
fix(chunk): prevent stale chunk_sent entries from blocking visibility

yunuservices:
This PR fixes intermittent missing chunk visibility while moving/flying.

  ### What changed

  - Updated ChunkManager::update_center_and_view_distance in pumpkin/src/entity/player.rs.
  - chunk_sent cleanup now uses current center + view distance bounds instead of only the unloading_chunks list.
  - Marked the now-unused parameter as _unloading_chunks.

  ### Why

  In some movement/update paths, unloading_chunks can be incomplete relative to actual out-of-range chunks.
  That leaves stale entries in chunk_sent, which can block re-queuing and prevent chunks from being sent to the client.

  ### Result

  - Reduces “holes” / not-loading chunk strips during movement.
  - Keeps chunk send state aligned with real view radius.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -195,7 +195,7 @@ impl ChunkManager {
         mut view_distance: u8,
         level: &Arc<Level>,
         loading_chunks: &[Vector2<i32>],
-        _unloading_chunks: &[Vector2<i32>],
+        unloading_chunks: &[Vector2<i32>],
     ) {
         view_distance += 1; // Margin for loading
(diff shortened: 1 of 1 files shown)


TheRedXD:
Tested this, works perfectly. @Snowiiii can we merge this in?

Q2297045667:
582d46595e8d8bc45ef6fecf8c23cb7a1b08a22e
The bug remains the same
image
image

yunuservices:
yeah, i experienced it too. planning to do some changes rn.

PR #1511, opened 2026-02-10
Prevent subtraction overflow in `translation_to_pretty`

### LINKED ISSUE #1507: `give player minecraft:lava_bucket` causes overflow crash

### Operating System

Linux  6.18.7-arch1-1

0.1.0-dev+1.21.11 (Commit: 5a5b9e4/debug)

### What happened?

When running the command `/give <player> minecraft:lava_bucket`, the server crashes. 

```
thread 'tokio-runtime-worker' (765582) panicked at pumpkin-util/src/translation.rs:161:25:
attempt to subtract with overflow
```

The culprit appears to be an error in the `translate_to_pretty` function in `pumpkin-util/src/translation.rs`. The line in question is the line `displacement += substitution.len() - range.len();`

I'm still wrapping my head around the translation system as I've never had to deal with one before. 

### To Reproduce

`cargo run`
In the server console, run `give @a minecraft:lava_bucket`
Crash results

### Expected behavior

Either the item being given, or the appropriate error (I need to sleep so I haven't looked into this yet).

tylercritchlow:
## Description
Prevent subtraction overflow in `translation_to_pretty`. fixes #1507 

## Testing

1. boot up pumpkin server
2. from console type `give @a minecraft:lava_bucket`
3. it should not panic because of an overflow error

### DIFF
--- a/pumpkin-util/src/translation.rs
+++ b/pumpkin-util/src/translation.rs
@@ -138,30 +138,31 @@ pub fn reorder_substitutions(
     }
     (substitutions, ranges)
 }
-
 pub fn translation_to_pretty<P: Into<Cow<'static, str>>>(
     namespaced_key: P,
     locale: Locale,
     with: Vec<TextComponentBase>,
 ) -> String {
-    let mut translation = get_translation(&namespaced_key.into(), locale);
+    let translation = get_translation(&namespaced_key.into(), locale);
     if with.is_empty() || !translation.contains('%') {
         return translation;
     }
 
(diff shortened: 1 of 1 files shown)


tylercritchlow:
> * What happens if range.start as i32 + displacement goes negative? The as usize cast would wrap and panic, should this use usize::try_from instead?
> 
> * Would building the output from parts instead of mutating in place be simpler and avoid displacement tracking entirely?

Good point it would be better to build from parts. I will edit the method i changed, and the one below it that does the same displacement stuff.

PR #1517, opened 2026-02-10
perf: packet encode, decode and compression

chocodev11:
## Description
- added normal + high-priority queues and batch sending (up to 64 packets) with one flush per batch.
- world/server Java broadcasts now bucket players by protocol version and serialize once per version.
- encoder now reuses zlib compressor state + scratch buffer instead of rebuilding per packet.
- decoder now reuses a payload buffer with size hints, reducing repeated allocations.
- added lightweight network metrics/traces (batch counts, packet/byte counters, and slow encode/decode/batch timings).

not tested yet but by calculate, the improvement is about 15-30% cpu for network

## Testing

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2135,7 +2135,6 @@ dependencies = [
  "bytes",
  "cfb8",
  "flate2",
- "log",
  "pumpkin-data",
  "pumpkin-macros",
  "pumpkin-nbt",
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -392,7 +392,8 @@ impl World {
     fn collect_java_recipients_by_version<'a>(
         players: impl Iterator<Item = &'a Arc<Player>>,
     ) -> BTreeMap<MinecraftVersion, Vec<&'a JavaClient>> {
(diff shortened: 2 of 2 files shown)


illyrius666:
your pr might be closed because you didnt provide a description (even if the title is self-explanatory)

Snowiiii:
I think having the separate queues for high and non high priority is not a good solution, not sure if it is needed at all but if we really decide to use a similar approche i would rather have a enum defining the Priority similar to scheduled ticks

chocodev11:
It's necessary and really improves performance, example disconnect packets should be sent before chunk packet. And 2 priority queues is fit for networking.

Purdze:
A few things:
  - The version-grouping logic in broadcast_packet_all, broadcast_platform_specific, and broadcast_packet_except is
  copy-pasted 3 times, worth extracting into a helper.
  - The ~13 global atomic counters and trace logging add a lot of code, feels like it belongs in a separate observability PR.
  - The log::debug! → log::error! change for packet read errors looks unrelated and will be noisy on client disconnects.
  - .or_insert(Vec::new()) → .or_default().
  
  Otherwise looks good :)

Purdze:
Looks good, thank you

PR #1519, opened 2026-02-10
Player samples

denys-shatin:
## Description

Implements player sample display in server status response.

### Changes:
- [x] Shows up to 12 players when hovering over player count in multiplayer menu
- [x] Dynamically updates on player join/leave
- [x] Removes redundant clone for better performance

### Testing
- [x] Tested player samples display in multiplayer menu
- [x] Verified players appear/disappear on join/leave
- [x] Passed clippy checks

### DIFF
--- a/pumpkin/src/server/connection_cache.rs
+++ b/pumpkin/src/server/connection_cache.rs
@@ -90,54 +90,54 @@ impl CachedStatus {
         CStatusResponse::new(self.status_response_json.clone())
     }
 
+    fn build_sample_list(&self) -> Vec<Sample> {
+        self.player_samples
+            .iter()
+            .take(MAX_SAMPLE_PLAYERS)
+            .map(|(id, name)| Sample {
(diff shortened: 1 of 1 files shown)


denys-shatin:
> This looks great! Just wondering if we should have an explicit option in the server config as to whether to show sample players that are online. I'll leave it for Alex to decide later.

i can make that if alex wants

Purdze:
the sample rebuild logic is duplicated in add_player and remove_player, extract it to a helper, also the old added_to_server_listing guard was protecting players.online from double-counting, and removing it means the online count can drift if add_player is called twice for the same player

denys-shatin:
> the sample rebuild logic is duplicated in add_player and remove_player, extract it to a helper, also the old added_to_server_listing guard was protecting players.online from double-counting, and removing it means the online count can drift if add_player is called twice for the same player

ok

PR #1520, opened 2026-02-10
Packet optimization

denys-shatin:
## Description
Optimizes network traffic by broadcasting packets only to players within configured view distance instead of all players in the world.

## Changes
- [x] Added `get_broadcast_radius()` helper method to `World` to calculate broadcast radius from server config
- [x] Replaced `broadcast_packet_all` calls with `get_nearby_players` filtering in:
  - [x] Entity status updates (`send_entity_status`)
  - [x] Mob effect removal (`send_remove_mob_effect`)
  - [x] Item pickup animations (`pick_up_item`)
  - [x] Hand swing animations (`swing_hand`)
- [x] Converts view_distance from chunks to blocks (chunks * 16)
- [x] Falls back to 64 blocks radius when server reference is unavailable

## Benefits
- Reduces unnecessary network traffic on larger servers
- Players only receive packets for entities they can actually see
- Improves performance for servers with many players spread across the world

## Testing
- [x] Tested with multiple players at various distances
- [x] Verified packets are sent within view distance
- [x] Confirmed fallback behavior works correctly

### DIFF
--- a/pumpkin-data/src/data_component_impl.rs
+++ b/pumpkin-data/src/data_component_impl.rs
@@ -4,8 +4,8 @@ use crate::attributes::Attributes;
 use crate::data_component::DataComponent;
 use crate::data_component::DataComponent::{
     AttributeModifiers, BlocksAttacks, Consumable, CustomData, CustomName, Damage, DeathProtection,
-    Enchantments, Equippable, Food, ItemName, JukeboxPlayable, MaxDamage, MaxStackSize,
-    PotionContents, Tool, Unbreakable,
+    Enchantments, Equippable, FireworkExplosion, Fireworks, Food, ItemName, JukeboxPlayable,
+    MaxDamage, MaxStackSize, PotionContents, Tool, Unbreakable,
 };
 use crate::entity_type::EntityType;
 use crate::tag::{Tag, Taggable};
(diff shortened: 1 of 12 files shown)


pumpkin/src/entity/living.rs
```diff
@@ -156,16 +156,23 @@ impl LivingEntity {
...
-            ))
-            .await;
+        let world = self.entity.world.load();
+        let packet = CTakeItemEntity::new(
+            item.entity_id.into(),
+            self.entity.entity_id.into(),
+            stack_amount.try_into().unwrap(),
+        );
+
+        // Use view_distance from config (in chunks) * 16 to get blocks
```
BjornTheProgrammer: Why `x16`?

pumpkin/src/world/mod.rs
```diff
@@ -308,22 +308,38 @@ impl World {
...
-        self.broadcast_packet_all(&CRemoveMobEffect::new(
-            entity.entity_id.into(),
-            VarInt(i32::from(effect_type.id)),
-        ))
-        .await;
+        let packet =
+            CRemoveMobEffect::new(entity.entity_id.into(), VarInt(i32::from(effect_type.id)));
+
+        // Use view_distance from config (in chunks) * 16 to get blocks
+        let radius = self.server.upgrade().map_or(64.0, |server| {
```
BjornTheProgrammer: This is repeated 4 times, maybe make it a function?

pumpkin/src/world/mod.rs
```diff
@@ -2107,8 +2123,19 @@ impl World {
...
+
+        let sleep_percentage = self
+            .level_info
+            .load()
+            .game_rules
+            .players_sleeping_percentage;
+        let required_sleeping =
+            ((player_count as f64 * sleep_percentage as f64) / 100.0).ceil() as usize;
+
+        sleeping_player_count >= required_sleeping
```
BjornTheProgrammer, line 2202: Probably should make this a seperate PR.

pumpkin/src/entity/living.rs
denys-shatin: 1 chunk = 16 blocks

pumpkin/src/world/mod.rs
denys-shatin: ok

pumpkin/src/world/mod.rs
denys-shatin: renamed this PR and made another for sleep percentage

Purdze:
The broadcast-to-nearby approach is the right idea, but the radius should probably be entity tracking distance, not view_distance * 16, that could broadcast 500+ blocks on high view distance servers. Also the pattern is repeated 4 times, and get_nearby_players doesn't exclude self, so players could receive their own swing/pickup animations

denys-shatin:
yes sir!
stock-vector-soldier-emoticon-saluting-saying-yes-sir-131260427

denys-shatin:
Forgot to say that i did thatimage

Purdze:
Looks good thank you

Purdze:
Rebase, then happy to merge

denys-shatin:
did that

pumpkin/src/world/mod.rs
```diff
@@ -264,6 +264,12 @@ impl World {
             .unwrap_or("world")
     }
 
+    /// Get the entity tracking distance for broadcasting packets.
+    /// Uses a fixed radius of 64 blocks, which is the standard entity tracking distance.
+    pub const fn get_broadcast_radius(&self) -> f64 {
+        64.0
```
BjornTheProgrammer, line 270: Do we actually want to pull this from a config value?

Snowiiii:
please choose an better title

Snowiiii:
Feel free to reopen

PR #1521, opened 2026-02-10
World names

denys-shatin:
## Description
Replaces hardcoded "world" string with actual world folder names in logs and query protocol responses.

## Changes
- [x] Modified player removal logs to use world folder name instead of hardcoded "world"
- [x] Updated Query protocol (both Full and Basic status) to return actual world folder name
- [x] Extracts world name from `level.level_folder.root_folder` path
- [x] Falls back to "world" if world name cannot be determined

## Implementation Details
Uses the world folder name (e.g., "world", "world_nether", "world_the_end") instead of dimension technical names (e.g., "minecraft:overworld"). This provides more meaningful information in server logs when players disconnect and in Query protocol responses for server monitoring tools.

## Example
Before: `Removed player id Steve from world world (5 chunks remain cached)` and Query response: `map = "world"`

After: `Removed player id Steve from world_nether (5 chunks remain cached)` and Query response: `map = "world_nether"`

## Testing
- [x] Tested with default world folders (world, world_nether, world_the_end)
- [x] Verified logs show correct world names
- [x] Confirmed Query protocol returns proper world names
- [x] Tested fallback behavior when world name unavailable

### DIFF
--- a/pumpkin/src/net/query.rs
+++ b/pumpkin/src/net/query.rs
@@ -150,14 +150,13 @@ async fn handle_packet(
                             hostname: CString::new(server.basic_config.motd.as_str())?,
                             version: CString::new(CURRENT_MC_VERSION)?,
                             plugins: CString::new(plugins)?,
-                            map: CString::new(server.worlds.load().first().map_or("world", |w| {
-                                w.level
-                                    .level_folder
-                                    .root_folder
-                                    .file_name()
-                                    .and_then(|n| n.to_str())
-                                    .unwrap_or("world")
-                            }))?,
+                            map: CString::new(
+                                server
+                                    .worlds
+                                    .load()
+                                    .first()
+                                    .map_or("world", |w| w.get_world_name()),
+                            )?,
                             num_players: server.get_player_count(),
                             max_players: server.basic_config.max_players as usize,
                             host_port: bound_addr.port(),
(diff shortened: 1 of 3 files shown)


pumpkin/src/net/query.rs
```diff
@@ -150,7 +150,13 @@ async fn handle_packet(
                             hostname: CString::new(server.basic_config.motd.as_str())?,
                             version: CString::new(CURRENT_MC_VERSION)?,
                             plugins: CString::new(plugins)?,
-                            map: CString::new("world")?, // TODO: Get actual world name
+                            map: CString::new(
+                                server
+                                    .worlds
+                                    .load()
+                                    .first()
+                                    .map_or("world", |w| w.dimension.minecraft_name),
```
BjornTheProgrammer: This uses the dimension name, not the actual world name. I'm not aware of on the top of my head on how to get that, but that is probably what we should be using instead.

denys-shatin:
bjorn thanks you for the review, i considered that and made update

denys-shatin:
also changed description of PR

Purdze (commented):
Looks good, however the same .file_name().and_then(...).unwrap_or("world") chain is
  repeated 3 times - worth extracting to a helper.

denys-shatin:
> Looks good, however the same .file_name().and_then(...).unwrap_or("world") chain is repeated 3 times - worth extracting to a helper.

implemented that! ty for the review

PR #1522, opened 2026-02-10
fix: collision

### LINKED ISSUE #1505: touching cactus doesnt deal damage

JustMarfix:
## Description

This fixes collision logic for cactus, campfire and probably some other blocks.

Fixes #1505

### DIFF
--- a/pumpkin/src/entity/item.rs
+++ b/pumpkin/src/entity/item.rs
@@ -351,12 +351,14 @@ impl EntityBase for ItemEntity {
         Box::pin(async move {
             // TODO: invulnerability, e.g. ancient debris
             loop {
-                let health = self.health.load(Relaxed);
-                if let Ok(new_health) =
-                    self.health
-                        .compare_exchange(health, health - amount, AcqRel, Relaxed)
+                let current = self.health.load(Relaxed);
+                let new = current - amount;
+                if self
+                    .health
+                    .compare_exchange(current, new, AcqRel, Relaxed)
+                    .is_ok()
                 {
-                    if new_health <= 0.0 {
+                    if new <= 0.0 {
                         self.entity.remove().await;
                     }
                     return true;


pumpkin-data/src/block_state.rs
```diff
@@ -89,6 +89,12 @@ impl BlockState {
         self.state_flags & IS_SOLID != 0
     }
 
+    pub const fn block_movement(&self) -> bool {
+        self.id != 129 && // cobweb
```
FMMazur: i think it would be better:
```rs
pub const fn block_movement(&self) -> bool {
        self.id != Block::COBWEB.id &&
        self.id != Block::BAMBOO_SAPLING.id &&
        self.is_solid()
    }
```

Purdze: there's already blocks_movement() in block_properties that does this same check, so neither the hardcoded IDs nor the Block::COBWEB.id version are needed, just use the existing function

JustMarfix: yeah, already figured it out, I somehow missed it lol
thx for comments

JustMarfix:
> Hardcoded state IDs (129, 789) will break on registry changes, use block constants instead. Also, removing the -0.001 bbox shrink and using a full-block AABB instead of actual collision shapes will cause false-positive collisions at block edges.

I brought back box shrink, yet made it 1.0E-5F as in vanilla code.

It also turns out vanilla uses `VoxelShapes.fullCube()` as the result of `getInsideCollisionShape(...)` for almost all blocks, except for powder snow, cauldron with lava and water, tripwire and end portal block. So I made a new function in BlockBehaviour to overwrite default behaviour if needed.

JustMarfix:
> self.health.store(self.health.load() - amount) in item damage is a race condition with concurrent damage sources, needs an atomic read-modify-write and the bbox shrink epsilon is 1.0E-5 but vanilla uses 1.0E-7

Done. Also implemented atomicf32, as discussed [here](https://discord.com/channels/1268592337445978193/1268611318617866261/1471242409139634456).

Purdze:
Looks good thank you

PR #1524, opened 2026-02-10
feat: Create bootsrapper crate for managing pumpkin with cargo

RodrigoAroeira:
# Add `pumpkin-build` standalone bootstrap binary for managing `pumpkin`

## Description

This PR introduces a new standalone crate, `pumpkin-build`, which acts as a bootstrap tool for the `pumpkin` binary. Its main goal is to allow `pumpkin` to be managed by cargo and be made globally available. 

## Changes in this PR

- **Standalone crate `pumpkin-build`:**
  - **Cargo.toml & Cargo.lock:** Standard crate setup, with a `bin` target named `pumpkin`.
  - **src/main.rs:** Implements bootstrap logic:
    - Detects if running under `cargo run` and warns that bootstrapping won't work
    - Locates the real `pumpkin` binary in `../target/release/`
    - Copies the real pre-built binary to a temporary file and replaces the bootstrap binary
    - Provides clear error messages for each possible failure
    - Prints success messages and instructions for normal usage once bootstrapped.

- **Workspace updates:**
  - `pumpkin-build` is excluded from the workspace to prevent accidental compilation as part of normal operations.

## Motivation

* Since `pumpkin` can't be installed directly with `cargo install`, it can't be just as easily added to path to be run in any directory. To circumvent this, I created the `pumpkin-build` directory with a crate named `pumpkin`, inspired by how [yazi-build](https://github.com/sxyazi/yazi/tree/main/yazi-build) is used to install it's necessary binaries. The crate was named `pumpkin` to facilitate `cargo uninstall` operations. 

* It would make it easier to have multiple servers, without having to clone repo multiple times, copy/link pre-compiled binary for each new server directly2 or manually put the binary in a directory in path to accomplish the same purpose.

> [!NOTE]
> Using the bootstrapper would be completely optional, and only for those who trully wish to do so.

## Testing

- Verified the bootstrap replaces itself with the real binary successfully.
- Checked messages and warnings when running under `cargo run`.
- Confirmed the binary works normally after bootstrapping.

https://github.com/user-attachments/assets/ec9df758-b4ed-4a14-8fd8-71f94508be61

## Limitations / Platform Support

- This bootstrap has **only been tested on Linux**
- Support for other platforms has not been verified and may require adjustments

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ members = [
     "pumpkin/",
     "pumpkin-data",
 ]
-exclude = ["pumpkin-codegen"]
+exclude = ["pumpkin-codegen", "pumpkin-build"]
 
 [workspace.lints.clippy]
 # Groups


BjornTheProgrammer:
I would like to review this. But I think it would be best to wait for Alex's thoughts about it first. It might be a week before we can get back to you on this.

Snowiiii:
I don't think this is necessary

RoosterBooster007:
Hi @RodrigoAroeira, thanks for taking the time to put this together. While we appreciate the effort, this isn't something we are currently looking for. So, I'm going to close this PR. 

If you're still interested in contributing, check out our available issues or chat with us in the [Discord](https://discord.gg/pumpkinmc). Thanks again for your interest in improving the project!

RodrigoAroeira:
What is the recommended way to install Pumpkin from source? Is copying the compiled binary into a directory on the PATH currently the only supported method?

PR #1518, opened 2026-02-10
fix: hunger and food behavior

RoosterBooster007:
## Description
Aiming to fix several food-related bugs.

**Changes:**
- [x] Improved exhaustion and saturation (for movement and player actions)
- [x] Working chorus fruit (without cooldown for now)
- [x] Food effects (like poison from spider eye, hunger from rotten flesh, etc.)
- [x] Regen hunger in peaceful mode (if switching with low hunger)

## Testing
- [x] Player gains exhaustion from activities
- [x] Chorus fruit tp works and is safe
- [ ] Hunger stays full/regens in peaceful mode

Issue #764, opened 2025-04-24
Effects


hthug06:
### Some effects are not working, here the full list :
- **speed** : :x: Don't affect the speed and the fov
- **slowness** : :x: Don't affect the speed and the fov
- **haste** : ✅ 
- **mining_fatigue** : ✅ 
- **strength** : ❓ Idk can't really try for the moment
- **instant_health** : :x: Don't heal you
- **instant_damage** : :x: Don't harm you
- **jump_boost** : ✅ You jump higher  :x You take normal fall damage
- **nausea** : ✅ 
- **regeneration** : ✅ The health bar is moving  :x: Don't heal you
- **resistance** : :x: No effect
- **fire_resistance** : ❓Idk you can't be on fire for the moment
- **water_breathing** : ❓can't try for now too
- **invisibility** : :x: You are still visible
- **blindness** : ✅ 
- **night_vision** : ✅ 
- **hunger** : ✅ Hunger bar moving  :x: You don't lose hunger
- **weakness** : ❓Can try for now 
- **poison** : :x: Don't affect your health
- **wither** : :x: same as poison
- **health_boost** : :x: Don't add you health
- **absorption** : :x: Same as health_boost
- **saturation** : :x: Nothing append 
- **glowing ** : :x: You simply don't glow
- **levitation** : ✅ 
- **luck** : ❓ Can't try for now
- **unluck** : ❓ Can't for now
- **slow_falling** : ✅ You fall slowly  :x: You take fall damage
- **conduit_power** : :x: No water_breathing, no night vision and no haste
- **dolphins_grace** : ✅ 
- **bad_omen** : :x:
- **hero_of_the_village** : ❓ But i think it's the same as bad_omen
- **darkness** : ✅ 
- **trial_omen** : ❓ Can't try for now
- **raid_omen** : ❓ Can't try for now
- **wind_charged** : :x: No effect when the entity die
- **weaving** : :x: Same as wind_charged
- **oozing** : :x: Same as wind_charged
- **infested** :  :x: Same as wind_charged

I tried every effect manually on the /effect command

If you found an effect i don't see or if mojang add a new one feel free to mention it

RoosterBooster007:
@hthug06, can you update this?

hthug06:
# Update for the effects (02/03/2026)

- **speed** : ✅ (now affect speed and FOV) 🆕
- **slowness** : ✅ (now affect speed and FOV) 🆕
- **haste** : ✅ 
[...]

PR #1526, opened 2026-02-11
fix: some entity effects and attributes system

RoosterBooster007:
## Description
Fixes several entity effects (should work with #1518). Added attributes system!

**Updated Effects:**
- [x] Speed
- [x] Slowness
- [x] Strength
- [x] Weakness
- [x] Regeneration
- [x] Resistance
- [x] Fire Resistance
- [x] Absorption
- [x] Instant Health
- [x] Health Boost
- [x] Instant Damage
- [x] Poison
- [x] Wither
- [x] Hunger
- [x] Saturation
- [x] Infested (unlimited exp, basically)
- [x] Invisibility (you disappear now)
- [x] Glowing (you glow now)

**Already working:**
- [x] Haste
- [x] Mining Fatigue
- [x] Jump Boost
- [x] Nausea
- [x] Water Breathing
- [x] Blindness
- [x] Night Vision
- [x] Levitation
- [x] Slow Falling
- [x] Dolphin's Grace

**Not added:**
- Conduit Power
- Bad Omen
- Hero of the Village
- Darkness
- Trial Omen
- Raid Omen
- Wind Charged
- Weaving
- Oozing
- Breath of the Nautilus
- Luck
- Bad Luck

I added an ``AttributeRegistry`` so that entities use their correct values (damage, speed, etc.). They can be defined in their respective impls with ``create_attributes()``.  This also allows effects like speed and slowness to work. The biggest changes were adding the attributes packet (``update_attributes.rs``) and manager (``attributes.rs``).

## Testing
- The altered effects above seem to behave correctly (with the correct attributes being applied and modified)

#764

### DIFF
--- a/pumpkin/src/entity/passive/cat.rs
+++ b/pumpkin/src/entity/passive/cat.rs
@@ -1,5 +1,7 @@
 use std::sync::{Arc, Weak};
 
+use crate::entity::attributes::AttributeBuilder;
+use pumpkin_data::attributes::Attributes;
 use pumpkin_data::{entity::EntityType, item::Item};
 
 use crate::entity::{
@@ -41,6 +43,14 @@ impl CatEntity {
         };
 
         mob_arc
+    }
+
+    #[must_use]
+    pub fn create_attributes() -> AttributeBuilder {
+        AttributeBuilder::new()
+            .add(Attributes::MOVEMENT_SPEED, 0.3)
+            .add(Attributes::ATTACK_DAMAGE, 3.0)
+            .add(Attributes::MAX_HEALTH, 10.0)
     }
 }
 
--- a/pumpkin/src/entity/passive/chicken.rs
+++ b/pumpkin/src/entity/passive/chicken.rs
@@ -3,6 +3,8 @@ use std::sync::{
     atomic::{AtomicI32, Ordering, Ordering::Relaxed},
 };
 
+use crate::entity::attributes::AttributeBuilder;
+use pumpkin_data::attributes::Attributes;
 use pumpkin_data::{entity::EntityType, item::Item};
 use pumpkin_world::item::ItemStack;
 use rand::RngExt;
@@ -60,6 +62,13 @@ impl ChickenEntity {
         };
 
         mob_arc
+    }
+
+    #[must_use]
+    pub fn create_attributes() -> AttributeBuilder {
+        AttributeBuilder::new()
+            .add(Attributes::MOVEMENT_SPEED, 0.25)
+            .add(Attributes::MAX_HEALTH, 4.0)
     }
 }
 
--- a/pumpkin/src/entity/passive/cow.rs
+++ b/pumpkin/src/entity/passive/cow.rs
@@ -1,5 +1,7 @@
 use std::sync::{Arc, Weak};
 
+use crate::entity::attributes::AttributeBuilder;
+use pumpkin_data::attributes::Attributes;
 use pumpkin_data::{entity::EntityType, item::Item};
 
 use crate::entity::{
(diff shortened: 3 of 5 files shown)


Purdze:
saturation formula is off - (amplifier + 1) * 2.0 should be (amplifier + 1)^2 * 2.0 (vanilla's eat() multiplies foodLevel * saturationModifier * 2.0 where both are amplifier + 1)

default_movement_speed changed to 0.1 in LivingEntity::new() which all mobs share, correct for players but will slow down every other mob too

send_attribute_update() fires on every effect add/remove, even ones that don't affect attributes (regen, poison, etc, should only trigger for speed/slowness changes

instant Health/Damage get stored in active_effects after applying, vanilla instant effects don't persist, they fire once and are done

I know its a draft though so dont worry

RoosterBooster007:
> saturation formula is off - (amplifier + 1) * 2.0 should be (amplifier + 1)^2 * 2.0 (vanilla's eat() multiplies foodLevel * saturationModifier * 2.0 where both are amplifier + 1)

[From the wiki](https://minecraft.wiki/w/Saturation_(effect)): "The Saturation effect instantly replenishes 1 (🍖) × level and 2 × level points of saturation‌" - isn't my current code for this correct? 

> default_movement_speed changed to 0.1 in LivingEntity::new() which all mobs share, correct for players but will slow down every other mob too

I wasn't sure how to implement this, as it used to be 0.25 before I changed it. How should I get/add this data for each entity type?

> send_attribute_update() fires on every effect add/remove, even ones that don't affect attributes (regen, poison, etc, should only trigger for speed/slowness changes

Fixed! I realized I could easily check ``effect.effect_type.attribute_modifiers``

> instant Health/Damage get stored in active_effects after applying, vanilla instant effects don't persist, they fire once and are done

Fixed

Purdze:
yeah you're right sorry..,

the movement speed point still applies though, currently it's hardcoded for all entities but only correct for players, a TODO is fine for now

Purdze:
entity_type.max_health is no longer used, so any mob without a create_attributes() registration defaults to 20 HP. The Wither's builder is also missing MAX_HEALTH (should be 300). You'll need to either add create_attributes() for all entities or fall back to entity_type.max_health when no attribute override exists

Strength/Weakness double-counting: add_effect() already generically applies attribute modifiers from effect.effect_type.attribute_modifiers, but try_attack() and Player::attack() also manually add +3/level and -4/level. If the data has attribute modifiers for these effects, the bonus gets applied twice. I think the attribute modifier system is the correct one

RoosterBooster007:
> entity_type.max_health is no longer used, so any mob without a create_attributes() registration defaults to 20 HP. The Wither's builder is also missing MAX_HEALTH (should be 300). You'll need to either add create_attributes() for all entities or fall back to entity_type.max_health when no attribute override exists

Yep, health will fall back to the default 20 (we should probably just have ``create_attributes()`` on all entities that need it). I added the wither's default ``MAX_HEALTH``. 

> Strength/Weakness double-counting: add_effect() already generically applies attribute modifiers from effect.effect_type.attribute_modifiers, but try_attack() and Player::attack() also manually add +3/level and -4/level. If the data has attribute modifiers for these effects, the bonus gets applied twice. I think the attribute modifier system is the correct one

Fixed this. It duped for non-player entities, and I simplified the player ``attack()`` code.

Purdze:
should the get_attribute_base check the local instance first?

Purdze:
In remove_effect, the comment says // Apply glowing effect but the block removes it

Snowiiii:
@RoosterBooster007 fix conflicts please

Snowiiii:

PR #1527, opened 2026-02-11
Sleep Percentage 

denys-shatin:
## Description
Implements the `playersSleepingPercentage` gamerule, allowing servers to configure what percentage of players need to sleep to skip the night.

## Changes
- [x] Modified `should_skip_night()` in `World` to use `players_sleeping_percentage` from game rules
- [x] Calculates required sleeping players: `ceil(player_count * percentage / 100)`
- [x] Returns false early if no players are online
- [x] Replaces hardcoded "all players must sleep" logic

## Behavior
- Default: 100% (all players must sleep, vanilla behavior)
- Can be configured to any value 0-100
- Example: With 50% and 10 players, only 5 need to sleep to skip night
- Uses ceiling to ensure minimum required players (e.g., 3.5 → 4 players)

## Testing
- [x] Tested with various player counts and percentages
- [x] Verified calculation rounds up correctly
- [x] Confirmed vanilla behavior with 100%

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -2115,9 +2115,11 @@ impl World {
             .level_info
             .load()
             .game_rules
-            .players_sleeping_percentage;
+            .players_sleeping_percentage
+            .clamp(0, 100);
         let required_sleeping =
             ((player_count as f64 * sleep_percentage as f64) / 100.0).ceil() as usize;
+        let required_sleeping = required_sleeping.max(1);
 
         sleeping_player_count >= required_sleeping
     }


Purdze:
when sleep_percentage is 0, required_sleeping becomes 0 and the night skips even if nobody is sleeping, need to require at least 1 player sleeping, also worth clamping the percentage to 0..=100 to handle negative or out-of-range values.

denys-shatin:
> when sleep_percentage is 0, required_sleeping becomes 0 and the night skips even if nobody is sleeping, need to require at least 1 player sleeping, also worth clamping the percentage to 0..=100 to handle negative or out-of-range values.

nice catch, thanks! I’ll fix that now.

PR #1530, opened 2026-02-11
ci: build Hello-Pumpkin plugin

FabseGP:
## Description

To ensure any changes pushed doesn't break the Hello-Pumpkin plugin example (or at least make us aware of a breakage and patch it), build the plugin on every pr

## Testing

WIP


Snowiiii:
This is not needed and we also soon will use WASM

PR #1531, opened 2026-02-11
Make pumpkin console colorful

GreenedDev:
## Description
Adds some colors to Pumpkin console.
image

### DIFF
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -95,28 +95,7 @@ async fn main() {
             "Release"
         }
     );
-
-    log::warn!(
-        "{}",
-        TextComponent::text("Pumpkin is currently under heavy development!")
-            .color_named(NamedColor::DarkRed)
-            .to_pretty_console(),
-    );
-    log::info!(
-        "Report issues on {}",
-        TextComponent::text("https://github.com/Pumpkin-MC/Pumpkin/issues")
-            .color_named(NamedColor::DarkAqua)
-            .to_pretty_console()
-    );
-    log::info!(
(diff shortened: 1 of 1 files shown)


Purdze:
Using TextComponent with legacy § codes for console logs feels like the wrong thing to do, it's built for minecraft chat, not terminal output, it also makes simple one line log statements much harder to read in code, lighter approach (direct ANSI codes or a terminal color crate) would be bwrrwe in my opinion. @BjornTheProgrammer thoughts on this?

BjornTheProgrammer:
I think that we should use TextComponents, since most of the input and output from the users in game will become TextComponents as well. Having good support for them in the console makes most sense to me. Additionally changing how TextComponents render will then change the rest of the terminal output.

Purdze:
If you go this route though, I'd use the builder api  (.color_named(...)) consistently instead of mixing it with from_legacy_string("§6...",  the builder is more readable and doesn't rely on legacy formatting codes for hardcoded strings

BjornTheProgrammer:
Forgot to mention it as well, but with ANSI being disabled by default in tracing due to security reasons, this could potentially be good segmentation for that as well https://github.com/tokio-rs/tracing/issues/3378.

BjornTheProgrammer:
I'll actually agree on the no `from_legacy_string`

Purdze:
Sounds good. If you can use the builder API instead, we can merge it

GreenedDev:
image

Purdze:
extracting those into a helper would keep main() cleaner, especially now that each log line is several lines with the builder calls

PR #1540, opened 2026-02-12
chore: replace deps + optimize dev dependencies

FabseGP:
## Description

cargo tree shows quite a few duplicated dependencies, while a few haven't been updated for a long time:

- make use of the new API in crossfire instead of the compability API
- replace take_mut with [replace_with](https://crates.io/crates/replace_with), a modern and faster alternative
- replace serde_json5 with [json5](https://crates.io/crates/json5), boasting higher throughput
- replace md5 with [md-5](https://crates.io/crates/md-5), a pure rust implementation
- don't use git-version of the rand-crate (not sure why that was done either, but rand have since released v0.10)

Certain dependencies also had features enabled that were unused and with a version bump as well, 10 fewer dependencies is now compiled in the dev-profile. Dependencies in the dev-profile are now compiled with opt-level 3 as well, which reduced the incremental compile time with about 20s on my machine (the initial compile time will take longer, but will benefit from [optimized macros](https://corrode.dev/blog/tips-for-faster-rust-compile-times/#avoid-procedural-macro-crates))

## Testing

Compiles and runs without any panics

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,12 +10,12 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aes"
-version = "0.8.4"
+version = "0.9.0-rc.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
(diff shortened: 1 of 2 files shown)


Purdze:
good work, signature = "2.2.0" in workspace deps also looks dead now

FabseGP:
> signature = "2.2.0" in workspace deps also looks dead now

fixed, I also noticed other dependencies that were quite old, but one of them (cfb8) is currently broken because one of its dependencies had a breaking change

setting this as a draft till cfb8 is updated

Snowiiii:
@FabseGP please split this up into multiple PRs, also why is this still a draft?

PR #1544, opened 2026-02-12
fix(chunk): resend replaced chunk data at same position

yunuservices:
This PR fixes the remaining chunk visibility issue still reproducible after #1510.

  ## Problem
  In some cases, a chunk gets regenerated/replaced while the player is still within view distance.
  The client then keeps showing missing terrain / void holes instead of receiving updated chunk data.

  ## Root Cause
  `ChunkManager::chunk_sent` only tracked chunk coordinates (`HashSet<Vector2<i32>>`).
  So if a new chunk instance appeared at the same position, resend was blocked because that position was already marked as sent.

  ## Changes
  - Changed `chunk_sent` from `HashSet<Vector2<i32>>` to `HashMap<Vector2<i32>, usize>`.
  - Added `should_enqueue_chunk(position, chunk)` helper.
  - Deduplication now checks both:
    - chunk position
    - chunk instance identity (`Arc` pointer)
  - If the chunk instance changes at the same position, it is enqueued and resent.
  - Applied this logic in both:
    - `pull_new_chunks`
    - `push_chunk`
  - Updated related `retain` / lookup logic for the new map structure.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -4,8 +4,8 @@ use std::f64::consts::TAU;
 use std::mem;
 use std::num::NonZeroU8;
 use std::ops::AddAssign;
-use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, AtomicI32, AtomicI64, AtomicU8, AtomicU32, Ordering};
+use std::sync::{Arc, Weak};
 use std::time::{Duration, Instant};
 
 use arc_swap::ArcSwap;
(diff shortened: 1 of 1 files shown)


Purdze (commented):
Looks good however a small note, if an Arc is dropped and a new one happens to get the same address, it'd be treated as the same chunk, but in practice that's very unlikely since loaded chunks are held alive

yunuservices:
> Looks good however a small note, if an Arc is dropped and a new one happens to get the same address, it'd be treated as the same chunk, but in practice that's very unlikely since loaded chunks are held alive

well, actually i really thought same. i will rewrite there, thanks!

PR #1545, opened 2026-02-12
Fixes #1513 (makes trapdoor placing more like vanilla and fixes it rotating incorrectly)

### LINKED ISSUE #1513: The trapdoor was placed facing the wrong direction.

### Operating System

Windows 11

a89fb93

### What happened?

on_place() block/blocks/trapdoor.rs
`let direction = args.player.living_entity.entity.get_horizontal_facing().opposite();`
`trapdoor_props.facing = direction;`
It causes

https://github.com/user-attachments/assets/7e149810-fc0e-4781-8a62-4b98e26a00a1

### To Reproduce

Place a trapdoor beside a block.

### Expected behavior

The trapdoor should face the same direction as the face of the block to which it is placed.

TheRealMusawer:
this fixes bug #1513 Pumpkin always set the trapdoor’s facing direction to the opposite of the player’s horizontal facing, ignoring the face the player actually clicked. This caused trapdoors to consistently rotate incorrectly compared to vanilla behavior. This fixes it

### DIFF
--- a/pumpkin/src/block/blocks/trapdoor.rs
+++ b/pumpkin/src/block/blocks/trapdoor.rs
@@ -4,7 +4,7 @@ use crate::block::{BlockBehaviour, BlockFuture, NormalUseArgs, OnNeighborUpdateA
 use crate::entity::player::Player;
 use crate::world::World;
 use pumpkin_data::BlockDirection;
-use pumpkin_data::block_properties::{BlockHalf, BlockProperties, HorizontalFacing};
+use pumpkin_data::block_properties::{BlockHalf, BlockProperties};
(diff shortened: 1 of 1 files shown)


TheRealMusawer:
for anyone reading this all the commits after like the first and i think second one are just trying to get the checks to be green.

TheRealMusawer:
to summarize everything up , the old code used args.player.living_entity.entity.get_horizontal_facing().opposite()
which when placing on side of block from like front messed it up, I changed it to args.direction.to_horizontal_facing().unwrap_or(player_facing) which should fix it.

Purdze:
Looks good thank you

TheRealMusawer:
your welcome :)

PR #1547, opened 2026-02-12
fix: sweet berry bush collision

JustMarfix:
## Description

- fixed player collision with berry bush
- fixed items destroying in berry bush

Also added tickLastPos functionality to update_last_pos, they're basically doing the same thing, except tickLastPos also sets last movement vector.

## Testing

Berry bush collision now works like in vanilla.

### DIFF
--- a/pumpkin/src/block/blocks/plant/crop/sweet_berry_bush.rs
+++ b/pumpkin/src/block/blocks/plant/crop/sweet_berry_bush.rs
@@ -116,7 +116,7 @@ impl BlockBehaviour for SweetBerryBushBlock {
             entity
                 .slow_movement(args.state, Vector3::new(0.8, 0.75, 0.8))
                 .await;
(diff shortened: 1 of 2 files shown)


Purdze (changes requested):
is_controlled_by_player() is just checking if the entity is a player, not if it's a vehicle ridden by a player (vanilla way) should rename it to is_player()

the spider/cobweb check is dead code since slow_movement is only called from sweet berry bush.

JustMarfix:
> the spider/cobweb check is dead code since slow_movement is only called from sweet berry bush.

yeah but it should be called from cobweb and some other blocks collision code. should I do a cobweb one in this pr?

Purdze:
keep this pr for sweet berry and do cobweb/other blocks in a follow up, then the spider check slow_movement becomes live code when cobweb is done

Issue #1546, opened 2026-02-12
Crash when take a firework


SparkleSavvy:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11 23h2

### Server Software Version/Commit

582d465

### What happened?

Exec start pumpkin-X64-Windows.exe
Found existing log file at 'logs\latest.log', gzipping it now...
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m main ThreadId(01) Starting ←[33mPumpkin←[0m ←[92m0.1.0-dev+1.21.11←[0m Minecraft (Protocol ←[34m774←[0m)
←[2m2026-02-12 21:46:14←[0m ←[33m WARN←[0m main ThreadId(01) ←[31mPumpkin is currently under heavy development!←[0m
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m main ThreadId(01) Report issues on ←[36mhttps://github.com/Pumpkin-MC/Pumpkin/issues←[0m
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m main ThreadId(01) Join our ←[34mDiscord←[0m for community support: ←[96mhttps://discord.gg/pumpkinmc←[0m
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m main ThreadId(01) Favicon usage is disabled.
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m main ThreadId(01) Starting parallel world load...
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m tokio-runtime-worker ThreadId(14) Loading ←[32mminecraft:overworld←[0m
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m tokio-runtime-worker ThreadId(15) Loading ←[32mminecraft:the_nether←[0m
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m tokio-runtime-worker ThreadId(16) Loading ←[32mminecraft:the_end←[0m
←[2m2026-02-12 21:46:14←[0m ←[32m INFO←[0m                 main ThreadId(01) All worlds loaded successfully.
$ 2026-02-12 21:46:14  INFO                 main ThreadId(01) Query protocol is enabled. Starting...
2026-02-12 21:46:14  INFO                 main ThreadId(01) Started server; took 6ms
[...]

PR #1548, opened 2026-02-13
feat(fireworks): implement firework data components

KsawierWilczynski:
## Description
Implemented `firework` and `firework_explosion` data components.

## Why was this done
This PR addresses and closes Issues #1546 
The server panicked when it encountered certain types of fireworks. Namely anything other than `flight duration: 1` fireworks.

## Testing
1. Open creative menu
2. Take out all types of fireworks
3. Server doesn't panic

### DIFF
--- a/pumpkin-nbt/src/tag.rs
+++ b/pumpkin-nbt/src/tag.rs
@@ -501,16 +501,13 @@ impl<'de> Deserialize<'de> for NbtTag {
                         }
                         Ok(NbtTag::ByteArray(vec.into_boxed_slice()))
                     }
-                    LIST_ID => {
+                    _ => {
                         let mut vec = Vec::new();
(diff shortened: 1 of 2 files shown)


Purdze (commented):
colors should be fixed-width 32-bit ints, not varInts, vanilla writes firework colors as 4-byte int values after a varInt-prefixed count. using varInt for the color values will produce corrupt packets.., also needs a length cap on colors_len/fade_colors_len/explosions_len during deserialization to prevent oom from malicious packets

KsawierWilczynski:
Forgot to `cargo fmt` mid re-base...
Changed colors to i32, added a 256-color limit which *should* be more than enough even considering Admins can spawn in more than that.
Also added a flight-duration check to prevent invalid NBT firework spawning/creating through any method.

PR #1549, opened 2026-02-13
fix: prevent mobs from attacking when already dead

PrintedScript:
## Description
Mobs like zombies and drowned were still able to deal damage to players even after being killed, this fixes that.

## Testing

Before:
https://github.com/user-attachments/assets/e88738ad-a568-4a5b-826c-85e6908f940c

After:
https://github.com/user-attachments/assets/9fb3c082-f71b-453f-ba29-932f07128574


Purdze (commented):
ideally the dead check would be higher up in the AI tick (so dead mobs don't even target/path) 

looks good though thank you

Issue #1402, opened 2026-01-29
Tracking: Redstone


Snowiiii:
## Tracking: Redstone

This issue tracks the implementation of Redstone components, signal calculations, and block updates. Implementation must ensure parity with the latest Java Edition behavior

### Core Logic & Engine
- [x] Redstone Wire signal propagation (Update order)
- [x] Redstone Tick System
- [x] Hard vs. Soft powering logic

### Power & Sensors
- [x] Pressure Plates (All variants)
- [x] Tripwire Hook
- [x] Trapped Chest
- [x] Sculk Sensors / Shriekers / Calibrated Sensors
- [x] Daylight Detector
- [x] Target Block
- [x] Redstone Torch / Levers / Buttons

### Logic & Timing
- [x] Comparator (Maintain state/modes)
- [x] Repeater (Tick delays/locking)
- [x] Note Block
- [x] Hopper (Item transfer & cooldowns)
- [x] Observer
- [x] Rails

### Mechanisms & Inventory
- [ ] Pistons (Sticky/Regular) & Block Swapping (Currently broken)
- [x] Dropper
- [x] Dispenser
- [x] Crafter (Latest version recipe handling)
- [x] Minecarts (Powering/Activating)
- [x] TNT (Ignition/Priming)

### Interactive & Utility
- [x] Bell
- [x] Redstone Lamp
- [x] Doors, Gates, and Trapdoors
- [x] Lectern (Signal strength based on page)
- [x] Composter (Signal strength based on fill)

WalterKruger:
I found a few more missing features:
- Comparators don't read the state/fullness of most blocks. (Excluding Composter and Chiseled Bookshelf)
- Hoppers don't collect item entities.
- Adding/Removing a power source to an already activated Powered/Activator Rail propagates that change further down the line. In vanilla, they need an additional external update to notice the change. (see https://bugs.mojang.com/browse/MC/issues/MC-957)
- Lightning Rod (non functional)

MiyaSHs:
i think it would be lovely for this to be configurable! choosing between full vanilla parity or a implementation like alternative current, having less block updates and better performance, altho im just asking i have no knowledge of how redstone works in current pumpkinmc and in minecraft servers in general

XtinywaffleX:
[...]

PR #1550, opened 2026-02-13
feat: add trapped chest block entity with redstone power support

aicheye:
## Description

Implements trapped chests and redstone emissions based on number of players viewing (#1402)

All features work except storage persistence across server restarts, which is because the implementation is not yet complete at the `BlockEntity` level. When the implementation is added, trapped chests should be persistent across wipes.

image

## Testing

Not sure if I should include tests in this PR, largely extends already existing features. If tests are preferred, high-level suggestions of what should be tested would be helpful.

### DIFF
--- a/pumpkin-world/src/block/entities/chest.rs
+++ b/pumpkin-world/src/block/entities/chest.rs
@@ -1,33 +1,15 @@
-use std::{
-    any::Any,
-    array::from_fn,
-    pin::Pin,
-    sync::{
-        Arc,
-        atomic::{AtomicBool, Ordering},
-    },
-};
+use std::sync::{Arc, atomic::AtomicBool};
 
-use pumpkin_data::{
-    HorizontalFacingExt,
-    block_properties::{BlockProperties, ChestLikeProperties, ChestType},
-    sound::{Sound, SoundCategory},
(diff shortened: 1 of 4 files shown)


aicheye:
> Massive code duplication here, TrappedChestBlockEntity and TrappedChestBlock are near-identical copies of their regular chest counterparts. The only real differences are the ID constant and update_neighbors in on_viewer_count_update. This should share code via composition or a shared impl rather than copy-pasting

Done, I also refactored wooden chests and copper chests to use a common implementation.

> Worth double-checking the strong power direction, the comment says "only emitted upward" but the code checks args.direction == BlockDirection::Down. Vanilla's getStrongRedstonePower receives the direction from which power is queried, so when the block below asks, it passes Direction.UP

I misread the wiki, the block below is supposed to be strongly powered, so `BlockDirection::Down` is the correct check. There was also a bug where the block below wouldn't be updated (i.e. no redstone updates emitted either), and I fixed that as well.

aicheye:
Oops sorry, just realized you were also referring to the block entity file as well. I'll fix that now.

Purdze:
Looks good, thank you!

Issue #1515, opened 2026-02-10
Zombies will attack the creative player and will not burn in the day.


DIOJOIO:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.11 (Commit: 5a5b9e4/debug)

### What happened?

Zombies attack players in Creative Mode and do not burn in sunlight

### To Reproduce

/gamemode creative
/time set day
/summon zombie
run around, the zombie will follow you

### Expected behavior

1. Zombies should ignore players in Creative Mode.
2. Zombies should burn in sunlight unless they have a helmet or the Fire Resistance effect.

Snowiiii:
Fixed

PR #1557, opened 2026-02-13
Prevent targeting of creative players for mobs and fixes 1 of the 2 bugs listed in bug #1515

TheRealMusawer:
Added a check to prevent targeting creative players. like for Hostile Mob's such as zombies

fixes  1 of the 2 bugs in #1515 

what changed: Inside TrackTargetGoal::can_track() a check was added to see if the player was creative and if they were it would ignore them like normal vanilla behavior


TheRealMusawer:
ok I'm gonna close this merge then make a new one with the  creative check in TargetPredicate::test()

PR #1558, opened 2026-02-13
fix(chunk): add fallback when chunk ACK stalls

yunuservices:
This PR adds a safe fallback path for chunk streaming when chunk batch ACKs are delayed or missing.

  ### What changed
  - Added ACK-stall fallback timing in `ChunkManager`.
  - Allowed chunk sending to continue if ACK window is closed for too long (`250ms` fallback delay).
  - Kept normal ACK-based flow control intact.
  - Reset fallback timing on cleanup/world change and after each sent batch.
  - Ensured at least one chunk can be sent per batch decision path (`chunks_per_tick.max(1)`).

  ## Why
  In some real-world cases, chunk visibility can stall at region borders even after prior fixes, likely due to ACK cadence issues.
  This fallback prevents the stream from getting stuck while still remaining rate-limited and conservative.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -348,7 +348,10 @@ impl ChunkManager {
     }
 
     pub fn next_entity(&mut self) -> Box<[SyncEntityChunk]> {
-        let chunk_size = self.entity_chunk_queue.len().min(self.chunks_per_tick);
+        let chunk_size = self
+            .entity_chunk_queue
+            .len()
+            .min(self.chunks_per_tick.max(1));
 
         let chunks: Box<[Arc<ChunkEntityData>]> = self
             .entity_chunk_queue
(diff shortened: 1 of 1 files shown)


Purdze (commented):
Looks good, j consider the count.add_assign(1) in next_chunk could theoretically overflow the u8 if ACKs never arrive. saturating_add(1) would be a safer choice

Purdze:
next_entity() needs to be updated to match the fallback changes in next_chunk()

Crash risk: BatchState::Waiting => unreachable!() will panic — the fallback now allows can_send_chunk() to return true while in Waiting state. Change to BatchState::Waiting => () like next_chunk() does

Missing .max(1): chunks_per_tick should use .max(1) here too, otherwise the fallback can produce empty batches that just burn a batch slot

Missing timestamp update: last_chunk_batch_sent_at isn't updated in next_entity(), so entity-only sends won't reset the fallback timer

PR #1559, opened 2026-02-13
feat(ai): vanilla-accurate player targeting, this also fixes 1 of the 2 bugs listed in bug #1515

TheRealMusawer:
Added logic to prevent targeting Creative-mode players.

I switched it from the old merge which used track_target.rs to the suggested target_predicate.rs :)

### DIFF
--- a/pumpkin/src/entity/ai/target_predicate.rs
+++ b/pumpkin/src/entity/ai/target_predicate.rs
@@ -1,5 +1,5 @@
-use crate::entity::living::LivingEntity;
 use crate::entity::EntityBase;
+use crate::entity::living::LivingEntity;
 use crate::world::World;
 use pumpkin_util::{Difficulty, GameMode};
 use std::future::Future;
@@ -98,7 +98,9 @@ impl TargetPredicate {
         target: &LivingEntity,
     ) -> bool {
         // 1. Equality check: A mob cannot target itself
-        if let Some(t) = tester && t.entity.entity_id == target.entity.entity_id {
+        if let Some(t) = tester
+            && t.entity.entity_id == target.entity.entity_id
+        {
             return false;
         }
 
@@ -139,7 +141,7 @@ impl TargetPredicate {
 
             // TODO: Visibility Modifier Logic (Sneaking = 0.8x, Invisibility = 0.07x)
             // Minecraft logic: max_dist *= target.get_visibility_modifier(tester);
-            
+
             let final_limit = dist_limit * dist_limit;
             let min_limit = MIN_DISTANCE * MIN_DISTANCE;
 


pumpkin/src/entity/ai/target_predicate.rs
```diff
@@ -106,6 +107,14 @@ impl TargetPredicate {
             return false;
         }
 
+        // Prevent targeting Creative‑mode players
+        if let Some(player) = target.entity.get_player() {
+            let gm = player.gamemode.load();
+            if gm == GameMode::Creative {
```
BjornTheProgrammer: The check shouldn't be if they are in creative mode. The check should actually be if they are invulnerable and invisible.

TheRealMusawer: i think for now it should just stay for creative since it was just a patch at zombies comming towards people in creative mode , and if we were doing it like normal minecraft we would have to factor in if they are crouching and how far away they are if they are invisible and even if they have mob head's so i think for now the gamemode thing should be like this.

BjornTheProgrammer: We don't have to implement everything. But I really do recommend at the very least checking for the invulnerability condition. Creative players have invulnerability, and it's easy to check on the abilities of the player.

TheRealMusawer: ok there I updated it.

TheRealMusawer:
updated to rely on target.can_take_damage(), ensuring Creative and Spectator players are excluded through the entity-level invulnerability flag.

BjornTheProgrammer:
I really recommend giving the Minecraft source code a once over on this specific section. If you read this [page](https://docs.pumpkinmc.org/developer/contributing), you'll be able to see more clearly what exactly I mean. Currently although technically correct. It still does deviate slightly from how Minecraft orginally did it.

TheRealMusawer:
ok il go and look at the real Minecraft code and make it like that.

TheRealMusawer:
Summary : switched Target_Predicate to be more in match with vanilla with all the stuff Pumpkin's engine currently has  
Vanilla Gamemode Filtering: Implemented checks for GameMode::Creative and GameMode::Spectator. Hostile AI will now correctly ignore players in these modes.

Difficulty Awareness Integrated Difficulty::Peaceful check within the targeting logic. Hostile entities will now return false for targets when the world is set to peaceful.

Equality Check Refactor:Replaced pointer-based equality with a more better entity_id comparison to ensure an entity cannot target itself.

Performance Optimization  Standardized the use of squared_distance_to_vec for range checks, matching Vanilla’s optimized mathematical approach.

Added todo comments for visibility scaling (sneaking/invisibility) and Line-of-Sight (raycasting) for whenever pumpkin has the stuff needed to implement.

Purdze:
closed as #1575 fixes this

PR #1560, opened 2026-02-13
feat: cobweb

JustMarfix:
## Description

Added cobweb block and its collision code

## Testing

Collision now works for entities, e.g. zombies are slowing down when going through.


JustMarfix:
done

PR #1562, opened 2026-02-14
fix: prevent memory leak when loading void chunks

### LINKED ISSUE #1555: Memory leak when loading void chunks

### Operating System

Arch Linux

Pumpkin 0.1.0-dev+1.21.11 (Commit: ba30b4e/release) - 181 Contributors

### What happened?

Image
Importing regular superflat maps will not cause any problems; only void maps will.

### To Reproduce

Create a void world in vanilla & copy it

### Expected behavior

working

Purdze:
## summary
- fixes #1555
- bound the `recv_chunk` channel in `GenerationSchedule` — was unbounded, now capped at `io_read_thread_count + gen_thread_count + 16`
- bound the global chunk listener channel to 512 with non-blocking `try_send()` — previously every loaded chunk cloned an `Arc<ChunkData>` into each player's unbounded channel, which for void worlds (where io is near-instant) caused thousands of arc references to pile up and prevent chunk unloading (`strong_count > 1`)
- remove empty `ChunkHolder` entries from `chunk_map` during unload — previously holders with `chunk: None` and no active tasks stayed in the map permanently

## note
importing a vanilla void world currently fails at startup with a `level.dat` deserialization error (`invalid type: map, expected a string`) before the chunk system is reached, so the fix could not be fully tested with an imported void world. the chunk system changes themselves are correct and the server compiles + passes clippy cleanly.

### DIFF
--- a/pumpkin-world/src/chunk_system/schedule.rs
+++ b/pumpkin-world/src/chunk_system/schedule.rs
@@ -71,7 +71,8 @@ impl GenerationSchedule {
         listener: Arc<ChunkListener>,
         thread_tracker: &mut Vec<thread::JoinHandle<()>>,
     ) {
-        let (send_chunk, recv_chunk) = crossfire::compat::mpmc::unbounded_blocking();
+        let (send_chunk, recv_chunk) =
+            crossfire::compat::mpmc::bounded_blocking(io_read_thread_count + gen_thread_count + 16);
 
         let (send_read_io, recv_read_io) =
             crossfire::compat::mpmc::bounded_tx_blocking_rx_async(io_read_thread_count + 5);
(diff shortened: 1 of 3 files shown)


pumpkin-world/src/chunk_system/schedule.rs
```diff
@@ -376,6 +377,9 @@ impl GenerationSchedule {
                 let mut tmp = None;
                 swap(&mut holder.chunk, &mut tmp);
                 let Some(tmp) = tmp else {
+                    // Holder has no chunk data and isn't occupied - remove the
+                    // empty entry so chunk_map doesn't grow without bound.
+                    self.chunk_map.remove(&pos);
```
sprinox, line 382: holder without chunk data only happen when it is being used for generation, and data will be returned after the generation task complete.
is it safe to remove it before the chunk data is written back to holder?

Purdze: good catch, i can add a guard

PR #1565, opened 2026-02-14
fix: replace oldest log file when log filename id has reached maximum

### LINKED ISSUE #1484: Can't generate a unique log filename after 100 attempts

### Operating System

Arch Linux

0.1.0-dev+1.21.11 (Commit: cef8a02/debug)

### What happened?

I was testing code changes and did lots of server startups and shutdowns... at some point 99 log files were created in total with the 100th failed to be created:

```rust
thread 'main' (18458) panicked at pumpkin/src/lib.rs:88:22:
Failed to initialize file logger.: "Failed to find a unique log filename for date 2026-02-07 after 100 attempts."
```

Deleting the 99th log file made the server start again

### To Reproduce

Start and stop the server quickly 99 times, the 100th time it'll fail to start

### Expected behavior

If the maximum number of allowed logfiles is reached, delete the oldest logfile

Solitarju:
## Description
Overwrites oldest log file when log filename id has reached maximum (100), instead of panicking.

fixes #1484 

## Testing
**Clippy:** Passed
**Cargo test:** Passed
**Behaviour tested and works as intended**

### DIFF
--- a/pumpkin/src/logging.rs
+++ b/pumpkin/src/logging.rs
@@ -81,7 +81,7 @@ macro_rules! plugin_log {
 }
 
 const LOG_DIR: &str = "logs";
-const MAX_ATTEMPTS: u32 = 100;
+const MAX_ATTEMPTS: u32 = 1000;
 
 /// A wrapper for our logger to hold the terminal input while no input is expected in order to
 /// properly flush logs to the output while they happen instead of batched


Solitarju:
I don't know how I would simplify the tracking with an iterator while keeping it to a single pass.

I've done the other two suggestions though.

Purdze:
Looks good thank you

BjornTheProgrammer:
Why don't we just go over 99? It seems like a bad practice to ever overwrite log files in general.

Solitarju:
> Why don't we just go over 99? It seems like a bad practice to ever overwrite log files in general.

Once MAX_ATTEMPTS is reached, the program will just panic for the rest of the calendar day, or until old logs are manually deleted. Increasing the value only softens this issue and doesn't truly resolve it.

BjornTheProgrammer:
If the max value for log id was a u32, then the max log value would be 4,294,967,295. Even if you started and stopped the server 49,000 times a second, you would never reach the max value. We could still have overwriting as a safety mechanism as well.

Solitarju:
Increasing the max value for log id to u32::MAX would introduce the possibility of creating u32::MAX separate files on disk, and having to check if up to u32::MAX files exist (and read their metadata) every time the application is executed.

Maybe something more moderate like 1000 before overwriting would be sufficient?

BjornTheProgrammer:
I've looked at Paper's code and it seems that they keep 1000 as the max as well before overwritting. Let's do that, and then this should be ready to merge. Thank you for your contribution!

PR #1570, opened 2026-02-14
fix: prevent chunk scheduler cpu spinning from orphaned implicit dependency tasks

### LINKED ISSUE #1541: Constant 1 CPU core load after player join

### Operating System

Windows 11

582d465/release

### What happened?

After player joins the server, Pumpkin starts to constantly load 1 CPU core and this problem persists after player leaving.

### To Reproduce

Open task manager, make sure pumpkin cpu usage is 0%, join server, leave server and take a look at cpu usage again.

### Expected behavior

Well, it definitely should not happen

Purdze:
## Summary
- Fixes #1541 — constant 1 CPU core load after player join (and persisting after disconnect)
- When the scheduler creates implicit dependency tasks for neighbor chunks (e.g., Lighting needs neighbors at Features), it now creates the **full prerequisite chain** (Empty → Biomes → ... → req_stage) instead of just the target stage, so tasks can actually progress through the generation pipeline
- Adds orphan detection: before re-queuing a task that failed runtime checks, checks if it still has valid dependents or is needed by the chunk's own target_stage — drops it if not, preventing infinite re-queue spinning

### DIFF
--- a/pumpkin-world/src/chunk_system/schedule.rs
+++ b/pumpkin-world/src/chunk_system/schedule.rs
@@ -63,6 +63,41 @@ pub struct GenerationSchedule {
 }
 
 impl GenerationSchedule {
+    /// Returns `true` if the task is orphaned: no longer needed by the chunk
+    /// itself AND no valid dependent tasks remain in the graph.
+    fn is_orphaned_task(&self, node_key: NodeKey, pos: ChunkPos, stage: StagedChunkEnum) -> bool {
+        // If the chunk's own target_stage still requires this task, keep it.
+        if let Some(holder) = self.chunk_map.get(&pos)
+            && holder.target_stage >= stage
+        {
+            return false;
+        }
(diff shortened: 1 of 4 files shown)


sprinox:
I don’t think this PR addresses the real issue. The lighting stage introduced in 8aeeacf wasn’t fully integrated across the entire codebase (partly my fault that I didn’t add enough comments, which left quite a few implicit rules for later contributors). Because lighting depends on neighboring features, the dependency radius effectively expanded by one chunk compared to before, so level_to_stage, FULL_RADIUS, and FULL_CHUNK_LEVEL all need to be updated accordingly.

Also, the “ready”-related logic added around L833–L875 is very strange. A task is never supposed to be enqueued unless its conditions are already satisfied (that was the original design), so this extra logic is completely unnecessary. It’s very likely that this is what’s causing the infinite loop.

If a new external mechanism is required to maintain the DAG, then the existing DAG logic is probably already in a messy state.

PR #1573, opened 2026-02-15
fix: resolve constant cpu core load after player join

### LINKED ISSUE #1541: Constant 1 CPU core load after player join

Purdze:
## Summary
Fixes #1541

- insert lighting into `level_to_stage` between full and features
- bump `FULL_RADIUS` from 2 to 3 to cover the full dependency chain
- bump `MAX_LEVEL` from 46 to 47 to accommodate the extra level
- remove redundant ready/missing_chunk requeue checks that caused the cpu spin (the dag already guarantees these invariants)
- add missing `send_change()` calls in `clean_up()` and `change_world()` so the scheduler is notified when chunk tickets are removed on player disconnect or world change


sprinox:
send_change() is actually not missing. Many changes can be merge and be sent in one send_change(). So change can be sent when other place call send_change(). I try to call send_change() as few as possible so that improve a bit perf. Definitely adding extra send_change() isn't incorrect.

Purdze:
> Contributor

Thank you

Purdze:
@BjornTheProgrammer take a look, not the proper fix we need but fixes the cpu core issue for now

Issue #264, opened 2024-11-13
Damage is not dealt by Blocks / Bushes


Rotstein007:
### I've searched existing issues and couldn't find a duplicate.

- [X] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

If you step on/in a block, that should deal damage, it doesn't.
![image](https://github.com/user-attachments/assets/d13c83f5-4c3b-4f3c-b6fc-51aa5c51022d)

### Describe the solution you'd like

Blue fire should deal more damage than "normal" fire. Wither Rose give an additional effect.

PR #1576, opened 2026-02-15
fix: match campfires with vanilla logic

GreenedDev:
## Description
Partially fixes https://github.com/Pumpkin-MC/Pumpkin/issues/264
fix: soul campfire deals 2 damage per half second instead of 1
fix: campfires dont deal damage if entity is wearing boots with frost walker or has fire resistance effect

Sources:
https://minecraft.wiki/w/Soul_Campfire#Damage
https://minecraft.wiki/w/Campfire#Damage

### DIFF
--- a/pumpkin/src/block/blocks/campfire.rs
+++ b/pumpkin/src/block/blocks/campfire.rs
@@ -44,7 +44,7 @@ impl BlockBehaviour for CampfireBlock {
                     .await
                     .is_some();
                 if has_frost_walker_enchantment || has_fire_res {
-                    //campfire burning doesn't work if entity's boots has frost walker enchantement or entity has fire res. source: https://minecraft.wiki/w/Campfire#Damage
+                    //campfire burning doesn't work if entity's boots has frost walker enchantment or entity has fire resistance. source: https://minecraft.wiki/w/Campfire#Damage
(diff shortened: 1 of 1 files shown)


Purdze (commented):
Use `if let` chains to avoid calling `get_living_entity()` twice (matches the style in cobweb block), and simplify the damage amount

if CampfireLikeProperties::from_state_id(args.state.id, args.block).lit
    && let Some(living_entity) = args.entity.get_living_entity()
{
    let has_frost_walker_enchantment = {
        let equipment = living_entity.entity_equipment.lock().await;
        let boots = equipment.get(&EquipmentSlot::FEET);
        let boots_stack = boots.lock().await;
        boots_stack.get_enchantment_level(&Enchantment::FROST_WALKER) != 0
    };
    let has_fire_res = living_entity
        .get_effect(&StatusEffect::FIRE_RESISTANCE)
        .await
        .is_some();
    if has_frost_walker_enchantment || has_fire_res {
        return;
    }
    let damage_amount = if args.block == &Block::SOUL_CAMPFIRE { 2.0 } else { 1.0 };
    args.entity
        .damage(args.entity, damage_amount, DamageType::CAMPFIRE)
        .await;
}

Issue #1274, opened 2026-01-12
Elytra Bug , Fire work 2&3 rocket crash the server , wind charge Not working ,Potions crash the server more idk


imon975:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 10

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.11

### What happened?

When i obtain some item like potion or firework 2&3 rocket server crash auto, elytra gliding bugged and stucked at gliding position u need to click shift to exit , wind charge are basicly useless

<img width="721" height="394" alt="Image" src="https://github.com/user-attachments/assets/de74cfbc-495a-41ce-b65d-508c351e52ea" />

left clicking on a spear kick the player , two item wont swap with the hotkeys

Nether is buggy mobs spawn on bedrock roof [Overworld mobs :] ]

<img width="643" height="337" alt="Image" src="https://github.com/user-attachments/assets/28421c4a-9e9d-4919-bbf3-c5d8df7de0f1" />

### To Reproduce

Glide with Elytra , take potion or firework 2or3 from Creative inv, left click with spear , enter the nether , throw wind charge

### Expected behavior

You guys doing great , keep it up , fix bugs im realy intrested to seeing how good and fast a minecraft server software Do in future.

PR #1582, opened 2026-02-16
fix: inventory saving/loading

RoosterBooster007:
## Description
- Container blocks like chests and furnaces now save and load items properly
- Placing potions in a chest will no longer cause a crash
- Fireworks level 2 and 3 no longer stack together (and can be kept in containers)

## Testing
Yes.

#1274


Purdze:
Furnace set_stack now calls mark_dirty() twice (the old conditional one + the new unconditional one), the first is redundant

Verify that barrel, dropper, shulker_box, and chiseled_bookshelf have clear_dirty overrides, otherwise their dirty flags never reset and the chunk re-saves every cycle

PR #1584, opened 2026-02-16
fix: item drop velocity

chocodev11:
## Description
Implemented vanilla landing motion handling so item bounce is driven by shared block physics, not item-specific hacks
- Added new block hook update_entity_movement_after_fall_on with default Y=0 behavior

- Added registry dispatcher for that hook

- Called the hook from common entity movement after vertical collision

- Added special bounce overrides for slime and bed

## Testing
yes

### DIFF
--- a/pumpkin/src/entity/item.rs
+++ b/pumpkin/src/entity/item.rs
@@ -275,7 +275,7 @@ impl ItemEntity {
         velo = velo.multiply(friction, 0.98, friction);
 
         if on_ground && velo.y < 0.0 {
-            velo = velo.multiply(1.0, -0.5, 1.0);
+            velo.y = 0.0;
         }
 
         entity.velocity.store(velo);


Purdze:
Loods good thank you

Couple of smaller things 

Bed and slime update_entity_movement_after_fall_on are nearly identical, only the bounce multiplier differs (0.66 vs 1.0). Could extract a shared bounce helper to avoid duplication

world in UpdateEntityMovementAfterFallOnArgs is unused by all three implementations (default, bed, slime). If you dont need it, we can just remove it

chocodev11:
> Loods good thank you
> 
> Couple of smaller things
> 
> Bed and slime update_entity_movement_after_fall_on are nearly identical, only the bounce multiplier differs (0.66 vs 1.0). Could extract a shared bounce helper to avoid duplication
> 
> world in UpdateEntityMovementAfterFallOnArgs is unused by all three implementations (default, bed, slime). If you dont need it, we can just remove it

the items are somehow still bouncy and glitchy when touching ground

PR #1589, opened 2026-02-17
feat: add igloo structure generation

mtthidoteu:
## Summary

Implements igloo generation for snowy biomes:

- Snow block dome with interior furnishings (bed, furnace, crafting table, redstone torch)
- Optional basement (50% chance) with ladder shaft (4-11 segments) and secret room
- Biome validation (snowy_taiga, snowy_plains, snowy_slopes)
- Terrain height adjustment via ShiftableStructurePiece
- Deterministic placement matching vanilla random call order

## Future work (requires infrastructure)

- Chest loot table
- Entity spawning (villager, zombie villager)

## Test plan

- [x] Build passes
- [x] Clippy passes
- [x] Format check passes
- [x] Test igloo generation in snowy biomes with known seed

**_Edit: I followed the swamp hut example. However, instead of hardcoding invidiual blocks. Why don't we load the same NBT files as Minecraft does? And then we can implement structures a lot easier_**

### DIFF
--- a/pumpkin-world/src/generation/structure/mod.rs
+++ b/pumpkin-world/src/generation/structure/mod.rs
@@ -9,7 +9,7 @@ use crate::{
         biome_coords,
         structure::structures::{
             StructureGenerator, StructureGeneratorContext, StructurePosition,
-            buried_treasure::BuriedTreasureGenerator, create_chunk_random,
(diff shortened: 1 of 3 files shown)


mtthidoteu:
Without the /locate command it was quite challenging to find a snow biome. However I found a seed and it generated correctly:
image
I will say I am not sure why but for some reason no snow is appearing on the island. Despite it being a snowy taiga biome

PR #1590, opened 2026-02-17
feat: wither rose collision

GreenedDev:
## Description
Add wither rose collision code.
Colliding with wither rose gives 2 seconds of wither effect and gives 1 damage per half second


Purdze (commented):
amplifier should be 0
remove the direct 1.0 wither damage, vanilla applies the effect, not damage
need a peacefull check
need wither immune entity check e.g. wither skeletons/boss

PR #1597, opened 2026-02-18
feat(event): add ArmorStand manipulate event

yunuservices:
This PR fixes missing plugin control for armor stand manipulation by adding a cancellable event hook.


yunuservices:
Addressed in df97f63b:
- armor_stand_item_key now reads from the actual armor stand equipment slot.
- Added typed slot fields (EquipmentSlot) for player hand slot and target armor stand slot.
- Removed raw-string slot usage from the event payload.

yunuservices:

PR #1598, opened 2026-02-18
feat(event): add bed enter/leave events

yunuservices:
This PR fixes the lack of lifecycle hooks for bed usage by adding enter and leave events.


yunuservices:

PR #1599, opened 2026-02-18
feat(event): add BlockCanBuild and BlockBurn events

yunuservices:
This PR fixes missing block placement validation and burn hooks by exposing both flows as events.


pumpkin/src/plugin/api/events/block/block_place.rs
```diff
@@ -26,6 +26,9 @@ pub struct BlockPlaceEvent {
     /// The position where the block is being placed.
     pub block_position: BlockPos,
 
+    /// Alias of `block_position` for newer call-sites.
+    pub position: BlockPos,
```
RoosterBooster007, line 30: Ideally, we should just use `block_pos` to match other events and keep things simple.

pumpkin/src/block/blocks/fire/fire.rs
```diff
@@ -167,10 +168,36 @@ impl FireBlock {
...
+                        block,
+                        block_pos: *pos,
+                        world_uuid: world.uuid,
+                        cancelled: false,
+                    };
+                    let event = server.plugin_manager.fire::<BlockBurnEvent>(event).await;
+                    if event.cancelled {
+                        return;
+                    }
+                }
```
RoosterBooster007, line 200: Looks good, but maybe make a util fn to simplify the event firing.

pumpkin/src/net/java/play.rs
```diff
@@ -2202,6 +2203,39 @@ impl JavaClient {
...
+                position: final_block_pos,
+                can_build: true,
+                cancelled: false,
+            };
+
+            let event = server.plugin_manager.fire::<BlockPlaceEvent>(event).await;
+            if event.cancelled || !event.can_build {
+                return Ok(false);
+            }
+        }
```
RoosterBooster007, line 2237: If these are cancelled, doesn't the client need some kind of block update?

yunuservices:

PR #1600, opened 2026-02-18
feat(event): add block damage events and hooks

yunuservices:
This PR fixes limited observability of block damage progression by adding dedicated damage events.

### DIFF
--- a/pumpkin/src/net/java/play.rs
+++ b/pumpkin/src/net/java/play.rs
@@ -1490,13 +1490,12 @@ impl JavaClient {
                         let (block, _state) =
                             world.get_block_and_state(&player_action.position).await;
                         let item_stack = player.inventory.held_item().lock().await.clone();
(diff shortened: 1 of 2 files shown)


pumpkin/src/plugin/api/events/block/damage_abort.rs
```diff
@@ -0,0 +1,51 @@
...
+use pumpkin_util::math::position::BlockPos;
+use pumpkin_world::item::ItemStack;
+use std::sync::Arc;
+
+use crate::entity::player::Player;
+
+use super::BlockEvent;
+
+/// An event that occurs when a player stops damaging a block.
+#[cancellable]
```
RoosterBooster007, line 12: Should the `BlockDamageAbortEvent` be cancellable? This seems like more of an informational event.

yunuservices:

PR #1601, opened 2026-02-18
feat(event): add BlockDispenseEvent for droppers

yunuservices:
This PR fixes the inability to intercept dropper dispense actions by firing a dispense event.

### DIFF
--- a/pumpkin/src/plugin/api/events/block/mod.rs
+++ b/pumpkin/src/plugin/api/events/block/mod.rs
@@ -1,8 +1,8 @@
 pub mod block_break;
 pub mod block_burn;
 pub mod block_can_build;
-pub mod dispense;
 pub mod block_place;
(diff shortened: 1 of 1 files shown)


Purdze (changes requested):
The unsafe pointer cast to get &'static Block is unnecessary, Block::from_id(args.block.id) returns &'static Block safely

Instead of:
```
let block_static: &'static pumpkin_data::Block =
    unsafe { &*std::ptr::from_ref::<pumpkin_data::Block>(args.block) };
```

Just use:
`let block_static = Block::from_id(args.block.id);`

yunuservices:
Addressed in 27dc1eed:
- Removed the unsafe pointer cast for block static reference.
- Switched to Block::from_id(args.block.id) as suggested.

yunuservices:

PR #1602, opened 2026-02-18
feat(event): add BlockDropItemEvent for block loot

yunuservices:
This PR fixes missing control over block loot drops by emitting a block drop item event.


yunuservices:
Addressed in d7d120ea:
- BlockDropItemEvent is now fired only when loot stacks are non-empty.
- If event is cancelled, item drops are suppressed but XP drop still proceeds.

yunuservices:

PR #1634, opened 2026-02-19
feat(event): add blockfadeevent for fire burnout

yunuservices:
## Title
This PR adds missing "BlockFadeEvent for fire burnout" to the pumpkinmc.

## Changes
- Adds `BlockFadeEvent` emission in fire burnout transitions.
- Respects cancellation before turning fire blocks into air.

## Notes


yunuservices:

PR #1639, opened 2026-02-19
feat(event): add block ignite event

yunuservices:
## Title
This PR adds missing "BlockIgniteEvent" to the pumpkinmc.

## Changes
- Adds ignition hook via `BlockIgniteEvent`.
- Applies cancellation-aware fire start behavior.

## Notes

### DIFF
--- a/pumpkin/src/plugin/api/events/block/mod.rs
+++ b/pumpkin/src/plugin/api/events/block/mod.rs
@@ -1,8 +1,8 @@
 pub mod block_break;
 pub mod block_burn;
 pub mod block_can_build;
-pub mod ignite;
 pub mod block_place;
+pub mod ignite;
 
 use pumpkin_data::Block;
 


pumpkin/src/item/items/ignite/ignition.rs
```diff
@@ -16,6 +16,7 @@ impl Ignition {
         location: BlockPos,
         face: BlockDirection,
         block: &Block,
+        cause: &str,
```
RoosterBooster007, line 19: Maybe make this an `IgniteCause` enum instead of using strings?

pumpkin/src/plugin/api/events/block/ignite.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::entity::player::Player;
+
+use super::BlockEvent;
+
+/// An event that occurs when a block is ignited.
+#[cancellable]
+#[derive(Event, Clone)]
+pub struct BlockIgniteEvent {
+    /// The player who ignited the block.
+    pub player: Arc<Player>,
```
RoosterBooster007, line 15: Other blocks like lava can cause fire, so maybe use an `Option<Arc<Player>>` and correct the ignite logic?

yunuservices:

PR #1642, opened 2026-02-19
feat(event): add block spread event

yunuservices:
## Title
This PR adds missing "BlockSpreadEvent" to the pumpkinmc.

## Changes
- Adds `BlockSpreadEvent` in spread/fire propagation logic.
- Destination spread application is gated by event result.

## Notes


yunuservices:

PR #1646, opened 2026-02-19
feat(event): add piston extend retract events

yunuservices:
## Title
This PR adds missing "piston extend/retract events" to the pumpkinmc.

## Changes
- Adds piston extend event dispatch.
- Adds piston retract event dispatch with cancellation support.

## Notes


yunuservices:

PR #1647, opened 2026-02-19
feat(event): add playerchangedmainhandevent

yunuservices:
## Title
This PR adds missing "PlayerChangedMainHandEvent" to the pumpkinmc.

## Changes
- Adds event emission when player main-hand preference changes.
- Exposes transition context to plugin layer.

## Notes
- Scope is limited to this event branch.

PR #1648, opened 2026-02-19
feat(event): add player channel events

yunuservices:
## Title
This PR adds missing "player channel events" to the pumpkinmc.

## Changes
- Adds channel register/unregister event hooks.
- Routes channel packet handling through event dispatch.

## Notes


yunuservices:

PR #1649, opened 2026-02-19
feat(event): add player drop item event

yunuservices:
## Title
This PR adds missing "PlayerDropItemEvent" to the pumpkinmc.

## Changes
- Adds event dispatch before final item drop spawn.
- Drop action now supports cancellation and payload changes.

## Notes


yunuservices:

PR #1651, opened 2026-02-19
feat(event): add playereggthrowevent

yunuservices:
## Title
This PR adds missing "PlayerEggThrowEvent" to the pumpkinmc.

## Changes
- Adds event in egg throw handling lifecycle.
- Throw behavior is now plugin-observable and cancellable.

## Notes


yunuservices:

PR #1654, opened 2026-02-19
feat(event): add playerfishevent

yunuservices:
## Title
This PR adds missing "PlayerFishEvent" to the pumpkinmc.

## Changes
- Adds fishing lifecycle event hooks.
- Cast/catch/reel behavior now dispatches plugin events.

## Notes
- Scope is limited to this event branch.

PR #1653, opened 2026-02-19
feat(event): add playerexpchangeevent

yunuservices:
## Title
This PR adds missing "PlayerExpChangeEvent" to the pumpkinmc.

## Changes
- Adds event around XP change application path.
- XP delta flow now supports cancellation and override.

## Notes
- Scope is limited to this event branch.

PR #1656, opened 2026-02-19
feat(event): add player item damage consume mend events

yunuservices:
## Title
This PR adds missing "player item damage/consume/mend events" to the pumpkinmc.

## Changes
- Adds hooks for item damage, consume, and mend pipelines.
- Item state mutations now pass through event-driven control.

## Notes


yunuservices:

PR #1657, opened 2026-02-19
feat(event): add playeritemheldevent

yunuservices:
## Title
This PR adds missing "PlayerItemHeldEvent" to the pumpkinmc.

## Changes
- Adds event on selected held-item slot changes.
- Slot transition is exposed before final sync/update.

## Notes


yunuservices:

PR #1658, opened 2026-02-19
feat(event): add player kick and level change events

yunuservices:
## Title
This PR adds missing "player kick and level change events" to the pumpkinmc.

## Changes
- Adds event hooks for player kick flow.
- Adds event hooks for level/world transition updates.

## Notes


yunuservices:

PR #1659, opened 2026-02-19
feat(event): add player resource pack status and respawn events

yunuservices:
## Title
This PR adds missing "player resource-pack status and respawn events" to the pumpkinmc.

## Changes
- Adds event coverage for resource-pack status callbacks.
- Adds respawn flow event dispatch for plugin interception.

## Notes


yunuservices:

PR #1660, opened 2026-02-19
feat(event): add player toggle and swap hand events

yunuservices:
## Title
This PR adds missing "player toggle and swap-hand events" to the pumpkinmc.

## Changes
- Adds toggle-state event hooks.
- Adds swap-hand event dispatch in hand-switch path.

## Notes


yunuservices:

PR #1661, opened 2026-02-19
feat(event): add plugin and service server events

yunuservices:
## Title
This PR adds missing "plugin and service server lifecycle events" to the pumpkinmc.

## Changes
- Adds lifecycle event coverage for plugin/service server operations.
- Routes lifecycle transitions through server event dispatch.

## Notes


yunuservices:

PR #1665, opened 2026-02-19
feat(event): add spawnchangeevent

yunuservices:
## Title
This PR adds missing "SpawnChangeEvent" to the pumpkinmc.

## Changes
- Adds event dispatch when spawn point changes.
- Spawn update flow now supports plugin cancellation/interception.

## Notes


yunuservices:
Batch queue reduction: keeping only #1666, #1654, #1653, #1647 for faster merge throughput. Will re-open/continue next batch after these land.

PR #1666, opened 2026-02-19
fix(event): avoid duplicate BlockBreakEvent and honor drop flag

yunuservices:
## Title
This PR adds missing "BlockBreakEvent duplicate-guard handling" to the pumpkinmc.

## Changes
- Prevents duplicate `BlockBreakEvent` dispatch scenarios.
- Ensures drop-flag behavior is honored consistently in break flow.

## Notes


Purdze:
PR title mentions "avoid duplicate BlockBreakEvent" but the code only addresses honoring the drop flag, there's no deduplication logic?

RoosterBooster007:
Thanks, @yunuservices! 🏗️

PR #1667, opened 2026-02-19
feat(event): fire BlockBreakEvent on player mining

yunuservices:
## Title
This PR adds missing "BlockBreakEvent dispatch on player mining" to the pumpkinmc.

## Changes
- Adds/fixes `BlockBreakEvent` dispatch in player mining path.
- Aligns mining break timing with expected event semantics.

## Notes


yunuservices:

PR #1669, opened 2026-02-19
tps command

Purdze:
## Summary
- Adds `/tps` command that displays server TPS (capped at 20.0) and MSPT
- Color-coded output: green (18+), yellow (15-18), red (<15)
- Requires OP level 2 permission (`pumpkin:command.tps`)


Q2297045667:
I think it's better to use the configuration in the config file as the upper limit.
 
image

Snowiiii:
LGTM. Thank you @Purdze :D

Issue #662, opened 2025-03-24
Switch to WASM for Plugins


Snowiiii:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

Currently we are using native-libraries for Plugins. They are really fast but have some problems.
- Nothing is sandboxed, Which is bad since we can't just trust every plugin
- Not cross-platform
- If a plugins crashes, it crashes the server
- Not really a problem, but its awesome that you can use many different programming languages
- We can easily validate if a WASM file is valid, which is super useful for the Marketplace and also in general for the server

### Describe the solution you'd like

We should consider using WebAssembly, We already tried to use Extism in the past but there were [certain problems](https://github.com/Pumpkin-MC/Pumpkin/issues/110) with that. Now im looking to may use [wasmtime](https://github.com/bytecodealliance/wasmtime) which is more lightweight.

4lve:
Ye, wasm looks like a great choice

suprohub:
Its cool, but there are caveats:
- The ability to generate random numbers, spawn threads, and check the time may introduce counterproductive nondeterminism.
Bruh just think: fast and ONLY asynx world edit, it kinda susck.

- Filesystem APIs, Socket APIs, etc. may allow access outside the sandbox if using an off the shelf default implementation of WASI, which may be a problem if your plugins are user generated content and thus potentially malicious.  Of course, you can always create your own WASI implementation, and limit access to your own virtualized filesystem / explicitly whitelisted network connections etc.

- Don't ask me what the state of the Wasm GC proposals are, and if they're actually able to support the more scripty style languages many prefer for game modding. 


Anyway.
If a mod can’t crash your OS is it really a mod?
Think about it :)

suprohub:
Will be cool support its as offical plugin. Not every server owner wants wasm, and it xan just waste binary. And it have really more caveats, but features too.

suprohub:
> Will native plugins still be supported? Personally I believe using a non sandboxed environment, and having access to native libraries can be useful when writing custom, server-specific plugins that may need to integrate with other systems. 

I think yes. Will be weird remove native support. Su better name is add wasm support, not switch.

Snowiiii:
> Will native plugins still be supported? Personally I believe using a non sandboxed environment, and having access to native libraries can be useful when writing custom, server-specific plugins that may need to integrate with other systems.
> 
[...]

PR #1675, opened 2026-02-20
feat: add wasm plugin api

BjornTheProgrammer:
## Description

Adds a wasm plugin interface to Pumpkin

## Testing

In the works

### DIFF
--- a/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
+++ b/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
@@ -24,7 +24,7 @@ pub enum PluginInitError {
     FailedToReadPayload(#[from] wasmparser::BinaryReaderError),
     #[error("failed to read plugin bytes")]
     FailedToReadPluginBytes(#[from] std::io::Error),
-    #[error("plugin failed to load")]
+    #[error("plugin failed to load with error: {0}")]
     PluginFailedToLoad(#[from] wasmtime::Error),
 }
 


pumpkin/src/plugin/loader/mod.rs
```diff
@@ -1,18 +1,16 @@
...
 pub type PluginLoadFuture<'a> = Pin<
     Box<
         dyn Future<
                 Output = Result<
-                    (
-                        Box<dyn Plugin>,
-                        PluginMetadata<'static>,
-                        Box<dyn Any + Send + Sync>,
-                    ),
+                    (Box<dyn Plugin>, PluginMetadata, Box<dyn Any + Send + Sync>),
```
vyPal, line 12: Since the new `PluginMetadata` struct uses `String` and `Vec` types which are allocated during runtime and can't be made static, you will also have to update the `pumpkin-api-macros` crate to expose the metadata some other way, and update the native loader to load them in that other way. Currently the native API exposes the `PluginMetadata` like this:
```rs
#[unsafe(no_mangle)]
pub static METADATA: pumpkin::plugin::PluginMetadata = pumpkin::plugin::PluginMetadata {...};
```
which will no longer be possible after this change

Q2297045667:
link 

https://github.com/Pumpkin-MC/Pumpkin/issues/662

Cargo.toml
```diff
@@ -171,3 +172,11 @@ arc-swap = "1.8"
 tokio-util = "0.7.18"
 toml = "1.0"
 ureq = "3.2.0"
+
+wasmtime = "41.0"
```
Snowiiii: -> "42.0"

Snowiiii:
LGTM. Awesome work. Thank you to everyone who worked on this :D

PR #1681, opened 2026-02-20
feat: implement new command system to achieve complex vanilla commands

Laptop59:
## Description

Adds the rest of the fundamentals like how *Brigadier* does it which allows us to create a new system revolving around one big tree.
This is separate from the current parsing system located in `pumpkin::commands::tree`.
> [!NOTE]
> I did not link the system's dispatcher to the server yet, but that can be done later in another PR, as I would have to find a  way to fallback command input properly between different systems. Maybe then commands and argument types can be reimplemented slowly in different PRs with respect to this system.  

### Things added
- A new `ArgumentType<T>` which parses a particular type `T`, and they all implement `AnyArgumentType` which is its dynamic version. Parsing of arguments can then be done solely on command parsing, and NOT on command execution.
- Core implementations of `ArgumentType` have also been added (from *Brigadier*)
- A new `CommandSender::Dummy` used in dummy `CommandSource`s (for testing purposes only) and maybe for running functions in datapacks soon.
- A `Tree`: the single tree which will contain all command nodes and their children. 
- Added nodes in two different forms:
  - `AttachedNode`: Represents a node already attached to a Tree. Input is parsed through these nodes.
  - `DetachedNode`: Represents a node which is not attached to a Tree yet. It can still be freely modified until it is moved to the `Tree`.
- There are four types of nodes:
  - Root: Does not have a parent. Exactly one instance of this type of node exists per `Tree`. Always identifiable by `ROOT_NODE_ID`. **Does not have its own `AttachedNode` variant**.
  - Command: Its parent must be the root node, and specifies the start of a command definition.
  - Literal: Accepts a particular constant word.
  - Argument: Parses and accepts a specific type of value. This is very dynamic.
  - Builders to create the new nodes have also been added.
- Identification of nodes:
  - `GlobalNodeId`: Every node, at runtime, has a unique identifier which is established. No two nodes can have the same `GlobalNodeId`, even if identical. 
  - `NodeId`: The unique ID of a node that is established when it is attached to a tree. This ID has different variants for different types of nodes.
- New kinds of executors, like `CommandExecutor` which takes a reference of a `CommandContext`. This has more than just the sender, even other details are stored in it. Commands like `/execute` when implemented can then hopefully use modifiers to 'modify' the `CommandSource` using the node builders.
- Structs for redirects and requirements.
- New `Suggestion`, `Suggestions`, and `SuggestionsBuilder` derived from *Brigadier*.
- A builder `CommandContextBuilder` that can create a `CommandContext`.
- A new structure `ContextChain` that chains together a list of `CommandContext`s that can own an `Arc` to the next.
- A new `CommandDispatcher` which acts very similarly to the *Brigadier*'s `CommandDispatcher`. This contains its `Tree`.

### Advantages over the current system
- A node is able to redirect to any other node, whether it be the root node or not. This makes commands like `/execute` feasible to recreate. This is because the current system has a tree for every non-alias command, while this system has one big tree, which allows easy redirection.
- The system of `CommandContext`s and their chaining is very similar to that of *Brigadier*, and so is the parsing & error messages.
- It is possible to add custom requirements and permission for each node, and not just per command. This allows us to have separate subcommands in a command having different permissions, used by `/random`.
- Argument types can be parsed better just like *Brigadier* using `StringReader` implemented before this PR. More complex types can be parsed easily.
- It is possible to show a context of where a command error occured.
- Recursion is possible via redirects.
- Forking is possible, which is used by the `/execute` command.
- Aliases are still possible by using redirection/sharing executors.

### Drawbacks
- For now, there is no way to unregister commands.

## Testing
Added a whole bunch of tests for the system.

### DIFF
--- a/pumpkin/src/command/context/command_context.rs
+++ b/pumpkin/src/command/context/command_context.rs
@@ -134,7 +134,8 @@ impl CommandContext<'_> {
     /// A simple example that takes two arguments specified from the node
     /// and returns their sum as the status output of the `Executor`:
     /// ```
-    /// use pumpkin::command::node::dispatcher::Executor;
+    /// use pumpkin::command::context::command_context::CommandContext;
+    /// use pumpkin::command::node::{CommandExecutor, CommandExecutorResult};
     ///
     /// struct Executor;
     /// impl CommandExecutor for Executor {


Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #1683, opened 2026-02-20
Fix block-break flickering again 

BraveDevelopment:
## Description

A fix for your fix. It now fixes the problem again.

## Testing

Tested locally and it works as intended.


Snowiiii:
Great, Thanks :D @BraveDevelopment

PR #1684, opened 2026-02-20
fix: visual daylight cycle (send dimension/timeline tag data)

RoosterBooster007:
## Description
- The proper dimension/timeline tag data is now sent on join so that each client's visual daylight cycle works as expected

## Testing
- The sun/moon and clouds interpolate correctly instead of lagging back


Snowiiii:
Works, Thank you @RoosterBooster007 :D

Issue #1619, opened 2026-02-19
Armor Stands don't have gravity


FireAcidCow:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

8bd5009

### What happened?

Armor stands don't fall when not supported by a block

### To Reproduce

1. Place an armor stand
2. Break the block beneath

### Expected behavior

The armor stand should fall like normal

FireAcidCow:
fixed by ca97b28

PR #1689, opened 2026-02-21
fix: make armor stands fall

Laptop59:
## Description

Implements `get_gravity` in `EntityBase` for `ArmorStandEntity` to make the armor stand fall instead of staying midair.
Fixes issue #1619.

## Testing

As expected, when testing in-game, the armor stand falls when a block is broken below it.

### DIFF
--- a/pumpkin/src/entity/decoration/armor_stand.rs
+++ b/pumpkin/src/entity/decoration/armor_stand.rs
@@ -437,7 +437,7 @@ impl EntityBase for ArmorStandEntity {
     }
 
     fn get_gravity(&self) -> f64 {
-        0.04
+        0.08
     }
 }
 


pumpkin/src/entity/decoration/armor_stand.rs
```diff
@@ -435,6 +435,10 @@ impl EntityBase for ArmorStandEntity {
             true
         })
     }
+
+    fn get_gravity(&self) -> f64 {
+        0.04
```
Snowiiii: it looks like vanilla uses the default attribute which is `0.08` ?

Laptop59: Yes, you're right. i also used the `/attribute` command in vanilla on an armor stand and got 0.08.

Snowiiii:
Thanks @Laptop59 :D

PR #1690, opened 2026-02-22
fix: use command sender dimension when tp

### LINKED ISSUE #1678: /tp Moves player to the overworld

### Operating System

Arch Linux

33cb7e5

### What happened?

When you run the /tp command with angle arguments, you get moved to the overworld. 

### To Reproduce

1. Go to the nether (beware of #1673
2. Run a teleport command with angle arguments (such as `/tp @s 0 90 0 -164.36 24.45`)
3. Observe that you are now in the overworld

### Expected behavior

/tp should always preserve the dimension you are in. Note that this doesn't occur when you don't have the angle arguments.

chocodev11:
Fixes #1678 

## Testing

### DIFF
--- a/pumpkin/src/command/commands/teleport.rs
+++ b/pumpkin/src/command/commands/teleport.rs
@@ -74,7 +74,11 @@ impl CommandExecutor for EntitiesToEntityExecutor {
             let targets = EntitiesArgumentConsumer::find_arg(args, ARG_TARGETS)?;
 
             let destination = EntityArgumentConsumer::find_arg(args, ARG_DESTINATION)?;
-            let pos = destination.get_entity().pos.load();
+            let destination = destination.get_entity();
+            let pos = destination.pos.load();
+            let yaw = destination.yaw.load();
+            let pitch = destination.pitch.load();
+            let world = destination.world.load_full();
             if !World::is_valid(BlockPos(pos.floor_to_i32())) {
                 return Err(CommandError::CommandFailed(TextComponent::translate(
                     "commands.teleport.invalidPosition",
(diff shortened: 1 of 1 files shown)


topi-banana (approved):

pumpkin/src/command/commands/teleport.rs
```diff
@@ -193,12 +193,11 @@ impl CommandExecutor for EntitiesToPosWithRotationExecutor {
             // For teleport, we use absolute values only (ignore relative flags)
             let (yaw, _, pitch, _) = RotationArgumentConsumer::find_arg(args, ARG_ROTATION)?;
 
-            // todo command context
-            let world = server.worlds.load().first().unwrap().clone();
             for target in targets {
+                let world = target.get_entity().world.load_full();
```
RoosterBooster007: This appears to work, but it breaks teleportations under certain circumstances. If a player in the nether tries to teleport a player (who's in the overworld), they will be teleported in the overworld and not to the player in the nether. Basically, the dimension the command was executed from matters.

Try something like this instead:
```rust
// Resolve world based on command sender context
let world = match sender {
    CommandSender::Rcon(_) | CommandSender::Console | CommandSender::Dummy => {
        server.worlds.load().first().unwrap().clone()
    }
    CommandSender::Player(player) => player.world().clone(),
    CommandSender::CommandBlock(_, w) => w.clone(),
};
```

RoosterBooster007 (approved):

pumpkin/src/command/commands/teleport.rs
```diff
@@ -109,14 +118,7 @@ impl CommandExecutor for EntitiesToPosFacingPosExecutor {
...
             let (yaw, pitch) = yaw_pitch_facing_position(&pos, &facing_pos);
-            //todo
-            let world = match sender {
-                CommandSender::Rcon(_) | CommandSender::Console | CommandSender::Dummy => {
-                    server.worlds.load().first().unwrap().clone()
-                }
-                CommandSender::Player(player) => player.world().clone(),
-                CommandSender::CommandBlock(_, w) => w.clone(),
-            };
+            let world = resolve_sender_world(sender, server);
```
RoosterBooster007, line 121: When teleporting to another entity/player, can you actually have it set the pos/world to the destination entity's? That way, teleportation works like in vanilla.

RoosterBooster007:
Thanks, @chocodev11! 📍

PR #1692, opened 2026-02-22
Finalize wither-rose implementation

GreenedDev:
## Description
Moved wither_rose.rs to plant/ folder
fix: wither roses can't be placed on any block.
fix: when block under wither rose breaks now the rose breaks too.


Snowiiii:

PR #1696, opened 2026-02-23
Fix bed obstruction check and guard counterpart break

gula00:
## Description
  - Fix bed obstruction validation in `BedBlock::normal_use` by checking both `bed_head_pos.up()` and
  `bed_foot_pos.up()`.
  - Add a counterpart existence/type guard in `BedBlock::broken` before breaking the other half.
  - This avoids false-negative obstruction checks and prevents accidental breaking of adjacent non-bed blocks when
  the counterpart half is already missing.

  ## Testing
  - `cargo check -p pumpkin`
  - Tested locally and it works as intended.

### DIFF
--- a/pumpkin/src/block/blocks/bed.rs
+++ b/pumpkin/src/block/blocks/bed.rs
@@ -150,6 +150,12 @@ impl BlockBehaviour for BedBlock {
                 args.position.offset(bed_props.facing.to_offset())
             };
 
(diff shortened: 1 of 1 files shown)


gula00:
You can reproduce a half-bed state using: `/setblock ~ ~ ~ red_bed strict`

Snowiiii:
LGTM. Thank you @gula00 :D

Issue #1617, opened 2026-02-19
Spectators mode players can pick up items


FireAcidCow:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

8bd5009

### What happened?

Spectator mode players can pick up items

### To Reproduce

1. Drop an item on the ground
2. Switch to spectator mode
3. Go to said item
4. Pick it up

### Expected behavior

Spectator mode players shouldn't be able to interact with the world

PR #1699, opened 2026-02-23
fix: spectators can no longer pickup items

kyotuca:
## Description
Simple fix for issue #1617, added a check for spectator mode on item pickup

## Testing
All tests passed successfully and no new clippy warning was found.
Manual testing was successful.

### DIFF
--- a/pumpkin/src/entity/item.rs
+++ b/pumpkin/src/entity/item.rs
@@ -441,11 +441,8 @@ impl EntityBase for ItemEntity {
         Box::pin(async {
             if self.pickup_delay.load(Ordering::Relaxed) > 0
                 || player.living_entity.health.load() <= 0.0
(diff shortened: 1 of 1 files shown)


Purdze (commented):
Is there already a dead player check above this? In vanilla, dead players also shouldn't pick up items

kyotuca:
> Is there already a dead player check above this? In vanilla, dead players also shouldn't pick up items

Yes, there is a dead player or pick up cooldown check just above this one. Should this be moved there?

Purdze:
fold the spectator check into the existing guard above it

```
if self.pickup_delay.load(Ordering::Relaxed) > 0
    || player.living_entity.health.load() <= 0.0
    || player.is_spectator()
{
    return;
}
```

PR #1700, opened 2026-02-23
feat: integrate command dispatchers + reimplement /help

Laptop59:
## Description

Now that we have the new command dispatcher, we would need to reimplement each command with the new dispatcher.
However, it would be a bad idea to make one big PR change all commands, so instead:

- The previous dispatcher has been embedded inside the new dispatcher *temporarily*.
- Some interface methods have been added to the dispatcher to interface between both systems.
- **Reimplementations of commands can be made in small PRs slowly instead of one big PR.**
- Added support for client suggestions for the new dispatcher (including redirects).
- Added an easy way to convert a `CommandSender` into a `CommandContext`.
- Also fixes the bug where typing just `/` in chat causes an error and kicks the player out of the server.
- The console allows up to one leading slash for commands (so `/stop` and `stop` would both work)

To further show that the dual-dispatching works properly, I have reimplemented the `/help` command under the new dispatcher system.

## Testing

Tested commands in the console and in the client:

*Note that* `/?` *is an alias for* `/help`, *so it technically redirects to the* `/help` *reimplementation*.
Screenshot_20260223_213738
Screenshot_20260223_220310
Screenshot_20260223_220826


Purdze (commented):
1. `usize` underflow in help header padding with long command names. `52 - header_text.len()` underflows on `usize` and panics when command name is ~44+ chars. Triggered via `/help <long_string>`. Use `52usize.saturating_sub(header_text.len())`

2. `usize` underflow when zero commands exist. If player has zero permitted commands, `total_pages` is 0, `page` is 0, `page - 1` panics on `usize`. Need early return when `commands.is_empty()`

**Logic bugs:**

3. Dead branch / missing multi-child handling in `get_usage_recursive`. The `else if !children.is_empty()` is inside the `children.len() == 1` guard, so it's always true. This means the multi-child case (`children.len() > 1`) is never handled. Commands with multiple children at a level will have incomplete usage strings

4. `resolve_node_id` relies on magic constant `-2`. Assumes non-root node IDs start at 2. If any node has internal ID 1 that isn't `ROOT_NODE_ID`, this underflows to `usize::MAX` and corrupts the command tree packet sent to clients. At minimum needs a `debug_assert!` or a named constant

Purdze:
@Snowiiii

Snowiiii:
@Laptop59 please fix conflicts

Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #1703, opened 2026-02-23
feat: Add support for multiple CPU targets in Rust workflow

Q2297045667:
Oh my god, it took 30 minutes

### DIFF
--- a/assets/translations/de_de.json
+++ b/assets/translations/de_de.json
@@ -1,0 +1,12 @@
+{
+    "commands.pumpkin.version": "Pumpkin %s\n",
+    "commands.pumpkin.description": "Ermöglicht es jedem, schnelle \nund effiziente Minecraft-Server zu hosten.\n",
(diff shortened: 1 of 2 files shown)


Purdze:
The Export executable (upload-artifact) step is missing the same if condition that's on Build Release and Prepare artifacts. This will fail CI for the skipped matrix combos since dist/ won't exist

The matrix creates 12 jobs but only 7 actually build. Consider using exclude in the matrix to avoid spinning  up runners that do nothing

The target-cpu mapping logic is duplicated in both the build and artifact steps, could be consolidated into a shared env var to make future changes easier

Q2297045667:
> The Export executable (upload-artifact) step is missing the same if condition that's on Build Release and Prepare artifacts. This will fail CI for the skipped matrix combos since dist/ won't exist
> 
> The matrix creates 12 jobs but only 7 actually build. Consider using exclude in the matrix to avoid spinning up runners that do nothing
> 
> The target-cpu mapping logic is duplicated in both the build and artifact steps, could be consolidated into a shared env var to make future changes easier

I'll deal with it later.

PR #1707, opened 2026-02-24
feat: Add support for multiple CPU targets in Rust workflow

Q2297045667:
For the x86 architecture, the specification with the name "standard" is x86-64-v1; for the Arm architecture, it defaults to normal optimization.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -72,25 +72,17 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm, windows-latest, windows-11-arm, macos-latest, macos-15-intel]
-        arch_variant: [standard, x86-64-v2, x86-64-v3]
+        arch_variant: [standard, x86-64-v3]
         toolchain: [stable]
         exclude:
           - os: ubuntu-24.04-arm
-            arch_variant: x86-64-v2
-          - os: ubuntu-24.04-arm
             arch_variant: x86-64-v3
           - os: windows-11-arm
-            arch_variant: x86-64-v2
-          - os: windows-11-arm
             arch_variant: x86-64-v3
-          - os: macos-latest
-            arch_variant: x86-64-v2
           - os: macos-latest
             arch_variant: x86-64-v3
           - os: macos-15-intel
             arch_variant: standard
-          - os: macos-15-intel
-            arch_variant: x86-64-v2
     steps:
       - uses: actions/checkout@v6
       - run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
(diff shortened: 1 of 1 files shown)


Purdze (commented):
Don't run tests across arch variants, same code, same results, just wastes CI time, The four release jobs can be collapsed into one with an include matrix

macos-15-intel lost its standard build, was this intentional?

Q2297045667:
> Don't run tests across arch variants, same code, same results, just wastes CI time, The four release jobs can be collapsed into one with an include matrix
> 
 ok

> macos-15-intel lost its standard build, was this intentional?

Yes, I think for the Intel version of macOS 15, a standard build is not needed.

Purdze:
Tests are still tripled for ubuntu-latest and windows-latest, the arch_variant matrix on build_and_test needs to go, only the release job should vary by CPU target

Snowiiii:
Looks like it uses -v2 as fallback, i would just prefer to use rusts default v1

Q2297045667:
> Looks like it uses -v2 as fallback, i would just prefer to use rusts default v1

I've checked the final generated files, and it doesn't seem to have any issues.
ARM platforms only use the standard architecture, while macOS on Intel utilizes v3 optimizations (x86_64-v3).

Snowiiii:
> > Looks like it uses -v2 as fallback, i would just prefer to use rusts default v1
> 
> I've checked the final generated files, and it doesn't seem to have any issues. ARM platforms only use the standard architecture, while macOS on Intel utilizes v3 optimizations (x86_64-v3).

v2 has less compatibility

Q2297045667:
> > > Looks like it uses -v2 as fallback, i would just prefer to use rusts default v1
> > 
> > 
> > I've checked the final generated files, and it doesn't seem to have any issues. ARM platforms only use the standard architecture, while macOS on Intel utilizes v3 optimizations (x86_64-v3).
> 
> v2 has less compatibility

 Do you mean to delete v2?

Snowiiii:
> > > > Looks like it uses -v2 as fallback, i would just prefer to use rusts default v1
> > > 
> > > 
> > > I've checked the final generated files, and it doesn't seem to have any issues. ARM platforms only use the standard architecture, while macOS on Intel utilizes v3 optimizations (x86_64-v3).
> > 
> > 
> > v2 has less compatibility
> 
> Do you mean to delete v2?

Yes, only provide support for v1 and v3

Q2297045667:
> > > > > Looks like it uses -v2 as fallback, i would just prefer to use rusts default v1
> > > > 
> > > > 
> > > > I've checked the final generated files, and it doesn't seem to have any issues. ARM platforms only use the standard architecture, while macOS on Intel utilizes v3 optimizations (x86_64-v3).
> > > 
> > > 
> > > v2 has less compatibility
> > 
> > 
> > Do you mean to delete v2?
> 
> Yes, only provide support for v1 and v3

Done

Snowiiii:
LGTM. Thank you @Q2297045667 :D, i think there should be a way to download them from the website

PR #1723, opened 2026-02-25
feat(config): hide verbose log metadata by default

Purdze:
## Summary
- add `target` config option to control module path display in logs
- default both `threads` and `target` to `false` for clean log output
- users can re-enable in `features.toml` under `[logging]`


Purdze:
image

Will not show metadata (e.g. `main ThreadId(01) pumpkin_world::level:`) if debug is set to false in features.toml

PR #1733, opened 2026-02-26
feat(event): add playeritemheldevent

yunuservices:
Revived from closed PR #1657. Rebased onto latest upstream master and verified.

### DIFF
--- a/pumpkin/src/net/java/play.rs
+++ b/pumpkin/src/net/java/play.rs
@@ -1958,12 +1958,13 @@ impl JavaClient {
             self.kick(TextComponent::text("Invalid held slot")).await;
             return;
         }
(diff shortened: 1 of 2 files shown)


pumpkin/src/net/java/play.rs
```diff
@@ -1957,6 +1958,22 @@ impl JavaClient {
             self.kick(TextComponent::text("Invalid held slot")).await;
             return;
         }
+        let previous_slot = player.inventory.get_selected_slot() as i32;
```
RoosterBooster007: Use `u8` instead for the slots, like `get_selected_slot` returns. Cast `held.slot` after the safety check.

RoosterBooster007:
Thanks, @yunuservices 🔥

Issue #1618, opened 2026-02-19
Lava no longer properly interacts with water


FireAcidCow:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

8bd5009

### What happened?

When lava flows on top of water, it acts as if water flowed on top of it

### To Reproduce

1. Place lava so that it will flow onto water. 
2. If the lava source is above water, it turns into obsidian. If flowing lava is above water, it turns into cobblestone
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/6f61e851-ff41-472d-8ad6-d96885db41d2" />

### Expected behavior

It should work how it did before 6e1bea7 to match vanilla

GreenedDev:
This is fixed. I checked with vanilla. @Snowiiii close

GreenedDev:
<img width="420" height="280" alt="Image" src="https://github.com/user-attachments/assets/bb575ef9-e4d5-4587-883c-7ee1ca751414" />

Behaviour matches vanilla

PR #1737, opened 2026-02-26
fix: allow lava to flow when above water block (#1618)

yuminstalljoe:
What was changed? Allow lava to flow if water is below.
Why were these changes necessary? Lava turning to obsidian when above water, not flowing.
What is the impact of this change? Mimics vanilla mechanic


Snowiiii:
LGTM. Thank you @yuminstalljoe :D

Purdze:
> LGTM. Thank you @yuminstalljoe :D

theres still a bug, return true exits the entire function early.., continue would skip just the current iteration

Issue #949, opened 2025-06-28
Large plants are always missing their top half


tn-lorenz:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 10

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.6 (d16b0f98)

### What happened?

Large fern, Rose Bush, Lilac and Peony can neither be generated, nor placed without them missing their top half.

### To Reproduce

.

### Expected behavior

.

tn-lorenz:
![Image](https://github.com/user-attachments/assets/2d229aec-8c91-4894-800c-fae89f5209fa)

GreenedDev:
This is still an issue.

RoosterBooster007:
This should be fixed in #1738. Closing.

PR #1738, opened 2026-02-26
fix: allow tall plants to place their upper half (#949)

yuminstalljoe:
What was changed? Implement "tall" variants top half placement
Why were these changes necessary? Previously did not place top half
What is the impact of this change? Allow tall plants to place their upper half 


RoosterBooster007:
Thanks, @yuminstalljoe! 🌻

Issue #1623, opened 2026-02-19
Mobs don't take knockback on the killing blow


FireAcidCow:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

8bd5009

### What happened?

When a mob gets killed, the final hit doesn't apply knockback to the mob. I can't reopen #1577 which is why I made a separate issue. I do recognize that a653726 fixed _a_ bug, it just wasn't this one.

### To Reproduce

1. Hit a mob. Knockback is applied
2. Kill a mob. Knockback isn't applied on the final blow.

https://github.com/user-attachments/assets/0630b072-45e4-443b-ba64-7df352af6b91

### Expected behavior

Knockback should be applied even on the final blow to a mob, to match vanilla.

PR #1741, opened 2026-02-26
fix: dying entity not ticking for movement (#1623)

yuminstalljoe:
What was changed? Allowed an entity to continue ticking for 20
Why were these changes necessary? This allows you to see the knockback during death animation
What is the impact of this change? Allows an entity to continue ticking for 20 during death animation


RoosterBooster007:
Thx, @yuminstalljoe! 🔥

PR #1742, opened 2026-02-26
feat: added unwrapping and wrapping for NBT lists to and from binary

SomeYellowGuy:
## Description
NBT lists do not support having more than one type of element inside. Therefore, during serialization and deserialization of NBT to and from the binary format, Minecraft internally:
- makes the list a *compound* tag list, and
- wraps every *non-compound* tag and *wrapper compound* tag in a compound.

During deserialization, Minecraft tries to "unwrap" every element in an NBT list tag.

A *wrapper compound* looks something like this:
```json
{
    "": <wrapped tag>
}
```

This PR handles this behavior while serializing and deserializing, and also renames `NbtCompound::put_component` to `NbtCompound::put_compound`.
## Testing

Added a test for serializing a list of different element types.

### DIFF
--- a/pumpkin-nbt/src/lib.rs
+++ b/pumpkin-nbt/src/lib.rs
@@ -418,8 +418,9 @@ mod test {
         vec.push(NbtTag::String("Not an int".to_string()));
         vec.push(NbtTag::Byte(2));
 
-        // This compound will not, since the list will already be a
-        // list of compound tags.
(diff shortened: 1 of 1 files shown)


SomeYellowGuy:
what is this CI?

Q2297045667:
> what is this CI?

Don't mind it

Snowiiii:

Issue #1698, opened 2026-02-23
Dropped items are not rendered correctly after re-entering the server


mingcha2333:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 10 22H2

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.11 (Commit: 32a5c52/release)

### What happened?

After throwing items on the ground and then rejoining the server, the dropped items are not rendered on the client side, even though they are still present (e.g., they can be picked up, and other players can see them).

A video demonstrating the issue:
https://github.com/user-attachments/assets/4d097c2d-aa04-410b-8b1e-79fbd0b03edb

### To Reproduce

1. Throw an item onto the ground.
2. Leave the server and rejoin.
3. Observe the area where the item was dropped – the item is invisible, though it is still there.

### Expected behavior

Dropped items should be visible to the player after rejoining the server, just as they are before reconnecting.

kyotuca:
After some tinkering I believe it is some kind of Nbt issue, as i read a:
`pumpkin_world::chunk::format: Entity in chunk 0,0 is missing UUID: NbtCompound { child_tags: [] }`
For every dropped item when I relog. While for other entities the tag array is full

This is also possibly related to #1329

Issue #1329, opened 2026-01-16
Mobs despawn on Server Leave


jay2510MC:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows

### Server Software Version/Commit

Pumpkin 0.1.0-dev+1.21.11

### What happened?

when you leave the Server mobs despawn

### To Reproduce

spawn mobs next to you and rejoin they despawn

### Expected behavior

They should stay there and NOT despawn

BraveDevelopment:
I don't think opening an issue like this makes any sense. I mean, if you join the server and take a look, you'll see that the entities aren't even moving. That suggests the entity code isn't even close to being finished. So this issue is completely pointless. It's not helpful to open an issue for every little thing that isn't working because its clearly unfinished. Especially since many of your issues are duplicates, and you don't seem to have really looked into the subject matter.
If i'm wrong @Snowiiii please correct me...

ghost:
They don't despawn, if you spawn a mob and let it sufffocate and then respawn, you still hear them suffocation and z.b. a panda drops a bambus after a while.

ghost:
So they still exists serverside but not clientside.

QuantumSegfault:
This should be fixed (for awhile now). Please verify with latest Pumpkin.

GreenedDev:
appears to still be an issue

thisissamiam:
> appears to still be an issue

can agree

Q2297045667:
This should now be fixed. Please verify again. If it still exists, please reopen

PR #1744, opened 2026-02-26
feat!: shulker (#1569)

yuminstalljoe:
## Description
Enables breaking/placing a shulker and retaining items/NBT resolving #1569. Serialization/deserialization adjusted to match vanilla.

## Testing
- [x] Place/Break Empty Survival (drop)
- [x] Place/Break Empty Creative (no drop)
- [x] Place/Break Survival/Creative with items (drop with NBT, does not consume in creative)
- [x] Break Survival/Creative (drop with NBT)
- [x] Insert item(s) with NBT data (retained)
- [x] Place shulker in shulker (rejected)
- [x] Automatically push shulker in shulker - hopper (rejected)

PR #1746, opened 2026-02-26
fix: render entities on login (#1698, #1329)

yuminstalljoe:
## Description
Send entities on login to player

## Testing
- [x] Drop items and relog (items appear)
- [x] Spawn mobs and relog (mobs appear)

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -1606,6 +1606,11 @@ impl World {
         // This is made before the player teleport so that the player doesn't glitch out when spawning
         chunker::update_position(player).await;
 
+        // Fix: Send entities from world.entities that may not be in loaded chunks
+        // This handles the case where entities exist in memory but their chunks were
+        // unloaded and not persisted to disk yet
+        self.send_in_memory_entities_to_player(player).await;
+
         // Teleport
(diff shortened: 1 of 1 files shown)


chocodev11:
as @kyotuca said in #1698, some entity is having empty serializers that writes empty compounds. And join-time entity spawn reads chunk NBT and skips entries without id. so your PR is fixing symptom, but not the root cause

yuminstalljoe:
@chocodev11 @kyotuca I did not experience that in testing before or after my changes. I also made changes to serialize/deserialize in #1744 that may alleviate it, but as of now I don't have those issues to reproduce.

kyotuca:
From my testing I saw that:
- items were serializing completely empty
- mobs were serializing with incorrect data types, (uuid was being serialized back to an array instead of a int_list)

I didn't dig any deeper though

yuminstalljoe:
@kyotuca Without being able to reproduce myself I can't test that specifically, but #1744 changes brings it more inline with vanilla serialization that I still believe will alleviate. It was needed to support shulker's since the server wasn't handling this properly.

Snowiiii:
This is wrong, entities should/are broadcasted from saved entity chunks

PR #1748, opened 2026-02-27
feat: monster room feature

RoosterBooster007:
## Description
- Adds monster/spawner rooms to world gen
- The chests and spawner are bugged rn (we need better loot table support and a way to set a spawner's entity)


Snowiiii:

PR #1749, opened 2026-02-27
fix(plugin): cdylib potion effects, shield deadlock, protocol fix

h0lybyte:
## Summary
- **PotionContents serialization**: `custom_name` was being serialized as `TextComponent` (NBT-encoded JSON) instead of `String` — vanilla protocol expects `Optional<String>`, the extra bytes caused client disconnect ("31 bytes extra")
- **cdylib patch component access**: Added `get_data_component_owned<T>()` to `ItemStack` with NBT round-trip fallback — `downcast_ref` silently fails for plugin-created patch components because `TypeId` differs across the cdylib boundary
- **Potion effect delivery**: Changed `self.add_effect()` (LivingEntity — stores only) to `player.add_effect()` (Player — sends `CUpdateMobEffect` packet to client) so status effect icons actually appear
- **Shield equip deadlock**: Added `Arc::ptr_eq` guard in `prepare_hand_item_for_use` to prevent re-entrant deadlock when a shield already in the offhand is right-clicked to block (`tokio::sync::Mutex` is not re-entrant)

## Test plan
- [ ] Connect with a cdylib plugin that creates custom potions with PotionContents in the patch — verify no client disconnect
- [ ] Drink a custom potion — verify status effect icons appear in HUD and effects apply
- [ ] Place shield in offhand, right-click to block — verify no deadlock/freeze
- [ ] Verify vanilla potions (creative mode) still work correctly

### DIFF
--- a/pumpkin-protocol/src/codec/data_component.rs
+++ b/pumpkin-protocol/src/codec/data_component.rs
@@ -196,10 +196,10 @@ impl DataComponentCodec<Self> for PotionContentsImpl {
             seq.serialize_field::<bool>("", &false)?;
         }
 
-        // Custom name (optional) — vanilla encodes as Optional<Component> (NBT-backed)
+        // Custom name (optional) — vanilla 1.21.11 encodes as Optional<String>
         if let Some(name) = &self.custom_name {
             seq.serialize_field::<bool>("", &true)?;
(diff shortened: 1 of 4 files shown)


GroobleDierne:
Was this closed on accident?

PR #1753, opened 2026-02-27
feat(event): add PlayerEggThrowEvent

yunuservices:
﻿## Title
This PR adds missing "PlayerEggThrowEvent" to the pumpkinmc.

## Changes
- Adds event in egg throw handling lifecycle.
- Throw behavior is now plugin-observable and cancellable.

## Notes

revived from #1651

### DIFF
--- a/pumpkin/src/entity/projectile/egg.rs
+++ b/pumpkin/src/entity/projectile/egg.rs
@@ -138,13 +138,17 @@ impl EntityBase for EggEntity {
                     player,
                     self.get_entity().entity_uuid,
                     hatching,
-                    to_spawn.min(u8::MAX as usize) as u8,
+                    to_spawn as u8,
                     hatching_type,
                 );
                 let event = server.plugin_manager.fire(event).await;
-                hatching = event.hatching && !event.cancelled;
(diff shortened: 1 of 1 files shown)


pumpkin/src/plugin/api/events/player/egg_throw.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub egg_uuid: uuid::Uuid,
+
+    /// Whether the egg should hatch.
+    pub hatching: bool,
+
+    /// The number of entities that should hatch.
+    pub num_hatches: u8,
+
+    /// The entity type that should hatch (registry key).
+    pub hatching_type: String,
```
RoosterBooster007: Can you change this to use `EntityType`?

Purdze:
to_spawn.min(u8::MAX as usize) as u8 (egg.rs L138) to_spawn is always 0, 1, or 4 at this point, so the .min(255) is a no-op, harmless but unnecessary

Cancelled but values still read, When cancelled, to_spawn and hatching_type are still updated from the event even though they won't be used (the if hatching && to_spawn > 0 guard prevents it). Not a bug, just slightly redundant

RoosterBooster007:
Thanks, @yunuservices! 🐣

Issue #1754, opened 2026-02-27
Cactus Flower not breaking when breaking cactus


yuminstalljoe:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

macOS Tahoe 26.0.1

### Server Software Version/Commit

9110727

### What happened?

Breaking a cactus does not break the flower. The flower remains in the air.

### To Reproduce

Break a cactus with cactus flower on it

### Expected behavior

Cactus flower breaks/drops when breaking cactus

PR #1755, opened 2026-02-27
fix: break cactus flower when cactus breaks (#1754)

yuminstalljoe:
## Description
Breaks cactus flowers when breaking cactus.

## Testing
- [x] Break cactus with flower above it

### DIFF
--- a/pumpkin/src/block/blocks/plant/cactus_flower.rs
+++ b/pumpkin/src/block/blocks/plant/cactus_flower.rs
@@ -5,7 +5,8 @@ use pumpkin_world::world::BlockFlags;
 use pumpkin_world::{BlockStateId, world::BlockAccessor};
 
 use crate::block::{
-    BlockBehaviour, BlockFuture, CanPlaceAtArgs, GetStateForNeighborUpdateArgs, OnScheduledTickArgs,
+    BlockBehaviour, BlockFuture, CanPlaceAtArgs, GetStateForNeighborUpdateArgs, OnPlaceArgs,
+    OnScheduledTickArgs,
 };
 
 #[pumpkin_macros::pumpkin_block("minecraft:cactus_flower")]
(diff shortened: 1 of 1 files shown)


Purdze (commented):
issues / questions:

1. `block_below` comparison uses reference equality (`&Block::CACTUS`) in `cactus_flower.rs:60`. this works if `get_block` returns a `&Block` and `Block` implements `PartialEq` by value. looks correct based on the pattern, but worth confirming it's consistent with how other blocks do this check (e.g., does the cactus block itself check for sand/red sand support the same way?).

2. missing support for other valid support blocks? in vanilla minecraft, a cactus flower should only be plantable on cactus. the check at line 60 restricts to `Block::CACTUS` only. verify this matches the vanilla spec. if cactus flowers can also sit on other blocks (e.g., sand when newly placed by a player), this would be incomplete.

3. `BlockDirection::Up` exclusion at `cactus_flower.rs:38`. the comment says "similar to ChorusFlowerBlock." this makes sense: if a block is placed above the flower, it shouldn't trigger a self-destruct check. but cactus flowers can't have anything above them in vanilla, so this guard is defensive. fine to keep.

4. no `placed_by` or `on_place` handler. the pr doesn't implement `on_place` to validate placement. if `can_place_at` is already called by the placement system upstream, this is fine. otherwise, a player could place a cactus flower on non-cactus blocks.

5. no tests. the testing section only mentions manual testing. if the project has integration/unit tests for block behaviors, it would be good to add a test case for breaking cactus dropping the flower above, and flower not being placeable on non-cactus blocks.

nit: the comment `// Cactus flower can be placed on cactus` at line 61 is redundant given the method name and the single-line check.

yuminstalljoe:
Thanks for the review @Purdze! I've addressed the feedback:

  1. Confirmed reference equality is consistent with other blocks in the codebase
  2. Updated `can_place_at_internal` to allow placement on cactus, farmland, and sturdy blocks (matches vanilla behavior after reviewing java source code)
  3. Removed `BlockDirection::Up` exclusion - verified vanilla doesn't exclude any direction in neighbor updates
  4. Added `on_place` handler to validate placement
  5. Pumpkin doesn't currently have block behavior unit tests
  6. Removed redundant comment

Snowiiii:
This seems to be wrong, should can just use the `PlantBlock` we already, have **look at vanilla code before submitting a PR**

PR #1757, opened 2026-02-27
fix: keep blockstate for falling blocks

### LINKED ISSUE #1756: Anvils always change their direction to face north when they start falling

### Operating System

Arch Linux

Pumpkin 0.1.0-dev+1.21.11 (Commit: c40ac36/debug)

### What happened?

Anvils make themselves face `north` when they start falling, no matter the direction they were facing before.

This is not specific to anvils. Currently, when a gravity-affected block is about to fall, it converts itself into a block entity with the default block state of that block.
This is fine for blocks like sand and gravel which have only one block state, but other blocks like anvils, which have multiple block states reset to their default block state upon conversion.

### To Reproduce

1. Join the server
2. Create a tall pillar of solid blocks like dirt on which anvils can be right-clicked to place them.
3. Place the anvil on all four sides of the pillar. You will notice that at some sides, the anvil suddenly changes its facing direction before actually falling. To be more specific, all anvils when falling change their blockstate property `facing` to `north`.

### Expected behavior

The anvils should keep the direction they were placed before they started falling and upon reaching the ground, should face the same direction as it did before falling.

Laptop59:
## Description

This PR makes it such that an anvil keeps their block state when it gets converted to a falling block entity, and upon reaching the ground, the direction it faces is the same as it did before it fell.

Fixes #1756.

## Testing

Tested anvils falling in-game.


RoosterBooster007:
Thanks, @Laptop59! 🔨

PR #1766, opened 2026-02-28
fix: boss bars & effects

GroobleDierne:
## Description
- Fix boss bar encoding
- Fix regressions regarding infinite effects and an integer overflow introduced in #1526 

## Testing

Tested in-game via the bossbar and effect commands as  #1526 seems to have broken plugin support.


RoosterBooster007:
Looks good! Thx, @GroobleDierne 🍾

PR #1770, opened 2026-03-01
chore: fix formatting of couple of blocks

GreenedDev:
## Description
fix formatting of some of the block structs (to match with other blocks' format)


Snowiiii:

PR #1772, opened 2026-03-01
Implement copper torches

GreenedDev:
## Description
Implements copper torches based on current implementation of other torches


Snowiiii:

PR #1787, opened 2026-03-03
fix: wrong piston break check

GreenedDev:
## Description
Fixes this bug:

https://github.com/user-attachments/assets/2fadf27a-77c5-4fbd-b8d8-dd4647141892


Snowiiii:

PR #1788, opened 2026-03-03
feat: identifiers and bossbar rework

Laptop59:
## Description

Adds a new struct `Identifier` which contains both a namespace and path.
Hopefully this can be used for datapacks, like identifying functions, and other user-defined things later when implemented.

- Also added the `StaticIdentifier` which can be queried at compile-time and can be converted to an `Identifier`.
- Replaced some usages of `ResourceLocation` with `Identifier`
- Did a rework with the `CustomBossbars`:
  - Bossbars are no longer identified by `String` and are now identified by `Identifier`.
  - The functions that modify a `CustomBossbar` from `CustomBossbars` have been moved to the `CustomBossbar` struct.
  - `CustomBossbars` can return a mutable reference to one of its bossbars by its identifier.
  - Modified the `/bossbar` command functionality to work with the above system.
- Added an `ArgumentType` implementation that parses `Identifier`s, which is called the `IdentifierArgumentType`.

## Testing

Added some very simple tests.

### DIFF
--- a/pumpkin-util/src/identifier.rs
+++ b/pumpkin-util/src/identifier.rs
@@ -602,3 +602,35 @@ impl PartialEq<Identifier> for StaticIdentifier {
         self.namespace == &*other.namespace && self.path == &*other.path
     }
 }
+
+#[cfg(test)]
+mod test {
+    use crate::identifier::{Identifier, IdentifierError};
+
+    #[test]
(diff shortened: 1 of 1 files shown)


Snowiiii:
Please split this into multiple PRs, also i don't see much benefit in using identifiers, we used an ResourceLocation struct before and the only results were more heap allocations

PR #1793, opened 2026-03-03
fix: offline player support for some command with usercache

### LINKED ISSUE #438: Keep track of offline users that joined previously into a word

chocodev11:
Reworked command selector/profile handling:
- Enforced strict player-type selectors for player-only args (`@e`/`@n` only valid with `type=player`).
- Rebuilt `game_profile` resolution order: selector -> online player -> UUID/name cache -> Mojang lookup (online mode) -> offline UUID fallback (offline mode).
- Added persistent `usercache.json` - UUID cache (TTL + MRU), updated on successful login/resolution.
- Added command-specific profile suggestions for `/ban`, `/op`, `/deop`, `/pardon`, and `/whitelist add/remove` (including selector suggestions where vanilla does).
- Fixed core selector execution behavior for `@p`, `@n`, `@r`, and `@s` with proper filter/sort/limit flow.
- Switched ban/whitelist membership checks to UUID-based matching; names are kept as metadata and refreshed.
- Added Brigadier syntax error propagation with cursor context (`...<--[HERE]`).
- Kept `/ban-ip` behavior unchanged as string target (IP or online player name).
may close #438

## Testing


pumpkin/src/command/args/gameprofile.rs
```diff
@@ -0,0 +1,280 @@
...
+        _input: &'a str,
+    ) -> SuggestResult<'a> {
+        Box::pin(async move {
+            let mut suggestions = vec![
+                CommandSuggestion::new("@s".to_string(), None),
+                CommandSuggestion::new("@p".to_string(), None),
+                CommandSuggestion::new("@r".to_string(), None),
+                CommandSuggestion::new("@a".to_string(), None),
+                CommandSuggestion::new("@e".to_string(), None),
+            ];
```
Laptop59: We should also add `@n` as a suggestion.
```suggestion
            let mut suggestions = vec![
                CommandSuggestion::new("@s".to_string(), None),
                CommandSuggestion::new("@p".to_string(), None),
                CommandSuggestion::new("@r".to_string(), None),
                CommandSuggestion::new("@a".to_string(), None),
                CommandSuggestion::new("@e".to_string(), None),
                CommandSuggestion::new("@n".to_string(), None),
            ];
```

pumpkin/Cargo.toml
```diff
@@ -76,7 +76,7 @@ tracing.workspace = true
 tracing-subscriber.workspace = true
 
 # Remove time in favor of chrono?
-time = { workspace = true, features = ["parsing", "macros"] }
+time = { workspace = true, features = ["parsing", "macros", "local-offset"] }
```
andreisugu: We really should consider using chrono instead
You literally had to write your own version of  "1 month diffrence" because we are still using "time"

pumpkin/src/net/authentication.rs
```diff
@@ -179,6 +181,57 @@ pub fn fetch_mojang_public_keys(
...
+    let dashed = format!(
+        "{}-{}-{}-{}-{}",
+        &id[0..8],
+        &id[8..12],
+        &id[12..16],
+        &id[16..20],
+        &id[20..32]
+    );
+    Uuid::parse_str(&dashed).ok()
+}
```
andreisugu: I dont think this is actually needed, UUID in rust natively parses correctly this in any format.

```
let parsed_uuid = Uuid::parse_str(&profile.id).map_err(|_| AuthError::FailedParse)?;
    Ok(Some((parsed_uuid, profile.name)))
}
```

Snowiiii:
Overall looks good. Thank you @chocodev11 :D, Awesome work

PR #1786, opened 2026-03-03
feat: implement complete Villager entity with trading, professions, and AI

WhiteProject1:
- Full vanilla-compatible Villager entity implementation with 14 professions, 7 biome types, 5 levels, and complete trading system
- Pathfinding performance optimization: ~70-85% CPU reduction via eliminating triple world queries, fixing hash collisions, flattening allocations, and smart re-pathing
- Complete AI goal system with schedule-based behavior (work, sleep, meet, idle), workstation claiming, gossip/reputation, iron golem spawning, and zombie villager curing

### Villager Entity & Metadata
- `VillagerEntity` struct with profession, biome type, level, experience, trade offers, gossips
- Metadata sync (`DATA_VILLAGER_DATA`) on spawn via new `mob_init_data_tracker` hook
- NBT persistence for all villager data (VillagerData, Xp, Offers, Gossips)

### Trading System
- `CMerchantOffers` / `SSelectMerchantTrade` packet handling
- `MerchantScreenHandler` with `MerchantOutputSlot` (prevents insertion, validates & consumes inputs on take)
- `as_any_mut()` added to `ScreenHandler` trait for mutable downcasting
- Complete trade tables for all 13 professions × 5 levels with random selection
- Level-up (XP thresholds), restock (max 2/day near workstation during work hours)

### Workstation & POI
- 14 workstation POI types + `POI_TYPE_HOME` for beds in `pumpkin-world/src/poi`
- `block_to_poi_type()` / `poi_type_to_profession()` mapping
- Automatic POI add/remove hooks in `set_block_state()` on block changes
- `ClaimWorkstationGoal`: unemployed villagers find and claim nearby workstations

### Schedule & AI Goals
- Schedule: Idle → Work → Meet → Rest (matching vanilla time ranges)
- `SleepInBedGoal`, `WorkAtStationGoal`, `GatherAtBellGoal` (POI/block-based navigation)
- Standard goals: Swim, EscapeDanger, WanderAround, LookAtEntity, LookAround

### Gossip & Reputation
- 5 gossip types (MajorPositive/MinorPositive/MinorNegative/MajorNegative/Trading)
- Per-player reputation affecting trade prices, daily decay, meeting-time sharing

### Iron Golem Spawning
- Conditions: 3+ villagers + 3+ beds + 5min cooldown + no golem within 16 blocks

### Zombie Villager Curing
- Golden apple + weakness effect → random 3-5 min conversion → VillagerEntity spawn
- Preserves profession and biome type through conversion

[...]

PR #1794, opened 2026-03-03
perf: optimize pathfinding system for ~70-85% CPU reduction

WhiteProject1:
- Replace triple async chunk lookups with single `get_block_state_id` + static array lookups
- Fix node hash collisions (Y masking) by using `Vector3<i32>` HashMap key
- Flatten `came_from` linked list (`Box<Node>`) to `Vector3<i32>` reference
- Replace `path_type_malus` HashMap with fixed-size array
- Simplify BinaryHeap from `Vec<Option<Box<Node>>>` to `Vec<Node>`
- Add reusable `get_neighbors` Vec + Navigator-level allocation caching
- Implement smart re-pathing with distance-proportional threshold + cooldown

This is part 1 of the villager implementation PR (#1786), split for easier review.


Snowiiii:
LGTM. Thank you @WhiteProject1 :D

PR #1796, opened 2026-03-03
feat: extend POI system with workstation and bed types

WhiteProject1:
- Add 14 workstation POI types (Armorer, Butcher, Cartographer, Cleric, Farmer, etc.)
- Add bed POI type (`POI_TYPE_HOME`) for villager sleeping
- Add `block_to_poi_type()` mapping from block names to POI types
- Add `poi_type_to_profession()` mapping from POI types to professions
- Add `is_workstation_poi()` helper for workstation detection

This is part 3 of the villager implementation PR (#1786), split for easier review. Extends the existing POI system (used by nether portals) with workstation and bed types needed for villager profession assignment.


Snowiiii:
It seems like this is not used anywhere ?

PR #1798, opened 2026-03-04
fix: reset terminal to initial state after stop command

rodrigosantosss:
## Description
This is my attempt to fix a bug where, when using rustyline, the terminal will not echo after the stop command is executed, because the console thread is not stopped before the program exits. Frankly, it's a janky solution, mainly because I can't interrupt the readline call, the solution I used was to look for the special case of the stop command and break out of the cycle, which is not so elegant, but it works. I also stored the join handle for the console thread wrapped by an Option in the Pumpkin Server instance, so the main function could join it.
## Testing
I ran it on my machine in debug mode and release mode and the fix worked.

Thank you for your attention and sorry that I couldn't do something better, it's mainly something I did for myself, but I opened this PR in case you found it useful.


rodrigosantosss:
This was already fixed on a different pull request, so I'm closing it. Thanks for fixing the issue.

PR #1799, opened 2026-03-04
feat: add vanilla behavior for eyeblossom

WCBBEX:
## Description
This PR implements the full gameplay mechanics for the eyeblossom and its potted variants. Key changes include:
1. Add specific particle effects and sound logic during state transitions.
2. Implemented the chain reaction logic and trigger nearby flowers to change state.
3. Open eyeblossoms now apply a poison effect to bees upon collision.
4. Add `spawn_particle_with_data` to both `World` and `Player` structs to support particles that require extra data payloads.

## Testing
Verified locally.


Snowiiii:
Looks like many things were hardcoded and also put into one already existing file which is really not great for readability

Snowiiii:
Feel free to reopen

PR #1801, opened 2026-03-04
fix: close stdin on shutdown to prevent terminal raw mode corruption

WhiteProject1:
Fixes terminal state corruption after server shutdown. When the server shuts down, the console thread remains blocked on `rl.readline()` (synchronous). The `Editor` holds the terminal in raw mode, so the process exits with the terminal corrupted.

- **`close_stdin()`**: Closes stdin fd/handle, causing `readline()` to return EOF. Cross-platform (Unix `FromRawFd`, Windows `GetStdHandle`/`CloseHandle`).
- **Shutdown retry**: Waits up to 1s for the console thread to return the Editor for proper restoration.
- **Signal guard**: Prevents redundant `stop_server()` when EOF is from shutdown itself.


Snowiiii:
This was recently fixed, also this looks way too hacky tbh

PR #1804, opened 2026-03-04
perf: optimize chunk loading throughput and memory usage

WhiteProject1:
- Scale IO read thread count with CPU cores (`(cores/6).clamp(2,4)`) instead of hardcoded 2 — conservative scaling since chunk reading is disk-bound, not CPU-bound
- Increase IO write queue from 500 to 1000 entries — moderate headroom to reduce generation thread stalls without excessive memory usage
- Enable `loaded_chunks` DashMap `shrink_to_fit` when over-allocated by 4096+ entries (was commented out), reclaiming memory after mass chunk unloads


WhiteProject1:
Closing — changes are too marginal to justify the PR.

PR #1805, opened 2026-03-04
Add Kelp block placement/break rules

GreenedDev:
## Description
Allow kelp to only be placed in full water or downward flowing water, on top of solid blocks.
when broken, if there is kelp below make it a kelp_plant (the pointy kelp)
when placed, if there is kelp below make it kelp (the not pointy one)


chocodev11:
can you make kelp immediately replaced by water on break too?

GreenedDev:
> can you make kelp immediately replaced by water on break too?

as you can see in broken() function, I already tried. It seems that the broken function is not called when block is broken by neighbor update check.

GreenedDev:
^ This should fix it

Snowiiii:

Issue #198, opened 2024-10-27
Its possible to create water pools due to conduits upon placing are placed already waterlogged


Z-fentom:
- [ ] I've searched existing issues and couldn't find a duplicate.

**Describe the bug**
Upon placing a conduit it places already in a waterlogged state

**To Reproduce**
1.Take a conduit
2.Place it

**Expected behavior**
A non waterlogged conduit should have been placed

**Environment (please complete the following information):**

- Operating System: Arch Linux
- Server Software Version/Commit: [e.g., 1.0.0/39b4cb3]

**Additional context**

lokka30:
I'm not sure if Pumpkin actually has code for waterlogging things, so it might just be placing the conduit in its default state, which might be waterlogged. Pure speculation but food for thought.

PR #1808, opened 2026-03-05
fix: conduit being waterlogged all the time

GreenedDev:
## Description
Fixes https://github.com/Pumpkin-MC/Pumpkin/issues/198

note:
MangroveRootsLikeProperties is used because it shares the same block properties


RoosterBooster007:
Thanks, @GreenedDev! 🌊

PR #1784, opened 2026-03-02
feat: Combat enchantments, sweep attack area damage, and critical particles

WhiteProject1:
Implements combat mechanics from #1404 and fixes #1620:

- **Enchantment damage**: Sharpness, Smite (vs undead), Bane of Arthropods (vs arthropods) using entity type tags
- **Knockback enchantment**: +0.5 per level to knockback strength
- **Fire Aspect**: 4 seconds per level
- **Sweeping Edge**: Scales sweep damage ratio (level/(level+1))
- **Sweep attack**: Area damage hitting nearby living entities with scaled damage and knockback
- **Critical/enchanted hit particles**: Crit on critical hits, EnchantedHit on enchantment bonus
- **AttackType fixes**: Added missing `!sprinting` and `on_ground` conditions for vanilla parity
- **Armor stand knockback**: Only from sprint attacks or Knockback enchantment (fixes #1620)
- **Knockback sync, resistance, cooldown, sprint reset** fixes

PR #1819, opened 2026-03-08
Fix: sweeping attack

### LINKED ISSUE #1622: Crits don't produce particles

### Operating System

Arch Linux

8bd5009

### What happened?

When you land a crit, it doesn't produce crit particles

### To Reproduce

1. Summon a cow or other mob
2. Hit with a netherite sword and kill it in two hits
3. Summon another cow
4. Crit with a netherite sword thereby killing it in one hit, but no crit particles are produced.

### Expected behavior

Crit particles should be produced when landing a crit.

1TheCrazy:
## Description
Adds the sweeping attack.
This does not fix checks for what `AttackType` should be used, it just provides a simple architecture for the sweeping attack (and possibly other attack types).
Adheres to vanilla behaviour
Also fixes #1622 

## Testing
- [x] Crosscheck with vanilla (to adhere to vanilla behaviour)
- [x] Ingame checks
   - [x] Mob masses are correctly damaged
   - [x] Individual crit still works
- [x] lint & test

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -754,10 +754,10 @@ impl Player {
         // TODO: Add other particles (e.g. crit)
         match attack_type {
             AttackType::Sweeping => {
-                combat::spawn_sweep_particle(&self.living_entity.entity, &world, &pos).await
+                combat::spawn_sweep_particle(&self.living_entity.entity, &world, &pos).await;
             }
             AttackType::Critical => {
-                combat::spawn_crit_particle(victim.get_entity(), &self.world()).await
+                combat::spawn_crit_particle(victim.get_entity(), &self.world()).await;
             }
             _ => {}
         }


pumpkin/src/entity/player.rs
```diff
@@ -754,23 +736,105 @@ impl Player {
...
+    }
+
+    async fn do_attack(&self, attack_type: &AttackType, damage: f64, victim: &Arc<dyn EntityBase>) {
+        let attacker_entity = &self.living_entity.entity;
+        let world = self.world();
+        let pos = attacker_entity.pos.load();
+        let did_damage: bool = match attack_type {
+            AttackType::Sweeping => self.do_sweeping_attack(damage, victim).await,
+            _ => self.do_simple_attack(victim, attack_type, damage).await,
+        };
```
andreisugu, line 751: You should likely call do_simple_attack on the primary victim first with standard damage. Then, inside do_sweeping_attack, filter out both the attacker and the primary victim from the entities list before applying the reduced sweep damage to the rest of the group.

Reason: Main Target SHOULD take FULL damage, not PARTIAL like the other targets.

pumpkin-util/src/math/boundingbox.rs
```diff
@@ -366,6 +366,24 @@ impl BoundingBox {
...
+        self.max.x - self.min.x
+    }
+
+    /// Returns the length of the Y-Side (height) of the bounding box
+    #[must_use]
+    pub fn get_y_length(&self) -> f64 {
+        self.max.y - self.min.y
+    }
+
+    /// Returns the length of the X-Side (depth) of the bounding box
```
andreisugu, line 382: not x-side. its z-side here

pumpkin/src/entity/player.rs
```diff
@@ -754,23 +736,105 @@ impl Player {
...
+                Some(self),
+            )
+            .await;
+
+        if victim.get_living_entity().is_some() {
+            let knockback_strength = if *attack_type == AttackType::Knockback {
+                2.0
+            } else {
+                1.0
+            };
```
andreisugu, line 830: vanilla: 
sweep targets get a fixed 0.4

So: 
```
let knockback_strength = match attack_type {
    AttackType::Knockback => 2.0,
    AttackType::Sweeping  => 0.8, // 0.8 × 0.5 = 0.4 — matches vanilla sweep knockback
    _                     => 1.0,
};

File: pumpkin/src/entity/player.rs [hunk 4]
```diff
@@ -754,23 +736,105 @@ impl Player {
...
+
+        // Spawn particles
+        // TODO: Add other particles (e.g. crit)
+        match attack_type {
+            AttackType::Sweeping => {
+                combat::spawn_sweep_particle(&self.living_entity.entity, &world, &pos).await;
             }
+            AttackType::Critical => {
+                combat::spawn_crit_particle(victim.get_entity(), &self.world()).await;
+            }
```
andreisugu, line 761: Particles are emitted before checking 'did_damage' . If the attack was blocked or failed (e.g. target is immune), crit particles still appear. In vanilla, the crit visual only fires when damage is actually applied.

pumpkin/src/entity/combat.rs
```diff
@@ -1,17 +1,17 @@
-use std::sync::atomic::Ordering;
-
+use crate::{
+    entity::{Entity, player::Player},
+    world::World,
+};
+use core::f64;
```
andreisugu, line 5: unused import

andreisugu:
To add:
This is a **Partial Dupliacte** of #1784

1TheCrazy:
Thanks for taking a look a this, I'll be closing this as the other PR mentioned (#1784 ) seems to be more dedicated and has more features implemented.

PR #1827, opened 2026-03-09
fix: resolve memory leaks from Arc cycles, stale refs, and unbounded collections

WhiteProject1:
Fixes multiple memory leaks discovered through static analysis of Arc reference patterns, collection growth, and resource cleanup paths.

### Arc Reference Cycle Leaks (Critical)

- **AI Goal self-referential cycles**: `ChasePlayerGoal`, `CreeperIgniteGoal`, `PickUpBlockGoal`, `PlaceBlockGoal`, and `TeleportTowardsPlayerGoal` held `Arc<EndermanEntity/CreeperEntity>` while being stored inside the entity's own `GoalSelector`. This created a cycle (`Arc<Mob> → GoalSelector → Goal → Arc<Mob>`) that **permanently leaked every despawned enderman and creeper**. Replaced with `Weak<Mob>` references.

- **Vehicle/passenger cycles**: `Entity.passengers` and `Entity.vehicle` hold `Arc<dyn EntityBase>` to each other. When entity A is B's vehicle and B is A's passenger, neither can be freed. Added cleanup in `remove_entity()` to break these references before removal.

### Stale Reference Leaks (High)

- **Mob target**: `MobEntity.target: Option<Arc<dyn EntityBase>>` kept dead entities alive in memory indefinitely after they were removed from the world. Added `is_removed()` check at the start of each mob tick to clear stale targets.

- **Screen handler listeners**: `ScreenHandlerBehaviour.listeners` accumulated `Arc<Player>` references every time a player opened a container, but `default_on_closed()` never cleared them. Players would never be deallocated. Added `listeners.clear()` and `sync_handler = None` on close.

### Unbounded Collection Leaks (Medium)

- **Plugin event handlers**: `HandlerMap` entries were never removed when plugins were unloaded, causing `Box<dyn DynEventHandler>` to accumulate. Added cleanup when no active plugins remain.

- **Query challenge tokens**: The `valid_challenge_tokens` HashMap had no size limit—only a 30s periodic clear. A malicious client could flood handshake packets to exhaust memory between clears. Added a 4096-entry cap.

### DIFF
--- a/pumpkin-inventory/src/screen_handler.rs
+++ b/pumpkin-inventory/src/screen_handler.rs
@@ -227,6 +227,16 @@ pub trait ScreenHandler: Send + Sync {
                 offer_or_drop_stack(player, cursor_stack_lock.clone()).await;
                 *cursor_stack_lock = ItemStack::EMPTY.clone();
             }
+
+            // Drop the lock before clearing listeners to avoid holding it unnecessarily
+            drop(cursor_stack_lock);
+
+            // Clear listeners and sync handler to prevent Arc reference leaks.
+            // Listeners hold Arc<Player> references that accumulate each time a
+            // container is opened, causing the Player to never be deallocated.
+            let behaviour = self.get_behaviour_mut();
+            behaviour.listeners.clear();
+            behaviour.sync_handler = None;
         })
     }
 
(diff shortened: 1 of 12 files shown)


Snowiiii:
Would still not consider that as one Feature, therefore still please use Atomic PRs

PR #1836, opened 2026-03-09
feat: add /function, /return, and /schedule commands

WhiteProject1:
- Implement `/function` command for running mcfunction files with name argument
- Implement `/return` command with value (integer) and `fail` subcommands for returning from function context
- Implement `/schedule` command with `function` (append/replace modes + time) and `clear` subcommands
- All commands registered with OP level 2 permissions matching vanilla behavior
- TODO markers for full functionality pending mcfunction parser and function scheduler


Snowiiii:
make Atomic PRs

Issue #1817, opened 2026-03-08
Terminal won't give back echoing after stopping the server


Barni-Mezei:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Endevour OS (Arch linux)

### Server Software Version/Commit

e312c4c/release

### What happened?

After closing the server, either with `Ctrl + C` or with the `stop` command, the server stops as expected, but the terminal won't give back echoing. So i could type stuff but it was not showing up on screen. `reset` solved everything.

I was using the default terminal in Endevour OS, called `Konsole`

### To Reproduce

Start the pumkin server, then type `stop` and press enter. After waiting a bit, the server stops and you can see, that you can't see what you are typing.

### Expected behavior

After stopping the server, it should give back echoing, so i could see what am i typing.

yuminstalljoe:
@Barni-Mezei I started experiencing this as well recently. I just made #1839 to resolve it.

PR #1839, opened 2026-03-10
fix: drop receiver on console exit (#1817)

yuminstalljoe:
## Description
Commit 49490237 replaced rustyline-async with rustyline. The old async version could be interrupted via tokio::select!, but the new version uses a blocking readline() in a spawned thread that cannot be interrupted at all times. I started experiencing the same issue frequently from #1817.

# Fix
Drop the channel receiver (rx) when the console task exits. This causes blocking_send() in the spawned thread to fail, breaking the loop and returning the editor so it can be properly dropped (restoring terminal state).

## Testing
- [x] Close terminal with stop
- [x] Close terminal with CTRL+C


Snowiiii:
LGTM. Thank you @yuminstalljoe :D

PR #1841, opened 2026-03-10
feat: add register permission to wasm

BjornTheProgrammer:
## Description

Adds the `register_permission` function to `Context`.
```rust
fn register_permission(&self, permission: &Permission) -> Result<(), String>
```

## Testing

Run on a test server to verify that it works

### DIFF
--- a/pumpkin-plugin-api/src/lib.rs
+++ b/pumpkin-plugin-api/src/lib.rs
@@ -13,7 +13,7 @@ pub mod command {
 
 pub use wit::pumpkin::plugin::{
     context::{Context, Server},
-    text,
+    permission, text,
 };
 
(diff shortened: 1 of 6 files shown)


illyrius666:
i was about to ask for this one haha, also can we have register_command at permission take Permission instead of &str?

BjornTheProgrammer:
> i was about to ask for this one haha, also can we have register_command at permission take Permission instead of &str?

This would make sense, I'll ask what the team thinks about changing it.

BjornTheProgrammer:
Okay, I asked @vyPal and it turns out that permissions are actually scoped, so it doesn't makes sense to register_command accept a Permission struct instead of a string.

PR #1856, opened 2026-03-12
perf: skip random tick on none ticking block

chocodev11:
## Description
- skip random tick for section that contain only none ticking block

## Testing

### DIFF
--- a/pumpkin-world/src/chunk/format/mod.rs
+++ b/pumpkin-world/src/chunk/format/mod.rs
@@ -335,7 +335,10 @@ impl ChunkEntityData {
                 if uuid.len() != 4 {
                     debug!(
                         "Entity in chunk {},{} has invalid UUID array length {}: {:?}",
-                        position.x, position.y, uuid.len(), entity_nbt
+                        position.x,
+                        position.y,
+                        uuid.len(),
+                        entity_nbt
                     );
                     continue;
                 }


RealVaprant:
ok

Snowiiii:
Code looks like a performance slowdown ngl

chocodev11:
> Code looks like a performance slowdown ngl

thanks for your comment, i mimic the "faster palette" from moonrise without considering how more expensive it is in Rust. i also put the has_only_air() out of loop. additionally, i fixed the issue when random tick select exact same 3 position per chunk

chocodev11:
> ## [Argus](https://argus.tyraisolutions.com) Code Review
> **Score:** 75/100
> 
> Reviewed 6 files. Found 4 issues. 1 observations. 1 high. Score: 75/100.
> 
> ### Additional Findings
> _These findings are on lines outside the diff and cannot be shown inline._
> 
> * **[HIGH] Panic when removing palette entry after swap_remove changes indices**
> 
> After swap_remove moves the last element, the code continues to use the original index to access counts array, potentially causing incorrect behavior or out-of-bounds access.
> 
> **Justification:** Line 49 decrements counts[original_index], then line 52 checks if counts[original_index] == 0. After swap_remove at lines 54-55, the element at original_index is replaced with the last element, so the zero check compares against the swapped-in count, not the decremented count.
> 
> **Validation:** Lines 49-56 show the sequence: decrement count, check if zero, then swap_remove. The index used after swap_remove refers to a different element than before the removal.
> 
> **Suggestion:** Store the count before decrementing and handle removal based on that stored value, or restructure logic to avoid index confusion after swap_remove. (`pumpkin-world/src/chunk/palette.rs:49`)
> 
> * **[MEDIUM] Integer overflow in block network serialization shift operation**
> 
> Left shift of u16 value by up to 45 bits could exceed i64 bounds, causing wrap-around and corrupting packed data representation.
> 
> **Justification:** Shift operation `(*value as i64) << (bits_per_entry as u64 * index as u64)` with bits_per_entry=15 and index=3 gives 45-bit shift. 65535 << 45 exceeds i64::MAX, causing wrap to negative value.
> 
> **Validation:** Line 389 shows the problematic shift. With max values, calculation produces 0x1FFFE00000000000 which exceeds i64::MAX (0x7FFFFFFFFFFFFFFF), wrapping to negative value.
> 
> **Suggestion:** Cast value to u64 before shifting, and mask result to ensure it fits within i64 word, or use u64 accumulator. (`pumpkin-world/src/chunk/palette.rs:388`)
> 
> * **[MEDIUM] Entity UUID parsing can panic on malformed input**
> 
> Direct indexing of UUID array without length validation can cause panic if array doesn't have exactly 4 elements, leading to chunk loading crash.
> 
> **Justification:** Lines 336-340 directly index uuid[0], uuid[1], uuid[2], uuid[3] without checking the vector length returned by get_int_array.
> 
> **Validation:** Code shows `let uuid = if let Some(uuid) = entity_nbt.get_int_array("UUID") { Uuid::from_u128((uuid[0] as u128) << 96 | ... ) }` - if uuid.len() != 4, indexing will panic.
> 
> **Suggestion:** Validate that the UUID int array has length 4 before accessing elements, and return an error or skip the entity if invalid. (`pumpkin-world/src/chunk/format/mod.rs:334`)
> 
> * **[MEDIUM] Unsafe unwrap on potentially None chat signature**
> 
> Direct unwrap on chat_message.signature without validation can panic if signature is None, causing denial-of-service for all players.
> 
> **Justification:** Line 517 calls `chat_message.signature.clone().unwrap()` without verifying chat_message.signature is actually Some. Comment claims safety but function doesn't validate.
> 
> **Validation:** Function broadcast_secure_player_chat (lines 482-531) never calls validate_chat_message or checks chat_message.signature.is_some() before unwrap.
> 
> **Suggestion:** Replace unwrap with proper error handling or match statement that validates signature is Some before using it. (`pumpkin/src/world/mod.rs:517`)

the first issue is false positive, the code checks before swap, then re-scans by value.

pumpkin-world/src/level.rs
```diff
@@ -464,32 +470,43 @@ impl Level {
...
                 .extend(chunk.block_entities.lock().unwrap().values().cloned());
 
+            // Acquire the read lock once per chunk to avoid per-section lock overhead
+            let sections = chunk.section.block_sections.read().unwrap();
+            let random_tick_sections = chunk.section.random_tick_sections.read().unwrap();
+            let min_y = chunk.section.min_y;
+
             for i in 0..section_count {
-                let y_base = i as i32 * 16;
+                if !random_tick_sections[i].is_randomly_ticking() {
```
Snowiiii, line 479: Instead of this, can't we just store random ticking blocks inside a Box and just tick them?

Snowiiii:
I also would like to know how much more memory this takes

pumpkin-world/src/level.rs
chocodev11: > Instead of this, can't we just store random ticking blocks inside a Box and just tick them?

it not vanilla behaviour. With cache, it take about 125–156kb per a thousand chunks

Snowiiii: I guess it would cause problem with the ticking logic since you still would need to loop trough the sections, so yeah i'm fine with leaving it just mentioned it, maybe could be useful in the future

Snowiiii:
Overall looks good. Thank you @chocodev11 :D

PR #1862, opened 2026-03-14
feat(world): Add distance-filtered broadcast API and migrate all world-level broadcasts

andreisugu:
### Summary

Introduces `broadcast_packet_nearby`, `broadcast_packet_nearby_except`, and `DEFAULT_ENTITY_TRACKING_DISTANCE_SQ` on `World`, and migrates all world-owned packet broadcasts to use distance filtering instead of sending to every connected player unconditionally.

---

### Problem

Every broadcast in the world module used `broadcast_packet_all` or `broadcast_packet_except`, which iterate over **all connected players** and send the packet regardless of where those players are in the world. This means:

- A block update at coordinate (0, 64, 0) is sent to a player at coordinate (10000, 64, 10000)
- An entity spawn packet is sent to every player in the world, even if they are thousands of blocks away and will never render that entity
- Sound effects are sent to players who are physically incapable of hearing them
- On servers with many players spread across a large map, this creates **O(players × events)** packet volume, most of which is noise

### Reason
This is being done because of the high number of packets clients recieve.

---

### Solution

**New API on `World`:**

```rust
pub const DEFAULT_ENTITY_TRACKING_DISTANCE_SQ: f64 = 128.0 * 128.0; // 128 blocks, 8 chunks

pub async fn broadcast_packet_nearby<P: ClientPacket>(
    &self,
    center: &Vector3<f64>,
    max_distance_squared: f64,
    packet: &P,
)

pub async fn broadcast_packet_nearby_except<P: ClientPacket>(
    &self,
    center: &Vector3<f64>,
    max_distance_squared: f64,
    except: &[uuid::Uuid],
    packet: &P,
)
```

Both methods filter `self.players` by squared distance before collecting recipients, then call the existing `broadcast_java_grouped` for version-batched delivery. Using squared distance avoids a `sqrt` per player per broadcast.

The constant `DEFAULT_ENTITY_TRACKING_DISTANCE_SQ` corresponds to 128 blocks (8 chunks), matching vanilla Minecraft's default entity tracking range.

**Sound distance is volume-aware:**

Rather than using the flat 128-block constant, sound broadcasts compute the audible range from volume, matching vanilla's formula:

```rust
let audible_distance = f64::from(volume.max(1.0)) * 16.0;
let max_distance_sq = audible_distance * audible_distance;
```

A sound with volume 1.0 reaches 16 blocks. Volume 4.0 reaches 64 blocks. This matches how vanilla clients decide whether to play a sound they receive.

---

### Migrated broadcasts

| Method | Old | New | Distance |
|---|---|---|---|
| `send_entity_status` | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |
| `send_remove_mob_effect` | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |
| `send_block_events` | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |
| `update_blocks` (single) | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |
| `update_blocks` (multi) | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |
| `spawn_particle` | unconditional loop | distance-filtered loop | 128 blocks |
| `play_sound` | `broadcast_packet_all` | `broadcast_packet_nearby` | volume × 16 |
| `play_sound_raw_expect` | `broadcast_packet_except` | `broadcast_packet_nearby_except` | volume × 16 |
| `spawn_entity` | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |
| `remove_entity` | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |
| `sync_world_event` | `broadcast_packet_all` | `broadcast_packet_nearby` | 128 blocks |

---

### Changes

| File | What changed |
|---|---|
| `pumpkin/src/world/mod.rs` | Added `DEFAULT_ENTITY_TRACKING_DISTANCE_SQ`, `broadcast_packet_nearby`, `broadcast_packet_nearby_except`; migrated all 11 world-level broadcast sites listed above |

---

### Notes

- `broadcast_packet_all` and `broadcast_packet_except` are intentionally kept — they are still needed for global broadcasts (chat, tab list, game state, etc.)
- This PR deliberately scopes to `world/mod.rs` only. Entity-level migrations are going to come in a future PR (Convert all entity broadcasts to distance-filtered and remove redundant enderman particles)
- The 128-block constant can be made configurable in a follow-up; hardcoding it here matches vanilla behavior and gives a correct baseline

---

### DIFF
--- a/pumpkin-codegen/src/attributes.rs
+++ b/pumpkin-codegen/src/attributes.rs
@@ -6,12 +6,16 @@ use std::collections::BTreeMap;
 use std::fs;
 use syn::LitInt;
 
+/// Raw deserialization shape for a single attribute entry from `attributes.json`.
 #[derive(Deserialize)]
 struct Attributes {
+    /// Numeric registry ID for this attribute.
     id: u8,
+    /// Default numeric value applied to entities that do not override this attribute.
     default_value: f64,
 }
 
+/// Generates the `TokenStream` for the `Attributes` struct and its associated constants.
 pub fn build() -> TokenStream {
     let attributes: BTreeMap<String, Attributes> =
         serde_json::from_str(&fs::read_to_string("../assets/attributes.json").unwrap())
--- a/pumpkin-codegen/src/biome.rs
+++ b/pumpkin-codegen/src/biome.rs
@@ -6,45 +6,69 @@ use quote::{format_ident, quote};
 use serde::Deserialize;
 use syn::LitInt;
 
+/// Raw deserialization shape for a single biome entry from `biome.json`.
 #[derive(Deserialize)]
 pub struct Biome {
+    /// Whether this biome has precipitation (rain or snow).
     has_precipitation: bool,
+    /// Base temperature of the biome affecting weather and foliage color.
     temperature: f32,
+    /// Downfall value affecting how wet the biome is (influences snow/rain).
     downfall: f32,
+    /// Optional modifier that changes how temperature is applied.
     temperature_modifier: Option<TemperatureModifier>,
     //carvers: Vec<String>,
+    /// Nested lists of feature resource-location strings applied during world generation.
     features: Vec<Vec<String>>,
+    /// Probability per chunk tick that a creature spawns, if not overridden per-biome.
     creature_spawn_probability: Option<f32>,
+    /// Spawn group data for each entity category in this biome.
     spawners: SpawnGroups,
+    /// Per-entity spawn cost budget entries, keyed by namespaced entity ID.
     spawn_costs: BTreeMap<String, SpawnCosts>,
+    /// Numeric registry ID assigned to this biome.
     pub id: u8,
 }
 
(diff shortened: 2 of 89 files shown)


andreisugu:
I will make this PR a draft for now, I cant help feeling I am way off.
I have to think about how vanilla does it. I remember something about each entity having their own range?

yuminstalljoe:
> I will make this PR a draft for now, I cant help feeling I am way off. I have to think about how vanilla does it. I remember something about each entity having their own range?

@andreisugu Vanilla maintains a "viewers" set per entity, not distance checks per packet as you mentioned.

andreisugu:
I'll close this PR for now. It will get superseded by another one. This one is too far gone

Issue #1369, opened 2026-01-24
Regression: Chunks will sometimes not load


FireAcidCow:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

afc4cb0

### What happened?

The bug that was patched in 6f18d16 was reintroduced in afc4cb0. The issue also still affects the latest version, 7b107c5

### To Reproduce

Fly in spectator mode until you notice some chunks that won't load

### Expected behavior

All chunks load, as it was in 6f18d16

lineality:
(using prism to play on self-hosted (same laptop) pumpkin server fedora links more specs below)

If this supports the above issue.

1. cargo cleaned and rebuilt after git fetch pull
2. new server made
3. To stress test I pushed video settings to more load 
4. No issues noticed at first, but as I walked around the island I was on, I eventually came to an area where chunks were either not visible or not loading. As before (for me) if I walked out into the nothing, sometimes it would load. But...this time I fell off the map into the under-ness... and started taking dammage but then glitchy-froze.
5. Looking up I could see the missing chunk areas.
6. I logged out and logged back in. I was still in the under-ness, BUT it looks like all the chunks were loading now.
7. But I could not move etc. 
8. Restarted server and same: looks like all chunks loading (but...underness...will need to make new-hello-world.

-- pic, initial load, missing chunks --
<img width="1384" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0b6df7bb-4517-48e3-8c23-5d3dc6c36a03" />

-- reload, chunks appear to be there --
<img width="1384" height="1080" alt="Image" src="https://github.com/user-attachments/assets/72ff9fd4-a27b-4287-9849-cd5ed25884c6" />

I think this version of The Pumpkin: [3776b22](https://github.com/Pumpkin-MC/Pumpkin/commit/3776b225e471084bce9cc33c02d849393565d767)
(just fetch pulled before I tested)
2026 01 27



# System Details Report
---

## Report details
- **Date generated:**                              2026-01-27 18:35:35

## Hardware Information:
- **Hardware Model:**                              Dell Inc. Inspiron 3501
- **Memory:**                                      12.0 GiB
- **Processor:**                                   11th Gen Intel® Core™ i5-1135G7 × 8
- **Graphics:**                                    Intel® Iris® Xe Graphics (TGL GT2)
[...]

PR #1868, opened 2026-03-16
fix(worldgen): Resolve random server stalls and chunk scheduler deadlocks

andreisugu:
> Should solve #1369 

### Why were these changes necessary?

The root cause of the chunk scheduler hangs and IO freezes stemmed from several critical concurrency and synchronization bugs within the chunk generation pipeline:

1.  **Silent IO Read Starvation:** The `io_read_work` loop was failing to properly wait for the `IOLock` when a chunk was currently being written. Instead of blocking and retrying, it would attempt to acquire the lock 10 times, yield, and then silently drop the request via a `continue` statement. Because the scheduler had already incremented its `running_task_count`, it would wait indefinitely for a response that would never arrive, eventually causing the scheduler and IO read watchdog to fail.
2.  **Unbounded Waiting Tasks (Orphaned Tasks):** Tasks waiting on neighbor chunk data were parked in a `waiting_for_chunks` set. If that neighbor data never arrived (due to the IO read drop mentioned above, or a generation panic), those tasks became permanent orphans. The scheduler's empty-queue logic would continuously spin while checking this set, causing a permanent stall.
3.  **Missed Wakeups During Empty Queue Waits:** While the scheduler was waiting for in-flight tasks to complete (when the queue was empty but tasks were running), it failed to poll the `send_level` channel. This meant that if a player moved during this wait phase, the level change was ignored until the wait condition exited, leading to unresponsive chunk loading.
4.  **Infinite Generation Retry Storms:** When a chunk generation task panicked or failed, the error handler would reset the chunk and immediately re-queue it with a priority boost. Consistently failing chunks (e.g., due to deterministic seed issues or corrupt data) would result in an infinite loop that starved healthy generation tasks.
5.  **Scheduler Starvation Under High Load:** The main scheduler loop drained incoming chunk results using an unbounded `while let Ok(...) = try_recv()` loop. On systems with many fast generation threads, this drain could take too long, starving the dispatch of new tasks and delaying the processing of level changes.

### What is the impact of this change?

* **Eliminated Scheduler Deadlocks:** By replacing the busy-wait in `io_read_work` with a proper `Condvar::wait_timeout` and sending a `GenerationFailure` on timeout, the scheduler correctly decrements its `running_task_count` and no longer hangs waiting for lost IO requests.
* **Prevented Orphaned Tasks:** Tracking the insertion time in `waiting_for_chunks` and expiring stale entries after 30 seconds ensures the scheduler will not spin forever on tasks whose dependencies are permanently lost. Downstream dependents are correctly unblocked or dropped.
* **Improved Responsiveness:** Polling `send_level.get()` during the scheduler's wait loop ensures player movements and level changes are processed promptly, even while waiting for other chunks to generate.
* **Stabilized World Generation:** Introducing a `retry_count` prevents infinite retry loops. Failing chunks are penalized with exponential backoff and eventually abandoned after 5 attempts, preventing broken chunks from halting the entire server's world generation.
* **Balanced Throughput:** Capping the `try_recv()` drain to 64 iterations prevents fast generation threads from starving the scheduler thread, ensuring a steady flow of both task dispatch and result processing.

### Are there any known issues or limitations?

Yes. While these changes resolve the immediate hangs and freezes, an underlying architectural limitation remains regarding the DAG (Directed Acyclic Graph) explosion caused by `StructureReferences`. (see my other PR #1888 )

Because `StructureReferences` requires a radius of 8, every chunk needing this stage pulls in a 17x17 grid (289 neighbor chunks) as dependencies. For multiple players with high view distances, this cascades into tens of thousands of DAG nodes and edges. This results in significant memory and CPU pressure during priority sorting and node traversal. This PR mitigates the symptoms (expiring stale tasks, bounded retries, responsive scheduler) but does not fix the root cause of the massive DAG growth, which will require a larger redesign of the structure placement strategy.

### DIFF
--- a/pumpkin-world/src/chunk_system/schedule.rs
+++ b/pumpkin-world/src/chunk_system/schedule.rs
@@ -441,7 +441,7 @@ impl GenerationSchedule {
         // 3. Process expired nodes (Re-queue them instead of abandoning)
         for &node_key in &expired {
             if let Some(node) = self.graph.nodes.get_mut(node_key) {
-                warn!(
+                debug!(
                     "Task for chunk {:?} stage {:?} timed out waiting for neighbors. Re-queueing.",
                     node.pos, node.stage
                 );


andreisugu:
As much as I like the fix, I'll keep improving a while longer (not much, just to make sure this won't be a problem)

andreisugu:
# Development Update: Scheduler Stability & Chunk Resilience

This update focuses on hardening the core world-generation scheduler, resolving critical hangs in the Directed Acyclic Graph (DAG) logic, and improving data safety during unexpected panics.

## Scheduler Resilience & Panic Handling
Improved the system's ability to recover from failures without losing progress or causing "cascading" crashes.
* **Panic Recovery:** The generation state now captures the `Cache` via `&mut` borrow during panics. This allows the system to restore chunk data into the `ChunkHolder`, preventing permanent data loss.
* **DAG Integrity:** Failed tasks are now re-queued instead of dropped. This prevents "shockwave" failures where a single node drop would break dependencies for all neighboring chunks.
* **Timeout Handling:** Eliminated the 30s worker thread freezes by implementing precise delivery hooks for neighbor boundaries and applying priority penalties to re-queued timed-out tasks.

## Core Scheduler & DAG Logic
Significant refactoring of how tasks are queued and dependencies are tracked.
* **Iterative Graph Traversal:** Replaced linear inner iterations in `resort_work` with a more efficient iterative `pending_deps` queue.
* **Dependency Chain Logic:** Updated `ensure_dependency_chain` to return a `Vec` of newly allocated tasks, enabling more robust recursive processing.
* **Edge Pruning:** Resolved infinite re-queueing loops by implementing logic to prune dead dependency edges.
* **Consistency:** Aligned `in_degree` counts through guaranteed edge assignments during task propagation, ensuring the scheduler never "abandons" dynamically scaled tasks.

## Performance & Sync Optimizations
Under-the-hood improvements to memory usage and thread synchronization.
* **Tokio Integration:** Replaced standard `Condvar` with `Tokio's Notify` for better integration with our asynchronous runtime.
* **IOLock Refactor:** Updated `IOLock` to use `usize` for chunk data management, reducing overhead.
* **Queue Management:** General optimizations to task queueing and IO handling to reduce contention.

---
### Commit Summary
* `78f3226`: fix(world): resolve scheduler panics and improve chunk generation resilience
* `5dc44b9`: fix(chunk): update IOLock to use usize for chunk data and replace Condvar
* `619092d`: fix(schedule): stop infinite re-queueing by pruning dead dependency edges
* `695cce0`: fix(chunk): improve IO handling and optimize task queue management
* `3c0effc`: fix: resolve chunk scheduler hang and missing temporal DAG edges

Q2297045667:
I'll find some friends from the Chinese community to help with the testing.

andreisugu:
> I'll find some friends from the Chinese community to help with the testing.

Sure! The state of the PR right now is pretty ok functionality-wise. As promised, no more timeouts/stalls or crashes.
After latest commit, there is but one issue: the scheduler is still triggering timeouts regarding StructureReferences.
No active memory/cpu leaks though. The high numbers of "Packets Received" I think is unrelated (It's related to mobs packets, a running problem in all recent pumpkin artifacts)

Morb0:
It seems that right now, the server isn't playable at all without this PR, since chunk generation and mobs just freezes. But I compiled locally this branch and tried playing - for some reason, chunk generation is really slow... Is that to be expected? (I was just expecting to see something like what's in the GIF in the README, but now it slwoer than on vanilla core)

andreisugu:
> It seems that right now, the server isn't playable at all without this PR, since chunk generation and mobs just freezes. But I compiled locally this branch and tried playing - for some reason, chunk generation is really slow... Is that to be expected? (I was just expecting to see something like what's in the GIF in the README, but now it slwoer than on vanilla core)

Sorry, this PR mostly solves the scheduler issues. The actual performance boost is in my new PR #1888 
So yea, when these 2 will be combined we will be basically good to go

Q2297045667:
According to the testing feedback from the Chinese community, the memory leak issue still persists. After the server is running and players join, even after players completely exit, the memory continues to grow slowly. After several hours of observation, the memory increases by several tens of MB per hour on average.

The issue of block freezing still persists.

andreisugu:
> According to the testing feedback from the Chinese community, the memory leak issue still persists. After the server is running and players join, even after players completely exit, the memory continues to grow slowly. After several hours of observation, the memory increases by several tens of MB per hour on average.
> 
> The issue of block freezing still persists.

You're completely right, and I really appreciate the Chinese community testing this so thoroughly. I got ahead of myself and over-promised on the memory and stall fixes... I apologize for jumping the gun there.

The reality is this PR suffered from major scope creep. I tried to rewrite the traversal logic when I should have just patched the crashes. I'm going to close this and open a new PR focused strictly on the minimum viable product: ensuring the server doesn't panic or permanently deadlock when chunk generation fails. Thank you for your patience and for holding me to a high standard!

PR #1872, opened 2026-03-17
feat: implement save-all/off/on, reload, debug, perf commands

WhiteProject1:
- `/save-all [flush]`: saves the server to disk, flush variant also saves entities
- `/save-off`: disables automatic chunk saving
- `/save-on`: re-enables automatic chunk saving
- `/reload`: placeholder for reloading loot tables, advancements, functions
- `/debug`: starts or stops a debug profiling session
- `/perf`: captures info and metrics about the server

### Infrastructure changes
- `Level.saving_disabled: AtomicBool` — controls autosave without blocking explicit saves
- `World::save_entity` made public for `/save-all flush`

| Command | OP Level |
|---------|----------|
| `/save-all` | 4 |
| `/save-off` | 4 |
| `/save-on` | 4 |
| `/perf` | 4 |
| `/reload` | 3 |
| `/debug` | 3 |

Partially addresses #15.


pumpkin/src/command/commands/reload.rs
```diff
@@ -0,0 +1,36 @@
...
+
+impl CommandExecutor for ReloadExecutor {
+    fn execute<'a>(
+        &'a self,
+        sender: &'a CommandSender,
+        _server: &'a crate::server::Server,
+        _args: &'a ConsumedArgs<'a>,
+    ) -> CommandResult<'a> {
+        Box::pin(async move {
+            // TODO: Actually reload datapacks when datapack system is implemented
```
SomeYellowGuy: What's the point of making a command that doesn't do anything? In my opinion, if a command cannot be implemented yet, don't create a file for it.

pumpkin/src/command/commands/perf.rs
```diff
@@ -0,0 +1,68 @@
...
+    fn execute<'a>(
+        &'a self,
+        sender: &'a CommandSender,
+        server: &'a crate::server::Server,
+        _args: &'a ConsumedArgs<'a>,
+    ) -> CommandResult<'a> {
+        Box::pin(async move {
+            let _mspt = server.get_mspt();
+
+            // TODO: track actual profiling duration and tick count
```
SomeYellowGuy: I'm not sure if this command is even _worth_ implementing, because this requires Minecraft's entire profiling system, which I'm pretty sure Pumpkin doesn't have.

pumpkin/src/command/commands/save.rs
```diff
@@ -0,0 +1,155 @@
...
+            for world in server.worlds.load().iter() {
+                world.level.should_save.store(true, Ordering::Relaxed);
+                world.level.level_channel.notify();
+            }
+
+            if flush {
+                // Flush: save entities and wait for chunk saver
+                for world in server.worlds.load().iter() {
+                    for entity in world.entities.load().iter() {
+                        world.save_entity(entity).await;
```
SomeYellowGuy: Instead of exposing `save_entity`, maybe it would be better to create a public dedicated `save_everything` kind of function, which could save all worlds, their chunks and entities.

pumpkin/src/command/commands/debug.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl CommandExecutor for StopExecutor {
+    fn execute<'a>(
+        &'a self,
+        sender: &'a CommandSender,
+        _server: &'a crate::server::Server,
+        _args: &'a ConsumedArgs<'a>,
+    ) -> CommandResult<'a> {
+        Box::pin(async move {
+            // TODO: track actual profiling duration and tick count
```
SomeYellowGuy: Same as `/perf`, there is no profiling system in Pumpkin, so this is not worth implementing.

Snowiiii:
make Atomic PRs

PR #1875, opened 2026-03-17
feat: implement clone, fillbiome, spreadplayers, random commands

WhiteProject1:
- `/clone`: copy blocks from one region to another with filtered/masked/replace modes
- `/fillbiome`: fill a region with a specific biome
- `/spreadplayers`: teleport entities to random surface locations in an area
- `/random`: value and roll subcommands with range parsing and broadcast

| Command | OP Level |
|---------|----------|
| `/random` | 0 |
| `/clone` | 2 |
| `/fillbiome` | 2 |
| `/spreadplayers` | 2 |

Partially addresses #15.

### DIFF
--- a/pumpkin-world/src/chunk/mod.rs
+++ b/pumpkin-world/src/chunk/mod.rs
@@ -408,7 +408,7 @@ impl ChunkSections {
     }
 
     pub fn set_relative_biome(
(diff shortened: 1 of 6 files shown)


Snowiiii:
make Atomic PRs

PR #1876, opened 2026-03-17
feat: implement execute, attribute, item commands + translation fixes

WhiteProject1:
- `/execute`: run commands with conditions (`if`/`unless`) and context modifiers
- `/attribute`: query, add, remove, or set entity attributes
- `/item`: manipulate items in entity/block inventories

### Translation fixes
- Replace hardcoded English error messages with proper translation key constants in:
  - `fill.rs` (pos out of bounds, too big, failed, success)
  - `setworldspawn.rs` (success message key)
  - `spawnpoint.rs` (success message key)
  - `teleport.rs` (invalid position, requires player)

| Command | OP Level |
|---------|----------|
| `/execute` | 2 |
| `/attribute` | 2 |
| `/item` | 2 |

Partially addresses #15.

### DIFF
--- a/pumpkin/src/command/commands/attribute.rs
+++ b/pumpkin/src/command/commands/attribute.rs
@@ -1,0 +1,289 @@
+use std::sync::Arc;
+
+use pumpkin_data::attributes::Attributes;
+use pumpkin_data::translation;
+use pumpkin_util::text::TextComponent;
+
+use crate::command::args::entity::EntityArgumentConsumer;
+use crate::command::args::simple::SimpleArgConsumer;
+use crate::command::args::{Arg, ConsumedArgs, FindArg};
+use crate::command::dispatcher::CommandError;
+use crate::command::tree::CommandTree;
+use crate::command::tree::builder::{argument, argument_default_name, literal};
+use crate::command::{CommandExecutor, CommandResult, CommandSender};
+use crate::entity::EntityBase;
+
+const NAMES: [&str; 1] = ["attribute"];
+
+const DESCRIPTION: &str = "Queries, adds, removes, or sets an entity attribute.";
+
+const ARG_ATTRIBUTE: &str = "attribute";
+const ARG_SCALE: &str = "scale";
+const ARG_VALUE: &str = "value";
+
+fn lookup_attribute(name: &str) -> Option<Attributes> {
+    let name = name.strip_prefix("minecraft:").unwrap_or(name);
+    match name {
+        "armor" => Some(Attributes::ARMOR),
+        "armor_toughness" => Some(Attributes::ARMOR_TOUGHNESS),
+        "attack_damage" => Some(Attributes::ATTACK_DAMAGE),
+        "attack_knockback" => Some(Attributes::ATTACK_KNOCKBACK),
+        "attack_speed" => Some(Attributes::ATTACK_SPEED),
+        "block_break_speed" => Some(Attributes::BLOCK_BREAK_SPEED),
+        "block_interaction_range" => Some(Attributes::BLOCK_INTERACTION_RANGE),
+        "burning_time" => Some(Attributes::BURNING_TIME),
+        "camera_distance" => Some(Attributes::CAMERA_DISTANCE),
+        "explosion_knockback_resistance" => Some(Attributes::EXPLOSION_KNOCKBACK_RESISTANCE),
+        "entity_interaction_range" => Some(Attributes::ENTITY_INTERACTION_RANGE),
+        "fall_damage_multiplier" => Some(Attributes::FALL_DAMAGE_MULTIPLIER),
+        "flying_speed" => Some(Attributes::FLYING_SPEED),
+        "follow_range" => Some(Attributes::FOLLOW_RANGE),
+        "gravity" => Some(Attributes::GRAVITY),
+        "jump_strength" => Some(Attributes::JUMP_STRENGTH),
+        "knockback_resistance" => Some(Attributes::KNOCKBACK_RESISTANCE),
+        "luck" => Some(Attributes::LUCK),
+        "max_absorption" => Some(Attributes::MAX_ABSORPTION),
+        "max_health" => Some(Attributes::MAX_HEALTH),
+        "mining_efficiency" => Some(Attributes::MINING_EFFICIENCY),
+        "movement_efficiency" => Some(Attributes::MOVEMENT_EFFICIENCY),
+        "movement_speed" => Some(Attributes::MOVEMENT_SPEED),
+        "oxygen_bonus" => Some(Attributes::OXYGEN_BONUS),
+        "safe_fall_distance" => Some(Attributes::SAFE_FALL_DISTANCE),
+        "scale" => Some(Attributes::SCALE),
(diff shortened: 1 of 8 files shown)


pumpkin/src/command/commands/execute.rs
```diff
@@ -0,0 +1,188 @@
...
+                format!("/{command}")
+            };
+
+            let dispatcher = server.command_dispatcher.read().await;
+            let mut success = 0i32;
+
+            for target in targets {
+                let entity = target.get_entity();
+                // Try to get player sender for this entity
+                if let Some(player) = entity.world.load().get_player_by_id(entity.entity_id) {
```
SomeYellowGuy, line 76: `/execute as` is **not limited to players**. It can be used on any entity.

pumpkin/src/command/commands/execute.rs
```diff
@@ -0,0 +1,188 @@
...
+        .then(literal("as").then(
+            argument(ARG_TARGETS, EntitiesArgumentConsumer).then(
+                literal("run").then(argument(ARG_COMMAND, MsgArgConsumer).execute(AsExecutor)),
+            ),
+        ))
+        .then(
+            literal("if").then(
+                literal("block").then(
+                    argument("pos", BlockPosArgumentConsumer).then(
+                        argument("block", crate::command::args::simple::SimpleArgConsumer)
```
SomeYellowGuy, line 163: We already have a `BlockPredicateArgumentConsumer`. Why is it not used here?

pumpkin/src/command/commands/execute.rs
```diff
@@ -0,0 +1,188 @@
...
+                    .await;
+            }
+            Ok(1)
+        })
+    }
+}
+
+pub fn init_command_tree() -> CommandTree {
+    CommandTree::new(NAMES, DESCRIPTION)
+        .then(literal("run").then(argument(ARG_COMMAND, MsgArgConsumer).execute(RunExecutor)))
```
SomeYellowGuy, line 153: When the command *to be run* is specified via the `run` subcommand, you treat it as a string. This loses many of the robust features of Brigadier that make Minecraft's command system very ergonomic. This is an issue with the current system, which does not allow any command-level nodes to be nested.

pumpkin/src/command/commands/execute.rs
```diff
@@ -0,0 +1,188 @@
...
+                literal("run").then(argument(ARG_COMMAND, MsgArgConsumer).execute(AsExecutor)),
+            ),
+        ))
+        .then(
+            literal("if").then(
+                literal("block").then(
+                    argument("pos", BlockPosArgumentConsumer).then(
+                        argument("block", crate::command::args::simple::SimpleArgConsumer)
+                            .then(
+                                literal("run").then(
```
SomeYellowGuy, line 165: `/execute` works very differently from other commands. Context modifiers can be *chained*. Take this command (which should work for your implemented part) for instance:
```
execute if block ~ ~ ~ air if block ~ ~-1 ~ glass run clear
```
This would not work with your implementation because I chained 2 `if` subcommands. In fact, I can put any valid `/execute` subcommand in Minecraft (`as`, `at`, `if`, `unless`, etc.) in any order (except `run` at the end, and it may not even be present), and it would still be valid.
Unfortunately, this is a fundamental issue with the currently (mostly) used command system that would make node redirection and forking very hacky.

pumpkin/src/command/commands/item.rs
```diff
@@ -0,0 +1,154 @@
...
+            Ok(changed)
+        })
+    }
+}
+
+pub fn init_command_tree() -> CommandTree {
+    CommandTree::new(NAMES, DESCRIPTION).then(
+        literal("replace").then(
+            literal("entity").then(
+                argument(ARG_TARGETS, PlayersArgumentConsumer).then(
```
SomeYellowGuy, line 142: Again, `/item`'s `target` argument is **not limited to players**. It works with any entity.

pumpkin/src/command/commands/item.rs
```diff
@@ -0,0 +1,154 @@
...
+                    argument(ARG_SLOT, SimpleArgConsumer).then(
+                        literal("with").then(
+                            argument(ARG_ITEM, SimpleArgConsumer)
```
SomeYellowGuy, line 145: We have `ItemArgumentConsumer`, why don't you use it?

pumpkin/src/command/commands/item.rs
```diff
@@ -0,0 +1,154 @@
...
+                                .then(
+                                    argument(ARG_COUNT, SimpleArgConsumer)
```
SomeYellowGuy, line 147: We have `BoundedNumArgumentConsumer` (limited from 1 to 99, inclusive). Why don't you use it?

pumpkin/src/command/commands/execute.rs
```diff
@@ -0,0 +1,188 @@
...
+use crate::command::args::position_block::BlockPosArgumentConsumer;
+use crate::command::args::{Arg, ConsumedArgs, FindArg};
+use crate::command::dispatcher::CommandError;
+use crate::command::tree::CommandTree;
+use crate::command::tree::builder::{argument, literal};
+use crate::command::{CommandExecutor, CommandResult, CommandSender};
+
+const NAMES: [&str; 1] = ["execute"];
+
+const DESCRIPTION: &str = "Executes a command.";
```
SomeYellowGuy: I feel this is a little vague. `/execute` allows changing the command context, storing values and adding preconditions.

pumpkin/src/command/commands/attribute.rs
```diff
@@ -0,0 +1,292 @@
...
+
+const ARG_ATTRIBUTE: &str = "attribute";
+const ARG_SCALE: &str = "scale";
+const ARG_VALUE: &str = "value";
+
+#[allow(clippy::match_same_arms)]
+fn lookup_attribute(name: &str) -> Option<Attributes> {
+    let name = name.strip_prefix("minecraft:").unwrap_or(name);
+    match name {
+        "armor" | "generic.armor" => Some(Attributes::ARMOR),
```
SomeYellowGuy: The `generic` prefix was removed in 1.21.2, so this is unnecessary.
I also think it's weird to have an attribute lookup in the command implementation, but if this was not implemented by anyone else, it's not your fault.

pumpkin/src/command/commands/attribute.rs
```diff
@@ -0,0 +1,292 @@
...
+
+            let value = living.get_attribute_value(&attribute) * scale;
+
+            sender
+                .send_message(TextComponent::translate(
+                    translation::COMMANDS_ATTRIBUTE_VALUE_GET_SUCCESS,
+                    [
+                        TextComponent::text(attr_name.to_string()),
+                        TextComponent::text(entity_name.to_string()),
+                        TextComponent::text(format!("{value:.2}")),
```
SomeYellowGuy: Why reduce the number of shown decimal digits to 2 at most? In Minecraft, there is no such decimal digit limit.

SomeYellowGuy (commented):
I mostly reviewed the `/execute` and `/item` commands, so I might have missed some things.

Snowiiii:
make Atomic PRs

Issue #690, opened 2025-04-01
Server sends SPlayerPosition to all players instead of only those who need it


BigPebble123:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Is your feature request related to a problem? Please describe.

https://github.com/Pumpkin-MC/Pumpkin/blob/2aa1ff6631c244bd7a55da8c0cd87e2b59c68dde/pumpkin/src/net/packet/play.rs#L209-L222

Here, when receiving `SPlayerPosition` from a player, we send an entity position update packet (`CUpdateEntityPos`) to all online players, even those who are not within the render distance.

This is bad for two reasons:
1. network usage will be incredibly high with a position packet that is sent a dozen times a second by a single connected player
2. it reveals the position of all players to the person who sniffs the network traffic (this can be bad on servers where not knowing the opponent's position is important)

### Describe the solution you'd like

Calculate which players are within the moving player's render distance and send `CUpdateEntityPos` packets only to them.

PR #1883, opened 2026-03-18
perf(network): migrate world broadcasts to O(V) chunk-based subscriptions

andreisugu:
> This PR supersedes #1862 as a corrected implementation of the proposed network optimization.
> Directly solves #690 

**What was changed?**
* Introduced `broadcast_to_chunk` and `broadcast_to_chunk_except` in `World` to utilize Chebyshev distance math (Chunk Grid) for highly optimized packet routing.
* Migrated block updates, block entity data, and synced block events to strictly broadcast to players with the target chunk loaded.
* Optimized entity network overhead by chunk-filtering movement, rotation, metadata, spawning, and equipment packets (in `mod_entity.rs` and `player.rs`).
* Refactored sound and particle events to use volume-aware, chunk-filtered loops instead of global server iterations.
* Handled complex edge cases, ensuring passenger mounting/dismounting, player bed animations, and intra-dimension teleportation utilize localized broadcasts.

**How does this achieve Vanilla Parity?**
* **Chunk Math:** Broadcasts now use Chebyshev distance (`max(abs(dx), abs(dz)) <= view_distance`), which perfectly matches Minecraft Java Edition's chunk grid loading and tracking logic.
* **Sound Distance:** Localized sounds now accurately calculate their audible range using Vanilla's exact formula (`volume.max(1.0) * 16.0` blocks), ensuring players only receive sound packets they can actually hear in-game.
* **Entity Tracking:** By bounding entity updates to the player's `watched_section` (Cylindrical view), we mimic Vanilla's `EntityTracker` viewership boundaries natively.

Previously, the server relied heavily on `broadcast_packet_all` and `broadcast_packet_except` for local world events. This meant that every time a zombie moved, a block was broken, or a sound was played, the server was forced to iterate over the entire global player list to either send the packet or perform floating-point distance math. This created an $O(P)$ (Players) scaling bottleneck per event, causing severe CPU overhead on servers with high player counts.

This drastically improves network and CPU performance for high-population servers. Local broadcasts now scale at $O(V)$ (Viewers). By checking against `player.watched_section` and calculating Chebyshev distance, the server only processes and sends packets to players who are physically close enough to care about the event.

* Global packets like `CPlayerInfoUpdate` (Tab list), `CRemovePlayerInfo` (Disconnects), and Chat messages are intentionally left as global broadcasts.
* **Future Optimization (Entity Viewers):** Currently, Pumpkin ties visibility to the player's `watched_section` (Cylindrical view). Because we aren't maintaining isolated `HashSet`s for individual entity viewers yet, this implementation relies on fast chunk-math filtering rather than a pure O(1) ECS lookup. 
    * *Note for future PRs:* Migrating from this cylindrical math-filtering approach to a dedicated, ECS-driven EntityTracker (where entities maintain exact lists of their viewers) would completely eliminate the math overhead, pushing this from O(V) filtering to true O(1) routing. However, the performance gains of the current approach over global iteration are still massive.

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -3041,12 +3041,12 @@ impl World {
                     broken_state_id.into(),
                     false,
                 );
-                let chunk_pos = position.chunk_position(); // Get chunk pos
+                let chunk_pos = position.chunk_position();
                 match cause {
                     Some(player) => {
                         self.broadcast_to_chunk_except(
                             chunk_pos,
-                            &[player.gameprofile.id],
+                            &[player.living_entity.entity.entity_uuid],
                             &particles_packet,
                         )
                         .await;


yuminstalljoe:
At scale, the constant costs of maintaining exact viewer sets can outweigh the lightweight Chebyshev math used in O(V) filtering. I think this should be the method going forward instead of O(1).


Approach | Strength | Cost
-- | -- | --
O(V) filtered broadcast | simple, Vanilla-aligned, easy to reason about | small per-event math/filter cost
O(1) viewer lookup | faster direct routing once viewer sets exist | constant maintenance overhead, more memory, more edge cases

andreisugu:
No reason going for O(1) right now anyway, this should suffice
I'll clean up the code and then I think this should be done

Snowiiii:
LGTM. Thank you @andreisugu :D

PR #1888, opened 2026-03-19
fix/perf(worldgen): eliminate DAG deadlocks and massively optimize structure references O(N^2) to O(1), and implement global stronghold cache

andreisugu:
**What was changed?**
* **DAG Lock Removal:** Reduced the `StructureReferences` stage dependency radius in `chunk_state.rs` from `8` to `0`.
* **Predictive Math:** Extracted Vanilla-accurate region seed math into a new public function `get_structure_chunk_in_region`.
* **Lazy Evaluation:** Introduced `lazily_generate_structure`, which calculates biomes mathematically using the `BiomeSupplier` and `MultiNoiseSampler` instead of querying generated chunks.
* **Refactored `set_structure_references`:** Completely removed the `GenerationCache` requirement. The function now iterates over a 3x3 mathematical region grid to find potential structures, evaluating their bounding boxes lazily.
* **Global Stronghold Cache:** Created a thread-safe `GlobalStructureCache` utilizing `std::sync::OnceLock` in `placement.rs` to mathematically pre-calculate global structure coordinates for `ConcentricRings` (Strongholds).
* **State Management:** Added `global_structure_cache` to the `VanillaGenerator` state so its lifetime persists across the entire world generation process. Threaded the cache reference through the chunk pipeline (`generation.rs`, `worker_logic.rs`, and `generation_cache.rs`).
* **Resolved Strongholds:** Updated `ProtoChunk::set_structure_references` and `ProtoChunk::set_structure_starts` to resolve `ConcentricRings` placements against the global cache.
* **Precise Reservoir Sampling:** Implemented the exact 112-block biome search grid and short-circuit evaluation logic used by Vanilla's `BiomeSource.findBiomeHorizontal`.

The structure generation system was triggering a massive DAG explosion and thread deadlocks. Because `StructureReferences` required a chunk radius of 8, a single chunk entering this stage forced the scheduler to lock and wait for a 17x17 grid (289 chunks) to reach the `StructureStart` stage. If multiple players moved, this cascaded into tens of thousands of required chunks, exhausting the thread pool, blowing up memory, and causing permanent server stalls. 

Resolving this required shifting from a neighbor-chunk-lookup approach to predictive mathematical bounds. However, `StructurePlacementType::ConcentricRings` (Strongholds) do not use localized grid spacing; they require calculating 128 positions across 8 expanding rings for the entire world. Without a centralized cache, evaluating Strongholds under the new 0-radius system would require either reintroducing the DAG thread deadlocks or performing expensive trigonometric calculations on every single chunk generation step.

This PR serves as a critical fix, a massive performance enhancement, and a parity restorer:
* **Eliminated Deadlocks & Increased Stability:** The circular dependencies in the chunk scheduler are completely severed. Chunks transition instantly into the `StructureReferences` stage without waiting on neighbors. The `tokio` and `rayon` thread pools will never stall waiting on stronghold boundaries.
* **Monumental Performance Boost (perf):** The algorithm shifts from $O(N^2)$ (instantiating and querying 289 neighboring chunks) to $O(1)$. It now performs just 9 fast mathematical region checks per standard structure type, and checks a cached `Vec` bounds filter for Stronghold intersections. The heavy trigonometry required for ring placement only executes exactly once per world lifecycle.
* **Memory Reduction:** The server no longer needs to hold thousands of `ProtoChunk` instances in memory just to resolve structure boundaries, significantly dropping RAM usage during worldgen.
* **Vanilla Logic Parity:** Mathematically restores Vanilla stronghold placement logic (128 strongholds in 8 rings, identical RNG branching, and identical target center points).

* **Minor Stronghold Offset (Terrain Data Desync):** While the core stronghold generation math (angles, distances, RNG sequence, 112-block search radius, and target chunk centers) is now 100% mathematically identical to Vanilla Java, the final generated coordinates can still be slightly off by a few chunks compared to Vanilla's `/locate` command. This is no longer a math or logic bug, but a terrain data desync. Because our `MultiNoiseSampler` output or `#minecraft:has_structure/stronghold` biome tags slightly disagree with Vanilla 1.21 data, the perfectly-synced Reservoir Sampling loop is occasionally forced to snap the stronghold to a different valid biome within the search grid. Achieving 1:1 absolute block parity will require a separate pass to perfectly sync our noise router parameters and biome maps with Vanilla JSONs. No further work will be done on this specific issue in this PR.


andreisugu:
I've tackled the Stronghold Regression and decided that it isnt worth another PR just to solve a problem I made.
As such, i've tacked on the fix too
Edit: i'm also hoping to achieve full stronghold placement parity..

andreisugu:
I have tried my best to get the stronghold generation as close as possible to actual Vanilla Minecraft. Right now we are still off by a little bit, but I cant seem to find out why, there might be something wrong to how we see biomes internally. I've followed the minecraft decompiled source code as closely as possible.
I hope this will suffice for now.

Snowiiii:
Definitly improved performance by a lot, Thank you @andreisugu :D

PR #1893, opened 2026-03-21
fix(worldgen): permanent server freeze and transition to async-native notification model

andreisugu:
> This is the precise fix that fixed #1868 , but without any of the failsafes or huge rewrites that PR had.
> Should solve #1369 

**What was changed?**
* **`schedule.rs`**: Added active polling via `self.resort_work(self.send_level.get())` inside the empty-queue wait loop of the scheduler.
* **`worker_logic.rs`**: Moved the initialization of the async message channel (`let (t_send, mut t_recv) = tokio::sync::mpsc::channel(1);`) inside the `while` loop within `io_read_work` so a fresh channel is created per chunk.

The server was experiencing a silent, permanent freeze in chunk generation after completing an initial batch, with no CPU spikes or error logs. This was caused by two intertwined deadlocks:
1. **Scheduler Sleep-Lock**: When the chunk queue emptied, the scheduler entered a `sleep` loop but failed to poll for new level changes. If a player moved after the queue was empty, the scheduler ignored the movement and stayed asleep.
2. **Worker Deadlock**: The `io_read_work` function reused a single Tokio `mpsc` channel across an infinite loop. If a chunk read was delayed or dropped, the `.recv().await` call would hang indefinitely, permanently trapping the worker thread and starving the server of IO workers.

* Restores continuous and reliable chunk generation as players explore the world.
* Eliminates silent thread deaths in the IO workers by ensuring a clean state (fresh channel) for every async fetch operation.
* Follows the recommended concurrent async patterns for `tokio::sync::mpsc` when bridging tasks.

None. This is a surgical fix specifically targeting the idle deadlocks without altering the broader priority or retry logic.


Q2297045667:
I will ask some friends from Chinese communities to help with testing, but I may not be able to reply promptly.

Also, how would this impact performance?

andreisugu:
All this is meant to impact is that the server doesn't timeout entirely and permanently anymore. Nothing more than that.

Q2297045667:
The issue still appears to exist, but according to tester feedback, it has significantly improved compared to before. I think this PR has partially addressed the problem, but it seems some issues remain. The test lasted 20 minutes.
Test Personnel @MDZZDYXCMDZZDYXC 

The card returned to normal after tens of seconds.
Image_1774156626296_786


Image_1774156469501_314

Image_1774156323763_109


The tree generation seems a bit odd. I'm not sure if it's caused by this PR.

Image_1774156318122_695
Image_1774156315155_415

Q2297045667:
In summary, I think this fix is quite successful. The next step is to wait for Alex's review.

Finally, thank you once again for your contribution to Pumpkin

Snowiiii:
Two things worth noting though:
- yield_now() won't help starve-prevention for generation threads if they run on a separate thread pool, it only yields to other async tasks on the same executor.
- block_in_place + `Condvar` for the lock is still blocking the async executor. Consider replacing with `tokio::sync::Notify` for a fully async-native solution.

andreisugu:
I mean sure, ill do that, though in the end its also just how i found the system, so i decided to be cautios xd
You're right though, we do need those fixes, first of all because condvar isnt even in the "contributing" section of pumpkin, so idk why we had that there

andreisugu:
@Snowiiii I have commited the proposed changes as follows
* Swapped out `std::sync::Condvar` for `tokio::sync::Notify` in the `IOLock`.
* Stripped out the blocking wrappers in `io_read_work` and replaced them with a proper `.notified().await` loop.
* Updated `io_write_work` to trigger `notify_waiters()` on release.

This should cleanly resolve the thread starvation issue without tying up the Tokio worker threads. Let me know if you spot anything else!
If nothing else, at least this is good practice and the proper way to do this.

MDZZDYXCMDZZDYXC:
The chunk glitch has been present for 10 minutes of map exploration. Flying with an elytra causes stutters in chunk loading and fireworks, and there’s no crash at all.
image
image

andreisugu:
Yes, you are quite right. The current pumpking server has 3 diffrent problems that stack upon each other:
- the starvation (HANDLED HERE): Once all chunks get generated, the scheduler stops checking for new chunks out of an logical error
- the erroring out chunk (partly handled here): because of the huge requirement the structures have for checking a 17x17 area for EVERY SINGLE CHUNK, chunks can and often do just error out. Thats why #1888 was born.
- the speed: also because of dag explosion, solved in #1888 

And just as a final note, this patch was meant as a band aid for the scheduler system. In no way shape or form did I have as my intentions to redo the scheduling system, It's a mess I dont feel qualified enough to fix. Once #1888 gets merged and this gets merged, everything should be muchhhhh much more smoother, just as can be seen in a build i also sent on discord: https://github.com/andreisugu/Pumpkin/actions/runs/23432200378#artifacts

EDIT: I'll investigate the tree problem separatedly. I don't think this is within the scope of the PR. And I don't want huge scope creep here..

Snowiiii:
@andreisugu not sure if you noticed but currently we allocate threads for every Level we create which results in many many mostly dead threads especially when there is no player in the world which often happens when having a level for every dimension (overworld, nether, end). This is a issue I already wanted to fix a long time ago, The best solution would be to move to a global thread pool, this would drastically increase performance and improve thread shutdown behavior

Q2297045667:
Can you provide a before-and-after performance benchmark test?
I'll do some testing later today to see how it performs.

IGN-Styly:
@MDZZDYXCMDZZDYXC could you try to repro the trees with master?

Snowiiii:
@andreisugu this seems to significantly slowed down chunk loading/generation especially in debug builds

MDZZDYXCMDZZDYXC:
> @MDZZDYXCMDZZDYXC could you try to repro the trees with master?

Sure, I'll give it a try.

MDZZDYXCMDZZDYXC:
> > @MDZZDYXCMDZZDYXC could you try to repro the trees with master?
> 
> Sure, I'll give it a try.

I spent 10 minutes exploring the map in Minecraft. Chunk generation is way too slow in debug mode, and in release mode, the game freezes immediately upon entering.
image

andreisugu:
Will be reverting all changes starting with [c5d06ba](https://github.com/Pumpkin-MC/Pumpkin/pull/1893/commits/c5d06bad3f2279b2157bd257bd770cbf9d9b60b8) . After which, I will only be implementing Task Dispatching without any other fluff.
My hope: to not introduce any more regressions. 
My plan -> ~~Tokio Global Thread Pool: I will be leveraging tokio::task::spawn_blocking for CPU-intensive generation and tokio::spawn for IO-bound reads. (Even If not our ideal Rayon goal this should still be a fine sweetspot to pave the way to a proper Rayon later down the line)~~ I will think about it
Should be way better than what we have now at way less risk than jumpin straight into Rayon. I am of the strong opinion that should be a separate PR (Rayon I mean).
I will start working as soon as I have the time.

PR #1894, opened 2026-03-21
feat: Twisting Vines block

GreenedDev:
## Description
Adds Twisting vine block, it's placement rules and break behavior.
image


Snowiiii:

PR #1898, opened 2026-03-22
feat(wasm-api): port i18n module for plugin and host use

TsubakiDev:
## Description
This PR introduces a dedicated i18n module to the WASM plugin API, allowing plugins to leverage the server's internal translation engine and manage their own localization.

  Key changes:
   - WIT Definition: Added i18n.wit providing translate (retrieve translated strings) and load-translations (load custom plugin language files).
   - Host Implementation: Implemented the i18n host bridge in pumpkin which connects WASM calls to the existing pumpkin-util::translation engine.
   - Locale Mapping: Added robust mapping between WIT-generated Locale enums and the server's internal Locale type using string-based normalization.
   - Plugin Integration: Updated the v0.1.0 plugin world to import the new i18n interface.

  This enables a consistent internationalization experience across both the core server and external plugins.

## Testing
- Verified with `cargo check && cargo clippy --all-targets --all-features` to ensure all WIT bindings and host implementations are correctly typed.
- Manually reviewed the wit_to_util_locale conversion logic to ensure compatibility with all supported Minecraft locales.

### DIFF
--- a/pumpkin-plugin-wit/v0.1.0/i18n.wit
+++ b/pumpkin-plugin-wit/v0.1.0/i18n.wit
@@ -1,0 +1,21 @@
+interface i18n {
+    use common.{locale};
+
+    /// Retrieves a translated string from the host.
+    ///
+    /// # Arguments
+    /// * `key`: The namespaced translation key (e.g. `pumpkin:my_key`).
+    /// * `locale`: The target locale.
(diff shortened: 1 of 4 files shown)


Snowiiii:
LGTM. Thank you @TsubakiDev :D

PR #1900, opened 2026-03-24
feat(codec): add either codecs

SomeYellowGuy:
## Description
This PR adds *either* codecs and map codecs, which are codecs that can use one of two sub-codecs and encodes/decodes an `either::Either` object of both the sub-codecs' types.

For example, take an `EitherCodec<IntCodec, StringCodec>`. It can either encode/decode an `i32` or a `String`, and if both occur somehow, the `i32` is preferred (it being the first sub-codec's type).

## Testing
The tests in `codecs/either` pass as expected.


Snowiiii:

PR #1902, opened 2026-03-24
feat: implement /fillbiome command

WhiteProject1:
- Implements `/fillbiome` command for filling regions with specific biomes


Snowiiii:
Too much not implemented, use Arg consumer like Vanilla does

PR #1903, opened 2026-03-24
feat: implement /spreadplayers command

WhiteProject1:
- Implements `/spreadplayers` command for teleporting entities to random locations


Snowiiii:
Too much not implemented, use Arg consumer like Vanilla does, Logic is wrong

PR #1904, opened 2026-03-24
feat: implement /random command

WhiteProject1:
- Implements `/random` command with value and roll subcommands
- OP level 0 (all players can use)


Snowiiii:
I think vanilla uses an Argconsumer for the Range, so should we

PR #1906, opened 2026-03-24
feat: implement /team command

WhiteProject1:
- Implements `/team` command (add/remove/join/leave/modify/list)
- `Teams` system with `CUpdateTeams` packet for client synchronization
- Teams are sent to players on join


Snowiiii:
Too much not implemented, use Arg consumer like Vanilla does

PR #1909, opened 2026-03-24
feat: implement /attribute command

WhiteProject1:
- Implements `/attribute` command for querying, adding, removing, and setting entity attributes


Snowiiii:
Too much not implemented, use Arg consumer like Vanilla does, also a bunch of stuff is hardcoded here

PR #1911, opened 2026-03-24
feat: implement /item command

WhiteProject1:
- Implements `/item` command for manipulating items in entity and block inventories


Snowiiii:
I would prefer to have the Slot as seperate Argument consumer and also most of the command functionaly is not implemented

PR #1912, opened 2026-03-24
fix: use translation constants instead of string literals

WhiteProject1:
- Replace hardcoded translation key strings with `pumpkin_data::translation` constants
- Affected commands: fill, setworldspawn, spawnpoint, teleport


Snowiiii:
LGTM, Thank you @WhiteProject1 :D

PR #1913, opened 2026-03-24
feat(api): added missing wit defined events

illyrius666:
## Description

Added missing events into the api defined in wit. Without this the standard way of hooking into events via the example in pumpkin docs wont work.

## Testing

### DIFF
--- a/pumpkin-plugin-api/Cargo.toml
+++ b/pumpkin-plugin-api/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pumpkin-plugin-api"
-# This should be different from the pumpkin server verion
+# This should be different from the pumpkin server version.
 version = "0.1.0"
 edition.workspace = true
 


Snowiiii:
I think it would be better to split them into separate files, this becomes really a mess

illyrius666:
> I think it would be better to split them into separate files, this becomes really a mess

i seperated it, and used the same structure as the events folder in pumpkin/src/plugin/api/events

pumpkin-plugin-api/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pumpkin-plugin-api"
-# This should be different then the pumpkin server verion
+# This should be different from the pumpkin server verion
```
illyrius666:
```suggestion
# This should be different from the pumpkin server version.
```

Snowiiii:
LGTM. Thank you @illyrius666 :D

PR #1914, opened 2026-03-24
feat(codec): add dispatch codecs and refactored struct codecs

SomeYellowGuy:
## Description
Adds both the `KeyDispatchMapCodec` and `DispatchedMapCodec`, which each do the following:
- `KeyDispatchMapCodec`: Used to encode objects of a type by giving each of its variants a type ID, usually stored under a `"type"` key and using it to then decode the appropriate data. Many Minecraft objects rely on it, like *predicates*, *click events and hover events* in text components and *dialogs*. In other words, this type of codec helps with **enum-like** values.
- `DispatchedMapCodec`: Used to encode a map for which the value's type of an entry *depends on its key*. A good example is data components. For example, `minecraft:damage` only encodes an `i32`, while `minecraft:enchantments` encodes a map of enchantment-level value pairs.
This PR also somewhat refactors the struct codec system to work with map codecs instead of directly using normal codecs for the inner `struct_n` factory functions.

## Testing
Added tests for both codec types mentioned above, which all pass.


SomeYellowGuy:
Going to try to split this instead.

PR #1918, opened 2026-03-25
feat(command): reimplement `/stop`

SomeYellowGuy:
## Description
Reimplements the `/stop` command to work with the new command system.

*As a convention, we should register commands in `mod.rs` in alphabetical order, similar to what Minecraft does.*

## Testing

Tried using `/stop` as intended and it works. I also tried placing some extra parameters after the `/stop`, which correctly shows errors.


SomeYellowGuy:
Fixed the merge conflict.

Snowiiii:
LGTM, Thank you @SomeYellowGuy :D

PR #1919, opened 2026-03-25
feat(command): reimplement `/seed`

SomeYellowGuy:
## Description
- Reimplements the `/seed` command to work with the new command system.
- I also fixed the message send by `/seed` by printing *square brackets* to contain the seed and fixed its `i32` result by removing the clamp.

## Testing

Tried using `/seed` as intended and it works.


Snowiiii:

PR #1925, opened 2026-03-27
feat(command): implement entity selectors and entity argument type for new command system

SomeYellowGuy:
## Description
This PR implements the entity selector argument type for the new command system:
- Added an `EntitySelector` (a replacement of `TargetSelector`) to select entities when needed. It stores specific parameters to only target the required entities. There are methods for it that can be called in order to get entities from it. It has a few smaller data structures:
  - `PositionFunction` to get the taken origin of the selector when needed.
  - `Order` to get the sorting order of the selector.
  - `EntitySelectorPredicate` to only select the entities that meet the selector's criteria (mostly used by options).
- Added a parser (`EntitySelectorParser`) to parse a selector. It takes a `StringReader` in order to parse from it.
- Added `EntitySelectorOption`s to help with parsing options of an entity selector (like `distance` as in `@a[distance=..10]`). The following are implemented:
  - `distance`
  - `level` for experience level checks
  - `x`, `y` and `z`
  - `dx`, `dy` and `dz`
  - `x_rotation` and `y_rotation` for yaw and pitch respectively
  - `limit` for entity/player limit
  - `sort`
  - `gamemode` for game mode checks
For the unimplemented ones, a warning is printed for their unimplementation.

The above will also help for other argument types that use entity selectors (like *game profile* and *message* argument types), and as such, they are not specific to the entity selector argument type.
- This PR also adds the utility `Bounds` object in `pumpkin_util` (along with `IntBounds`, `DoubleBounds` and `FloatDegreeBounds`) to check whether a number is in some range. The three in brackets can be parsed from a `StringReader`. This can be followed up later with the `RangeArgumentType`s later.

Finally, the `EntityArgumentType` can be used for entities. It can be chosen to only include players or to only include a single entity. An `EntitySelector` can then be parsed from it, which can be used to fetch the appropriate entities.

Overall, this entity selector system is more accurate to what *Minecraft* does.

## Testing

Tested some selectors and options by creating a test command (locally), and they all seem to be working fine.

### DIFF
--- a/pumpkin/src/command/argument_types/entity_selector/option.rs
+++ b/pumpkin/src/command/argument_types/entity_selector/option.rs
@@ -25,11 +25,6 @@ pub const SORT_UNKNOWN_ERROR_TYPE: CommandErrorType<1> =
 pub const GAMEMODE_INVALID_ERROR_TYPE: CommandErrorType<1> =
     CommandErrorType::new(translation::ARGUMENT_ENTITY_OPTIONS_MODE_INVALID);
 
-/// Temporary command error for an unsupported option.
-/// TODO: Remove this when all entity selector options are implemented.
-pub const UNIMPLEMENTED_OPTION_ERROR_TYPE: CommandErrorType<1> =
-    CommandErrorType::new(translation::ARGUMENT_ENTITY_OPTIONS_UNKNOWN);
-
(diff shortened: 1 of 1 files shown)


Snowiiii:

PR #1926, opened 2026-03-28
fix(entity): use correct hurt sound for small slimes

### LINKED ISSUE #1847: Slime hit sound is not correct

### Operating System

Endevour OS

7e1e411

### What happened?

When i hit a naturally generated small slime (the smallest stage) it played the attack sound of a zombie

### To Reproduce

Join to the pumkin dev server and hit a small slime

### Expected behavior

The slime should sound slimy

CompileRider:
## Summary
- resolve slime hurt sounds in the living damage path
- use the small slime hurt sound when the saved slime size is 1
- keep the change scoped to slime hurt sound handling only

## Testing
- cargo fmt --check
- cargo test

Fixes #1847


CompileRider:
Closing this in favor of #1935. I consolidated slime plus the shared enderman/zombie-family/skeleton-family hurt sound parity work into one final PR so living.rs has one correct final integration point and slime remains preserved as the stateful special case.

PR #1928, opened 2026-03-28
fix(entity): You can now land while flying on elytra

Mrpupsik228:
# Not tested for Minecraft bedrock players 

## Description
fall_flying was set to true when elytra flight started but never reset to false when landing. Now fall_fling sets to false when landing.

## Testing
```sh
cargo fmt --check
cargo clippy --workspace --all-targets --all-features -- -D warnings
cargo test
```

### DIFF
--- a/pumpkin/src/net/java/play.rs
+++ b/pumpkin/src/net/java/play.rs
@@ -337,7 +337,11 @@ impl JavaClient {
                     player.jump().await;
                 }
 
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM. Thank you @Mrpupsik228 :D

Mrpupsik228:
@Snowiiii it looks like after recent updates my fix do not work anymore

Mrpupsik228:
Never mind

PR #1930, opened 2026-03-28
fix(entity): use correct hurt sound for endermen

CompileRider:
## Summary
- use the enderman hurt sound in the shared living damage path
- keep the fix scoped to hurt sound parity only
- add a narrow test for the enderman mapping and generic fallback

## Testing
- cargo fmt --check
- cargo test


CompileRider:
Closing this in favor of #1934. I consolidated enderman, zombie-family, and skeleton-family hurt sound parity into one shared-dispatch PR so living.rs has one helper, one call site, and one clean review path.

PR #1932, opened 2026-03-28
fix(entity): use correct hurt sounds for skeleton family

CompileRider:
## Summary
- use entity-specific hurt sounds for skeleton family variants in the shared living damage path
- keep the fix scoped to skeleton, bogged, parched, wither skeleton, and stray hurt sound parity only
- add a narrow mapping test and generic fallback coverage

## Testing
- cargo fmt --check
- cargo test

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -126,6 +126,21 @@ impl LivingEntity {
         &Block::SLIME_BLOCK,
     ];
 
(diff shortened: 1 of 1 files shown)


CompileRider:
Closing this in favor of #1933. I regrouped the zombie-family and skeleton-family hurt sound fixes into one PR to remove overlapping helper-introduction noise in pumpkin/src/entity/living.rs.

PR #1933, opened 2026-03-28
fix(entity): use correct hurt sounds for undead families

CompileRider:
## Summary
- use entity-specific hurt sounds for zombie-family and skeleton-family variants in the shared living damage path
- keep the scope limited to hurt sound parity plus mapping tests
- avoid repeated helper-introduction noise across separate family PRs

## Testing
- cargo fmt --check
- cargo test

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -126,6 +126,25 @@ impl LivingEntity {
         &Block::SLIME_BLOCK,
     ];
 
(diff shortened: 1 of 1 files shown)


CompileRider:
Closing this in favor of #1934. I consolidated enderman, zombie-family, and skeleton-family hurt sound parity into one shared-dispatch PR so living.rs has one helper, one call site, and one clean review path.

PR #1934, opened 2026-03-28
fix(entity): use correct hurt sounds for shared dispatch families

CompileRider:
## Summary
- use one shared hurt-sound selector for enderman, zombie-family, and skeleton-family entities
- keep slime separate because its hurt sound depends on slime size state
- keep the scope limited to hurt sound parity plus mapping tests in pumpkin/src/entity/living.rs

## Testing
- cargo fmt --check
- cargo test


CompileRider:
Closing this in favor of #1935. I consolidated slime plus the shared enderman/zombie-family/skeleton-family hurt sound parity work into one final PR so living.rs has one correct final integration point and slime remains preserved as the stateful special case.

PR #1935, opened 2026-03-28
fix(entity): consolidate hurt sound parity for slime and shared dispatch families

### LINKED ISSUE #1847: Slime hit sound is not correct

CompileRider:
## Problem

Pumpkin was missing correct hurt sound parity across several entity paths. Small slimes need a stateful size-based hurt sound path, while enderman, zombie-family, and skeleton-family entities should not fall back to the generic hurt sound in the shared living damage dispatch.

## Why this shape

Slime stays separate because its hurt sound depends on entity state and size. Enderman, zombie-family, and skeleton-family fit better in one shared selector inside `living.rs`, which keeps the final integration point singular and avoids overlapping helper/call-site duplication across multiple PRs.

## What changed

- `pumpkin/src/entity/living.rs` — one final shared hurt-sound selector for enderman, zombie-family, and skeleton-family, plus one final integrated `hurt_sound(&self)` path
- `pumpkin/src/entity/mob/slime.rs` — preserved slime-specific size-based hurt sound logic
- `pumpkin/src/entity/mob/mod.rs` — preserved the slime module integration needed by the final shape
- `pumpkin/src/entity/type.rs` — preserved the slime integration needed by the final shape

## Validation

I verified the final integrated branch locally, and the relevant hurt sounds were also checked in-game.

- [x] `cargo fmt --check`
- [x] `cargo test`
- [x] Manual in-game verification for slime, enderman, zombie-family, and skeleton-family hurt sounds

Fixes #1847


Snowiiii:
@CompileRider this looks quiet hard coded which is an anti pattern, not sure how easy or hard it would be to extract the sound from the Extractor?

CompileRider:
Thanks, that makes sense.

I agree the current selector in `living.rs` is drifting toward a hardcoded vanilla mapping, which is not the best long-term shape if this data can be sourced from the extractor/codegen side instead.

I kept this first pass small to confirm the parity gaps and behavior in-game, but I think your point is right for the final direction. I’m now checking whether the hurt sound mapping can be derived from the extractor/generated data path instead of continuing to grow the runtime selector by hand.

If there’s a specific extractor output you had in mind for entity sound mappings, feel free to point me at it and I’ll align the follow-up with that.

Snowiiii:
Looking at the solution now, this looks much better, maybe it is also worth adding getDeathSound at least for the future, also don't forget to open a PR in the Extractor repo :D

CompileRider:
Opened the Extractor-side companion PR for this generated hurt-sound mapping work:

Pumpkin-MC/Extractor#43

That PR adds the narrow `entity_hurt_sound` extraction source used by this Pumpkin change. I kept `slime` out of the extracted mapping because its hurt sound is still size-dependent at runtime, and I left `getDeathSound` as follow-up work.

Snowiiii:
using a separate Extractor for this is unnecessary, just add a field into the existing Entity extractor

CompileRider:
Updated this branch to match the Extractor ownership change.

The companion Extractor PR now emits `hurt_sound` through `entities.json` from the existing entity extractor instead of a separate extractor file, and this Pumpkin branch now derives the generated `hurt_sound_for_entity_type(...)` helper from that owner-aligned source.

`slime` remains runtime-only because its hurt sound is still size-dependent, and `getDeathSound` is still left as follow-up work.

pumpkin-codegen/src/entity_hurt_sound.rs
```diff
@@ -0,0 +1,39 @@
+use std::{collections::BTreeMap, fs};
```
Snowiiii: Well i guess this file can be deleted now?

Snowiiii: and the generated as well

CompileRider: Yes, I agree.

Now that `hurt_sound` is sourced from `entities.json`, the standalone codegen file and generated file look unnecessary. I’ll fold that helper generation into the existing entity codegen path and keep the runtime shape unchanged.

`slime` will stay runtime/stateful, and `death_sound` will stay out of scope for this PR.

CompileRider: Folded this into the existing entity codegen path now. The standalone `pumpkin-codegen/src/entity_hurt_sound.rs` file and the separate generated file were removed, while `living.rs` stays dispatch-only and slime remains the runtime size-based exception.

pumpkin-codegen/src/entity_type.rs
```diff
@@ -220,6 +231,8 @@ pub fn build() -> TokenStream {
         use crate::tag::Taggable;
         use crate::tag::RegistryKey;
         use crate::attributes::Attributes;
+        #[cfg(feature = "entity_hurt_sound")]
```
Snowiiii: please remove the feature entirly

pumpkin-data/src/generated/entity_type.rs
```diff
@@ -9163,6 +9165,22 @@ impl EntityType {
...
+        id if id == EntityType::HUSK.id => Some(Sound::EntityHuskHurt),
+        id if id == EntityType::PARCHED.id => Some(Sound::EntityParchedHurt),
+        id if id == EntityType::SKELETON.id => Some(Sound::EntitySkeletonHurt),
+        id if id == EntityType::STRAY.id => Some(Sound::EntityStrayHurt),
+        id if id == EntityType::WITHER_SKELETON.id => Some(Sound::EntityWitherSkeletonHurt),
+        id if id == EntityType::ZOMBIE.id => Some(Sound::EntityZombieHurt),
+        id if id == EntityType::ZOMBIE_VILLAGER.id => Some(Sound::EntityZombieVillagerHurt),
+        _ => None,
+    }
+}
```
Snowiiii: just put it into the EntityType field, no need for an ugly extra function

pumpkin-codegen/src/entity_type.rs
CompileRider: Done. I removed the `entity_hurt_sound` feature entirely so this no longer needs a separate feature gate.

pumpkin-data/src/generated/entity_type.rs
CompileRider: Done. `hurt_sound` now lives directly on generated `EntityType`, and the extra helper function is gone. `living.rs` still stays dispatch-only and `slime` remains the runtime size-based exception.

Snowiiii:
Looks good now, Thank you @CompileRider :D

Issue #1853, opened 2026-03-11
Adding more candle to the burning candle caused it to go out.


RegulusSource:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

windows11

### Server Software Version/Commit

ee3aa76c2b1a526a22eb39631473f6859ae52ce3

### What happened?

Adding more candle to the burning candle caused it to go out.
The number of candles did not increase.

### To Reproduce

Use a candle and click on the burning candle.

### Expected behavior

The candles increase in number but do not go out.

PR #1936, opened 2026-03-28
Fix: Preserve lit state when adding candles to an already lit candle (#1853)

TheRealMusawer:
Fix candle blocks losing their lit state when adding another candle.

## Description
When adding a candle to an already lit candle, the block was recreated without keeping the lit value, which caused the candle to go out. The fix simply stores the previous properties.lit value and restores it after increasing the candle count. This matches vanilla behavior — adding a candle no longer extinguishes it.


TheRealMusawer:
fixes #1853

Snowiiii:
LGTM. Thank you @TheRealMusawer :D

TheRealMusawer:
Yw

PR #1937, opened 2026-03-29
fix: Skip deactivated plugins during server startup

### LINKED ISSUE #1929: Server doesn't skip plugin when disabled

### Operating System

Fedora 43

Pumpkin 0.1.0-dev+1.21.11 (Commit: 49b3d25/release)

### What happened?

Image
When you disable a plugin in smashed pumpkin, it gets the suffix .deactivated.
If you start / restart the server after disabling the plugin, you get two plugin not found errors.

### To Reproduce

1. Start pumpkin (via smashed pumpkin) with at least one plugin.
2.1. If you use smashed pumpkin: deactivate the plugin
2.2. If you don't: Re- name the plugin and add .deactivated.
3. Restart pumpkin. 

### Expected behavior

Pumpkin shouldn't miss the plugin and shouldn't throw any [ERR].

Exteron01:
## Summary

  - Skip files with `.deactivated` extension in `PluginManager::load_plugins()` before attempting loader matching
  - Prevents `PluginNotFound` errors being logged for intentionally disabled plugins
  - Deactivated files are never added to the `unloaded_files` retry queue

  Fixes #1929. When plugins are disabled, the server logged errors on every restart instead of silently skipping them.

  ## Test plan

  - `cargo clippy --all-targets` passes
  - Place a `.deactivated` file in `./plugins/`, start the server, no error logged
  - Active plugins still load normally

### DIFF
--- a/pumpkin/src/plugin/mod.rs
+++ b/pumpkin/src/plugin/mod.rs
@@ -311,6 +311,14 @@ impl PluginManager {
                 continue;
             }
 
+            if path
+                .extension()
+                .and_then(|ext| ext.to_str())
+                .is_some_and(|ext| ext.eq_ignore_ascii_case("deactivated"))
+            {
+                continue;
+            }
+
             // Start loading plugin concurrently
             match self.start_loading_plugin(&path).await {
                 Ok(task) => load_tasks.push(task),


Snowiiii:
LGTM. Thank you @Exteron01 :D

PR #1941, opened 2026-03-29
feat(command): add time argument

SomeYellowGuy:
## Description
Adds the time argument, which specifies some time duration in ticks. It can either be in ticks, seconds or Minecraft days.

## Testing
Added tests for the argument type, which all pass.


Snowiiii:

PR #1943, opened 2026-03-30
feat(command): implement `Vec3` argument type

SomeYellowGuy:
## Description

Adds the `Vec3` argument type, which specifies a 3-dimensional vector. It also adds `Coordinates`, both *world* and *local* ones. 

- *world*: Accepts fixed (no prefix) and relative coordinates (starting with `~`)
- *local*: Accepts local coordinates (starting with `^`)

The implemented methods will be useful for other coordinate-like argument types.

## Testing

Added tests for the `Vec3` argument type, which all pass.


Snowiiii:

PR #1945, opened 2026-03-31
fix: end portal crash and island setup

RoosterBooster007:
## Description
- Fixed the portal crash (using `noise_sampler`) and another chunk gen crash
- You now always teleport to the end platform when entering the end via the portal
- Fixed the pillars so they aren't chopped anymore (literally)
- Started the dragon fight logic with boss bar and end crystal spawns

## Testing

Note: there's some console logging going on, which should be fine since the end isn't finished yet, and it should help with implementing the dragon AI anyway.


Snowiiii:
LGTM. Awesome work, Thank you @RoosterBooster007 :D

PR #1946, opened 2026-03-31
refactor: add tracked absorption id

RoosterBooster007:
## Description
- This fixes the hardcoding (TODO) for the absorption id in `living.rs` by moving it into the tracked JSON files instead.

## Testing
- I didn't use the Extractor, so idk the correct value for `v26_1` for absorption... so rn it doesn't show the yellow hearts right on the client.


Snowiiii:
We don't modify this files by hand, this will break on the next update and is also not really the intent way, Not sure where the absorption id is in vanilla but these variables often have different names

PR #1951, opened 2026-04-01
feat: implement /random command

Pa-dej:
## Description
Implements the vanilla `/random` command.

**Supported syntax:**
- `/random value <range>` - result visible only to sender
- `/random roll <range>` - result broadcast to all players
- `/random value <range> <sequence>` - sample from named sequence
- `/random roll <range> <sequence>` - same with broadcast
- `/random reset <sequence> [seed] [includeWorldSeed] [includeSequenceId]`
- `/random reset *` - reset all sequences

**Implementation notes:**
- Sequence names validated as namespaced IDs (lowercase only), 
  error position matches vanilla behavior.
- Tab completion suggests existing sequence names
- Sequences persisted in level.dat via RandomSequence in LevelData
  (same pattern as GameRules - vanilla uses separate random_sequences.dat,
  can be migrated in follow-up PR once world/data/ infrastructure exists)
- xoroshiro128 RNG for deterministic sequence generation.

## Testing
- Valid/invalid range inputs tested manually in-game
- Sequence reset + deterministic replay verified.
- `cargo fmt`, `cargo clippy`, `cargo test` - all pass.

### DIFF
--- a/pumpkin/src/command/commands/random.rs
+++ b/pumpkin/src/command/commands/random.rs
@@ -6,6 +6,7 @@ use std::{
 use pumpkin_data::{translation, world::SAY_COMMAND};
 use pumpkin_util::{
     PermissionLvl,
+    math::bounds::IntBounds,
     permission::{Permission, PermissionDefault, PermissionRegistry},
     random::{RandomImpl, xoroshiro128::Xoroshiro},
     text::TextComponent,
@@ -14,27 +15,28 @@ use pumpkin_world::world_info::RandomSequence;
 use rand::RngExt;
 use tokio::sync::Mutex;
 
-#[cfg(test)]
-use crate::command::dispatcher::CommandError;
 use crate::command::{
     argument_builder::{ArgumentBuilder, argument, command, literal},
     argument_types::{
-        argument_type::{ArgumentType, JavaClientArgumentType},
-        core::{bool::BoolArgumentType, long::LongArgumentType, string::StringArgumentType},
+        core::{
+            bool::BoolArgumentType, long::LongArgumentType,
+            resource_location::ResourceLocationArgumentType,
+        },
+        range::IntRangeArgumentType,
     },
-    context::{command_context::CommandContext, command_source::CommandSource},
+    context::command_context::CommandContext,
     errors::{
         command_syntax_error::{CommandSyntaxError, CommandSyntaxErrorContext},
         error_types,
     },
     node::{CommandExecutor, CommandExecutorResult, dispatcher::CommandDispatcher},
-    string_reader::StringReader,
     tree::RawArg,
 };
-use pumpkin_protocol::java::client::play::StringProtoArgBehavior;
 
 const DESCRIPTION: &str = "Generates a random integer, or controls random sequences.";
 const PERMISSION: &str = "minecraft:command.random";
+const PERMISSION_SEQUENCE: &str = "minecraft:command.random.sequence";
+const PERMISSION_RESET: &str = "minecraft:command.random.reset";
 
 const ARG_RANGE: &str = "range";
 const ARG_SEQUENCE: &str = "sequence";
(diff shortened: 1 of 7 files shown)


Q2297045667:
We now have a new command system.
Please refer to this PR to reimplement the command.
https://github.com/Pumpkin-MC/Pumpkin/pull/1918/

Pa-dej:
Ported to the new command system as requested.

Snowiiii:
Hey, Thank you for your PR
1. Please use Arg consumers which are already implemented, you should not implement them in the Command File
2. Don't use Literals use translations

I think all Arg consumers you need are already implemented

Pa-dej:
Fixed both review points:
- Replaced custom arg consumers with existing ones 
  (ResourceLocationArgumentConsumer, BoundedNumArgumentConsumer, BoolArgConsumer)
- All messages now use vanilla translation keys from en_us.json

SomeYellowGuy:
Doesn't this go back to the old command dispatcher?

pumpkin/src/command/commands/random.rs
```diff
@@ -126,7 +120,74 @@ struct InclusiveRange {
...
+            );
+        }
+
+        Ok((min, max))
+    }
+
+    fn client_side_parser(&'_ self) -> JavaClientArgumentType<'_> {
+        JavaClientArgumentType::String(StringProtoArgBehavior::SingleWord)
+    }
+}
```
Laptop59: You should remove this part of the code as the `RangeArgumentType` has been implemented in the `argument_types` module, which can also be used by other commands.

pumpkin/src/command/commands/random.rs
```diff
@@ -164,9 +225,10 @@ impl InclusiveRange {
     }
 }
```
Laptop59: This should be removed as it's an artifact of the previous commits in this PR, and that we already have the argument type required.

pumpkin/src/command/commands/random.rs
```diff
@@ -0,0 +1,822 @@
...
+
+#[cfg(test)]
+fn parse_i32_range_bound(raw: &str) -> Result<i32, CommandError> {
+    raw.parse::<i32>().map_err(|_| {
+        CommandError::CommandFailed(TextComponent::translate(
+            translation::PARSING_INT_INVALID,
+            [TextComponent::text(raw.to_string())],
+        ))
+    })
+}
```
Laptop59: Uses the older `CommandError`, was this supposed to be removed?

pumpkin/src/command/commands/random.rs
```diff
@@ -0,0 +1,822 @@
...
+fn get_optional_argument<'a, T: 'static>(
+    context: &'a CommandContext,
+    name: &str,
+) -> Result<Option<&'a T>, CommandSyntaxError> {
+    if context.arguments.contains_key(name) {
+        context.get_argument(name).map(Some)
+    } else {
+        Ok(None)
+    }
+}
```
Laptop59: A utility method for optionally parsing an argument is 'very hacky'. The executor should know what arguments to parse and what arguments not to parse after command parsing from its stored values

pumpkin/src/command/commands/random.rs
```diff
@@ -0,0 +1,822 @@
...
+    let sequence = context.get_argument::<String>(ARG_SEQUENCE)?;
+    let raw_arg = RawArg {
+        value: sequence,
+        start: 0,
+        end: sequence.len(),
+        input: sequence,
+    };
+
+    validate_sequence_name(raw_arg)
+}
```
Laptop59, line 394: The `sequence` argument is an identifier (formerly resource location) containing a namespace and path, not a string. We don't have parsing for identifiers yet. A way to parse them will be required before this argument can be implemented properly.

pumpkin/src/command/commands/random.rs
```diff
@@ -0,0 +1,822 @@
...
+
+const fn seed_argument_type() -> LongArgumentType {
+    LongArgumentType::any()
+}
+
+fn level_two_requirement(
+    source: &CommandSource,
+) -> std::pin::Pin<Box<dyn std::future::Future<Output = bool> + Send + '_>> {
+    Box::pin(async move { source.output.has_permission_lvl(PermissionLvl::Two) })
+}
```
Laptop59: Instead of checking if the sender has permission level 2, we should create a permission and check if it has *that*, so that the permission level of it can be changed as desired (by server owners, for example) without conflicts.

pumpkin/src/command/commands/random.rs
```diff
@@ -0,0 +1,822 @@
...
+        ));
+    }
+
+    #[test]
+    fn reject_too_large_range_size() {
+        assert!(matches!(
+            InclusiveRange::parse("-2147483648..2147483647"),
+            Err(CommandError::CommandFailed(_))
+        ));
+    }
```
Laptop59: The `range` module in `argument_types` has tests for parsing different types of ranges, so we don't need this part of the code.

pumpkin/src/command/dispatcher.rs
```diff
@@ -288,8 +288,16 @@ impl CommandDispatcher {
...
+            let mut path_raw_args = raw_args.clone();
+            match Self::try_find_suggestions_on_path(
+                src,
+                server,
+                &path,
+                tree,
+                &mut path_raw_args,
+                cmd,
+            )
+            .await
```
Laptop59: These changes look like an artifact of previous commits of this PR with the older dispatcher. `raw_args` was made `mut`able during those changes and probably should be reverted to be immutable again.

Laptop59 (commented):
Overall, this PR still needs some work to be done before it has the potential to be merged.

pumpkin/src/command/commands/random.rs
Pa-dej: Removed, switched to shared IntRangeArgumentType. Kept the /random-specific size checks (>= 2, <= 2_147_483_646) in the executor since those are command constraints, not general range constraints.

pumpkin/src/command/commands/random.rs
Pa-dej: Removed.

pumpkin/src/command/commands/random.rs
Pa-dej: Removed, was leftover code.

pumpkin/src/command/commands/random.rs
Pa-dej: Removed. ResetExecutor now uses explicit ResetMode variants wired from the command tree, so each path knows exactly which arguments to parse.

pumpkin/src/command/commands/random.rs
Pa-dej: Added ResourceLocationArgumentType to argument_types/core and used it for ARG_SEQUENCE. Client-side parser is now ResourceLocation. validate_sequence_name() stays in the executor for server-side validation.

pumpkin/src/command/commands/random.rs
Pa-dej: Replaced with explicit permissions minecraft:command.random.sequence and minecraft:command.random.reset, both registered with Op(Two) and checked via .requires(...)

pumpkin/src/command/commands/random.rs
Pa-dej: Removed, range tests are covered in argument_types/range.

pumpkin/src/command/dispatcher.rs
Pa-dej: Reverted dispatcher.rs to master baseline.

Snowiiii:
@Pa-dej There should be now an Identifer in master already, also please fix the conflict

Snowiiii:
@Pa-dej are you planning on finishing this?

Pa-dej:
Yes, sorry for the wait, my exams are starting.

Q2297045667:
> Yes, sorry for the wait, my exams are starting.

No worries about the PR. Take it easy and good luck with your results!

Snowiiii:
Feel free to reopen

PR #1953, opened 2026-04-01
fix chunk issues

IGN-Styly:
## Description
Fixes an issue where chunk generation only did surface generation in 1 chunk and applied features after, causing exposed ores in the surface. This patch adds a test to make sure this isn't regressed in later patches, and rewrites pars of chunk_state.rs and generation.rs. So that it also generates surface of chunks with a radius of 1 thus being a closer match to minecraft.

DISCLAIMER, I USED OPENAI CODEX TO GENERATE THE TEST.
DISCLAIMER, I HAVE NO IDEA WHAT I'M DOING PLEASE DOUBLE CHECK.

## Testing
Comparing seed `-7057863872970073962` at cord `-1000 100 450` and looking for surface exposed ore features. Use `F3+G` to see it better.

### DIFF
--- a/pumpkin-world/src/chunk_system/chunk_state.rs
+++ b/pumpkin-world/src/chunk_system/chunk_state.rs
@@ -106,9 +106,14 @@ impl StagedChunkEnum {
 
     /// Total number of state values (0 = None … 9 = Full).
     pub const COUNT: usize = Self::Full as usize + 1;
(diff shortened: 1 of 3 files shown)


Snowiiii:
> DISCLAIMER, I HAVE NO IDEA WHAT I'M DOING PLEASE DOUBLE CHECK.

I see that, please don't submit Pull Requests then, the soultion is really not good

PR #1969, opened 2026-04-04
feat(command): implement range argument types

Laptop59:
## Description

Adds two range argument types:
- `IntRangeArgumentType` to parse integers (`i32`), and
- `FloatRangeArgumentType` to parse floating point values (`f64`).

They will be required for the `/execute` and `/random` commands.

## Testing

Added some tests for each argument type, and they pass, as expected.


Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #1972, opened 2026-04-05
chore(command): add argument type helpers

SomeYellowGuy:
## Description
In Minecraft, there can be many ways to parse a single `ArgumentType`. Minecraft provides certain "helper" functions to abstract the argument parsing methods away from command implementations.

This PR adds helper functions for all currently existing argument types and can be more convenient. For example, for the `EntityArgumentType`, there are direct methods to get entities/players from an argument instead of working with an `EntitySelector`.

In addition: 
- The `/setidletimeout` command has been changed to use the `IntegerArgumentType` helper.
- The `/kill` command has been changed to use the `EntityArgumentType` helper.

## Testing
The `/setidletimeout` and `/kill` commands still work as expected.


Snowiiii:

PR #1987, opened 2026-04-08
fix(entity): persist dropped items across reconnect

### LINKED ISSUE #1698: Dropped items are not rendered correctly after re-entering the server

CompileRider:
## Problem
Dropped item entities become invisible to a reconnecting player even though they still exist and can be picked up. The reconnect/load path was not preserving the item-specific entity payload needed to respawn dropped items correctly from chunk entity data.

## What changed
- added real `ItemEntity` NBT round-trip persistence for the dropped item payload and related state
- rehydrate `EntityType::ITEM` through `ItemEntity` in the entity type restore path
- added a focused regression test covering UUID and item payload round-trip for dropped items

## Testing
- [x] `cargo fmt --check`
- [x] `cargo test`
- [x] `cargo audit check`

Fixes #1698


pumpkin/src/entity/type.rs
```diff
@@ -76,6 +77,15 @@ pub async fn from_type(
...
         }
+        id if id == EntityType::ITEM.id => Arc::new(
+            ItemEntity::new_with_velocity(
+                entity,
+                pumpkin_data::item_stack::ItemStack::EMPTY.clone(),
+                Vector3::new(0.0, 0.0, 0.0),
+                0,
+            )
+            .await,
+        ),
```
Snowiiii: kinda ugly to have this here

Snowiiii:
@CompileRider the nbt system changed quiet a bit, you would need to adjust

Q2297045667:
Resolved conflicts and switched to our PBNT.

Q2297045667:
https://youtu.be/VdAgWdlirMY
This PR hasn't covered everything yet.
Joining the server replays the drop animation and creates a dupe glitch.


image
Uncollectable items spawn during multiplayer sessions.

Q2297045667:
I'm not entirely sure if these match vanilla behavior. I'll test it again when I get a chance. By the way, we rolled back the PNBT changes, so you may need to adapt to that.

Q2297045667:
image


After reconnecting to the server
image
There is still an issue.

CompileRider:
That’s odd. I tested it locally and it seems to work, but I’ll investigate further to understand why it’s not working on your end.

Q2297045667:
> That’s odd. I tested it locally and it seems to work, but I’ll investigate further to understand why it’s not working on your end.

Maybe it's due to some network packet changes made to the master branch?

Q2297045667:
When is completion expected?

CompileRider:
> When is completion expected?

Hi, sorry if I'm making you wait, I'm checking everything carefully, as much as possible during my schedule

CompileRider:
Okay so I went through the commit history to figure out why these bugs are still happening. Here's the deal.
The original fix was split in three:

1. NBTStorage for ItemEntity (write_nbt/read_nbt_non_mut) — item.rs
2. EntityType::ITEM dispatch in from_type() — type.rs  
3. UUID dedup + velocity clear on reconnect — world/mod.rs

Parts 1 and 2 landed in e25fdf34 and cbe0fefc. Then the branch got merged with master like 7 times and somewhere along the way item.rs got overwritten with master's version. Master just has an empty impl NBTStorage for ItemEntity {}, so all the NBT persistence code just vanished. Tests too.

Here's what's actually in the branch right now:

- world/mod.rs lines 2730-2768 and 3201-3211 — the UUID dedup and velocity clear. Those survived because they were added after all the merges.
- type.rs line 240-242 — the ITEM dispatch is there but it came from master, so it uses ItemEntity::new() with random velocity instead of a proper restore constructor.
- item.rs line 364 — impl NBTStorage for ItemEntity {} is empty. No write, no read. Items don't save anything to chunk NBT.

Which explains all three bugs:
- Drop animation replay — no NBT restore means the stale velocity from the original drop gets sent to the client again
- Ghost items — from_type() creates an ItemEntity with AIR instead of whatever was actually dropped
- Uncollectable items — the UUID dedup can't match entities that were never created properly in the first place

I'm putting the NBTStorage implementation back and will post a build to test.

Snowiiii:
LGTM. Thank you @CompileRider :D

PR #1992, opened 2026-04-09
feat(api): Advanced Packet System

TsubakiDev:
## Description

  Implemented an end-to-end raw packet pipeline across:

  `WIT -> Host -> Plugin API`

  Plugins can intercept packets before protocol decoding with access to:
  - direction
  - connection state
  - raw packet ID and payload
  - cancellable / mutable event data

  Supported behaviors:
  - pass through unchanged
  - cancel propagation
  - replace payloads deterministically at the event boundary

  ### Packet API

  Added `PacketWrapper` over `RawPacketEventData` with:
  - `id()`, `direction()`, `state()`
  - `reader()`
  - `replace_payload(...)`
  - `cancel()`
  - `protocol_version()`
  - `minecraft_version()` for Java
  - `resolve_packet_id(...)`
  - `java_packet_id(phase, name)`
  - `java_packet_name()`
  - `java_packet_key()`
  - `java_packet_id_for_key(...)`
  - `decode_java_registered_schema()`

  Also added safe reply helpers for packet hooks:
  - `reply_system_message(...)`
  - `reply_raw_packet(...)`
  - `reply_packet_on_channel(...)`

  ### Codec / Typed Helpers

  Added symmetric packet utilities:
  - `PacketReader`
  - `PacketWriter`

  Supported primitives include:
  - VarInt / VarLong
  - bounded strings
  - fixed-width integers
  - UUID bytes
  - raw byte slices

  Added generated typed Java packet keys for all supported packets:
  - `packet::java::typed::JavaPacketKey`

  Added schema support for field-level parsing / encoding:
  - `packet::java::schema::PacketSchema`
  - `FieldType`
  - `FieldValue`
  - global schema registry keyed by `JavaPacketKey`

  ### Versioned Packet Metadata

  Added a version-aware Java packet catalog backed by embedded packet assets:
  - all supported Java versions
  - clientbound and serverbound
  - all protocol phases

  Exposed APIs:
  - `get_id(version, direction, phase, name)`
  - `get_name(version, direction, id)`
  - `list(version, direction)`

  Exported the full generated packet ID table to WASM via:
  - `packet_ids_full`

  Added:
  - `player.get_protocol_version()`
    - Java: negotiated protocol version
    - Bedrock: `CURRENT_BEDROCK_MC_PROTOCOL`

  ### Deferred Effects

  Reworked WASM packet-event side effects to use a deferred effect queue instead of immediate execution inside packet hooks.

  This avoids recursive raw-packet reentry when plugins send packets from within packet handlers, and fixes hangs caused by re-entering the same WASM store during
  packet processing.

  Deferred operations now include:
  - player system messages
  - player custom payload sends
  - player raw packet sends
  - server broadcast payload sends
  - server broadcast raw packet sends

  Effects are queued during event execution and flushed after the current event returns.

### Testing

```bash
cargo check
cargo fmt
cargo clippy --all-targets --all-features


Snowiiii:
In best case we would also add something like Packet Wrapper, basically our current Packet Read and Write implementation similar to packet events. So plugin devs can easily get packet information and even with multi version support

TsubakiDev:
> In best case we would also add something like Packet Wrapper, basically our current Packet Read and Write implementation similar to packet events. So plugin devs can easily get packet information and even with multi version support

Yes, perhaps I should take further steps here next

Snowiiii:
@TsubakiDev when do you plan to finish this ?

TsubakiDev:
> @TsubakiDev when do you plan to finish this ?

I will complete this pr within this month

Snowiiii:
CI @TsubakiDev

QuantumSegfault (commented):
Looks pretty good overall. 

One relatively minor short coming is that you expose `packet_ids_full` only to Rust, by copying out of `pumpkin-data`. This means e.g. Python won't have access to them. There might also be a couple more helpers that could be better suited to language-agnostic WIT. I don't think it's worth holding up this PR though. This situation can be improved later.

Besides that, Bedrock doesn't have a nice ID mapping exposed to plugins like there is for Java. But that's another thing where other work needs to be done before this is convenient to include.

Other than that, I think this is a great step forward

Snowiiii:
@TsubakiDev I'm sorry but i think the recent permission addtions caused merge conflicts

TsubakiDev:
> @TsubakiDev I'm sorry but i think the recent permission addtions caused merge conflicts

@Snowiiii I've fixed it

Snowiiii:
@TsubakiDev thanks for your work, but i decided to implement this myself

PR #2000, opened 2026-04-10
feat(command): identifier parsing & argument type

Laptop59:
## Description

This PR aims to create an argument type `IdentifierArgumentType` that can parse identifiers.
However, since we don't actually have a structure with namespace and path separated, this PR also adds an new struct `Identifier` in `pumpkin-util`.

`Identifier` consists of both a namespace and a path, each stored as a `Cow<'static, str>`. Not for registries, but for identifying user-generated content (like bossbars) which cannot be fully represented by `&'static str`s.

We need an argument type for this as commands like `/bossbar`, `/random`, and `/waypoint` take an identifier which is user-generated and not associated with any registry.

Maybe if we want to parse item IDs later for example in a new argument type, we can first parse an `Identifier` using `from_reader` and then check if an item under that identifier does exist.

## Testing

Added new tests for creating and parsing `Identifier`s, and they all pass.


Snowiiii:
@Laptop59 fix conflicts plis

Snowiiii:
LGTM, Thank you @Laptop59 :D

PR #2003, opened 2026-04-10
fix: make 256-character limit apply on UTF-16 chars instead of bytes

### LINKED ISSUE #2002: Kicked for `too large: string` when chatting more than 256 bytes of `char`s

### Operating System

Arch Linux

0.1.0-dev+26.1 (Commit: b8c7067/debug)

### What happened?

When a player types more than **256 bytes'** worth of `char`s into the chat *without exceeding* the 256-character limit, the server kicks them out for `Error while handling incoming packet too large: string`.

### To Reproduce

1. Join the server
2. Type the following 65 pumpkin symbols into the chat: 🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃🎃

### Expected behavior

It should show that I chatted the 65 pumpkin symbols to everyone without kicking me out of the server.

Laptop59:
## Description

Fixes #2002 

Vanilla checks if the string exceeds a length of 256 UTF-16 characters instead of bytes, so we should replicate that behavior.

## Testing

It now shows that I chatted the 65 pumpkin symbols to everyone without kicking me out of the server.

### DIFF
--- a/pumpkin-protocol/src/ser/mod.rs
+++ b/pumpkin-protocol/src/ser/mod.rs
@@ -162,9 +162,9 @@ impl<R: Read> NetworkReadExt for R {
         // 1 Java `char` takes a maximum of 3 bytes in UTF-8:
         let maximum_utf8_bytes = bound.saturating_mul(3);
         if bytes_len > maximum_utf8_bytes {
-            return Err(ReadingError::TooLarge(
-                format!("string has too many bytes ({bytes_len} > {maximum_utf8_bytes})"),
-            ));
+            return Err(ReadingError::TooLarge(format!(
+                "string has too many bytes ({bytes_len} > {maximum_utf8_bytes})"
+            )));
         }
 
         let data = self.read_boxed_slice(bytes_len)?;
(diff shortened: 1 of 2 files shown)


pumpkin-protocol/src/ser/mod.rs
```diff
@@ -154,13 +154,20 @@ impl<R: Read> NetworkReadExt for R {
     }
 
     fn get_string_bounded(&mut self, bound: usize) -> Result<String, ReadingError> {
-        let size = self.get_var_uint()?.0 as usize;
-        if size > bound {
-            return Err(ReadingError::TooLarge("string".to_string()));
+        let bytes_len = self.get_var_uint()?.0 as usize;
```
chocodev11, line 157: in vanilla, it also check if bytes_len is greater than max byte the message can contain before read it, which is `bound.saturating_mul(3) `

Laptop59: Fixed it according to the `read` method (i looked at the `write` one by mistake)

Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #2004, opened 2026-04-10
feat(command): add suggestion providers and their support in argument nodes and fix `list_suggestions` method

SomeYellowGuy:
## Description
This PR adds `SuggestionProvider`s for specific arguments that may require them.
- The `suggests` and `suggests_arc` methods can be used on an argument node to add a custom `SuggestionProvider` for it. This is used quite a lot: some commands that use this include `/op`, `/execute`, `/bossbar`, `/scoreboard` and `/tick`.
- Added support for it in the command dispatcher.
- Fixed the `list_suggestions` method to take a `SuggestionBuilder` rather than a *mutable reference* to one

## Testing

- Tested (locally) adding custom suggestions `1` to `5` for the integer argument of `/setidletimeout`, and it works for the server and the client!

### DIFF
--- a/pumpkin/src/command/argument_builder.rs
+++ b/pumpkin/src/command/argument_builder.rs
@@ -6,6 +6,7 @@ use crate::command::node::detached::{
 use crate::command::node::{
     Command, CommandExecutor, RedirectModifier, Redirection, Requirement, Requirements,
 };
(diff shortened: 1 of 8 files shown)


Snowiiii:
LGTM, Thank you @SomeYellowGuy :D

PR #2005, opened 2026-04-10
feat(command): implement game profile argument type

SomeYellowGuy:
## Description
Implements the *game profile* argument type with the previously made `EntitySelector`s.

Note that no `GameProfileSuggestionMode` object was added for this argument (as was in the `args`' `GameProfileArgumentConsumer`) because custom suggestions can be controlled by *custom suggestion providers* (#2004).

## Testing

Created the test command locally, which printed the resulting `GameProfile`s to the command source, and it works!


Snowiiii:

PR #2012, opened 2026-04-11
fix: (player) potion effect crash

RoosterBooster007:
## Description
- Falls back to the `attribute.default_value` if no attribute base is found for a `LivingEntity`, preventing a panic.

## Testing
- Doesn't crash, but the default speed attribute (0.7) is used for players rn, causing extreme speed effects (instead of the correct 0.1). The right attribute defaults aren't present for players (in `entity_type.rs`) atm.


Snowiiii:
is this the same as https://github.com/Pumpkin-MC/Pumpkin/pull/1990 ?, does it fix the issue?

PR #2024, opened 2026-04-14
fix(worldgen): prune inactive chunk holders

### LINKED ISSUE #1780: Memory leak triggered after player disconnect

### Operating System

Fedora 43

17118be

### What happened?

Memory usage starts to grow indefinitely after a player leaves the server. Based on profiling, the issue appears to be located in `pumpkin_world::chunk_system::schedule::GenerationSchedule::ensure_dependency_chain`.

Image

### To Reproduce

1. Start the server and join.
2. Move around.
3. Disconnect from the server.
4. Observe memory usage.

### Expected behavior

Memory usage should stabilize or decrease when a player leaves the server.

mw2000:
- prune dependency-only chunk holders once they become fully inactive instead of keeping them in the scheduler forever
- clear any leftover occupied-by edge chains when dropping those holders so graph bookkeeping is released too
- reduce chunk-system memory growth after player disconnects when chunk demand falls back to zero

Fixes #1780


Snowiiii:
@mw2000 chunk system changed a bit, is this still relevant ?

Q2297045667:
Could you please pull the latest code and retest to see if it works?

Q2297045667:
fix ci

Q2297045667:
After repeated testing, the memory leak bug can no longer be reproduced in the latest version.

PR #2025, opened 2026-04-14
feat: implement coral plants, coral fans and fix coral feature

GreenedDev:
## Description
Implemented coral plants and coral fans, their placing rules and behavior (dying without water)
Fix coral feature (placing)


pizzapunt53820:
The coral in this pull request doesn't replicate vanilla 26.1 yet in this seed.

Schermafdruk van 2026-04-14 15-22-19
Left: Vanilla

Right: Pumpkin

Seed: 873465098712865
Coordinates: 2698.5 51 -2699.5 facing north

GreenedDev:
> doesn't replicate vanilla 26.1

this PR didn't add that behavior. It was already there. This pr is only supposed to fix coral blocks hanging in water without support blocks.

pizzapunt53820:
Ok, then i read the title wrong. When i break a coral block, all connected coral fans or coral plants disappear. There are also no hovering coral fans generated in the same seed. All coral plants can be placed in the same places as in vanilla. The only issue i have is that specifically dead coral fans cannot be placed on the side of blocks, so only on top of blocks. This seems to work correctly on the normal branch. 

Example of what is possible in vanilla: 
Schermafdruk van 2026-04-14 17-59-02

pumpkin/src/block/blocks/coral/coral_fan.rs
```diff
@@ -0,0 +1,272 @@
...
+    } else if block == &Block::BRAIN_CORAL_FAN {
+        Some(Block::BRAIN_CORAL_WALL_FAN)
+    } else if block == &Block::BUBBLE_CORAL_FAN {
+        Some(Block::BUBBLE_CORAL_WALL_FAN)
+    } else if block == &Block::FIRE_CORAL_FAN {
+        Some(Block::FIRE_CORAL_WALL_FAN)
+    } else if block == &Block::HORN_CORAL_FAN {
+        Some(Block::HORN_CORAL_WALL_FAN)
+    } else {
+        None
```
RoosterBooster007, line 229: Adding the dead `CORAL_FAN` variants here should fix @pizzapunt53820's issue. This block could also be written cleaner as a `match` statement.

GreenedDev:
should be fixed now

pizzapunt53820:
The dead coral fan placement works now, but it seems that this pull request makes the server hang much more frequently than the main branch. For me, this pull requests works for a few seconds before the server stops responding and the main branch server keeps responding for multiple minutes. I don't know if this is caused by the changes or by the different build and if it is relevant.

GreenedDev:
> The dead coral fan placement works now, but it seems that this pull request makes the server hang much more frequently than the main branch. For me, this pull requests works for a few seconds before the server stops responding and the main branch server keeps responding for multiple minutes. I don't know if this is caused by the changes or by the different build and if it is relevant.

i haven't noticed that

pizzapunt53820:
> i haven't noticed that

Then i guess the server hang in my case is not caused by this pull request

pumpkin/src/block/blocks/coral/coral_fan.rs
```diff
@@ -0,0 +1,247 @@
...
+        Some(Block::DEAD_BRAIN_CORAL_WALL_FAN)
+    } else if block == &Block::DEAD_BUBBLE_CORAL_FAN {
+        Some(Block::DEAD_BUBBLE_CORAL_WALL_FAN)
+    } else if block == &Block::DEAD_FIRE_CORAL_FAN {
+        Some(Block::DEAD_FIRE_CORAL_WALL_FAN)
+    } else if block == &Block::DEAD_HORN_CORAL_FAN {
+        Some(Block::DEAD_HORN_CORAL_WALL_FAN)
+    } else {
+        None
+    }
```
Snowiiii, line 230: There has to be a better way, you know i hate this stuff

DaWaffles: I'm in no way intelligent when it comes to this stuff, so I have a question.
Why would making use of a ton of if statements be better than a match here?
I'm not familiar with the language functionality, but wouldn't a match be far cleaner and efficient?

GreenedDev: Because i can't match Block struct

GreenedDev: @Snowiiii If you don't have any suggestions on how to improve this I'll mark it as resolved ig.

Snowiiii:
Yah, maybe we find a better solution in the future, Still thank you @GreenedDev :D

PR #2033, opened 2026-04-15
fix(networking): add login timeout for java and bedrock

mw2000:
- add a shared `networking.login_timeout` setting for login/configuration flows
- disconnect Java and Bedrock clients that remain stuck before entering play
- apply existing auth connect/read timeouts to Mojang HTTP calls so login auth cannot hang indefinitely


Q2297045667:
Please fix CI.

Snowiiii:
Feel free to reopen

PR #1790, opened 2026-03-03
feat: crash reports with graceful shutdown

Laptop59:
## Description

Adds crash reports to Pumpkin.

When a thread panics while the server tries to start up or is running, a crash report is generated in the `crash-reports` directory. This PR also adds graceful shutting down when a panic is detected (sometimes the server takes too much time to stop, in that case `CTRL+C` should terminate it)

The crash report includes the time of panic, panic payload, the thread that panicked and its backtrace, and some server and system information.

## Testing

Added `panic!` temporarily in the code to test out the crash reports, but I removed them after testing.

### Upon panic when the server is running
Screenshot_20260301_201812

### Upon panic when the server is starting
Screenshot_20260303_174557

### Backtrace in the crash report (debug mode)
Screenshot_20260303_175351

### Backtrace in the crash report (release mode)
Screenshot_20260303_175232

### Server and system information in the crash report
Screenshot_20260303_175701

PR #2035, opened 2026-04-15
fix: add panic location to crash reports

Laptop59:
## Description

While making the crash reports PR (#1790), it looks like I forgot to put the panic location in the generated crash report file itself, even though it is printed on the console.
As this is a very key detail to diagnosing a crash, I have created this PR to add it near the top of the file:

## Testing
Screenshot_20260415_130305

As observed in the above screenshot, it has now been added.
(Don't mind the `stop.rs` panic location, I just added a `panic!()` macro call in it for testing. That's not included in the PR.)


Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #2039, opened 2026-04-16
Use existing impl for entity's nearest facing directions

GreenedDev:
## Description
instead of writing a new function, use a function that is already in Player


Snowiiii:

PR #2041, opened 2026-04-16
feat: Attack packet validation

chocodev11:
makes combat packet handling reliable by adding shared pre-hit validation for both attack packet paths, including target validity, border/reach checks, and minimum attack charge gating. It also adds missing item data component support (attack_range, minimum_attack_charge, piercing_weapon)

## Testing


Snowiiii:
I kinda think this is too much, would rather just prefer having the data component logic added, i don't think we necessarily need to do much validation

PR #2043, opened 2026-04-17
feat(plugin-api): TextComponent add color functions

FlorianLang06:
## Description
This PR adds two functions to TextComponent in the plugin API.

1. function `text-colored` creates a TextComponent and color it with a NamedColor.
2. function `text-rgb` creates a TextComponent and color it with a RGB color.

Both methods make it easier to create text components with color, so the devs don't ned a speared line to color the component. Unfortunately functions with same name are not possible in WIT. If you wish to also have this functions for translate please let me know (I was not sure if this is needed). 

## Testing
- `cargo test` passes
- `cargo clippy --all-targets`: no warnings 
- Used the function in [this branch[(https://github.com/FlorianLang06/PumpkinPing/tree/color-fn) of my plugin and it worked.


Snowiiii:
tbh, i kinda feel this is unnecessary, since chaining support is planned to be extended

PR #2044, opened 2026-04-18
Fix drowning damage and air supply handling

### LINKED ISSUE #2029: Drowning immediately

### Operating System

bluefin linux (fedora silverblue)

0.1.0-dev+26.1 (Commit: 1d70ec5/release)

### What happened?

Drowning ignores air remaining and starts immediately after entering water after drowning and then not respawning for a bit of time

https://github.com/user-attachments/assets/b1e8d6ea-d773-45bd-bd17-591d65e62b30

### To Reproduce

1. Start server
2. Drown once
3. Wait ~1 minute before respawning
4. Enter water again
5. Success?

### Expected behavior

Drowning should only start after all air bubbles have been depleted.

maxer137:
## Description
This PR fixes several issues with the breath/drowning system in breath.rs.
Such as magic numbers, dead entities still being handled, and air supply going far beyond 0.

### What was changed

- Introduced named constants (`MAX_AIR`, `AIR_RECOVERY_RATE`, `AIR_DEPLETION_RATE`, `DROWNING_INTERVAL`, `DROWNING_DAMAGE`) to replace magic numbers throughout the breath tick logic.
- Added an early return to skip breath processing entirely for dead entities.
- Fixed a bug where `air_supply` was never stored back after being clamped to `0`, causing `fetch_sub` to drive it arbitrarily negative. This meant recovery after leaving the water would take far longer than intended. 
- Updated air metadata serialization type to `INT` for Minecraft 26_1 protocol compatibility.

### Why were these changes necessary?
Players were instantly drowning after death. The root cause was that `air_supply` was going negative during drowning, meaning that when they left the water, it would first have to count back all the way to the positive before actually being restored. Additionally, dead players were unnecessarily running through breath logic.

Together, these two would lead to a situation where, if a player were to drown and not respawn, the `air_supply` would continue to go far into the negative.  

**Impact:**

Drowning damage now applies correctly. Air supply recovers immediately from `0` upon leaving water.

**Known issues/limitations:**

None.

Closes #2029

### DIFF
--- a/pumpkin/src/entity/breath.rs
+++ b/pumpkin/src/entity/breath.rs
@@ -74,6 +74,7 @@ impl BreathManager {
                 .fetch_sub(AIR_DEPLETION_RATE, Ordering::Relaxed);
             let new_air = (prev - AIR_DEPLETION_RATE).max(0);
             if new_air != prev {
+                self.air_supply.store(new_air, Ordering::Relaxed);
                 self.send_air_supply(player).await;
             }
 
--- a/pumpkin/src/entity/breath.rs
+++ b/pumpkin/src/entity/breath.rs
@@ -13,6 +13,12 @@ use pumpkin_util::math::position::BlockPos;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicI32, Ordering};
 
+pub const MAX_AIR: i32 = 300;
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM. Thank you @maxer137 :D

PR #2046, opened 2026-04-18
Rework the versioning logic in the Wasm plugin loader

QuantumSegfault:
## Description

This PR reworks the plguin versioning system. Instead of custom metadata with strict behavior, we rely on the Wasm component model's own concept of versioning. This locks the plugin system in to semantic versioning.

Firstly, the v0_1_0 API has been renamed across the board to just v0_1. Only major versions (0.2.x, or 1.x.y, 2.x.y, etc.; v0_2, v1, v2) need to have separate implementation.

All major APIs are provided to a single `Linker` instance. This has the effect that plugins may mix and match major APIs at whim. During plugin load, the imports are first checked the linker, and will result in linker failure if one or more imports of the plugin can't be satisfied.

After this initial verification pass, the plugin is then checked against the declared exports of each major API in reverse order (e.g. when v0_2 is added, it will be checked before v0_1). Once a match is found (instantiates without error), the plugin is marked as that major version. Failing to conform to any version of the exports will yield in the old `ApiVersionMismatch`.

These changes have a number of effects:
- In theory, plugins may freely mix and match API versions. How practical this is in reality is a different story.
- The system is backwards compatible, and somewhat "forward compatible". Plugins targetting 0.1.0 can be loaded in Pumpkin implementing 0.1.4. The reverse also applies, where Pumpkin API 0.1.0 will accept 0.1.4 plugins, so long as the 0.1.4 plugin does not use any features not present in 0.1.0
- Plugins made for 0.2.0 (new major version) will be rejected by old Pumpkin, but old major versions will still be supported, at least for a time
- There is something of a divide in what the versioning means. A plugin's API "version" is detected by what version of the exports/callbacks it implements (and thus what the system invokes; e.g. onLoad), not what APIs it consumes.
- Changes to any particular major version can happen freely, so long as the changes are backwards compatible (no breaking changes). Once a breaking change is required (removing deprecated functionality, renaming, etc.) a new major version has to be implemented in its entirety.
- Hypothetically, different parts of the API can be versioned independently.

## Testing

`cargo test`, `cargo clippy --all-targets`, and `cargo fmt` all pass.

Pumpkin does not have multiple major API versions at present, so testing thoroughly is tricky

I've done some theoretical experiments at https://github.com/QuantumSegfault/wit-plugin-versioning-experiments

I've also done some testing here at Pumpkin fudging the version numbers in plugin bindins to use a 0.1.4 and 0.2.0 API (both otherwise identical to 0.1.0). 0.1.0 and 0.1.4 loaded sucessfully, while 0.2.0 was rejected (as desired).

Further testing and tweaks will have to wait until a a new major version (e.g. 0.2) of the API is released.


QuantumSegfault:
@Snowiiii @BjornTheProgrammer

Snowiiii:

PR #2050, opened 2026-04-20
feat(command): reimplement `/banlist`

Laptop59:
## Description

Reimplements the `/banlist` command with the new command system, along with a few other changes:
- Instead of using a tuple `(String, String, String)`, we switch to a new struct `BanListEntry` for more clarity.
- 'Unifies' both the previous executors to only one new `BanListCommandExecutor` by storing `bool`s in the executor itself. This also means we don't need to declare another function outside anymore to print the entries.
- Description is more descriptive.
- Instead of using argument nodes, we use literal nodes.

## Testing

Tested the command in-game, and it works as intended.


vyPal (approved):

Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #2051, opened 2026-04-20
fix(codec): replace codec and map codec structs with `Encode` and `Decode` traits and their field versions

SomeYellowGuy:
## Description
Removes codecs in favor of the `Encode` and `Decode` traits (similar to a codec), which is more Rust-idiomatic for serialization.
- This PR adds built-in support for implementing both traits for:
  - Primitive DFU types (all Java numbers, `String`, `bool`, `IntStream`, `LongStream` and `ByteBuffer`) and *unsigned* integer types
  - Lists (`Vec` or `BoundedVec` for specifying a size limit)
  - Unbounded `HashMap`s
  - `Either`s
- I also added `FieldEncode` and `FieldDecode` (similar to a map codec) to encode/decode specific fields of a map.
Defaulted fields (fields that are optional and have a default value) can also be encoded/decoded via `FieldEncode` and `FieldDecode`.

  Technically, `Option`s don't implement `Encode` and `Decode`, neither do they implement `FieldEncode` nor `FieldDecode`; they only implement special traits `OptionalFieldEncode` and `OptionalFieldDecode`, which both provide methods to add optional fields.

- Added `FlatTryFrom` and `FlatTryInto`, which are similar to Rust's `TryFrom` and `TryInto`, except that they return a `DataResult` instead of a `Result`.

I will add derive macros for `Encode` and `Decode` and remake the `NbtOps` tests in a later PR.

## Testing

Added tests for primitive types, unsigned number types, lists, maps and eithers, which all pass.


Snowiiii:

PR #2053, opened 2026-04-21
perf: enable opt-level=3 and lto=fat for release profile

shulng:
perf: enable opt-level=3 and lto=fat for release profile


Snowiiii:
This changes nothing since both are already default

PR #2055, opened 2026-04-21
feat(codec): add derive macros for `Encode` and `Decode`

SomeYellowGuy:
## Description
Adds derive macros for `Encode` and `Decode` (in a new crate `pumpkin_codecs_macros`). They currently support *structs* and *enums*. Here are the cases of the derive macros, looking for the first satisfying condition:
- Unit struct: Simply encodes an *empty map* and directly decodes the struct.
- Any other struct: Encodes every required field and decodes every required field.
- Enum whose variants have no fields: Encodes each enum variant as a *string*.
- Any other enum: Adds a *differentiator entry* (e.g. `"type": "multi_action"`) to determine the type of variant for decoding, and adds the required fields of the encoded variant.

In addition, this PR also adds `MapEncode` and `MapDecode` to act as the equivalent of a `MapCodec` from which a `Codec` may be adapted.

## Testing

Added tests for struct and enum cases.


Snowiiii:
Way too much code in one PR, and also i'm not sure if this is necessary

SomeYellowGuy:
It would be better to write some derive macros for `Encode` and `Decode` (analogous to codecs) instead of manually writing each of them. Also, in terms of code, there isn't much. This is a little more than the bare minimum to get the macros working. Maybe the tests were a little big?

PR #2060, opened 2026-04-23
fix: Jungle Tree Trunks generation

GreenedDev:
## Description

Before:

After:

I also fixed a typo in dirt placing (below a giant tree trunk) where it was doing .south() twice instead of south().east()


Snowiiii:

PR #2062, opened 2026-04-25
Feat(storage) dynamic Storage types

topi-banana:
## Description

This PR introduces an in-memory storage backend by abstracting every persistent-data domain behind its own trait in a new `pumpkin-storage` crate. The server now holds storage as `Arc<dyn ...Storage>` trait objects, so any implementation that satisfies the trait can be plugged in at construction time.

### What changed

- **New crate `pumpkin-storage`** — owns the trait definitions, error type, and shared domain types (`LevelData`, `BannedPlayerEntry`, `BannedIpEntry`, `UserCacheEntry`, `PoiEntry`).
- **Domain traits** — one trait per persistence concern:
    - `LevelInfoStorage` — world metadata (`level.dat`)
    - `PlayerDataStorage` — per-player NBT (`playerdata/<uuid>.dat`)
    - `ChunkStorage<T>` — chunk payloads, generic over the chunk type so the same trait is reused for blocks and entities
    - `BannedPlayerStorage`, `BannedIpStorage`, `OpStorage`, `WhitelistStorage` — server config files (`banned-players.json`, `banned-ips.json`, `ops.json`, `whitelist.json`)
    - `UserCacheStorage` — name/UUID cache (`usercache.json`)
    - `PoiStorage` — Points of Interest (`poi/r.X.Z.mca`)
- **Three concrete backends:**
    - `VanillaStorage` — filesystem-backed, vanilla-compatible on-disk layout. Split into a per-world directory (`world_dir`) and a server-wide data directory (`server_data_dir`).
    - `MemoryStorage` — fully in-memory, format-agnostic. Holds domain values directly with no serialization or disk I/O. `MemoryChunkStorage<T>` is exposed separately because chunks are generic.
    - `NullStorage` — discards every write and returns `NotFound` for reads. Used when player-data persistence is disabled and as a placeholder where a real backend is not desired.
- **`StorageError`** — single error enum surfaced by every trait. Carries optional path context for I/O errors and exposes `is_not_found()` plus a `severity()` helper that maps each variant to a `tracing::Level` (NotFound → DEBUG, IO → WARN, format/version errors → ERROR).
- **Backend selection via config** — `AdvancedConfiguration` gains a `StorageConfig` enum (`Vanilla(LevelConfig)` | `InMemory`). `Vanilla` keeps the existing per-world settings (chunk format, lighting, autosave); `InMemory` carries no file-layout options and synthesises a default `LevelConfig` for the cross-cutting fields shared with the file backend.
- **Server wiring** — `Server` now stores its storages as `Arc<dyn ...Storage>` and constructs them once from `StorageConfig`. The on-disk session lock and the `region/`, `entities/` directory creation are skipped for the in-memory backend since they have no meaning there. `ServerPlayerData` becomes a thin wrapper around `Arc<dyn PlayerDataStorage>`, falling back to `NullStorage` when persistence is disabled.
- **Chunk I/O bridge** — `FolderBoundFileIO<T>` adapts the legacy `pumpkin-world` `FileIO` trait (which takes `&LevelFolder` on every call) into the new folder-aware `ChunkStorage<T>`, so call sites only need to know about `Arc<dyn ChunkStorage<T>>`.
- **Lock guard generalisation** — `LockGuard` becomes a marker trait on the released-on-drop guard, allowing the server to hold `Box<dyn LockGuard>` regardless of the concrete locker (and to hold `None` for in-memory mode).
- **Command paths and player flows** — `ban`, `pardon`, `op`, `deop`, `whitelist`, `banlist`, `pardonip`, `banip`, the gameprofile arg consumer, the join gate (`can_not_join`), `Player::ban`, `Player::ban_ip`, and the nether-portal POI lookups all go through the new traits instead of the previous `server.data.*` config wrappers.
- **Removed legacy modules** — `pumpkin/src/data/{banlist_serializer,banned_ip,banned_player,op,usercache,whitelist,mod}.rs`, `pumpkin-world/src/data/player_data.rs`, `pumpkin-world/src/poi/`, `pumpkin-world/src/world_info/anvil.rs`. Their responsibilities are now covered by the trait implementations in `pumpkin-storage`.

### Why

- Lets users choose ephemeral, persistence-free servers (tests, scratch worlds, embedded scenarios) without touching disk.
- Decouples domain logic from on-disk format; future backends (alternative on-disk layouts, network-backed stores, etc.) can be added by implementing the traits without changing call sites.
- Centralises error reporting and log-level classification across every persistence domain.

### Authoring note

This PR was drafted in collaboration with Claude (LLM-assisted coding). All resulting code was reviewed line-by-line and rewritten by a human author before being committed; the LLM output was used as a starting point rather than as the final artefact.

### Future work

- Add a comprehensive compatibility test suite that drives the full server through `MemoryStorage`, exercising end-to-end gameplay flows (login, world load, command execution, ban/whitelist/op semantics, chunk save/load, POI lookup) against the in-memory backend so the trait surface stays behaviour-compatible with the file-backed backend.

### Compatibility / impact

- Default behaviour is unchanged: `StorageConfig` defaults to `Vanilla`, which uses the same on-disk layout, locking, and command behaviour as before.
- `features.toml` users who customised chunk/lighting/autosave settings keep working — those fields now live under `[storage]` (`type = "vanilla"`).
- Selecting `type = "in_memory"` skips the session lock, the `region/` and `entities/` directory creation, and all on-disk I/O for level metadata, player data, banlists, ops, whitelist, user cache, POIs, and chunks.

## Testing

- `cargo test -p pumpkin-storage` — conformance tests in `pumpkin-storage/src/conformance_tests.rs` exercise each domain trait against `VanillaStorage`, `MemoryStorage`, and `NullStorage`. The same scenarios run for every backend to confirm round-trip semantics, "not found" handling, expired-ban filtering, MRU eviction in user cache, POI radius queries, and persistence across `VanillaStorage` instances.
- Existing unit tests in `pumpkin-storage/src/vanilla/level_info.rs` continue to verify `level.dat` (de)serialization against the bundled `assets/level_1_21_4.dat` / `assets/level_1_20.dat` fixtures (including the unsupported-version error path).
- `cargo clippy --all-targets` — fixed clippy pedantic warnings across the new crate.
- Manual smoke test:
    - [x] Start the server with default config (`type = "vanilla"`); confirm world loads, ops/whitelist/banlist commands persist across restart.
    - [x] Switch to `type = "in_memory"` in `features.toml`; confirm server starts, player can connect, no `world/` or `data/` files are written, state resets on restart.


Q2297045667:
This PR is too large, which makes it hard to review. It would be better to split it into multiple smaller PRs.

PR #2064, opened 2026-04-25
Fix/structure block direction

GreenedDev:
## Description
This pr is supposed to fix wrong stair directions in structures. should work for other blocks too.


GreenedDev:
Accessing BlockRegistry (or WorldPortalExt) in pumpkin-world/src/generation/structure/structures/mod.rs add_block() function remains problematic.

I need BlockRegistry so that i can call block-behavior specific .mirror() and .rotate() for each block (these functions should be added in each block's BlockBehaviour implementation).

GreenedDev:
@Snowiiii any ideas on how may I approach this?

GreenedDev:
I'm no longer working on this.

PR #2066, opened 2026-04-26
fix(plugin): fix native plugin registration hang on macOS

faustyu1:
## Problem

On macOS, native plugins are dylibs that statically link their own copy of pumpkin — and therefore their own copy of hashbrown. When plugin code called `HashMap::entry()` or `tokio::sync::RwLock::write()` on data structures owned by the **server** binary, two things went wrong:

1. **HashMap infinite loop** — hashbrown uses `Group::static_empty()` as a sentinel in its probe sequence. Each binary has its own copy at a different address. Plugin code probing a server-owned map sees the wrong sentinel and loops forever.
2. **RwLock permanently contended** — same binary boundary issue causes `try_write()` to always return `Err` on a lock created by the server binary.

The symptoms were: `register_event` and `register_command` called from `on_load` would hang indefinitely, preventing the plugin from ever finishing initialisation.

## Fix — two-stage registration for `on_load`

Add `register_command_sync` / `register_event_sync` to `Context` that push into plain `Vec` queues (no HashMap involved → no cross-binary issue).

After `on_load` returns, `spawn_plugin_initialization` drains those queues **inside the server binary**, where the correct hashbrown copy is in scope, and does the actual HashMap insertions there.

The existing async `register_command` / `register_event` are unchanged and remain correct for registrations that happen after startup (e.g. from event handlers at runtime).

## Fix — native loader metadata

The loader read the `METADATA` symbol as `*const PluginMetadata`. That only works when the static *is* the pointer. The idiomatic way to export metadata from Rust is `LazyLock<PluginMetadata>` (required because `PluginMetadata` has `String` fields), but reading a `LazyLock` as `*const T` gives garbage — the first bytes are the `AtomicUsize` init state, not a pointer to the value.

Add support for a `get_metadata() -> *const PluginMetadata` function export. When present it is preferred over the raw `METADATA` static, letting the plugin initialise the lazy cell and return a stable pointer to the inner value. The old `METADATA` path is kept for backwards compatibility.

## Plugin-side usage

```rust
// Export so the loader can read LazyLock<PluginMetadata> correctly
#[unsafe(no_mangle)]
pub extern "C" fn get_metadata() -> *const PluginMetadata {
    &*METADATA as *const PluginMetadata
}

impl Plugin for MyPlugin {
    fn on_load(&mut self, ctx: Arc<Context>) -> PluginFuture<'_, Result<(), String>> {
        Box::pin(async move {
            // Safe to call from on_load — queued, no cross-binary HashMap ops
            ctx.register_command_sync(my_command_tree(), "myplugin:cmd");
            ctx.register_event_sync(Arc::new(MyHandler), EventPriority::Normal, true);
            Ok(())
        })
    }
}
```

## Testing

Verified on macOS (arm64) with a native dylib plugin. Before this fix the server hung indefinitely inside `on_load`. After, the plugin loads cleanly and commands + handlers are registered correctly.

### DIFF
--- a/pumpkin/src/plugin/api/context.rs
+++ b/pumpkin/src/plugin/api/context.rs
@@ -22,6 +22,9 @@ use tracing_subscriber::{fmt, layer::SubscriberExt, util::SubscriberInitExt};
 
 use super::{EventPriority, Payload};
 
+type PendingCommands = Arc<std::sync::Mutex<Vec<(crate::command::tree::CommandTree, String)>>>;
+type PendingHandlers = Arc<std::sync::Mutex<Vec<(&'static str, Box<dyn DynEventHandler>)>>>;
+
 /// The `Context` struct represents the context of a plugin, containing metadata,
 /// a server reference, and event handlers.
 ///
@@ -47,10 +50,8 @@ pub struct Context {
     // The fix: plugin code only pushes to these plain `Vec`s (no hashmap involved);
     // the loader drains them *after* `on_load` returns, doing all hashmap work inside
     // the server binary where the right hashbrown copy is in scope.
-    pub(crate) pending_commands:
-        Arc<std::sync::Mutex<Vec<(crate::command::tree::CommandTree, String)>>>,
-    pub(crate) pending_handlers:
-        Arc<std::sync::Mutex<Vec<(&'static str, Box<dyn DynEventHandler>)>>>,
+    pub(crate) pending_commands: PendingCommands,
+    pub(crate) pending_handlers: PendingHandlers,
 }
 impl Context {
     /// Creates a new instance of `Context`.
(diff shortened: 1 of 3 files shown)


Laptop59:
You should run `cargo fmt` and check clippy warnings, so that the CI can succeed.

Snowiiii:
I'm not a big fan of the pending Vecs, would prefer a more direct way

faustyu1:
Changed it to register commands and handlers directly through server-owned callbacks instead of collecting them in pending Vecs. This keeps the HashMap work on the server binary side while avoiding the extra pending queues.

faustyu1:
@Snowiiii

Snowiiii:
@faustyu1 i would close this for now because it seems to much work for just one platform, especially because native is not supported as 1 tier plugin platform

PR #2072, opened 2026-04-30
fix(codegen): Fix clippy warnings in `pumpkin-data`'s `block.rs`

QuantumSegfault:
## Description

Tweaks some things in `pumpkin-codegen/src/block.rs` so that the emitted `pumpkin-data/src/generated/block.rs` no longer trips `cargo clippy`.

The diff is smaller than it should be, as `cargo clippy --fix` was ran on `master`. This is more about fixing warnings for new runs/regens of `pumpkin-codegen`, then actually solving them on `master`.

## Testing

This PR passes `cargo clippy`. It is otherwise an NFC.


QuantumSegfault:
@Snowiiii 

Requesting review

Snowiiii:

PR #2075, opened 2026-04-30
Fix: truncate query packet buffer to actual UDP packet length

KurtThiemann:
## Description

In the Minecraft query protocol, the only difference between a basic stat request and a full stat request is 4 bytes of padding at the end of the packet. `SStatusRequest` therefore uses the buffer length to determine which one it is and returns an error if the length matches neither of those types. Currently, however, the length of the buffer is always 1024, as allocated in `start_query_handler`. Therefore, neither basic stat nor full stat requests work at the moment.

This PR uses the actual packet size from the recv_result and truncates the buffer to that length before passing it to the handler function.


Snowiiii:
LGTM. Thank you @KurtThiemann :D

PR #2088, opened 2026-05-03
feat(command): implement SNBT parser

Laptop59:
## Description

**SNBT** is a **S**tringified representation of **NBT** tags.

This PR mainly creates a SNBT parser that hopes to recreate the same errors and suggestions as that of Java Edition, at the same positions, given the same SNBT strings. It probably doesn't act exactly like Vanilla 100% of the time (*yet*), but surprisingly even though they work differently, they produce the same error messages in many edge cases, so it actually is really close. I implemented it in this way to make it much more faster than Vanilla.

One thing that hasn't been implemented yet here is `\N{ ... }` in strings, which take a Unicode character name and return the character who has that name, mainly because it's probably not used much and would take effort to generate a constant compile-time map.

This will be required for parsing of arguments of:
- NBT tags
- Text Components
- Block States
- Block Predicates
- Items
- Item Predicates
- Particles
- Predicates in `/execute`

A few other fundamental changes have also been made:
- Made NBT tag `ByteArray` store a `Vec<i8>` instead of `Box<[u8]>` for consistency with `IntArray` and `LongArray`.
- NBT Compounds are now stored as a `HashMap<String, NbtTag>` instead of `Vec<(String, NbtTag)>` to ensure that duplicate keys cannot occur and also matches Vanilla behavior.
- Suggestions now keep insertion order before sorting case-insensitively.
- `read_unquoted_string` no longer returns an `Option` as it cannot fail.

## Testing
Added a separate `tests` module in the `snbt` module to ensure the parser acts very similarly to that of Java Edition, and all tests pass.


Snowiiii:
LGTM, Thank you @Laptop59 :D

PR #2091, opened 2026-05-04
fix: tall plant block update check

GreenedDev:
## Description
if you broke the support block, the tall grass used to not break.


Snowiiii:

PR #2097, opened 2026-05-05
chore: only fetch contributors once

awildergoose:
## Description

This makes it so the contributors from the `/pumpkin` command are only fetched once, on startup.

## Testing

Works as expected, ran command, checked GlassWire and there were no requests :)

### DIFF
--- a/pumpkin/src/command/commands/pumpkin.rs
+++ b/pumpkin/src/command/commands/pumpkin.rs
@@ -13,13 +13,15 @@ const NAMES: [&str; 2] = ["pumpkin", "version"];
 
 const DESCRIPTION: &str = "Display information about Pumpkin.";
 
-struct Executor;
+struct Executor {
+    contributors: Vec<Contributor>,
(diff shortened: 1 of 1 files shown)


Laptop59:
I'm a bit worried about the server startup time being affected due to this. If that's the case, the issue can easily be resolved by using a `OnceLock`.
So when the command is run for the first time, it gets initialized only once, and then later in subsequent times, it will be the same.

awildergoose:
> I'm a bit worried about the server startup time being affected due to this. If that's the case, the issue can easily be resolved by using a `OnceLock`.
> So when the command is run for the first time, it gets initialized only once, and then later in subsequent times, it will be the same.

That sounds like a great idea, I'll do that when I get the free time to, thanks!

Snowiiii:
Yes you should either use LazyLock or OnceLock

Snowiiii:
Feel free to reopen

PR #2101, opened 2026-05-05
feat(command): implement `list_suggestions` for some argument types

Laptop59:
## Description

This PR implements the `list_suggestions` method for the few `ArgumentType`s we currently have that should give suggestions. This allows the server console to get proper, argument-specific suggestions!
It also reinforces the permission for using selectors, by checking if the source has permission to use them.

## Testing

Tested some selectors in the `/kill` command, and it works as intended:
Screenshot_20260505_235153


Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #2106, opened 2026-05-07
feat(command): make `GameProfileArgumentType` search and reimplement `/op` command

SomeYellowGuy:
## Description
This PR makes the `GameProfileArgumentType` able to search for online players (by copying the existing code from the equivalent `args` argument).

Along with that, it also reimplements the `/op` command with the new argument type.

## Testing

Tested the command, and it seems to work properly with online, offline and non-existent players.


Snowiiii:

PR #2109, opened 2026-05-08
feat(command): reimplement `/tick` command

SomeYellowGuy:
## Description
Reimplements the `/tick` command with the newly made custom suggestion providers (which replace the hardcoded literal arguments from earlier.)

This also fixes a regression that prevented the client from asking the server for custom suggestions.

## Testing
Tested the command, and the suggestions (along with the command itself) work.


Q2297045667:
It's super important. My ticks are all over the place during tests—they just aren't stable.

Snowiiii:

PR #2117, opened 2026-05-09
feat(command): implement UUID argument type

Laptop59:
## Description

Implements the argument type `UuidArgumentType`, which parses UUIDs like Java.
To do this, I decided to move the methods and tests responsible for parsing UUIDs from the `pumpkin` crate to the `pumpkin-util` crate, so that it can be used by some other Pumpkin crates as well, in addition to using it in the argument type.

Also fixes the argument type numeric IDs so that UUIDs display correctly client-side.

## Testing
Tested the UUID argument type in a local-only test command, and it works as intended.


Snowiiii:
LGTM. Thank you @Laptop59 :D

Issue #2105, opened 2026-05-07
Picking up items inside of a boat does not work


impresson:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Server: Debain 13 Client: Macos 26.3

### Server Software Version/Commit

 63d39f4

### What happened?

Players inside of a boat do not pick up items when near them

### To Reproduce

drop a couple items on the ground and then get inside of a boat and head towards the items and it does not pick up

### Expected behavior

player inside of the boat is expected to pick up the items

SepehrRajabi:
Hello.
The cause of the bug was that the world collision scan used the player entity’s own bounding box for item pickup. When a player rides a boat, vehicle movement updates the boat’s position/bounding box, while the passenger’s entity box can remain at the mount/start position. The item collision callback never fired near the boat.

i was able to fix the bug here: https://github.com/SepehrRajabi/Pumpkin/commit/5921597554789f89c319f45869b4adaeafa8085d

I also added item drop for whenever the boat was hit enough times.

However, there is one thing i have not been able to implement yet.
- the wobble animation for when the player hits the boat.

I also realized that the world state does not save correctly.
I am running minecraft version: 26.1 on macos 26.4

GreenedDev:
fixed by https://github.com/Pumpkin-MC/Pumpkin/pull/2166

PR #2118, opened 2026-05-09
Fix for Picking up items inside of a boat

SepehrRajabi:
I Fixed the player not being able to pick up items when they are in a boat. https://github.com/Pumpkin-MC/Pumpkin/issues/2105

The cause of the bug was that the world collision scan used the player entity’s own bounding box for item pickup. When a player rides a boat, vehicle movement updates the boat’s position/bounding box, while the passenger’s entity box can remain at the mount/start position. The item collision callback never fired near the boat.

I will add a quick recording of it


Snowiiii:
You changed a lot of unnecessary files

PR #2123, opened 2026-05-10
fix(entity): mobs no longer burn at night

Giems:
## Description

Fixed sun-sensitive mobs burning to death during the night by adding day/night gate from vanilla. For now hardcoded values from official .json and left TODO.

## Testing

Manually verified in-game by setting time at boundary ticks using '/time' command. Rain still blocks burning via the existing check.


Snowiiii:
LGTM. Thank you @Giems :D

PR #2127, opened 2026-05-12
refactor(pumpkin-util): gate codegen deps behind feature flag

topi-banana:
## Summary

- Make `syn`, `quote`, and `proc-macro2` optional in `pumpkin-util`, exposed via a new `codegen` feature.
- `pumpkin-codegen` opts into the feature explicitly.
- Downstream consumers of `pumpkin-util` no longer pay the proc-macro build cost when they don't need the `ToTokens` impls.

The `ToTokens` impls in `pumpkin-util` (in `loot_table.rs`, `math/experience.rs`, `math/float_provider.rs`, `math/int_provider.rs`) are only consumed by `pumpkin-codegen`. Within the workspace nothing else imports them, so gating them behind a feature is a no-op for runtime crates while keeping codegen functionality intact.

## Note on API surface

This moves the existing `ToTokens` impls behind the new `codegen` feature. For workspace consumers this is transparent (only `pumpkin-codegen` uses them, and it now enables the feature). External downstream users that depended on these impls would need to enable `features = ["codegen"]` on `pumpkin-util`.


Snowiiii:
LGTM. Thank you @topi-banana :D

PR #2128, opened 2026-05-12
feat(plugins): Integrate `wasmtime_wasi_http`

QuantumSegfault:
## Description

Integrates `wasmtime_wasi_http` into the plugin loader to allow plugins to access [`wasi:http`](https://github.com/WebAssembly/WASI/tree/main/proposals/http), enabling for easier access to usable HTTP client/request implementations for plugins.

Instead of having to find a working, feature-complete (e.g. TLS is a tough one) port of an existing HTTP request library that works in their language over `wasi:sockets` (and dealing with the associated bloat), slimmer libraries using `wasi:http` can be utilized (which moves the bulk of the work over to the host instead). Even in languages that don't have a neat wrapper library for `wasi:http`, raw bindings to the interface can be used with relative ease.

## Testing

`cargo fmt` and `cargo clippy` pass.

Testing was done with a simple Rust plugin using [`waki`](https://crates.io/crates/waki) (which wraps `wasi-http` in a simple API) to fetch some data from the Pumpkin Market API. See the gist for source: https://gist.github.com/QuantumSegfault/21fa165ff2ccd9dd1ada358656d34740

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1539,8 +1539,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed5909b6e89a2db4456e54cd5f673791d7eca6732202bbf2a9cc504fe2f9b84a"
 dependencies = [
  "foldhash 0.2.0",
- "serde",
- "serde_core",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


samolego:
Tried and it works :D

Snowiiii:

PR #2131, opened 2026-05-14
feat(plugins): Enable WasmGC

QuantumSegfault:
## Description

Flips some switches in the `Engine` config for the Wasm plugin loader to allow the use of newer GC and exception handling features in plugins.

This is required to be able to build Pumpkin Wasm plugins with some languages, e.g. Kotlin.

## Testing

`cargo clippy` and `cargo fmt` both pass

I don't have a publicly available reproducer right now, but I tested this building and running a dummy plugin in Kotlin (just logs on load and unload). Kotlin uses WasmGC and exception handling features when targetting Wasm with Kotlin/Wasm.


Snowiiii:

PR #2132, opened 2026-05-14
fix: orientation placing block in-place of existing one 

### LINKED ISSUE #2112: Stairs rotate incorrectly when placed on a snow layer

### Operating System

Macos Tahoe 26.3

45867d3

### What happened?

Placing stairs on a snow layer above any block causes the rotation to be wrong.

**Also happens with:**

- Slabs
- End rods
- Trapdoors
- Buttons & Levers don't place

### To Reproduce

1. Get a block
2. Place a snow layer on top of that block
3. Place stairs on the top side of the snow layer

### Expected behavior

The stairs should be rotated correctly according to the players direction

AurelienFT:
Fix https://github.com/Pumpkin-MC/Pumpkin/issues/2112

## Description

When a block is placed on a replaceable block (snow, grass, etc.), the clicked face was passed as-is to the placement logic, unlike the non-replaceable path which uses `face.opposite()`. This caused directional blocks (stairs, slabs, trapdoors, buttons) to compute incorrect orientation. Fixed by inverting the face in the replaceable case to match the existing convention.

## Testing

Tested manually :
- Stairs on snow ✅ 
- Lever on snow ✅ 
- Stairs on water ✅

### DIFF
--- a/pumpkin/src/net/java/play.rs
+++ b/pumpkin/src/net/java/play.rs
@@ -2667,7 +2667,7 @@ impl JavaClient {
 
         let (final_block_pos, final_face, replacing) =
             if let Some(replacing) = replace_clicked_block {
-                (clicked_block_pos, face, replacing)
+                (clicked_block_pos, face.opposite(), replacing)
             } else {
                 let block_pos = BlockPos(location.0 + face.to_offset());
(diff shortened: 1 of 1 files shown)


Snowiiii:

PR #2134, opened 2026-05-14
fix: Place leaves around the trunk instead of 0,0

### LINKED ISSUE #2126: Mangrove trees generate without leaves

### Operating System

Windows 11

 0.1.0-dev+26.1 / c8633b7

### What happened?

Image

Mangrove trees generate without leaves

### To Reproduce

Create a new pumpkin server with the seed 1776928931536313616
then teleport to coordinates x: 1189 z: 1783

There will be a mangrove swamp with trees that have missing leaves

### Expected behavior

mangrove trees should generate with all leaves

AurelienFT:
Fixes https://github.com/Pumpkin-MC/Pumpkin/issues/2126

## Description

Leaves generated by `random_spread.rs` was generated around 0,0 coordinates instead of being around of trunk coordinates.

## Testing

I have tested with the world provided in the bug report.

When we look at the world, we see more leaves than we usually see on a minecraft vanilla world and this is due to the mangrove being only trunks with no roots. I haven't made the root generation in this MR because it's just a bugfix, it would added hundreds of changes to the PR and I think it can be merged without fixing roots. (I will try to do it later on)

### DIFF
--- a/pumpkin-world/src/generation/feature/features/tree/foliage/random_spread.rs
+++ b/pumpkin-world/src/generation/feature/features/tree/foliage/random_spread.rs
@@ -1,6 +1,6 @@
 use pumpkin_data::BlockState;
 use pumpkin_util::{
-    math::{int_provider::IntProvider, position::BlockPos},
+    math::{int_provider::IntProvider, position::BlockPos, vector3::Vector3},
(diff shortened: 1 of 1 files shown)


Snowiiii:

PR #2136, opened 2026-05-14
Add vines decorator to leaves

AurelienFT:
## Description

Implements the `leave_vine` tree decorator (was a no-op stub) so `jungle`, `mega_jungle`, `swamp_oak`, `mangrove` and `tall_mangrove` trees get hanging vines on their leaves matching vanilla.

Foliage placers now return the positions of placed leaf blocks; these are threaded through `TreeFeature::generate_main` to `TreeDecorator::generate` via a new `foliage_positions` parameter. `LeavesVineTreeDecorator` is ported from vanilla. For each leaf, each horizontal direction has `probability` chance to spawn a vine that hangs down up to 4 air blocks. The decorator's `probability` field (0.125 for mangroves, 0.25 for the others) is now parsed by the codegen instead of being silently dropped.

## Testing

I have re-used the world from : https://github.com/Pumpkin-MC/Pumpkin/issues/2126
I can see leaves on jungle tree and mangrove tree with leaves (after https://github.com/Pumpkin-MC/Pumpkin/pull/2134)

leaves


Snowiiii:

PR #2141, opened 2026-05-15
feat: implement plugin whitelisting and configuration persistence

NicolaiVdS:
## What was changed?
- `pumpkin-server`:
  -  Converted `PluginManager::ask_permission_confirmation` from a static method to an `async` instance method (`&self`).
  - Implemented a whitelist check: If a plugin's name and version match an entry in `advanced_config.plugins.whitelist`, the permission prompt is skipped automatically.

Currently, server administrators are forced to manually approve plugin permissions on every single startup. This creates friction for production environments and stable setups. This PR provides a "trust" mechanism where known-good plugins can be authorized once and remembered.

Administrators gain a smoother workflow for managing plugins.

- The current implementation matches on both **Name** and **Version**. If a plugin is updated, the admin will be prompted again to ensure they review any new permissions required by the new version.

---

# Example of the new `pumpkin.toml` structure:
```toml
[plugins]
blocked_permissions = []

[[plugins.whitelist]]
name = "Jack O Map"
version = "0.1.0"
```

### DIFF
--- a/pumpkin-config/src/plugins.rs
+++ b/pumpkin-config/src/plugins.rs
@@ -5,4 +5,12 @@ use serde::{Deserialize, Serialize};
 pub struct PluginsConfig {
     /// List of permissions that are globally blocked for all plugins.
     pub blocked_permissions: Vec<String>,
+    /// List of plugins that are whitelisted.
+    pub whitelist: Vec<WhitelistedPlugin>,
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct WhitelistedPlugin {
+    pub name: String,
+    pub version: String,
 }
--- a/pumpkin/src/plugin/mod.rs
+++ b/pumpkin/src/plugin/mod.rs
@@ -468,7 +468,10 @@ impl PluginManager {
 
     /// Ask the server owner if they allow the permissions requested by a plugin
     #[expect(clippy::print_stdout)]
-    fn ask_permission_confirmation(metadata: &PluginMetadata) -> (bool, std::time::Duration) {
+    async fn ask_permission_confirmation(
+        &self,
+        metadata: &PluginMetadata,
+    ) -> (bool, std::time::Duration) {
         use colored::Colorize;
         use rustyline::DefaultEditor;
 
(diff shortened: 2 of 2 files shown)


QuantumSegfault:
When we discussed this a little while back, there was talk of possibly matching on a hash of the Wasm file as well. That way it's not so easy to spoof plugins and gain permissions.

This should probably be made automatic as well. Automatically populate the whitelist upon permissions accept.

Also, perhaps generalize it to mark which plugins are known, whitelisted or blacklisted, and remember the choice, whatever it was. 

------

However there is an increased security hazard now, since we store this as a plain file, and it's possible to grant plugins access to this config file (with the blanket `fs.read` permission), meaning plugins can mess with the permissions of other plugins. So now one plugin could install another, and grant it permissions as long as it has `fs.write` perms.

NicolaiVdS:
> When we discussed this a little while back, there was talk of possibly matching on a hash of the Wasm file as well. That way it's not so easy to spoof plugins and gain permissions.
> 
> This should probably be made automatic as well. Automatically populate the whitelist upon permissions accept.
> 
> Also, perhaps generalize it to mark which plugins are known, whitelisted or blacklisted, and remember the choice, whatever it was.
> 
> However there is an increased security hazard now, since we store this as a plain file, and it's possible to grant plugins access to this config file (with the blanket `fs.read` permission), meaning plugins can mess with the permissions of other plugins. So now one plugin could install another, and grant it permissions as long as it has `fs.write` perms.

I’m in favor of maintaining a list of pre-approved plugins based on their Wasm hash, I think that’s a strong step forward from a security standpoint. My main question is where this list would be sourced from. Would it come from a public API, a curated registry, or be locally maintained? The source of truth here affects both trust and implementation complexity.

Regarding automatically populating the whitelist upon permission acceptance that makes sense to me. If a user explicitly approves a plugin, persisting that decision (along with its hash) would provide both usability and integrity guarantees.

I also agree that generalizing the system to track plugin state (e.g., known, whitelisted, blacklisted) and remembering the user’s choice is a good direction. That would make permission handling more explicit and auditable.

That said, I share the concern about the current security model around `fs.read` and `fs.write`. Granting blanket filesystem access introduces a significant risk particularly because plugins could modify the permissions config itself. In that scenario, a malicious plugin with write access could elevate the permissions of other plugins or install and authorize additional ones, which undermines the trust model.

Instead of exposing raw filesystem access, I think a safer long-term approach would be to introduce a dedicated API layer in the plugin system. For example:
- Provide controlled APIs for accessing core server data (world data, configs, etc.).
- Run plugins in a sandboxed environment.
- Restrict file access to scoped, isolated storage per plugin.
- Prevent direct access to the permissions configuration entirely.

If we can move toward that model, then the risks introduced by this PR would be mitigated significantly, and the overall system would be more robust.

NicolaiVdS:
@QuantumSegfault i have made a post about this topic on Discord: https://discord.com/channels/1268592337445978193/1505160986145984582

NicolaiVdS:
closing this since https://github.com/Pumpkin-MC/Pumpkin/commit/c786bb0dc2c14a1ee52f676ef96e6ab1653105cb has been added

PR #2142, opened 2026-05-15
Grab block after a slimeblock with sticky piston

AurelienFT:
Fixes : https://github.com/Pumpkin-MC/Pumpkin/issues/1981 

## Description

Fix sticky piston chain walking: the block variable was shadowed inside the while loop body so the condition kept checking the seed block (always sticky), causing the chain to walk past non-sticky blocks until hitting `MAX_MOVABLE_BLOCKS`. Now block is reassigned at each iteration so the chain correctly stops at the first non-sticky block, matching vanilla behavior.

## Testing

https://github.com/user-attachments/assets/387837ae-a5eb-44c5-9574-cbf018b44d04


Snowiiii:

PR #2148, opened 2026-05-17
feat: Implement Bedrock NBT (network variant)

QuantumSegfault:
## Description

Adds support to `pumpkin-nbt` for Bedrock's NBT format, specifically the network variety. Main differences from Java are use of little-endian, use of varint/varlong instead of fixed length i32/i64, and the encoding of string (normal UTF-8; var u32 length, rather than fixed u16).

This also ports the one existing use of Bedrock NBT in `pumpkin-codegen` to use `pumpkin-nbt`, instead of rolling it's own encoder. This also revealed and fixed a subtle bug with the existing implementation (treating NBT Int as unsigned vu32 instead of ZigZag encoded vi32), which should improve the block ID mappings for a few blocks.

## Testing

`cargo fmt`, `cargo clippy`, and `cargo test` all pass. This includes the existing tests for Java NBT, as well as new equivalents of each tests made for Bedrock. Bedrock is covered by tests about as well as Java right now. There is room for improvement for test coverage (for both variants), but that's outside the scope of this PR.

Each of the fuzzers have been ran for about a minute each, with no crashes. A bug with one of the fuzzers was fixed (assert failures causing NaN != NaN) in the process (in order to accomplish this result).

Additionally, testing of Bedrock deserialization (raw; via `Nbt`; not serde) has been by reworking the `blocks.rs` generator to read use the new NBT decoder, instead of manual reading.


Snowiiii (approved):

Snowiiii:

PR #2158, opened 2026-05-18
fix: snowy grass

### LINKED ISSUE #2113: Grass block doesn't update texture when removing snow layer.

### Operating System

Macos Tahoe 26.3

45867d3

### What happened?

When in a snowy biome and you remove the snow layer on top of a grass block, it causes the block to not update to be normal grass. This also happens when you place a block over it and remove it; it still does not cause it to update.

### To Reproduce

1. Go to a biome where there is a grass block and snow that has been naturally generated.
2. Remove the snow layer.

### Expected behavior

The texture of the grass block should be updated.

GreenedDev:
## Description
fix: https://github.com/Pumpkin-MC/Pumpkin/issues/2113
image


Snowiiii:

PR #2161, opened 2026-05-19
fix: gravity for couple of projectile entities

GreenedDev:
## Description
wind charge now goes in a straight line. also fixed gravity for couple other projectile entities.


Snowiiii:

PR #2166, opened 2026-05-20
fix chunks not loading when riding a vehicle

Nrleryxx:
fix handle_move_vehicle not triggering chunk loading when player rides a boat/horse. the function was updating the vehicle position but never updating the player entity position or calling chunker::update_position so new chunks never loaded until the player dismounted


Snowiiii:
LGTM. Thank you @RuleGaed :D

PR #2144, opened 2026-05-16
Fix caves generate in the ocean

Giems:
## Description

Fixes [2092](https://github.com/Pumpkin-MC/Pumpkin/issues/2092) by adjusting carver logic. I have not noticed any obvious artifacts while looking around the overworld.

image2

image

PR #2168, opened 2026-05-21
fix: implement ocean carver logic using fluid levels

royvortex:
Makes caves and canyons generate more like Vanilla Minecraft, especially under oceans.

Key Changes:

Added a proper fluid sampler system so carved areas can now correctly become air, water or lava depending on the biome and depth.
Fixed canyons not generating through ocean biomes because of old hardcoded checks.
Removed some old unused parameters (has_grass, is_nether) and cleaned up the carving logic with a more dynamic aquifer based system.


pumpkin-world/src/generation/carver/cave.rs
```diff
@@ -363,36 +337,22 @@ impl CaveCarver {
...
-
-            //     if below_block.id == pumpkin_data::Block::DIRT.id {
-            //         // TODO: Java uses Biome top material here, defaulting to Grass for now
-            //         let top_material =
-            //             BlockState::from_id(pumpkin_data::Block::GRASS_BLOCK.default_state.id);
-            //         chunk.set_block_state(x & 15, local_y - 1, z & 15, top_material);
-            //     }
-            // }
+                if below_block.id == pumpkin_data::Block::DIRT.id {
+                    let top_material =
```
chocodev11: why do you hardcode?

pumpkin-world/src/generation/carver/canyon.rs
```diff
@@ -285,23 +286,12 @@ impl CanyonCarver {
         let state_id = chunk.get_block_state_raw(x & 15, local_y, z & 15);
         let block = pumpkin_data::Block::from_state_id(state_id);
 
-        if block.id == pumpkin_data::Block::WATER.id || block.id == pumpkin_data::Block::LAVA.id {
```
chocodev11, line 288: what? didnt you said using tag? why did you delete it?

pumpkin-world/src/generation/carver/cave.rs
royvortex: fixed

pumpkin-world/src/generation/carver/canyon.rs
royvortex: Vanilla does not have this hardcoded check. It uses Tags.

chocodev11: > Vanilla does not have this hardcoded check. It uses Tags.

ok great, where is the tag?

royvortex: The tag is handled through the CarverConfig::replaceable

royvortex: the tag was already there

Nrleryxx:
A couple of things I noticed during the review:

1. In `get_top_material`, hardcoding biome IDs (14, 15, etc.) might be hard to maintain. It would be cleaner to pull the top material directly from the biome data in `pumpkin_data` if possible.
2. I noticed some index calculation changes in `ProtoChunk::get_biome_id`. Could the `biome_coords::from_block` conversions be causing the index panic in tests? It might be worth double-checking the map bounds there.
3. For the carver logic, are we planning to extend this to full aquifer support (local water tables) in the future? This PR seems like a perfect bridge for that.

Snowiiii:
so @royvortex is this fully working correctly and tested?

royvortex:
no it is not here the get top block was need to be cherry picked
https://github.com/Pumpkin-MC/Pumpkin/pull/2144 & added with the pr 2144

Snowiiii:
> no it is not here the get top block was need to be cherry picked #2144 & added with the pr 2144

So should this be closed or made draft?

royvortex:
Yes

PR #2170, opened 2026-05-21
feat(command): slot argument types

Laptop59:
## Description

Implements the following argument types:
- **`SlotArgumentType`**: Parses only one slot from its name and returns its slot index. Example: `enderchest.7` parses as `207`.
- **`SlotsArgumentType`**: Parses one or more slots from the name of the range containing them and returns a slice of their slot indices. Example: `hotbar.*` parses as `[0, 1, 2, 3, 4, 5, 6, 7, 8]`.

## Testing

Tested both argument types locally, and they work as intended.


Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #2171, opened 2026-05-21
Add armor reduction EPF and weapon enchantments to damage system

Nrleryxx:
Implement missing vanilla 1 21 damage calculations
- Armor reduction with armor toughness formula according to modern vanilla
- Enchantment Protection Factor EPF for Protection Fire Protection Blast Protection Projectile Protection Feather Falling
- Weapon enchantment damage Sharpness Smite Bane of Arthropods
- Thorns enchantment reflection damage
- Fixed last_damage_taken to store final damage instead of raw amount
- Fixed EPF to check bypasses_enchantments instead of bypasses_armor
- Fixed armor formula to use post Resistance amount with floor
- Fixed EPF cap to 20 with divide by 25 for modern formula
- Fixed Sharpness to use 1 0x level for all weapons

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -30,6 +30,7 @@ use crate::entity::{EntityBaseFuture, NbtFuture};
 use crate::server::Server;
 use crate::world::loot::{LootContextParameters, LootTableExt};
 use crossbeam::atomic::AtomicCell;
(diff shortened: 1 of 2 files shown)


Snowiiii:
There has to be a better way than this, way too much hardcoding, please look at how vanilla does it

PR #2172, opened 2026-05-21
feat: Add piston push_entites to match vanilla behaviour 

### LINKED ISSUE #2114: Placing blocks doesnt work when pushing an end crystal with a piston.

### Operating System

macOS Tahoe 26.3

45867d3

### What happened?

May be related to #1981 

When you push an end crystal with a piston and try to place blocks where that end crystal's hitbox was, it causes it to not place the block.

### To Reproduce

1. Place a piston.
2. Place an end crystal in front of the piston and activate the piston.
3. Once the end crystal has been pushed, try placing blocks where the end crystal’s hitbox was.

### Expected behavior

Placing blocks where the end crystal was should work.

AurelienFT:
Fix https://github.com/Pumpkin-MC/Pumpkin/issues/2114

## Description

Implements vanilla's piston entity-push behavior (`PistonBlockEntity.pushEntities`) so entities like end crystals get moved out of the way during piston animation instead of being left at their original server-side position (which caused ghost-block placement rejection).
Each tick of the moving-piston animation, `push_entities` builds the swept AABB of the block's current visual position, finds entities intersecting it via `get_entities_at_box`, and shifts each one by `min(intersection, delta) + 0.01` in the motion direction (calling `set_pos + send_pos` to update both server state and clients). For the piston head (`source=true`) the AABB is clipped to a 0.25-block slab on the facing side (`head_shape_aabb`) to match the real `PistonHeadBlock` collision shape — without this, a full-cube AABB drags entities backward during retraction and they stick to the piston. On retraction `push_out_of_piston_body` also shoves any entity that ended up inside the piston body cube back out, mirroring vanilla's `push` helper.

## Testing

https://github.com/user-attachments/assets/14039156-eb34-4041-9a8c-dd372d13a7b2

### DIFF
--- a/pumpkin/src/block/entities/piston.rs
+++ b/pumpkin/src/block/entities/piston.rs
@@ -1,9 +1,10 @@
+use std::sync::atomic::Ordering;
 use std::{pin::Pin, sync::Arc};
 
 use crossbeam::atomic::AtomicCell;
 use pumpkin_data::{Block, BlockDirection, BlockState};
 use pumpkin_nbt::compound::NbtCompound;
-use pumpkin_util::math::position::BlockPos;
+use pumpkin_util::math::{boundingbox::BoundingBox, position::BlockPos, vector3::Vector3};
 
 use crate::world::{BlockFlags, World};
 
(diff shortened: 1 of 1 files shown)


Snowiiii:

PR #2177, opened 2026-05-22
Fix/world serialization fix wich resulted in void/air chunks

NicolaiVdS:
## Description
Fixes multiple bugs causing chunk data loss (void world after reload) and generation assertion panics.
- **Serialization mismatch**:  removed `#[serde(rename_all = "PascalCase")]` from `ChunkSectionNbtRef` and added inline rename to match `ChunkSectionNbt` so block palettes are correctly deserialized from disk.
 - **Stage initialization**: Set `proto_chunk.stage = StagedChunkEnum::from(chunk_data.status)` in `from_chunk_data`, preventing scheduler from re‑generating already‑completed stages.
- **Generation order**:
  - Moved Biomes before StructureStart in generate_single_chunk stages.
  - Added missing Carvers stage.
 - **step_to_***: added/updated asserts to reflect new order 

## Testing
- Fresh world: placed blocks, stopped server (/stop), restarted -> blocks persisted.
- Teleported to far coordinates (1000, 1000) and back to origin -> chunks reloaded with correct terrain and lighting.
- Verified that caves (carvers) generate in new chunks.
- Confirmed no generation panics (Surface/Noise, Carvers/Surface) occur.


Snowiiii:
LGTM. Thank you @NicolaiVdS :D

PR #2181, opened 2026-05-23
fix(codegen): correct tracked data index lookup for Java protocol versions

NicolaiVdS:
## Problem

The tracked data codegen (`pumpkin-codegen/src/tracked_data.rs`) produces
incorrect `TrackedId` constants for all Java protocol versions, causing entity
metadata packets to be silently dropped by clients.

Root cause: the codegen iterates keys from the Bedrock JSON (`26_1`) as the
canonical source. Bedrock prefixes field names with `DATA_` (e.g. `DATA_ITEM`),
but Java JSONs use the bare name (`ITEM`). The per-version lookup always used
the Bedrock key, which doesn't exist in Java JSONs, so `unwrap_or(255)` fired
for every Java version.

The sentinel value `255` in `Metadata::write` means "skip this entry" so the
`CSetEntityMetadata` packet was sent with an empty payload for every Java client,
making dropped items (and potentially other entities) invisible while still
existing server-side with a working hitbox.

## Changes

- **Fix DATA_ prefix mismatch**: fall back to the stripped name when the
  original key isn't found in a version's data, so Java versions resolve
  correctly.

- **Fix Java-only constants missing**: iterate the union of all version keys
  rather than only the Bedrock latest, so Java-exclusive tracked fields get
  constants generated.

- **Fix wrong fallback version in `TrackedId::get()`**: unknown/future Java
  versions now fall back to the latest Java version instead of the Bedrock
  version, preventing silently wrong indices.

- **Warn on name collisions**: replace the silent `continue` with a logged
  warning so prefix-collision drops are visible during codegen.

## Testing

- Mined blocks now show the dropped item entity visually in-game.
- Re-ran codegen and verified `TrackedData::ITEM` emits `8u8` for all Java
  versions.


NicolaiVdS:
this also makes changes to the WIT https://github.com/Pumpkin-MC/pumpkin-plugin-wit/pull/2

NicolaiVdS:
this make bug #2182 a bit more "stable" and it somewhat playable

PR #2185, opened 2026-05-25
fix(protocol): remap item/block state IDs and entity metadata for older Java clients

NicolaiVdS:
## Problem

The server was using reversed ViaVersion mappings (designed for old→new) to downgrade packets for older clients, which produced ID collisions and incorrect fallbacks for 26.1-exclusive items and block states.

## Changes

- Replace ViaVersion with ViaBackwards mapping files for item ID and block state remapping, providing the correct new→old direction with proper closest-item fallbacks for 26.1-exclusive content
- Fix \`CSetEntityMetadata\` to remap \`ItemStack\` item IDs per client version, matching the version-aware handling already done for inventory packets and block state metadata
- Fix \`TrackedData\` codegen to resolve Java JSON keys correctly by trying both bare names and \`DATA_\`-prefixed names, and iterate the union of all version keys instead of only the latest (Bedrock) version
- Fix \`LootCondition::MatchTool\` to default to \`false\` instead of \`true\`, preventing silk-touch loot table branches from always winning and causing ores to drop themselves instead of their correct items

## Fixes
- Block drops invisible on older Java clients
- Leather showing as milk bucket on 1.21.11
- Coal ore dropping coal ore instead of coal
- 26.1-exclusive blocks showing as wrong blocks on older clients

## Testing
Tested on 1.21.11 and 26.1 clients:
- Block drops now visible and correctly remapped on older clients
- Coal ore drops coal
- Cow drops leather and beef correctly
- 26.1-exclusive items/blocks show as closest equivalent on older clients


chocodev11:
LGTM, but i cant merge :Đ

Q2297045667:
> LGTM, but i cant merge :Đ

Could you help test it?

chocodev11:
> > LGTM, but i cant merge :Đ
> 
> Could you help test it?

work as expect, i would make another pr for nametag and new version's item handler in chat

NicolaiVdS:
> work as expect, i would make another pr for nametag and new version's item handler in chat

ok once this PR is merged il start looking into nametag and new version's item handler in chat

Snowiiii:
LGTM, Thank you @NicolaiVdS :D

PR #2198, opened 2026-05-29
fix(protocol): Fix RakNet frame split threshold.

QuantumSegfault:
## Description

Fixes an issue with RakNet frame splitting, where the packet content did not account for the header overhead of UDP + RakNet. Packet content was split into chunks of MTU, and in each frame, content length declared as MTU. However for clients that strictly adhere to the MTU, this would cause errors as there wouldn't be enough space in one UDP datagram to fit the whole frame. Strict clients will read MTU (1400) bytes, read the header, then try to read 1400 additional bytes after that from the datagram (but after some 52 bytes, there are only 1348 bytes left in the datagram, less than declared by the frame).

## Testing

`cargo fmt`, `cargo clippy` and `cargo test` all pass.

This problem was discovered with, and tested against, a bot client created with `gophertunnel`. Large split packets are now transmitted properly without `gophertunnel` silently erroring out and dropping N-1 frames of any packet split into N frames (and so completely missing large enough packets).


QuantumSegfault:
Ping

@Snowiiii

Snowiiii:

PR #2199, opened 2026-05-29
fix: item pickup animation

GreenedDev:
## Description
Item Pickup packet was only being sent to the collector player which meant that only he was getting the pickup animation. in POV of other players standing beside him the picked up item just disappeared.


Q2297045667:
I'll test this branch today.

Q2297045667:
LGTM，The animation is now displaying correctly.
I think it's ready to merge
@Snowiiii

PR #2202, opened 2026-05-30
fix(math): prevent NaN from zero-length normalization

Tunar-ssp:
## Description

`Vector2::normalize()` and `Vector3::normalize()` could return non-finite values (`NaN`/`inf`) when called on zero-length vectors.

## Why

A non-finite vector can silently propagate into movement, knockback, rotation, and other calculations, making debugging difficult.

## Impact

- No behavior change for normal vectors
- Zero / near-zero vectors now return a zero vector instead of non-finite values

### DIFF
--- a/pumpkin-util/src/math/vector2.rs
+++ b/pumpkin-util/src/math/vector2.rs
@@ -88,10 +88,17 @@ impl<T: Math + Copy + Float> Vector2<T> {
         self.length_squared().sqrt()
     }
 
-    /// Returns a normalized version of the vector with length 1.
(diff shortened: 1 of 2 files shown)


Snowiiii:
I really don't like the unwrap here, i also think we instead should just check for `is_finite` similar to glam: https://github.com/bitshifter/glam-rs/blob/29b3b2863960aacce2d1d515fbf2d38c98c3dd46/src/f32/vec2.rs#L536

Tunar-ssp:
Like this? I updated the implementation.

Snowiiii:
LGTM. Thank you @Tunar-ssp :D

PR #2149, opened 2026-05-17
fix(command): make commands with slash not crash

Laptop59:
## Description

Fixes #2147 

Should fix the crash described in the above issue, where just typing a command with a slash in the beginning would crash the entire server due to the index being out of bounds.

## Testing

Tested a bunch of commands (including `/banlist`, `/difficulty`, `/list`, `/setidletimeout`, and `/tick`), with and without slashes, and the server does not crash.

PR #2207, opened 2026-05-31
fix(command): prevent SuggestionsBuilder panic on out-of-bounds start

Marshall-67:
### Summary

During my first run of pumpkin on MacOS i ran into a panic when trying to figure out the right command to op myself, when entering a sequences of text and pressing tab it caused the server to panic.

### Description/Changes 

Clamp the start position to the input length in `remaining()` and 
`remaining_lowercase()` at suggestions.rs to prevent an unbounded slice 
panic (`&self.input[self.start..]`) when `start` exceeds `input.len()`.

Also, use `saturating_sub(1)` in the entity_selector parser to 
prevent a panic in debug mode when `start` is 0.

Fixes panic from:
- Console completer passing cursor positions without adjusting for a stripped '/'
- `create_offset()` constructing a builder with `start > input.len()`
- `entity_selector` parser underflowing on `builder.start - 1`

### DIFF
--- a/pumpkin/src/command/argument_types/entity_selector/parser.rs
+++ b/pumpkin/src/command/argument_types/entity_selector/parser.rs
@@ -475,7 +475,7 @@ impl EntitySelectorParserSuggestions {
         match self {
             Self::Nothing => {}
             Self::Selector => {
-                let mut sub = builder.create_offset(builder.start - 1);
+                let mut sub = builder.create_offset(builder.start.saturating_sub(1));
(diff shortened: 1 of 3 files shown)


gloryrunner:
#2149 is currently open which should fix this issue.

Marshall-67:
> #2149 is currently open which should fix this issue.

this patch covers more areas and actually prevents a panic, but they don't conflict so both should be merged if they look good

Snowiiii:
LGTM. Thank you @Marshall-67 :D

PR #2214, opened 2026-06-02
Fix stale proto chunk generation tasks

shaches:
## Summary
- Drop all satisfied proto chunk generation task nodes when a returned proto chunk is already at an advanced stage
- Remove those completed task nodes from the waiting set to prevent stale queued generation work
- Add a regression test for stale lower-stage task cleanup

## Testing
- cargo fmt --check
- cargo test -p pumpkin-world chunk_system::tests
- cargo build -p pumpkin


Snowiiii:
LGTM. Thank you @shaches :D

PR #2215, opened 2026-06-02
feat(bedrock): Initial inventory setup

QuantumSegfault:
## Description

Implements `CItemRegistry` and `CInventoryContent` packets. Sends them both on player spawn.

A large portion of this PR is the initial work to remap Java `ItemStack`s to Bedrock ones. Part of this is also introducing `uid` to the server `ItemStack` to keep track of stack identity as necessary for Bedrock server-authoritative inventory.

## Testing

Testing was done to make sure basics work with Bedrock client. Player can join, and after running a few `give` commands (particular ones like `red_bed` which isn't a 1 : 1 mapping to Java; gets collpased into `bed` + data), player can leave and rejoin, and expected inventory contents will be visible in the intended order.

At the surface level, everything works as intended at this somewhat early stage.


assets/bedrock/README.md
```diff
@@ -0,0 +1,15 @@
+# Bedrock Assets
+
+This directory contains a number of different data files used to help support connecting via Bedrock Edition clients (including Java => Bedrock remapping).
+
+- `block_states.nbt`
+    - mined from BDS (taken from where? `BetterAltay`? `dragonfly`?)
```
QuantumSegfault: Where do we source this? I don't remember

Snowiiii: https://github.com/pmmp/BedrockData/blob/master/canonical_block_states.nbt

Snowiiii:

Issue #2225, opened 2026-06-04
Surface slime spawning


Barni-Mezei:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Server: Alpine Linux v3.23 (aarch64) running on a Raspberry Pi 4B

### Server Software Version/Commit

a2f0122/release

### What happened?

Small slimes are spawning on the surface, and they move wierdly

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/a528bbff-2187-4bf1-8e12-f512686eddb2" />

### To Reproduce

Start a world with seed: 1779920288596261407
Teleport to: /execute in minecraft:overworld run tp @s 732.80 90.87 635.23 -1594.20 36.60

### Expected behavior

I beleive slimes should spawn only certain light levels

Snowiiii:
Should be fixed

PR #2227, opened 2026-06-04
 fix(natural-spawner): gate slime and magma cube spawns behind vanilla rules

CompileRider:
## Summary
The `check_spawn_rules` function in `pumpkin/src/entity/type.rs` had no input for the slime or the magma cube. Both were included in the general `true` return condition, so the natural spawn loop accepted them anywhere. This is the surface spawning bug from bug #2225: slimes were spawning above Y=40 and outside of any swamp, where the base game would never place them.

## Changes
Mojang 1.21+ spawn rules have been adapted:
- **SlimeEntity::check_spawn_rules**: Ignored in peaceful environments or outside the main world, and the position is accepted if it is a slime fragment below Y=40 (with the correct base game polynomial + Java LCG), or a swamp/mangrove biome between Y=51 and 69 with the light check `getMaxLocalRawBrightness(pos) <= nextInt(8)`.

- **MagmaCubeEntity::check_spawn_rules** — Nether only, excluding peaceful mode.

- **Slime Fragment Algorithm** — Exactly mirrors the original: 32-bit integer arithmetic for polynomial terms, XOR with the decoration seed `0x3ad8025f`, Java LCG initialization + `nextInt(10)`. The previous implementation omitted the XOR and LCG, producing slime fragments different from the original for the same seed.

- **Swamp Biome Check** — Compares by biome ID (not by registry_id string) to avoid namespace prefix discrepancies. `pumpkin-data` stores "swamp" without the `minecraft:` prefix.

- **Sky Darkening** — `get_max_local_raw_brightness` now subtracts the cached value of `sky_darken` (0..11) from the sky brightness, matching the original behavior for nighttime spawn checks. Previously, daylight was used directly (always 15 on the surface), which prevented the formation of swamp mud at night.

- **Proofs** — 5 unit proofs covering the polynomial, determinism, and distribution of the mud fragments. The fragment Seed=0 (0,0) is correctly identified as NOT a mud fragment according to the original version.

Closes problem #2225.


Atlas16A:
commit description says 1.18 rules but code comments say 1.21?

CompileRider:
> commit description says 1.18 rules but code comments say 1.21?

You're right, that's inconsistent. The "1.21+" came from the code comments first because the check itself is pulled from Mojang's current Slime.java (1.21+ mappings). The "1.18+" in the description was my own leftover from when the surface-spawn rulewas first introduced in 1.18 (Caves & Cliffs). PR description just got updated to match the code.

Atlas16A:
Where the spawn rules changed since 1.21? When I consulted the mc wiki some details about the rules in the description didn't match the wiki, light level for instance for below y40 in slime chunks doesn't matter at all.

CompileRider:
> Where the spawn rules changed since 1.21? When I consulted the mc wiki some details about the rules in the description didn't match the wiki, light level for instance for below y40 in slime chunks doesn't matter at all.

You're right — light level doesn't matter for slime chunks below Y=40. The code is correct on this point (light check is only in the swamp branch), but the PR description is misleading by listing "require sky light at most 7" before the branch split. I'll update the description.

Separately, while verifying against the wiki's Java reference code, I found that is_slime_chunk is missing the ^ 0x3ad8025fL decoration-seed XOR and uses direct % 10 instead of Java's LCG nextInt(10). The ~10% distribution holds (test passes) but specific chunk assignments will differ from vanilla. I'll push a fix for both.

CompileRider:
I encountered more issues while exploring the swamp spawn path.

`is_swamp_biome` is being compared to `minecraft:swamp`, but the `registry_id` in `pumpkin-data` is simply `swamp` without the namespace prefix. The same applies to `mangrove_swamp`. Therefore, this check always returns `false`, and the swamp spawn path is never executed. It's an easy fix: simply compare by ID instead of by string.

Additionally, `get_max_local_raw_brightness` has an outstanding issue regarding sky darkening. It currently reads the raw sky light, which is always 15 on the surface. Vanilla subtracts approximately 11 at night to ensure the light check is correct. Without this, swamp slimes cannot spawn at night, even after correcting the biome check.

I will upload both fixes along with the polynomial changes. The lunar phase can be a continuation, as it only requires a small addition to `LevelTime`.

@Atlas16A  Thanks for bringing this up; your question about the discrepancy in the wiki led me to find all this.

Snowiiii:
AI slop i would say, also i recently implemented the slime chunk check

PR #2234, opened 2026-06-05
feat: adding codegen fro advancements

aure31:
## Description
contains only the advancements codegen of the closed pull request
## Testing
generate the code as intended and pass clippy
Please follow our [Coding Guidelines](https://github.com/Pumpkin-MC/Pumpkin/blob/master/CONTRIBUTING.md#coding-guidelines)

PR #2235, opened 2026-06-05
feat: player advancement

aure31:
## Description
adding the player advancement so the advancement are player dependent and progress can be save and load between session
## Testing
implementing some bear-bone test for saving loading and of the different functions


Q2297045667:
This is too big; it's better to split it into 3 PRs.

aure31:
> This is too big; it's better to split it into 3 PRs.

fo the other one i see how but here it could be a little bit complicate to split in 2 but 3 i just don't see how i can do it. could you maybe help me on how I could cut it into 3 prs ? (this one is much smaller then the first pull request)

Laptop59:
> > This is too big; it's better to split it into 3 PRs.
> 
> fo the other one i see how but here it could be a little bit complicate to split in 2 but 3 i just don't see how i can do it. could you maybe help me on how I could cut it into 3 prs ? (this one is much smaller then the first pull request)

I think the problem is that this PR includes the commits from the previous PR #2234 you opened for the first part. So it's a way bigger PR than it should really be.

aure31:
> > > This is too big; it's better to split it into 3 PRs.
> > 
> > 
> > fo the other one i see how but here it could be a little bit complicate to split in 2 but 3 i just don't see how i can do it. could you maybe help me on how I could cut it into 3 prs ? (this one is much smaller then the first pull request)
> 
> I think the problem is that this PR includes the commits from the previous PR you opened for the first part. So it's a way bigger PR than it should really be.

Yes, but how could I split it if it depends on the other one? like they are all from the same feature globally

aure31:
maybe i didn't understand when you were saying it to be split. did I misunderstand something ?

Laptop59:
> > > > This is too big; it's better to split it into 3 PRs.
> > > 
> > > 
> > > fo the other one i see how but here it could be a little bit complicate to split in 2 but 3 i just don't see how i can do it. could you maybe help me on how I could cut it into 3 prs ? (this one is much smaller then the first pull request)
> > 
> > 
> > I think the problem is that this PR includes the commits from the previous PR you opened for the first part. So it's a way bigger PR than it should really be.
> 
> Yes, but how could I split it if it depends on the other one?



> maybe i didn't understand when you were saying it to be split. did I misunderstand something ?

Since each PR is dependent on the previous PR (if not, you can just open them in parallel), usually the previous PR has to be merged first before the current PR can be merged. So you need to wait for that PR to be merged first. Might be better to open one at a time if each one depends on the previous (at least here) so that confusion doesn't arise

aure31:
so it's means that I'm forced to wait for the code to be merged every little feature i code ? so should I close this pr ?

Laptop59:
> so it's means that I'm forced to wait for the code to be merged every little feature i code ? so i should close this pr ?

Unfortunately, yeah I think you have to, so you can close this for now, but I think you should be able to re-open it later

aure31:
> > so it's means that I'm forced to wait for the code to be merged every little feature i code ? so i should close this pr ?
> 
> Unfortunately, yeah I think you have to, so you can close this for now, but I think you should be able to re-open it later

thanks for the advice 👍

PR #2236, opened 2026-06-05
feat: implementing the base generation of advancement

aure31:
## Description
generate the advancement code to be use afterward
## Testing
cargo clippy and generate the code as intended


aure31:
I removed a big chunks of the code is this fine this times or should i reduce it even more ? @Q2297045667

aure31:
but i still don't know why the checks doesn't pass here

Q2297045667:
> but i still don't know why the checks doesn't pass here

[dependencies]
proc_macro2 = "1.0"
quote = "1.0"

Q2297045667:
> I removed a big chunks of the code is this fine this times or should i reduce it even more ? @Q2297045667

I think it looks good.

aure31:
> proc_macro2 = "1.0"
> quote = "1.0"

I mean yes but like isn't this lines already do it ? and other module uses them in the crate and they don't have any issue

image

Laptop59:
> > proc_macro2 = "1.0"
> > quote = "1.0"
> 
> I mean yes but like isn't this lines already do it ? and other module uses them in the crate and they don't have any issue
> image

Honestly, I really have no idea, it should work... but it looks like your branch is behind upstream master. Maybe try updating it to the latest commit of upstream?

aure31:
updating doesn't seams to fix it. I really don't understand what's going on.

Laptop59:
> updating doesn't seams to fix it. I really don't understand what's going on.

Maybe you can try moving that `ToTokens` somewhere in the codegen crate? Or just map the identifiers manually and maybe change this
```
minecraft_namespaces.extend(quote! { #identifier,})
```
to
```
minecraft_namespaces.extend(identifier.map(|identifier| quote! {
    // Do the mapping here?
}))
```

Laptop59:
I know this probably isn't the best approach but I can't think of anything else (and don't forget the comma)

aure31:
> I know this probably isn't the best approach but I can't think of anything else (and don't forget the comma)

Ok I will try that. but what is really strange is that locally i doesn't have the error.

Laptop59:
> > I know this probably isn't the best approach but I can't think of anything else (and don't forget the comma)
> 
> Ok I will try that. but what is really strange is that locally i doesn't have the error.

Yeah, I know, i even tried your branch locally but no errors. So I'm guessing that this is some kind of cached thing in CI

aure31:
and I cannot move it inside the codegen crate because it's not the same crate and in rust you cannot implement a type that hasn't been define in the crate. or I just do the mapping manually but it's not as clean as directly implementing the ToToken

aure31:
Could I just let the CI error ?

Laptop59:
> Could I just let the CI error ?

Problem with that is that the upstream repo also has CI checks after a PR is merged

Laptop59:
> and I cannot move it inside the codegen crate because it's not the same crate and in rust you cannot implement a type that hasn't been define in the crate. or I just do the mapping manually but it's not as clean as directly implementing the ToToken

It's not a clean solution, but not much code is really required to map the `Identifier`

Laptop59:
> and I cannot move it inside the codegen crate because it's not the same crate and in rust you cannot implement a type that hasn't been define in the crate. or I just do the mapping manually but it's not as clean as directly implementing the ToToken

Oh and it's also possible to create a wrapper type around the foreign type and then make an impl on it

aure31:
> > and I cannot move it inside the codegen crate because it's not the same crate and in rust you cannot implement a type that hasn't been define in the crate. or I just do the mapping manually but it's not as clean as directly implementing the ToToken
> 
> Oh and it's also possible to create a wrapper type around the foreign type and then make an impl on it

okay i will do that

Snowiiii:
LGTM. Thank you @aure31 :D

PR #2237, opened 2026-06-05
feat(command): implement column pos argument type

Laptop59:
## Description

Implements the **`ColumnPosArgumentType`** which can parse the newly added `ColumnPos` struct. It represents two-dimensional integral coordinates in the X and Z axes.

## Testing

Tested the argument type locally and it works as intended.


Snowiiii:
LGTM. Thank you @Laptop59 :D

PR #2238, opened 2026-06-05
fix(player): sync zero-health metadata to fix death animation (#2205)

### LINKED ISSUE #2205: Player death animation is broken

### Operating System

CachyOS

Pumpkin 0.1.0-dev+26.1 (Commit: fa5e07f/debug) - 221 Contributors

### What happened?

Player death animation isn't played.
Player entity still exists when player is in respawn screen.

https://github.com/user-attachments/assets/c144e159-d236-4e53-9bf4-460021c458c1

### To Reproduce

kill a player

### Expected behavior

player should die correctly

isaac2343:
### Description
This pull request resolves an issue where the vanilla Minecraft client fails to play the death animation sequence for player entities, leaving them standing frozen upright while the respawn screen overlay is visible.

### Root Cause
The vanilla Minecraft client skips or ignores clientbound entity status death packets (status `3`) if the server's tracked entity metadata properties still report the player's health as greater than zero. Because the player's health was not explicitly zeroed out in memory at the exact moment of the atomic death transition, the client dropped the animation sequence.

### Solution
* **Zero Health State Sync**: Explicitly clear the internal entity health pool state (`self.health.store(0.0)`) immediately upon processing the atomic death state swap sequence inside `on_death`. This forces the entity metadata tracking loop to broadcast an accurate zero-health state to the client, allowing the client-side rendering pipeline to successfully execute the falling-over animation.

### Testing & Verification
- [x] Code passes `cargo check` and `cargo clippy` without warnings.
- [x] Verified locally using third-person front view (F5) that player entities flash red and tilt 90 degrees onto the ground upon taking fatal damage.
- [x] Validated that the respawn overlay screen displays concurrently without locking or freezing the client UI state.

Closes #2205


Snowiiii:
Fails CI, feel free to reopen

PR #2239, opened 2026-06-05
chore: TakeItem packet tweaks.

GreenedDev:
Implement TakeItemActor for bedrock.
Only send TakeItem packet for tracking players.

source for TakeItemActor packet:
https://github.com/Mojang/bedrock-protocol-docs/blob/main/docs/TakeItemActorPacket.html

I haven't tested if the packet works for bedrock. I don't think it's possible at the time


Snowiiii:

PR #2241, opened 2026-06-06
feat: Only send some packets to players who are tracking the chunk

GreenedDev:
## Description
This PR includes packet broadcast optimizations.
it also replaces `player.living_entity.entity` with `player.get_entity()`


Snowiiii:

PR #2244, opened 2026-06-07
Init submodules from build.rs

vyPal:
## Description
Since the `pumpkin-plugin-wit` was moved to a git sub-module, people keep running into the same issue of their local copy not building since the wit folder is missing. The fix is easy, but cannot be easily determined from the error message unless the person already knows a bit about submodules and rust errors.

This PR calls the `git submodule update --init` command from the build.rs file of the main pumpkin crate, as well as adds a new build.rs file to the `pumpkin-plugin-api` crate, as it is outside the main dependency tree (doesn't depend on the `pumpkin` crate) and can therefore be built before the build.rs of the `pumpkin` crate runs.

## Testing

### DIFF
--- a/pumpkin-plugin-api/build.rs
+++ b/pumpkin-plugin-api/build.rs
@@ -1,0 +1,13 @@
+use std::process::Command;
+
+fn main() {
+    // Sync submodules
+    Command::new("git")
+        .args(["submodule", "update", "--init"])
+        .current_dir("..")
+        .output()
+        .expect("Failed to update submodules");
+
+    println!("cargo::rerun-if-changed=../.git/HEAD");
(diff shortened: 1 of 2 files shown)


Snowiiii:
This kinda seems not too complicated to the only purpose to eliminate confusion, I think it might be better to change the `git clone <repo>` command to include `--recurse-submodules` in the docs?

vyPal:
Yeah I guess that would be an option as well. Although it seems most semi-technical users don't seem to read the docs though, and just clone the repo normally

Snowiiii:
Also this makes adding stuff in wit locally a pain, i often have wit changes in Pumpkin which are not in the wit repo for testing, so i think this does not make much sense debugging wise

PR #2245, opened 2026-06-07
chore: code tweaks

GreenedDev:
## Description
To improve code readability

introduce is_sneaking() is_sprinting() and is_fall_flying() functions so that we don't have to write "".load(Ordering::Relaxed) every time.

remove duplicate shooter ids in projectile

load pitch and yaw using the designed function (.rotation())


Snowiiii:
lgtm, tanks  @GreenedDev :D

PR #2246, opened 2026-06-07
fix: allow snow layers to be placed on top of leaves

### LINKED ISSUE #2218: Snow Layer can't be place on leaves

### Operating System

macOS Thaoe

0.1.0/0fd6fef

### What happened?

when i try to place snow layer on leaves they are prevent by the server to be place and when i update the ones of the generation they are destroy

### To Reproduce

take snow layer and try to place it on leaves

### Expected behavior

should be placeable on leaves

perplexinglysimple:
## Description

Snow layer placement checked `is_side_solid` (`isSideSolidFullSquare` / support shape), which leaves override to be empty — so snow could never be placed on top of leaves.

Vanilla's `SnowBlock#canPlaceAt` instead:
1. Rejects blocks tagged `minecraft:cannot_support_snow_layer` (ice, packed ice, barrier).
2. Allows blocks tagged `minecraft:support_override_snow_layer` (honey block, soul sand, mud) regardless of their collision shape.
3. Otherwise checks whether the block's collision shape fully covers the top face (`Block.isFaceFullSquare`). Leaves have a full-cube collision shape (only their support/side-solid shape is empty), so they pass this check and can support snow layers.
4. Also allows stacking on top of a fully-layered (8-layer) snow block.

This PR ports that logic to `LayeredSnowBlock::can_place_at`.

Fixes #2218

## Testing

Manually verified in-game against a local server:
- Snow layers can now be placed on oak/spruce/azalea leaves and stack up to 8 layers.
- Snow still cannot be placed on ice, packed ice, or barrier (full collision shape, but tagged `cannot_support_snow_layer`).
- Snow can be placed on honey block, soul sand, and mud (tagged `support_override_snow_layer`, despite not having a full top-face collision shape).
- Snow still cannot be placed on glass, fences, slabs, stairs, carpets, torches, etc. (no full top-face collision, not in the override tag).
- Snow can be placed on top of a fully-layered (8-layer) snow block.

Also ran the full local CI suite:
- `cargo fmt --check`
- `cargo clippy --all-targets --all-features` (debug & release, `-Dwarnings`)
- `cargo test --workspace`

### DIFF
--- a/pumpkin/src/block/blocks/snow.rs
+++ b/pumpkin/src/block/blocks/snow.rs
@@ -1,5 +1,6 @@
 use pumpkin_data::block_properties::BlockProperties;
-use pumpkin_data::{Block, BlockDirection, block_properties::SnowLikeProperties, item::Item};
+use pumpkin_data::tag::Taggable;
+use pumpkin_data::{Block, block_properties::SnowLikeProperties, item::Item, tag};
 use pumpkin_macros::pumpkin_block;
 use pumpkin_util::math::position::BlockPos;
 use pumpkin_world::{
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM. Thank you @perplexinglysimple :D

PR #2247, opened 2026-06-07
fix: melt snow layers near light sources

### LINKED ISSUE #2179: Torches dont melt snow layer

### Operating System

Server: Debian 13 Client macOS Tahoe 26.5

4aa7849

### What happened?

When you place a torch next to a snow layer it should slowly melt the snow layer around it.

EDIT: Also applies to Campfires, Lanterns, Fire, Lava, & Glowstone. ETC
All Sources can be found here: https://minecraft.wiki/w/Snow#Melting

### To Reproduce

1. Place snow layers down.
2. Place torch near snow layers.

### Expected behavior

The snow layers should melt.

perplexinglysimple:
## Description

Snow layers (`minecraft:snow`) had no `random_tick` behavior, so they persisted indefinitely even when placed right next to bright light sources. In vanilla, `SnowLayerBlock#randomTick` removes the block once the surrounding **block light** level exceeds 11:

```java
if (level.getBrightness(LightLayer.BLOCK, pos) > 11) {
    dropResources(state, level, pos);
    level.removeBlock(pos, false);
}
```

This PR ports that behavior to `LayeredSnowBlock`:
- Added a `random_tick` implementation that checks `world.get_block_light_level` (block light only — sunlight/sky light does **not** melt snow in vanilla, which is why snow persists fine in sunny snowy biomes).
- Extracted the threshold check into a small pure helper `melts_at_light_level` for testability, with unit tests covering the boundary (11 = no melt, 12 = melt).
- Removal mirrors the existing "unsupported snow" cleanup already in `on_scheduled_tick` (`break_block` with no item context — matches vanilla, which doesn't drop snowballs from melting since the loot table requires a shovel).

Light sources that trigger melting include torches, campfires, lanterns, fire, lava, and glowstone — see https://minecraft.wiki/w/Snow#Melting.

Notes on related blocks:
- A layered snow stack melts away **all at once** regardless of its layer count (1–8) — vanilla removes the whole block, not layer by layer.
- The solid **Snow Block** (`minecraft:snow_block`) is a separate block class with no `randomTick` override, so it does not melt — this PR does not touch it.

Fixes #2179

## Testing

- `cargo check -p pumpkin`
- `cargo test -p pumpkin snow::` — new unit tests `does_not_melt_at_or_below_threshold` and `melts_above_threshold` pass
- Manually verified in-game: a stack of snow layers placed near a torch/lava disappears completely after a short time, while a solid Snow Block in the same spot remains unaffected.

### DIFF
--- a/pumpkin/src/block/blocks/snow.rs
+++ b/pumpkin/src/block/blocks/snow.rs
@@ -119,7 +119,7 @@ fn can_place_at(block_accessor: &dyn BlockAccessor, position: &BlockPos) -> bool
 /// Snow melts away when exposed to a block light level greater than 11, e.g. from
 /// torches, campfires, lanterns, fire, lava, or glowstone.
 /// See <https://minecraft.wiki/w/Snow#Melting>.
-fn melts_at_light_level(block_light_level: u8) -> bool {
+const fn melts_at_light_level(block_light_level: u8) -> bool {
     block_light_level > 11
 }
 
--- a/pumpkin/src/block/blocks/snow.rs
+++ b/pumpkin/src/block/blocks/snow.rs
@@ -10,7 +10,7 @@ use pumpkin_world::{
 
 use crate::block::{
     BlockBehaviour, BlockFuture, GetStateForNeighborUpdateArgs, OnPlaceArgs, OnScheduledTickArgs,
-    UseWithItemArgs, registry::BlockActionResult,
+    RandomTickArgs, UseWithItemArgs, registry::BlockActionResult,
 };
 
 #[pumpkin_block("minecraft:snow")]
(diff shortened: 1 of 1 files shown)


pumpkin/src/block/blocks/snow.rs
```diff
@@ -104,3 +115,29 @@ fn can_place_at(block_accessor: &dyn BlockAccessor, position: &BlockPos) -> bool
     let state = block_accessor.get_block_state(&position.down());
     state.is_side_solid(BlockDirection::Up)
 }
+
+/// Snow melts away when exposed to a block light level greater than 11, e.g. from
+/// torches, campfires, lanterns, fire, lava, or glowstone.
+/// See <https://minecraft.wiki/w/Snow#Melting>.
+const fn melts_at_light_level(block_light_level: u8) -> bool {
```
Snowiiii, line 122: I think it would be better to have that as a constant above instead of this function

pumpkin/src/block/blocks/snow.rs
```diff
@@ -104,3 +115,29 @@ fn can_place_at(block_accessor: &dyn BlockAccessor, position: &BlockPos) -> bool
...
+    block_light_level > 11
+}
+
+#[cfg(test)]
+mod tests {
```
Snowiiii, line 127: Also these tests are kind nonsense

PR #2248, opened 2026-06-07
Remove unnecessary shell specification in rust.yml

isaac2343:
## Description

This PR fixes Windows runner CI failures in the GitHub Actions workflow by removing problematic explicit `shell: bash` declarations that were causing bash startup failures on Windows ARM runners.

## Changes

- Removed `shell: bash` from `build_release` job step (line 129)
- Removed `shell: bash` from `prepare_artifacts` job step (line 136)
- Fixed YAML formatting to ensure multiline `run:` directives are properly indented

The workflow was failing on Windows runners (especially ARM) with: `"install-action: installation failed due to bash startup failure"`. This is a known issue with GitHub Actions partner runner images. By removing the explicit `shell: bash` declaration, the runner defaults to PowerShell on Windows (which handles the scripts properly) and bash on Linux/macOS.

## Impact of this change

- ✅ Fixes CI failures on Windows runners (windows-latest, windows-11-arm)
- ✅ Maintains existing functionality on Linux and macOS runners
- ✅ Simplifies the workflow by letting each OS use its native shell
- ✅ Resolves YAML parsing errors in the workflow file

## Related issues

Fixes CI failures mentioned in #2238

## Notes

The shell changes are compatible with existing scripts since Rust's `cargo` commands work the same across PowerShell, bash, and other shells.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -126,14 +126,12 @@ jobs:
       - run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
       - uses: Swatinem/rust-cache@v2
       - name: Build Release
-        shell: bash
         run: |
           if [[ "${{ runner.arch }}" == "X64" ]]; then
             export RUSTFLAGS="$RUSTFLAGS -C target-cpu=x86-64-v3"
           fi
           cargo build --verbose --release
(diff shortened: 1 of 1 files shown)


Snowiiii:
Yeah they don't seem unnecessary, CI is failing

PR #2254, opened 2026-06-11
fix: prevent spectators from opening doors, trapdoors and fence gates

### LINKED ISSUE #2253: You can open doors in spectator

### Operating System

CachyOS

Pumpkin 0.1.0-dev+26.1 (Commit: 5b8eab0/release) - 225 Contributors

### What happened?

You can open fence gates, trapdoors, and doors in spectator

### To Reproduce

go in spectator and try to open them

### Expected behavior

you shouldn't be able to

faustyu1:
## Summary

Spectators were able to open/close doors, trapdoors and fence gates by right-clicking them, which should not be possible.

This adds a `GameMode::Spectator` check at the start of each block's `normal_use` handler, returning `BlockActionResult::Pass` for spectators. This matches vanilla behaviour and the existing pattern already used by other interactable blocks (e.g. `candle_cakes.rs`, `chests.rs`).

Affected blocks:
- Doors (`doors.rs`)
- Trapdoors (`trapdoor.rs`)
- Fence gates (`fence_gates.rs`)

## Test plan

- [x] `cargo check -p pumpkin` passes
- [x] `cargo fmt` applied
- [ ] In-game: enter spectator mode, attempt to open doors/trapdoors/fence gates — they should no longer toggle

Fixes #2253


faustyu1:
the failing CI checks are not related to my changes

chocodev11:
not vanilla align

faustyu1:
> not vanilla align

why?

faustyu1:
> not vanilla align

this matches vanilla behaviour and the existing pattern already used by other interactable blocks

chocodev11:
> > not vanilla align
> 
> this matches vanilla behaviour and the existing pattern already used by other interactable blocks

of cource it matches the behaviour, but not vanilla structure. as you can see here https://mcsrc.chocodev.workers.dev/1/26.2-rc-2/net/minecraft/server/level/ServerPlayerGameMode , it blocks the click from the root which should be in pumpkin\src\net\java\play.rs and some work in pumpkin\src\block\mod.rs. and what is the parttern you mentioned?

chocodev11:
do you mean the chest? we should change it too

faustyu1:
> do you mean the chest? we should change it too

yes

faustyu1:
Moved the spectator check up into `handle_use_item_on` instead of doing it per-block, like vanilla does in `ServerPlayerGameMode#useItemOn`. A spectator can only open a block that has a menu provider (chests, furnaces, anvils, etc), everything else just passes.

Did it with a `has_menu_provider` method on `BlockBehaviour` (defaults to false), overridden on the blocks that actually open a screen, plus a small helper on `BlockRegistry`. The root gate covers doors/trapdoors/fence gates and also buttons, levers and the rest.

chocodev11:
with some block, normal_use is more than opening an ui so you need to seperate it with somth like open_menu_provider. also, some block is only able to open ui sometime like chest so fn has_menu_provider(&self) -> bool is not enough

faustyu1:
Split the menu opening out of `normal_use` into a dedicated `open_menu` method on `BlockBehaviour`, so blocks whose `normal_use` does more than open a screen aren't dragged into the spectator path. The spectator gate now calls `open_menu` directly and `normal_use` just delegates to it.

Also dropped the static `has_menu_provider() -> bool` for `open_menu(args) -> BlockActionResult` (defaults to `Pass` = no provider), so it's world/player-aware: a chest still decides at call time whether it can actually open (obstructed, pending loot for a spectator, etc.) instead of being a fixed yes/no.

Snowiiii:
Actually fixed that already recently

faustyu1:
bruh

PR #2258, opened 2026-06-12
Batch blocks update implementation

justxraf:
## Description
This replaces the old per-block update path with a batch block update API (World::set_block_staes, I have it ready, but I am not sure about the structure of this project so please let me know if I should send new code snippets to you on discord or in comments here), designed for large-scale edits like a WorldEdit-style plugin pasting millions of blocks at once. - Important caveat to note here is that actual millions of blocks cannot currently go from a plugin to the engine as the file is too large, so anyone using this API has to make smaller calls to the engine - It is still incredibly fast.

**Problems with current implementation:**
Current implementation allows individual chunk updates or block updates, with re-light for each block as far as I am aware. Client-visible updates are flushed a few sections at a time per tick, so a large paste trickles visually over many ticks.

**What this PR is solving with bulk API:**

- `World::set_block_states(changes, flags)` takes a `Vec<(BlockPos, BlockStateId)>`, dedupes, validates state IDs, and writes all blocks first before running any side effects.
- Chunk writes are parallelized across a bounded task pool (`MAX_BULK_BLOCK_WRITE_TASKS`, scaled to CPU cores) — each chunk only locks its own data, so distinct chunks don't contend.
- Per-block side effects (neighbor updates, block-entity callbacks, redstone, etc.) run on a separate, smaller bounded pool (`MAX_BULK_BLOCK_SIDE_EFFECT_TASKS`), yielding periodically
- Relighting for the whole batch runs once, off the runtime via `update_lighting_bulk`, instead of once per block.
- New chunk write primitve (`pumpkin-world/src/chunk/mod.rs) - `ChunkData::set_blocks_absolute_y` write many blocks while holding the section/heightmap/random-tick locks once, instead of re-acquiring them per block.
**lighting**
- `update_lighting_bulk` seeds light checks for every position first, thgen drains the propagation queues to convergence a single time for the whole batch.
- Added a per-colum "highest opaque block" cache so the open-sky check doesn't rescan the same column repeatedly.
- Propagation queues now drain brightest-first (`drain_highest_first`) instead of FIFO, so each cell is written its final value on the first pass instead of being revisited up to 15 times.
- Added a `#[ignore]` d micro-benchmark test for measuring bulk lighting performance manually.
**Client-side flush changes** (pumpkin/src/world/mod.rs)
- flush_block_updates now drains its entire pending queue each tick (previously throttled), and for chunk columns with ≥ CHUNK_RESEND_BLOCK_THRESHOLD (4096) queued changes, resends the whole chunk via CChunkData/CLevelChunk instead of many per-section CMultiBlockUpdate packets.
- Since the background relight can finish after a dense-column resend (baking stale/black lighting into the packet), affected columns + their neighbors are resent again once the bulk relight converges.
- Fires the ChunkSend plugin event once per resent chunk (broadcast semantics — cancel skips the resend for everyone).
## Testing
`set_block_state` now validates the block state ID and logs+ignores invalid IDs instead of potentially panicking/corrupting state.
Added `Level::new_for_lighting_bench` to support the new lighting benchmark.

I tested the new changes with my worldedit-rs prototype, everything seems to be working just fine, with amazing results as for the speed. The only caveat worth mentioning is when I place or remove very large amounts of blocks - some clusters of blocks seem to have a problem with loading so they're essentially invisible, you can fly closer to them and they appear back again - Relog helps with it so I am guessing it is a client-side issue that I could not resolve due to my limited knowledge of the engine.

### DIFF
--- a/pumpkin-plugin-api/src/lib.rs
+++ b/pumpkin-plugin-api/src/lib.rs
@@ -51,10 +51,10 @@ pub mod command {
 pub use wit::pumpkin::plugin::{
     bedrock_packets, block_entity, boss_bar, command as command_wit, common,
     context::{Context, Server},
-    entity,
+    data_components, entity,
     entity_types::EntityType,
-    gui, i18n, java_dialogs, java_packets, particles, permission, player, scoreboard, server, text,
-    world,
+    gui, i18n, item_stack, java_dialogs, java_packets, particles, permission, player, scoreboard,
+    server, text, world,
 };
 
 pub mod java_dialog {
--- a/pumpkin-protocol/src/java/client/play/chunk_data.rs
+++ b/pumpkin-protocol/src/java/client/play/chunk_data.rs
@@ -11,13 +11,104 @@ use pumpkin_world::chunk::format::LightContainer;
 use pumpkin_world::chunk::{ChunkData, palette::NetworkPalette};
 use std::io::Write;
 
+pub struct ChunkBlockEntityData {
+    pub position: pumpkin_util::math::position::BlockPos,
+    pub type_id: i32,
+    pub nbt: pumpkin_nbt::compound::NbtCompound,
+}
+
 /// Sent by the server to provide the client with the full data for a chunk.
 ///
 /// This includes heightmaps, the actual block and biome data (organized into sections),
 /// block entities (like signs or chests), and the light level information for both
 /// sky and block light.
 #[java_packet(PLAY_LEVEL_CHUNK_WITH_LIGHT)]
-pub struct CChunkData<'a>(pub &'a ChunkData);
+pub struct CChunkData<'a> {
+    chunk: &'a ChunkData,
+    live_block_entities: &'a [ChunkBlockEntityData],
+}
+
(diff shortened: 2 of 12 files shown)


justxraf:
world.wit:

 ```rs
/// A block state change for bulk world updates.
    record block-change {
        pos: block-pos,
        state: u16,
    }
    
    
 /// Sets multiple block states with the same update flags.
        set-block-states: func(changes: list<block-change>, update-flags: block-flags);
```

**Review by justxraf — COMMENTED** [review]
Needs to be added to the block-entity.wit

```
    use item-stack.{item-stack};
    resource chest-block-entity {
        get-block-entity: func() -> block-entity;
        viewer-count: func() -> u32;
        size: func() -> u32;
        get-item: func(slot: u32) -> option<item-stack>;
        set-item: func(slot: u32, item: option<item-stack>);
    }
```

Snowiiii:
CI is unhappy

justxraf:
Good afternoon Snowiii,
It will be resolved in the coming days, please allow me some time. 
Thank you, Raf

Snowiiii:
Slop

PR #2264, opened 2026-06-14
docs: document local setup requirements

Jan1k1:
Adds the missing local setup notes for contributors: recursive submodules for the plugin WIT files and the linker requirement needed before running cargo build/run.\n\nVerified with:\n- cargo +stable-x86_64-pc-windows-gnu build -p pumpkin --quiet


Snowiiii:
That's something for the docs

PR #2265, opened 2026-06-14
fix: avoid readline for redirected console streams

Jan1k1:
Avoids enabling rustyline unless both stdin and stdout are attached to a terminal. This prevents Windows launches with redirected output from hitting the invalid console handle path and losing console input setup.\n\nVerified with:\n- cargo fmt --check\n- cargo +stable-x86_64-pc-windows-gnu test -p pumpkin uses_readline_only_when_tty_is_enabled_and_streams_are_terminal\n- cargo +stable-x86_64-pc-windows-gnu build -p pumpkin --quiet\n- launched target/debug/pumpkin.exe with stdout/stderr redirected on Windows and confirmed it no longer logs Error reading console input: The handle is invalid.


Snowiiii:
Please remove this nonsense tests

PR #2266, opened 2026-06-14
feat: add zstd packet compression support for the Java client

Q2297045667:
## Description

Add zstd packet compression support for the Java client

## Testing client zstd mod:

- [ ] [PumpkinClient](https://github.com/Q2297045667/PumpkinClient)


chocodev11:
feel like it is quite overkill and should be done with a plugin

Snowiiii:
The vanilla client does not support zstd

Q2297045667:
> The vanilla client does not support zstd

I made a simple zstd support mod for third-party clients just to test out these features

PR #2269, opened 2026-06-15
fix: send air supply metadata across versions

Jan1k1:
Air supply was only sent through AIR_SUPPLY_ID. That maps to 255 for the normal 1.21 client versions, so the metadata write gets skipped and the client HUD keeps showing full bubbles while the server still drains air and applies drowning damage.

This uses AIR for the 1.21 mappings and AIR_SUPPLY_ID for 26.1, then adds a regression test so supported versions keep a real metadata id.

Tested:
- cargo fmt --all --check
- cargo test -p pumpkin entity::breath::tests::air_supply_metadata_is_available_for_supported_versions --quiet

### DIFF
--- a/pumpkin/src/entity/breath.rs
+++ b/pumpkin/src/entity/breath.rs
@@ -5,7 +5,7 @@ use pumpkin_data::effect::StatusEffect;
 use pumpkin_data::meta_data_type::MetaDataType;
 use pumpkin_data::tag;
 use pumpkin_data::tag::Taggable;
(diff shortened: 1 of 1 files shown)


Jan1k1:
Tested this in game on 26.1.2 as well. I joined the patched server, went underwater in survival, and the air bubbles now drain on the client instead of staying full.

Snowiiii:
ugly hardcoded and unnecessary tests, screems ai slop

Issue #2262, opened 2026-06-13
water bouncing


Alfiestuff:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

https://github.com/Pumpkin-MC/Pumpkin/commit/68bae5899ef0741075063cb8d97ab30b05e5c543

### What happened?

when i swim down into the water i start bouncing up and down.

https://github.com/user-attachments/assets/4952be74-2e71-4370-8f67-f13795bba11f

### To Reproduce

go for a swim!

### Expected behavior

i should be able to swim normally like vanilla minecraft.

gloryrunner:
Unable to reproduce this on the latest commit. Are you still having this issue?

Alfiestuff:
> Unable to reproduce this on the latest commit. Are you still having this issue?

sorry for the late reply this issue is fixed!

PR #2270, opened 2026-06-15
Added fix for water bouncing.

SepehrRajabi:
## Fix for water bouncing: #2262 
---
### Description:

The issue was that the the water “step out” boost still fired while a player would sneak to descend.
The water “step out” boost now skips players who are sneaking while in fluid, so swimming downward should no longer get converted into an upward bounce.

---

here is a recording of me testing the changes, and there seems to be no bounce. (the recording was too large to upload here)
https://drive.google.com/file/d/1DqLUmaQTvr562D_72JY-_9MPwSvRtVdj/view?usp=drive_link

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -991,7 +991,12 @@ impl LivingEntity {
 
         let mut velo = self.entity.velocity.load();
 
+        let descending_in_fluid = self.entity.entity_type == &EntityType::PLAYER
+            && self.entity.is_sneaking()
+            && (self.entity.touching_water.load(SeqCst) || self.entity.touching_lava.load(SeqCst));
+
         if self.entity.horizontal_collision.load(SeqCst)
+            && !descending_in_fluid
             && !self
                 .entity
                 .world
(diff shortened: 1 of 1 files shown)


GreenedDev:
is this vanilla?

SepehrRajabi:
Yes. Version 26.1 on macos

GreenedDev:
> Yes. Version 26.1 on macos

I mean does this code contain vanilla logic?

SepehrRajabi:
i think so, yes
It implements the same behavior as vanilla: if an entity in water/lava hits a horizontal collision while not sneaking as a player, it gets a small upward boost (velo.y = 0.3).

> > Yes. Version 26.1 on macos
> 
> I mean does this code contain vanilla logic?

Snowiiii:
Yeah i don't think vanilla does `self.entity.entity_type == &EntityType::PLAYER`, for such things you should look at vanilla code, please confirm its matching

chocodev11:
i think you should look for collision check rather than sneaking

SepehrRajabi:
It think this change now would suffice. I don't know why the builds are failing though.

chocodev11:
hmm seem like this one didnt fix anything. maybe collision is not the problem.

SepehrRajabi:
when i test it, i do not get bouned. perhaps i am reproducing it wrong?
I also realized that the oxygen bar does not decrease in the ui. but maybe that's its own issue.

Snowiiii:
CI is also still failing, Feel free to reopen

PR #2271, opened 2026-06-15
fix: prevent statistic overflow panics

Jan1k1:
Reproduced on latest master (8634d874) by running /kill on a player, which panicked at player/statistics.rs while incrementing a stat.\n\nThis changes statistic increments to saturate instead of overflowing. Stats are counters, so hitting an integer bound should not take the server down.\n\nTests:\n- cargo test -p pumpkin increment_saturates_when_stat_is_at_i32_ --quiet


GreenedDev:
the tests are unnecessary. this is literally a one-line fix

Jan1k1:
Trimmed it down to the actual fix.

Snowiiii:
LGTM. Thank you @Jan1k1 :D

PR #2272, opened 2026-06-15
fix: encode item custom names as text components

Jan1k1:
Reproduced on latest master (8634d874) with an anvil rename. The client disconnected while decoding container_set_slot after the result item was sent.\n\nThe anvil path was storing the rename text as a raw string in minecraft:custom_name. This stores item custom names as text component JSON, reads old raw strings as plain text, and keeps name tags/WASM item-stack access using TextComponent values.\n\nTests:\n- cargo test -p pumpkin-data set_custom_name_ --quiet\n- cargo test -p pumpkin-data custom_name_survives_item_stack_nbt_round_trip --quiet\n- cargo check -p pumpkin --quiet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2760,6 +2760,7 @@ dependencies = [
  "pumpkin-util",
  "rand",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


Snowiiii:
solved in https://github.com/Pumpkin-MC/Pumpkin/commit/2fa7a1829b20b340273f0ac6c2c9db16743158ea

PR #2273, opened 2026-06-15
fix(plugin): `inherit_stdout` and `inherit_stderr` in Wasm plugins.

QuantumSegfault:
## Description

Enables `inherit_stdout` and `inherit_stderr` to allow plugins to successfully write to standard WASI stdout/stderr (without Pumpkin logging API).

Plugin authors don't always have control to force everything through the logging API, so this allows stdout/stderr writes outside of their control to work properly.

## Testing

Testing was done by creating some example Python plugins. Using `print` during `get_metadata` (early, before permissions) and `on_load` (after permissions), which writes to stdout. Also, an error (missing parameter for PluginMetadata constructor) was introduced, to verify that the Python error backtrace was correctly dumped to the terminal before aborting.


Snowiiii:

PR #2275, opened 2026-06-16
Add a binary (non compiling) version of the pterodactyl egg

amir16yp:
## Description

The current pterodactyl egg requires compiling the pumpkin binary during the installation step. 

The average user wanting to run pumpkin on Pterodactyl/Pelican might not care at all about this step, and would want to simply download and run the nightly precompiled binary. 

This version of the egg json does exactly that: the installation phase uses curl to download the nightly binary make it executable, configure the pumpkin.toml according to the various variables specified. (online mode, max players, gamemode, seed, etc)

I used a custom version of the debian yolk docker image, the official one in pterodactyl's repos uses debian 11, which has outdated glibc causing the binary to not run, i simply created a [yolk docker image](https://github.com/amir16yp/docker-images/tree/main/debian-trixie-yolk) that is the exact same but uses debian 13 trixie. i wouldn't mind if you change it, as long you test that it works.

## Testing

I've had a lot of trial and error getting this to run on Pterodactyl. glad to say it works decently.
The only functionality that is missing is enabling bedrock, which might need additonal allocated ports in the configuration, which i cant be arsed unfortunately, though its functional for java, 100%. Screenshot attached

image

### DIFF
--- a/egg-bin-pumpkin.json
+++ b/egg-bin-pumpkin.json
@@ -15,7 +15,7 @@
     "file_denylist": ["pumpkin"],
     "startup": ".\/pumpkin",
     "config": {
-        "files": "{}",
+        "files": "{\r\n    \"pumpkin.toml\": {\r\n        \"parser\": \"file\",\r\n        \"find\": {\r\n            \"motd =\": \"motd = \\\"{{server.build.env.MOTD}}\\\"\",\r\n            \"seed =\": \"seed = \\\"{{server.build.env.SEED}}\\\"\",\r\n            \"online_mode =\": \"online_mode = {{server.build.env.ONLINE_MODE}}\",\r\n            \"default_difficulty =\": \"default_difficulty = \\\"{{server.build.env.DIFFICULTY}}\\\"\",\r\n            \"default_gamemode =\": \"default_gamemode = \\\"{{server.build.env.GAMEMODE}}\\\"\",\r\n            \"force_gamemode =\": \"force_gamemode = {{server.build.env.FORCE_GAMEMODE}}\",\r\n            \"max_players =\": \"max_players = {{server.build.env.MAX_PLAYERS}}\"\r\n        }\r\n    }\r\n}",
         "startup": "{\r\n    \"done\": \"Server is now running.\"\r\n}",
         "logs": "{}",
         "stop": "stop"
--- a/egg-bin-pumpkin.json
+++ b/egg-bin-pumpkin.json
@@ -95,7 +95,7 @@
             "default_value": "10",
             "user_viewable": true,
             "user_editable": false,
-            "rules": "required|integer|max:100",
+            "rules": "required|integer",
             "field_type": "text"
         }
     ]
--- a/egg-bin-pumpkin.json
+++ b/egg-bin-pumpkin.json
@@ -12,7 +12,7 @@
     "docker_images": {
         "debianTrixie": "ghcr.io\/amir16yp\/debian-trixie-yolk:latest"
     },
-    "file_denylist": [],
+    "file_denylist": ["pumpkin"],
     "startup": ".\/pumpkin",
     "config": {
         "files": "{}",
--- a/egg-bin-pumpkin.json
+++ b/egg-bin-pumpkin.json
@@ -1,0 +1,102 @@
+{
+    "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
+    "meta": {
+        "version": "PTDL_v2",
+        "update_url": null
+    },
+    "exported_at": "2026-06-16T13:37:05+00:00",
+    "name": "Pumpkin",
+    "author": "cheaplydothost@proton.me",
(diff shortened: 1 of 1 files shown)


amir16yp:
removed the silly max 100 requirement from the max players field

Snowiiii:
what do you think @vyPal ?, also shouldn't we rather replace/edit the original egg file?

vyPal:
I'd definitely replace the original egg, much less messy than having two separate ones. Although the approach in this PR should work, I'd consider making a second Dockerfile that would be built alongside the main one in github actions, but the image would be tagged with pterodactyl to tell them apart, this way we could build the image on an alpine base and it would use a lot less storage. Also idk if depending on some random docker image as the base for this egg is the best idea

amir16yp:
> I'd definitely replace the original egg, much less messy than having two separate ones. Although the approach in this PR should work, I'd consider making a second Dockerfile that would be built alongside the main one in github actions, but the image would be tagged with pterodactyl to tell them apart, this way we could build the image on an alpine base and it would use a lot less storage.

> Also idk if depending on some random docker image as the base for this egg is the best idea

Correct, I think it would be better if you guys would make a repo for the yolk docker image and use a copy of mine instead of mine directly

also regarding alpine: as far as I know alpine uses musl and not glibc since we only compile for glibc alpine would not work for running the image. The difference in overhead between trixie-minimal and alpine is negligible imo

vyPal:
> also regarding alpine: as far as I know alpine uses musl and not glibc since we only compile for glibc alpine would not work for running the image. The difference in overhead between trixie-minimal and alpine is negligible imo

While you are right that there are no prebuilt binaries of Pumpkin with musl support, since we'd be building our own docker image, we could build Pumpkin into the image and build it to target the specifics of the base docker image, similar to how our normal docker image works. 

As for the overhead, the trixie-minimal image (at least based on what I was able to find) takes up around 60mb compressed and up to around 180mb uncompressed. A base alpine image takes up around 5mb. This difference is negligible for smaller hosts, it can add up on larger hosts, especially if many servers use different versions of the image.

amir16yp:
> While you are right that there are no prebuilt binaries of Pumpkin with musl support, since we'd be building our own docker image, we could build Pumpkin into the image and build it to target the specifics of the base docker image, similar to how our normal docker image works.

this is a silly way of going about it, it’s why I made this change in the first place. A host wouldn’t want to dedicate 4gb of ram, time, cpu to compile the whole program every time an instance is set up.

> As for the overhead, the trixie-minimal image (at least based on what I was able to find) takes up around 60mb compressed and up to around 180mb uncompressed. A base alpine image takes up around 5mb. This difference is negligible for smaller hosts, it can add up on larger hosts, especially if many servers use different versions of the image.

This is fair. What I’m thinking is why not just add a musl binary to the nightly build process? That wouldn’t be too difficult considering it’s a change within GitHub actions workflow?

vyPal:
> this is a silly way of going about it, it’s why I made this change in the first place. A host wouldn’t want to dedicate 4gb of ram, time, cpu to compile the whole program every time an instance is set up.

It seems like you don't quite understand what I meant. The current approach uses a pretty much blank docker image besides having rust installed and compiles pumpkin on initial startup using the install command. What I proposed above would move the build from the hosting servers to our CI/CD pipeline. We'd build a specific docker image based off of alpine that would come with a Pumpkin binary compiled for alpine and musl. This would be essentially the same as our existing docker image, but modified to work with pterodactyl (as it requires some non-standard steps in the docker image creation as you are probably aware).

> This is fair. What I’m thinking is why not just add a musl binary to the nightly build process? That wouldn’t be too difficult considering it’s a change within GitHub actions workflow?

While it wouldn't be a complicated change, the isn't much of a need for musl builds as there aren't many use cases for musl besides Alpine. It would just add extra CI/CD runtime and storage requirements

amir16yp:
> > this is a silly way of going about it, it’s why I made this change in the first place. A host wouldn’t want to dedicate 4gb of ram, time, cpu to compile the whole program every time an instance is set up.
> 
> It seems like you don't quite understand what I meant. The current approach uses a pretty much blank docker image besides having rust installed and compiles pumpkin on initial startup using the install command. What I proposed above would move the build from the hosting servers to our CI/CD pipeline. We'd build a specific docker image based off of alpine that would come with a Pumpkin binary compiled for alpine and musl. This would be essentially the same as our existing docker image, but modified to work with pterodactyl (as it requires some non-standard steps in the docker image creation as you are probably aware).
> 
> > This is fair. What I’m thinking is why not just add a musl binary to the nightly build process? That wouldn’t be too difficult considering it’s a change within GitHub actions workflow?
> 
> While it wouldn't be a complicated change, the isn't much of a need for musl builds as there aren't many use cases for musl besides Alpine. It would just add extra CI/CD runtime and storage requirements

Oh ok I understand now.

amir16yp:
Setting something like that up is beyond me, considering publishing a docker image as a package is not something a non repo maintainer can do (?) you reuse some parts of my egg json for that I presume

Snowiiii:
@amir16yp so is this ready?, you still did not replace the original egg file

PR #2280, opened 2026-06-19
fix(player): fix newly joined players not seeing existing players in `tab list`

Q2297045667:
## Description

Fixed a bug where newly joined players could not see previously joined players in the tab list.

## Mismatch between Action Order and Bitmask Bit Order
The Minecraft protocol requires that the action data for each player in a `CPlayerInfoUpdate` packet must be strictly ordered according to the bitmask bit sequence (bit 0 → bit 7). The client parses the data based on this bit order; if the server writes the actions in a different sequence, the client will read corrupted data.

In the code responsible for batch-sending existing player information to new players (`spawn_java_player`):
```rust
// PlayerInfoFlags bitmask:
// ADD_PLAYER(0x01) | UPDATE_GAME_MODE(0x04) | UPDATE_LISTED(0x08) | ... = 0x5D
// Bit order: bit0 → bit2 → bit3 → bit4 → bit6

// ❌ Old Code - Action order mismatch:
let mut player_actions = vec![
    AddPlayer,          // bit 0 ✓
    UpdateListed,       // bit 3 ✗ (Client expects bit 2: UpdateGameMode)
    UpdateLatency,      // bit 4 ✗
    UpdateListOrder,    // bit 6 ✗
    UpdateGameMode,     // bit 2 ✗ (Placed at the end, client no longer expects this bit)
];
```

Fix:
```rust
// ✅ New Code - Action order strictly matches bitmask bit order:
let mut player_actions = vec![
    AddPlayer,          // bit 0
    UpdateGameMode,     // bit 2
    UpdateListed,       // bit 3
    UpdateLatency,      // bit 4
    UpdateListOrder,    // bit 6
];
```

## `tab_list_listed` Field Initialized to  `false`  and Never Updated
When a new player joins, there are two distinct logic paths in  `spawn_java_player`:

- Broadcasting the new player to existing players: Hardcodes  `UpdateListed(true)`  to instruct existing clients to list the new player.

- Sending the existing player list to the new player: Reads the existing players'  `tab_list_listed` field. However, this field defaults to  `false`  and is never updated to `true` by any code. Consequently, the new player receives  `UpdateListed(false)` for existing players, causing the client to hide them from the tab list according to the protocol.

Fix: Changed the default value of `tab_list_listed` from `false` to `true`, ensuring consistency with the hardcoded `true` used in the broadcast logic.

Behavior After Fix

Scenario | Before | After
-- | -- | --
Player B joins, Player A sees B in tab | ✅ | ✅
Player B joins, B sees A in tab | ❌ | ✅
Player C joins, C sees A and B in tab | ❌ | ✅

## Testing

image
image


Snowiiii:
LGTM. Thank you @Q2297045667 :D

PR #2281, opened 2026-06-19
refactor(pumpkin-data/compile-times): replace (some) const arrays with statics

PigTurtleDev:
the `const [T; 29873]`s for block(state) mapping were being inlined ~300 times just to be deduplicated by LLVM. They are now statics but are accessed via const fns. This improves compile times by roughly 12 minutes.

## Description
Fixes a compilation-performance regression introduced in 5b8eab0f56070d77cba8e950e1886e94584d2168.
On current master, recompilation of the `pumpkin` crate alone takes 800s + 350s (bin), compared to 200s + 350 beforehand.

This is caused by the compiler inlining a `const [&'static BlockState; 29873]` across 151 implementations of `BlockBehaviour`. Said inlined code is later deduplicated by LLVM and never makes it to the binary.
<sub>Note: This is an educated guess. The compiler spends 550s on `codegen_module`, which is a step that produces LLVM IR but is not monomorphization.</sub>

<details>
<summary> Why not fix just that? </summary>

# Alternatives 
There is an alternative quick-fix to this PR: marking these functions as `#[inline(never)]` in [./pumpkin-data/src/blocks.rs](https://github.com/Pumpkin-MC/Pumpkin/blob/master/pumpkin-data/src/blocks.rs#L163-L179)

```rs
impl Block
    #[must_use]
    pub fn mirror(
        &self,
        state_id: u16,
        _mirror: crate::block_rotation::Mirror,
    ) -> &'static BlockState {
        BlockState::from_id(state_id)
    }

    #[must_use]
    pub fn rotate(
        &self,
        state_id: u16,
        _rotation: crate::block_rotation::Rotation,
    ) -> &'static BlockState {
        BlockState::from_id(state_id)
    }
}
``` 

this does however not address the underlying issue of needlessly duplicating that large array for callers of `BlockState::from_id()`, which is slowing down compilation in other places too (most notably in `pumpkin-world`).

</details>

This PR reduces build times from 21min to 9min on my device.

## Changes

- replaced the `const array` (and 3 more like it) with a `static array` \
  and moved them to their own module `mappings` (You can't tie statics to structs like you can with consts)
- marked the functions accessing these arrays const

Marking the functions const required me to manually implement unchecked indexing as a replacement for `unsafe fn &[T].get_unchecked()`. The replacement is
```rs
const unsafe fn unchecked_index<T: Copy, const N: usize>(array: &[T; N], index: usize) -> T {
    unsafe { *(array.as_ptr().add(index)) }
}
```
Calling this function with `index >= N` is unsound. 
`BlockState::from_id()` calls this function without checking that bound.
It did so with `.get_unchecked()`, apparently due to slowdowns in debug builds.

We absolutely should switch to an u16 wrapper type that checks BlockState id validity on (const) initialization, but that's out of scope for this PR.

## Testing

- [x] `cargo check && cargo clippy`
- [x] `cargo test`
- `cargo bench` expectedly reports (mostly) noise.

This PR shouldn't (negatively) affect runtime performance, the `const fn`s still allow the compiler to optimize for compile-time-known calls and the arrays were almost definitely staticized before too.
Startup time is unaffected. Disregarding the generated block.rs the actual diff for this PR is (+40, -32)


Snowiiii:
LGTM. Thank you @PigTurtleDev :D

PR #2284, opened 2026-06-20
chore(pumpkin/compile-times): remove unused module declarations

PigTurtleDev:
## Description
`main.rs` currently contains duplicates of all `mod x` statements from `lib.rs`.
The code is unused but cargo doesn't throw a warning.

As such, analysis and codegen happen twice for the `pumpkin` crate, once for its `lib` and once for its `bin`.

This takes about a minute extra of compile time for both debug and release builds, 
though for debug builds incremental compilation can stop it from rerunning if no change is made to `pumpkin`.

## Changes
- removed the unused modules from the binary half of pumpkin (`main.rs`)
- removed an unused static from the binary half of pumpkin (`main.rs`)

Compilation takes a minute less.

## Testing
- [x] `cargo fmt && cargo check && cargo clippy`

This change has no effect on the actual binary, all removed code was dead to begin with.

### DIFF
--- a/pumpkin/src/main.rs
+++ b/pumpkin/src/main.rs
@@ -1,6 +1,5 @@
 // Don't warn on event sending macros
 #![recursion_limit = "512"]
-#![expect(unused_labels)]
 
 #[cfg(target_os = "wasi")]
 compile_error!("Compiling for WASI targets is not supported!");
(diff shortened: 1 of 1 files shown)


Snowiiii:
Oh, yeah, LGTM, Thank you @PigTurtleDev :D

PR #2287, opened 2026-06-23
feat(world): add SlimeWorld (.slime) chunk format

dongzh1:
## Summary

Adds the SlimeWorld (`.slime`) world format as a selectable, **read-write** chunk
format, addressing part of #118. Targets the current AdvancedSlimePaper format
(`SLIME_VERSION = 13`). Set `chunk.type = "slime"` in a world's config to load
the first `*.slime` file in that world folder.

## What this does

- **Container parser** (`chunk::format::slime::reader`): decodes the `.slime`
  binary container — `0xB10B` magic, version, the two zstd-compressed blocks
  (chunks + world extra), and each chunk's sections / light / heightmaps /
  block entities / entities.
- **Conversion** (`…::convert`): resolves the vanilla string block-state palette
  to Pumpkin global state ids (`Block::from_registry_key` +
  `from_properties`/`to_state_id`) and biome names to ids, rebuilds the section
  block / biome / light palettes into `ChunkData`, and maps block entities and
  entities (keyed by UUID) onto `ChunkData` / `ChunkEntityData`.
- **Writer** (`…::writer`): the inverse — section palettes become vanilla
  string-paletted NBT, light is written as raw nibble arrays, block entities and
  entities are emitted, and the container is framed and zstd-compressed like
  ASP's `SlimeSerializer`.
- **IO backends** (`…::io` + `…::store`): both the chunk and entity `FileIO`
  backends share one `SlimeWorldStore` holding the whole world in memory. Reads
  are served from the in-memory maps; a save from either side re-serializes the
  whole world back to the single file. Because SlimeWorld is a "whole world in
  one file, in RAM" format by design, this implements `FileIO` directly instead
  of going through the region-oriented `ChunkFileManager`, so there is zero
  impact on the Anvil / Linear / Pump hot path.
- **Wiring**: `ChunkConfig::Slime`, dispatched in `Level::from_root_folder`.

## Scope / limitations

- Targets `SLIME_VERSION = 13` (ASP `dev/26.2`); legacy v1–v9 layouts are out of
  scope for now.
- On save the whole file is rewritten (fine for the small worlds SlimeWorld is
  meant for). POI, block-tick and fluid-tick segments are not round-tripped (the
  `flags` byte is written as `0`), and `level.dat` metadata from the slime
  `properties` (spawn / environment) is not yet synthesized.

## Testing

- 12 unit tests: the container parser (round-trip, light arrays, bad magic /
  newer version / truncation), the converter (block-state + property resolution
  against the real registry, biome lookup, UUID decode), and a write→read
  round-trip that rebuilds a chunk through the writer and reader.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features`
  (clean), and `cargo test -p pumpkin-world` (80 passing) are all green.

The byte layout was verified against the AdvancedSlimePaper `dev/26.2` sources
(`SlimeSerializer`, `v13SlimeWorldDeSerializer`).

Part of #118.

### DIFF
--- a/pumpkin-config/src/chunk.rs
+++ b/pumpkin-config/src/chunk.rs
@@ -4,7 +4,8 @@ use serde::{Deserialize, Serialize};
 
 /// Configuration for chunk storage format.
 ///
-/// Supports multiple chunk formats, currently `Anvil` and `Linear`.
+/// Supports multiple chunk formats: `Anvil`, `Linear`, `Pump`, and the
+/// read-only `Slime` (AdvancedSlimePaper `.slime`) format.
 #[derive(Deserialize, Default, Serialize, Clone)]
 #[serde(tag = "type")]
 pub enum ChunkConfig {
@@ -18,6 +19,12 @@ pub enum ChunkConfig {
     #[serde(rename = "pump")]
     #[default]
     Pump,
+    /// SlimeWorld (`.slime`) format, loaded read-only (AdvancedSlimePaper v13).
+    ///
+    /// The whole world is read into memory from the first `*.slime` file in the
+    /// world folder; runtime changes are not written back.
+    #[serde(rename = "slime")]
+    Slime,
 }
 
 /// Configuration for Anvil chunk storage.
--- a/pumpkin-config/src/chunk.rs
+++ b/pumpkin-config/src/chunk.rs
@@ -5,7 +5,7 @@ use serde::{Deserialize, Serialize};
 /// Configuration for chunk storage format.
 ///
 /// Supports multiple chunk formats: `Anvil`, `Linear`, `Pump`, and the
-/// read-only `Slime` (AdvancedSlimePaper `.slime`) format.
+/// read-only `Slime` (`AdvancedSlimePaper` `.slime`) format.
 #[derive(Deserialize, Default, Serialize, Clone)]
 #[serde(tag = "type")]
 pub enum ChunkConfig {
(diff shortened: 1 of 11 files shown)


dongzh1:
Update: the format is now **read-write**, not read-only. Both the chunk and entity `FileIO` backends share a single in-memory `SlimeWorldStore`, so saving from either side re-serializes the whole world back to the `.slime` file (writer added in `chunk::format::slime::writer`). Tests, clippy (workspace, all-targets/all-features) and fmt are green.

dongzh1:
Rebased onto current `master` and fixed the issues that were causing CI to fail before pedantic lints even ran:

- `chunk_to_entity_data` built a `HashMap<Uuid, NbtCompound>`, but `ChunkEntityData::data` is actually `Mutex<Vec<NbtCompound>>` (same shape the Anvil format uses) — this was both a type error and a correctness bug, since entities missing a well-formed `UUID` int-array tag were silently dropped on import. Entities are now kept as a flat list, matching vanilla's on-disk representation.
- Block-state ids moved to the `BlockStateId` newtype upstream (#2328) since this branch was opened; `convert.rs`/`writer.rs` are updated accordingly.
- The module never actually got far enough in CI to hit the workspace's `pedantic`/`nursery`/`redundant_clone` deny-lints (it always failed to compile first); fixed doc-markdown backticks, `pub(crate)` items inside already-private modules, non-`const` constructors, a `Vector2` taken by reference instead of by value, and a redundant `Arc` clone in `level.rs`.

`cargo fmt --check`, `cargo clippy --all-targets --all-features` (workspace-wide, `-D warnings`), and `cargo test -p pumpkin-world` (97 passing, incl. a new regression test for the entity-list fix) are all clean locally.

Snowiiii:
Slop

PR #2292, opened 2026-06-24
fix(entity): preserve infinite status effect durations

### LINKED ISSUE #2289: Infinite potion effects turn negative after rejoin

### Operating System

Linux Mint 22.3 - Xfce 64-bit

0.1.0-dev+26.1/461a5eb

### What happened?

I gave myself infinite night vision
I rejoined the server
The potion effect duration turned to 00:-3
(this also sometimes causes screen flicker in the distance)
(i used a java 26.2 client)

### To Reproduce

give yourself any infinite potion effect using /effect give @p minecraft:night_vision infinite (1-255) (true/false) 

### Expected behavior

potion effect stays infinite and does not turn negative

NicDevTV:
## Description

Fixes #2289

Fixes infinite status effects being decremented after they are loaded from persisted entity data.

Vanilla uses `-1` as the sentinel duration for infinite status effects and keeps that value unchanged while ticking. Pumpkin already saved and loaded infinite effects as `-1`, but `LivingEntity::tick_effects` decremented every active effect duration unconditionally. After rejoining, an infinite effect could therefore continue as `-2`, `-3`, etc., which caused invalid negative timers such as `1:-01` on the client.

This change keeps `-1` unchanged and uses the entity age for periodic effect tick checks when the effect is infinite, matching vanilla `StatusEffectInstance` behavior.

## Testing

- `cargo fmt --check`
- `cargo check -p pumpkin`

## Notes

This is my first contribution to Pumpkin, so feedback is welcome.


NicDevTV:
And now the branch update broke the checks. 😅

Snowiiii:
LGTM. Thank you @NicDevTV :D

PR #2293, opened 2026-06-24
fix(entity): split the types of wind charges

SomeYellowGuy:
## Description
Minecraft has two separate entities for wind charges: normal ones from players and dispensers, and breeze ones from breezes. Only normal wind charges have a deflection cooldown, so some slightly different tick logic is required.

## Testing
Verified that both types exist independently using the `/summon` command.


Snowiiii:

PR #2294, opened 2026-06-24
fix(world): persist end portal block entities

### LINKED ISSUE #2231: End portal vanishes after rejoin

### Operating System

Fedora 44

Pumpkin 0.1.0-dev+26.1 (Commit: 463c25d/release) - 224 Contributors

### What happened?

The endportal vanished after rejoining the server.

### To Reproduce

Build an endportal and activate it. Leave and join the server.

### Expected behavior

It should stay visible.

NicDevTV:
## Description

Fixes #2231.

Fixes end portals losing their rendered portal surface after a player rejoins.(only the render, the block is still there)

End portal blocks render through their block entity renderer, so the block entity must be present when the chunk is sent back to the client. This stores the minimal end portal block entity NBT when an end portal block is created, while keeping the runtime block entity registration intact.

Also strips save-only block entity fields (`id`, `x`, `y`, `z`) from Java chunk block entity payloads, since the chunk packet already sends type and position separately.

I'm new to the project, so I'm happy to adjust this if there is a preferred pattern for handling block entity persistence.

## Testing

- `cargo clippy --all-targets`
- `cargo test`

### DIFF
--- a/pumpkin-protocol/src/java/client/play/chunk_data.rs
+++ b/pumpkin-protocol/src/java/client/play/chunk_data.rs
@@ -213,7 +213,13 @@ impl ClientPacket for CChunkData<'_> {
             });
 
             write.write_var_int(&VarInt(id as i32))?;
-            write.write_nbt(nbt.clone().into())?;
+
+            let mut client_nbt = nbt.clone();
+            client_nbt.child_tags.remove("id");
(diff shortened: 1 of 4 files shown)


Snowiiii:
LGTM. Thank you @NicDevTV :D

PR #2300, opened 2026-06-25
feat(codec): add barebones derive macros for `Encode` and `Decode`

SomeYellowGuy:
## Description
This PR adds support to derive `Encode` and `Decode` implementations for simple and unit structs (which don't require attributes). This is the equivalent of implementing `Codec`s in Minecraft, which are used everywhere, and this is supposed to have a similar effect to the `RecordCodecBuilder` system in Java with the Rust idiom.

For example, a `Codec` for a record `Book` in Java which would be written like

```java
public static final Codec<Book> CODEC = RecordCodecBuilder.create(instance -> instance.group(
		Codec.STRING.fieldOf("name").forGetter(Book::name),
		Codec.STRING.fieldOf("author").forGetter(Book::author),
		ExtraCodecs.NON_NEGATIVE_INT.fieldOf("pages").forGetter(Book::pages)
).apply(instance, Book::new));
```

could simply have its traits derived like so:

```rust
#[derive(Encode, Decode)]
pub struct Book {
    name: String,
    author: String,
    pages: u32,
}
```

This should hopefully minimize manually implementing these traits for many structs.

_Support for attributes, enums and tuple structs can be added in a later PR, but I removed them from this branch to make the PR smaller._

## Testing

Added tests for deriving the above traits for some example structs, which all pass.


Sawlyer:
Two blockers: the branch conflicts with master; the macros crate imports proc-macro2 via proc_macro_error2::__export::proc_macro2, a private internal re-export that can break in any patch release, and proc-macro2 is already a workspace dependency, so depend on it directly (at which point proc-macro-error2 is entirely unused). Two optional asks: a test pinning the corrected omit-when-default behavior, and a clearer error for zero-field structs. Heads-up: if #2410 merges first, this new crate needs hakari wiring.

PR #2301, opened 2026-06-25
fix(protocol): encode and decode renamed item names

### LINKED ISSUE #2223: Crahs when renaming things in an anvil

### Operating System

Server: Alpine Linux v3.23 (aarch64) running on a Raspberry Pi 4B

a2f0122/release

### What happened?

I tried renaming a nametag in an anvil (the error happens with any item) and when i modify the text in any way (delete a character, add a character) then i get disconnected

Image

Image

### To Reproduce

Place an anvil
Put an item into the first slot
Click in the name field at the top
Press any letter key, or backspace

### Expected behavior

It should not kick me

NicDevTV:
## Description

Fixes item stack custom name handling for items renamed in anvils.

Renamed items use Minecraft's text component codec for the `custom_name` data component. Pumpkin previously treated this value like a plain string in parts of the item stack protocol path, which could produce invalid component payloads and disconnect or crash clients when typing an anvil rename.

This also fixes the related serverbound creative slot path by decoding item stacks with the length-prepended optional item stack format used by Minecraft.

Closes #2223

**I'm somehow new to the Pumpkin Community so please be nice. I am open for feedback**

## Testing

- `cargo fmt --all`
- `cargo check -p pumpkin-protocol -p pumpkin`
- `cargo clippy -p pumpkin-protocol -p pumpkin --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/pumpkin-data/src/data_component_impl.rs
+++ b/pumpkin-data/src/data_component_impl.rs
@@ -62,6 +62,7 @@ pub fn read_data(id: DataComponent, data: &NbtTag) -> Option<Box<dyn DataCompone
         PotionContents => Some(PotionContentsImpl::read_data(data)?.to_dyn()),
         Fireworks => Some(FireworksImpl::read_data(data)?.to_dyn()),
         FireworkExplosion => Some(FireworkExplosionImpl::read_data(data)?.to_dyn()),
+        CustomName => Some(CustomNameImpl::read_data(data)?.to_dyn()),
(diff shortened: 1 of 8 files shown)


NicDevTV:
So overall, I can say I completed the TODO to change it from a String to a TextComponent.

Snowiiii:
LGTM, Awesome!. Thank you @NicDevTV :D

PR #2307, opened 2026-06-27
Fix: set client player before spawn to prevent invisible chunks on first join

ZlordHUN:
## Description
The client.player was set after spawn_java_player, causing the first few chunk sends to silently drop because the client didn't know about the player yet — and since those chunks were marked as sent, they were never re-delivered.

### DIFF
--- a/pumpkin/src/lib.rs
+++ b/pumpkin/src/lib.rs
@@ -471,11 +471,15 @@ impl PumpkinServer {
                                      .add_player(Arc::new(ClientPlatform::Java(java_client)), profile, Some(config))
                                           .await
                                 {
+                                    // Set the player on the client BEFORE spawning so that chunk
+                                    // sends during spawn_java_player don't get silently dropped.
+                                    if let ClientPlatform::Java(client) = player.client.as_ref() {
+                                        *client.player.lock().await = Some(player.clone());
+                                    }
                                     world
                                         .spawn_java_player(&server_clone.basic_config, &player, &server_clone)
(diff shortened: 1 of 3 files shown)


dongzh1 (commented):
The reordering looks right for the Java first-join case — setting `client.player` before `spawn_java_player`, so the initial chunk sends aren't silently dropped and then marked as sent, is the correct fix. Nice catch on the "marked sent but never re-delivered" detail.

A couple of things worth considering:

1. The reorder is applied only to the Java path in `lib.rs`, but the new `warn!` was added to **both** `JavaClient::send_chunks` and `BedrockClient::send_chunks`. Does the Bedrock first-join flow also set `player` after spawn? If so, Bedrock could still drop those first chunks (just logged now rather than fixed) — might be worth fixing symmetrically, or noting it as out of scope for this PR.
2. The `warn!("send_chunks: player not set yet ...")` fires at WARN on a path that, after this fix, should normally never trigger. As a tripwire that's fine, but if any legitimate early call remains it could get spammy — `debug!` might be safer.

Otherwise LGTM.

ZlordHUN:
@dongzh1 from what I can tell bedrock doesnt have this bug. I implemented your second suggestion, thank you for the feedback!

dongzh1 (commented):
Thanks for checking the Bedrock path and downgrading the log — both make sense. LGTM 👍

Snowiiii:
LGTM, Thank you @ZlordHUN :D

PR #2308, opened 2026-06-27
fix: creative fire ticks

AnthonyDuong1:
## Description
1. Moved set_on_fire_for and set_on_fire_for_ticks to EntityBase trait so player.rs can override set_on_fire_for_ticks with its own logic. Also changed lava and fire collision to call set_on_fire_for and set_on_fire_for_ticks through EntityBase so player entities can use its own function.
2. This is what vanilla does, and currently, even when players are in creative, fire ticks is set to the default 160 ticks, when it should be 1.
3. Players in creative mode now correctly have fire ticks set to 1.

## Testing
I ran and passed cargo clippy --all-targets, cargo test, and cargo fmt


Snowiiii:
LGTM. Thank you @AnthonyDuong1 :D

PR #2314, opened 2026-06-28
fix(protocol): bound bedrock RakNet split-fragment and ACK allocations

dongzh1:
## Summary

Fixes two unbounded-allocation (memory-exhaustion) DoS vulnerabilities in the Bedrock RakNet read path inside `pumpkin-protocol`. Both are reachable from a single, unauthenticated UDP datagram, before any login handshake completes.

## Bug A 鈥 unbounded `split_size` in `Frame::read`

`pumpkin-protocol/src/bedrock/frame_set.rs` reads the split header field with no validation:

```rust
if split {
    frame.split_size = u32::read_be(reader)?; // attacker-controlled, unbounded
    ...
}
```

The consumer in `pumpkin/src/net/bedrock/mod.rs` (`handle_frame`) then pre-allocates from that value **before** the fragment-index bounds check:

```rust
let entry = compounds.entry(compound_id).or_insert_with(|| {
    let mut vec = Vec::with_capacity(frame.split_size as usize);
    vec.resize_with(frame.split_size as usize, || None);
    vec
});
if fragment_index >= entry.len() { /* checked only AFTER the alloc */ }
```

A datagram declaring the split flag with `split_size` near `u32::MAX` makes the server try to allocate ~4 billion `Option<Frame>` slots (tens of GB) 鈫 OOM / crash.

**Fix:** reject `split_size == 0` (a split frame must declare at least one fragment) or `split_size > MAX_FRAGMENTS` right after reading it, returning `InvalidData` via the file's existing `Error::other(..)` style.

## Bug B 鈥 range expansion in `Acknowledge::read`

`pumpkin-protocol/src/bedrock/ack.rs` caps the record **count** but not the expanded total:

```rust
if size > MAX_ACK_RECORDS { return Err(..); } // count bounded to 4096
...
let start = u24::read(reader)?.0;
let end = u24::read(reader)?.0;
for i in start..=end { sequences.push(i); } // 24-bit range 鈫 up to ~16.7M each
```

Each of up to 4096 range records can span the full 24-bit sequence space (~16.7M values), so a few records push hundreds of millions of `u32` into `sequences` 鈫 OOM.

**Fix:** introduce `MAX_ACK_SEQUENCES` and track the running expanded total, validating each record's span **before** growing the vector (so the huge allocation never happens, not just detected afterward). `end < start` is handled gracefully (empty range).

## Constants chosen

- `MAX_FRAGMENTS = 1024`: RakNet fragments carry at most `SPLIT_FRAME_MAX_CONTENT` (~1346) bytes, so 1024 fragments allow a ~1.3 MB reassembled message. This is the inbound (client鈫抯erver) path, where messages are small (movement, commands, acks); 1.3 MB is comfortably above any legitimate value while capping the pre-allocation to a few tens of KB. Conservative and tunable.
- `MAX_ACK_SEQUENCES = 4096`: matches the spirit of `MAX_ACK_RECORDS` and bounds the vector to ~16 KB per packet. RakNet's in-flight/sliding window is far smaller than this in practice, so legitimate ACK/NACK ranges are unaffected.

## Why it's low-risk

- Read-path only; no behavioral change for well-formed packets within the (generous) bounds.
- Errors use the same `Error::other` / `ErrorKind::InvalidData` types already returned in these functions, so callers handle them unchanged (the connection's existing error path).
- No new dependencies; all imports already present.

## Verification

`cargo -p pumpkin-protocol build` / `clippy` (human to run).

### DIFF
--- a/pumpkin-protocol/src/bedrock/ack.rs
+++ b/pumpkin-protocol/src/bedrock/ack.rs
@@ -1,6 +1,11 @@
 use std::io::{Error, ErrorKind, Read, Write};
 
 const MAX_ACK_RECORDS: u16 = 4096;
+/// Upper bound on the total number of sequence numbers a single ACK/NACK
+/// packet may expand to. Range records (`start..=end`) cover 24-bit values, so
+/// without this cap a handful of records could expand to hundreds of millions
+/// of entries and exhaust memory.
+const MAX_ACK_SEQUENCES: usize = 4096;
 
 use crate::{
     codec::u24,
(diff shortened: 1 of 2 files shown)


Sawlyer:
Verified both vulnerabilities against master: handle_frame pre-allocates from attacker-controlled split_size before the fragment-index check, and Acknowledge::read caps record count but not range expansion. The fix is minimal, checks bounds before allocating, and matches the existing error style. Two asks: add unit tests for both bounds so this can't regress silently; confirm a login carrying a large persona skin stays under the 1024-fragment (likely 1.3 MB) inbound cap.

PR #2328, opened 2026-06-30
refactor(data) BlockId and BlockStateId wrapper types

PigTurtleDev:
## Description
This PR introduces two bounded transparent u16 wrapper types: `BlockId` and `BlockStateId`.
By ensuring all Block(State)Ids are always valid, we can avoid revalidating them outside of (plugin-) IO.

This should also help to prevent people accidentally comparing `BlockId`s with `BlockStateId`s or even item ids,
and may make it more obvious that checking whether a `BlockStateId` is part of a BlockTag is wrong.

## Changes
- added the two structs as public with a private wrapped field: `pub struct BlockId(u16)`.
  - they are `repr(transparent)` and effectively just a single `u16` so I made them `Copy` and every method take ownership of `self`.
- pumpkin-codegen/src/block.rs now adds a `const BLOCK_COUNT | STATE_COUNT: u16` to the impls for `BlockId` and `BlockStateId` respectively
  - the constructors for the structs check this bound.
  - added a comment above them asking not to violate the assumption that the contained value is bounded to said constant

### `BlockId`
- (currently) bounded to < 1196u16

Similarly to `Block`, contains named constants for every block in the game.
because the struct is `derive(PartialEq)`, pattern matching against the constants is allowed (even in const contexts!)
<details>
<summary>Example</summary>

```rs
const fn get_dead_coral_block_type(id: BlockId) -> Option<&'static Block> {
    match id {
        BlockId::BRAIN_CORAL_BLOCK => Some(&Block::DEAD_BRAIN_CORAL_BLOCK),
        BlockId::BUBBLE_CORAL_BLOCK => Some(&Block::DEAD_BUBBLE_CORAL_BLOCK),
        BlockId::FIRE_CORAL_BLOCK => Some(&Block::FIRE_CORAL_BLOCK),
        BlockId::HORN_CORAL_BLOCK => Some(&Block::DEAD_HORN_CORAL_BLOCK),
        BlockId::TUBE_CORAL_BLOCK => Some(&Block::DEAD_TUBE_CORAL_BLOCK),
        _ => None,
    }
}
```

note that stable constification of PartialEq is currently in a weird state:

```rs
// this is ok
const { matches!(block_id, BlockId::AIR) }

// this is not
const { block_id == BlockId::AIR }

// this is ok again
const { block_id.as_u16() == BlockId::AIR.as_u16() }
```

</details>
Also implements `PartialEq<Block>`, but intentionally not `PartialEq<u16>`.
may be converted to a `&'static Block` via `block_id.to_block()` or `Block::from_id(block_id)`.

### `BlockStateId`
- (currently) bounded to < 32366u16

Replacement for `pumpkin_world::BlockStateId`/`pumpkin_world::RawBlockState`/(unused)`pumpkin_data::BlockStateRef`

Allows conversion to `&'static BlockState`, `BlockId`, `&'static Block`, using the same methods as `RawBlockState`:
`to_state()`, `to_block()` and `to_block_id()`.

---

### Bound checks

Since all Block(State)Ids are always valid, we can use them to safely index into the arrays in `pumpkin-data/src/generated/block.rs` without checking bounds.
From my testing the compiler struggles to reason about that, so I added `unsafe { assert_unchecked() }` expressions for them.

In `profile.dev` builds (opt-level=0 and debug_assertions), these expressions get compiled to actual (panicking) UB-checks, reintroducing the bound checks they remove. 
Interestingly this behavior isn't different to `&[T].get_unchecked()` that was used before I recently replaced it with a `const fn unchecked_index()`; Notably though the compiler liked my function much less, the UB checks for validating alignment, pointer arithmetic and derefs in a non-inlined (despite the attribute) function are far more than just one bounds check. The more you know...

In release builds (or any build with opt-level=1 or debug_assertions=false) these checks disappear entirely and the compiler accepts the hint.
The unsafe blocks are annotated for Safety along with this info.

### Note
I've tried to split this PR into reviewable commits:
1. just replaces `impl BlockMetadata for T` with `#[pumpkin_block("id_of_t")]`

<details>
<summary>
2. adds `BlockId`
</summary>

- touches `generated/block.rs`: 
  - +-1196 loc `Block { id: <changed> }`
  - +1196 loc BlockId consts
  - +32366, - ?? loc switching `mappings::BLOCK_ID_FROM_STATE_ID` from `[u16; 32366]` to `[BlockId; 32366]`
- touches Cargo.lock: added `heck` dependency to `pumpkin-macros` to make generated `BlockMetaData` impls use `BlockId::SOME_CONST`
</details>

<details>
<summary>
3. adds `BlockStateId`
</summary>

- touches `generated/block.rs`:
  - adjusted signature of BlockProperties trait
  - +-32366 loc `BlockState { id: <changed> }`
</details>

4. tons of changes from `pumpkin_world::BlockStateId` to `pumpkin_data::BlockStateId`. It's the same type, 3. adds it as a reexport to pumpkin-world.

5. refactors of the unsafe blocks like previously mentioned.

Sadly changes to the generation of `block.rs` always end up blowing up PR size... 

## Further Work
Fluid ids and Item ids are still plain `u16` values; they could use their own wrappers.

I don't like that `Tag`s contain raw u16 and str values. It might be worth to change them to a more type safe data structure
<details>
<summary>Just one Idea</summary>

```rs
pub trait Taggable {
    type RegistryId;
    /* methods */
}

pub struct Tag<T: Taggable>{
    ids: &'static [T::RegistryId];
    values: &'static[&'static T];
}
```

</details>

But that's out of scope for this PR.

## Testing
- [x] `cargo check && cargo clippy`
- [x] `cargo test` passes
- [x] `cargo bench`marks are much too noisy for me to say with certainty, but world generation seems to have improved by 1-5%?

There is no observable change to compile times.


Snowiiii:
LGTM. Thank you @PigTurtleDev :D

Issue #36, opened 2024-08-19
World Generation


Snowiiii:

Bryntet:
Idea, maybe a first start could be to implement superflat, we also should probably have an enum with the different world-gen options (noise, debug, superflat, normal)

Snowiiii:
> Idea, maybe a first start could be to implement superflat, we also should probably have an enum with the different world-gen options (noise, debug, superflat, normal)

Good idea

CoolSlimbo:
To add, slime chunks are also needed to be generated.

NiLSPACE:
Another resource that could be useful : http://cuberite.xoft.cz/docs/Generator.html

It documents the way Cuberite's world generator works. I believe Feather also based their generator on it.

roobscoob:
> Idea, maybe a first start could be to implement superflat, we also should probably have an enum with the different world-gen options (noise, debug, superflat, normal)

Rather than an enum, you should probably create a `WorldGenerator` trait. You could create a `SuperflatGenerator` that implements `WorldGenerator`. This makes it super easy for third-parties to implement their own generators.

Snowiiii:
> > Idea, maybe a first start could be to implement superflat, we also should probably have an enum with the different world-gen options (noise, debug, superflat, normal)
> 
> Rather than an enum, you should probably create a `WorldGenerator` trait. You could create a `SuperflatGenerator` that implements `WorldGenerator`. This makes it super easy for third-parties to implement their own generators.

Hey, I think this is actual what we are ended doing now

dbwrush:
This thread is, as far as I know, the only mention of "XOROshiro48" on the open internet. @Snowiiii did you mean XOROshiro64 or XOROshiro32?

suprohub:
Yeah

rice7th:
[...]

PR #2329, opened 2026-06-30
feat: Implement Ancient Cities

ZlordHUN:
## Description
- This PR adds ancient cities and a stability guard for schedule.rs without it my Pumpkin server crashes every time I go near an ancient city.
- If the code is adequate another item can be checked off here: https://github.com/Pumpkin-MC/Pumpkin/issues/36

## Testing
- For testing I resorted to implementing /locate again which I removed after testing was completed. Located coordinates for ancient cities, teleported there. Initially each teleport crashed the server (Multi chunk ancient city generation caused the server to panic) which is why I had to modify the schedule.rs file.


Snowiiii:
LGTM, Thank you @ZlordHUN :D

Issue #2330, opened 2026-06-30
Panic: "acacia_fence is not a valid block for WallTorchLikeProperties"


lineality:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

fedora linux 

### Server Software Version/Commit

af712b7 (today)

### What happened?

```
2026-06-30 18:25:29 ERROR            main ThreadId(01) pumpkin::crash: Pumpkin has encountered a panic!
2026-06-30 18:25:29 ERROR            main ThreadId(01) pumpkin::crash: 
2026-06-30 18:25:29 ERROR            main ThreadId(01) pumpkin::crash: Thread 'tokio-rt-worker' with ThreadId(3) panicked at pumpkin-data/src/generated/block.rs:548993:13:
2026-06-30 18:25:29 ERROR            main ThreadId(01) pumpkin::crash: acacia_fence is not a valid block for WallTorchLikeProperties
2026-06-30 18:25:29 ERROR            main ThreadId(01) pumpkin::crash: 
2026-06-30 18:25:29 ERROR            main ThreadId(01) pumpkin::crash: The full backtrace will be printed to the crash report.

```

Found odd village, punched a fence... on no, panic!

Crash Report:
[crash-2026-06-30_22.25.29.txt](https://github.com/user-attachments/files/29525297/crash-2026-06-30_22.25.29.txt)

```
2026-06-30 18:08:58  INFO main ThreadId(01) pumpkin: Starting Pumpkin 0.1.0-dev+26.2 Minecraft (Protocol 776)
2026-06-30 18:08:58  WARN main ThreadId(01) pumpki


n: Pumpkin is currently under heavy development!
2026-06-30 18:08:58  INFO main ThreadId(01) pumpkin: Report issues on https://github.com/Pumpkin-MC/Pumpkin/issues
2026-06-30 18:08:58  INFO main ThreadId(01) pumpkin: Join our Discord for community support: https://discord.gg/wT8XjrjKkf
2026-06-30 18:08:58  WARN main ThreadId(01) pumpkin::server: Failed to get level_info, using default instead: Info not found!
2026-06-30 18:08:58 ERROR main ThreadId(01) pumpkin::server: Failed to save level.dat: Info not found!
2026-06-30 18:08:58  INFO main ThreadId(01) pumpkin::server: Enabled dimensions: ["minecraft:overworld", "minecraft:the_nether", "minecraft:the_end"]
[...]

PR #2331, opened 2026-07-01
fix: fences crash

DeskTecc:
## Description
1. Changed ids() in glazed_terracotta.rs from **tag::Item** to **tag::Block**.

2. This change was necessary because in the tag.rs, inside the mod Item in the const C_GLAZED_TERRACOTTAS, all the ID's of items are fences and this is the reason of crash.

3. For 26.1+ clients, players will now be able to place/break fences without crash.

4. I saw this issue [#2330](https://github.com/Pumpkin-MC/Pumpkin/issues/2330) and I've decided to test it. Not only the acacia fences but all non oak fences was panicking.

## Testing
I know it's just a simple bug of misstype but just to follow guidelines, I've ran and passed all the tests: cargo clippy, cargo test and cargo fmt.


Snowiiii:
LGTM, Thank you @DeskTecc :D

PR #2332, opened 2026-07-01
fix: Made (CUSTOM_)NAME_VISIBLE work across versions

knownasred:
## Description

I'm trying to experiment with pumpkin plugins on an old idea of mine, and I had issues making name tags on armor stands work. After some looking around, it seems that the metadata name changed between 1.21 and later versions, as they are now supported by more entities.

This PR is a workaround to make it work on older versions.

In relation with the other PR in this set (#2333), it should make this feature work.

## Testing
- [x] Tested on version 1.21.8 (originally had the issue, fixed after changes)
- [x] No regressions on 26.2


Snowiiii:
Quick and dirty fix, i guess its fine for now but we should consider something better in the future, Thanks @knownasred :D

PR #2333, opened 2026-07-01
feat(plugin-api): Forced use of setters

knownasred:
## Description
As far as my testing goes, the Plugin API doesn't send the metadata to the client, which causes the existing clients to not get the updated custom name.

This uses the setters which also sends the associated metadata, even if it is in two separate packets (seems related to the fact that we cannot provide multi-type packets for entity metadata updates)

## Testing
- Tested on 26.2 and 1.21.8 with #2332


Snowiiii:
LGTM, Thank you @knownasred :D

PR #2335, opened 2026-07-01
Speed up chunk generation (structure starts + light engine)

TheDarkSword:
This speeds up chunk generation quite a bit, mostly by removing redundant work in structure placement and in the light engine. Generation output is unchanged — all `pumpkin-world` tests, including the fixed-seed ancient-city parity test, still pass.

### Results

Measured with the `chunk_gen` / `chunk_gen_concurrent` benches (seed 42, overworld):

| benchmark | before | after | speedup |
|---|---|---|---|
| full chunk generation | 103 ms | 52 ms | 2.0× |
| region — 16×16 chunks, 4 threads | 730 ms | 225 ms | 3.2× |
| structure references stage | 342 µs | 105 µs | 3.3× |
| lighting stage | 65 ms | 22 ms | 3.0× |

### Changes

**1. Cache computed structure starts**

`set_structure_references` runs for every chunk and, for each nearby structure candidate, recomputed the structure's placement from scratch. For jigsaw structures (villages, ancient cities, …) that means re-running the full jigsaw expansion for every chunk whose references overlap the structure — so a single structure ends up being expanded many times over as its neighbours generate. This got noticeably more expensive once the ancient-city jigsaw landed.

A structure's placement only depends on its start chunk and the world seed: the random source is seeded from the start chunk, `min_y` is constant per dimension, and the surface-height estimate it uses is position-independent. So the result is now memoized in `GlobalStructureCache` (which already existed for strongholds) and reused, and the `context` is only built on a cache miss.

**2. Use a fast hasher in the generation light engine**

The BFS light propagator's `visited` / `shadow_cache` / `pending_updates` maps were aliased to `std::collections::HashSet`/`HashMap` — i.e. SipHash — despite being named `FastHashSet`/`FastHashMap`. They're probed on every neighbour of every propagated block, so the hash function dominates. Pointing the aliases at rustc-hash's `FxHash` (already a dependency) is a drop-in change.

**3. Propagate light through storage instead of a shadow cache**

On top of the hashing, the propagator kept a hashed shadow cache of in-flight light values plus a per-chunk batched write buffer, layered over the actual light storage. But the storage is itself a simple array lookup, so that extra bookkeeping cost more than it saved. The BFS now reads and writes the storage directly and treats it as the single source of truth, which removes the per-neighbour hash-map work and the temporary allocations.

### Correctness

These are all output-preserving refactors — no change to generated terrain, structures or light. Verified with the existing `pumpkin-world` test suite, including `configured_seed_generates_vanilla_ancient_city_chunk` and `generate_chunk_should_return`.


Q2297045667:
Looks good! Are you all done with it? If possible, I'd like to test it out later today.

TheDarkSword:
Thank you! Yes all done, you can test when you want

Q2297045667:
I can confirm that this PR does not break the final terrain generation. Waiting for @Snowiiii  final review.
I can confirm that these optimizations are indeed effective. The terrain generation speed is noticeably faster. LGTM!

TheDarkSword:
Pulled the latest master in. The only thing that clashed was the import block in the generation light engine.

master had made that surface-height map explicit about its hasher with `RandomState`, but this PR already replaces that map with `FxHashMap` (that's the whole point of the fast-hasher change). `FxHashMap::with_capacity_and_hasher` expects an `FxBuildHasher`, not a `RandomState`, so I kept the Fx path (passing `rustc_hash::FxBuildHasher` straight in) and dropped the now-unused `HashMap`/`HashSet`/`RandomState` imports. Everything else merged clean.

TheDarkSword:
I think CI fail is github problem

Q2297045667:
> I think CI fail is github problem

Yes, signature verification for Windows has been added in the latest version. This PR, https://github.com/Pumpkin-MC/Pumpkin/pull/2365, will resolve the issue.

Snowiiii:

PR #2337, opened 2026-07-02
Test framework for vanilla-parity checks (WIP)

TheDarkSword:
## What this is

An early, work-in-progress test framework for checking that Pumpkin behaves like a vanilla Minecraft: Java Edition server. The goal is to make it easy to add tests at increasing levels of fidelity:

- **L1 — in-memory game logic:** spin up a real `Server` in RAM, attach a player, drive real actions and assert on world/entity state. No sockets, no sleeps, deterministic.
- **L2 — protocol conformance:** feed serverbound packets through the real decode/handle path and check the clientbound packets that come back.
- **L3 — end-to-end:** a headless bot speaking the real protocol against a running server.
- **L4 — differential:** run the same scenario against Pumpkin and a real vanilla server and diff the results.

I'm opening this as a **draft** to share the direction early and get feedback on the approach before building out the higher levels. It's nowhere near finished.

## What's here so far (L1 + L2)

A new `pumpkin-test` crate with an in-memory harness:

- Boots a real `Server` on a temp dir with a fixed (signed) world seed, offline mode, overworld only.
- Attaches players backed by a real `JavaClient` over a loopback socket; clientbound frames are captured instead of going to the wire.
- Player actions go through the **real public serverbound handlers** (`handle_player_action`, `handle_use_item_on`, `handle_position`, creative slot set, …), not shortcuts.
- Deterministic ticking by calling the server tick directly (no `Ticker`, no sleep).
- Fast fixtures: an all-air chunk can be inserted instantly instead of paying for full world generation (~0.1s per test instead of ~13s), with `load_chunk` still there when real terrain is actually needed.
- A `#[pumpkin_test]` attribute macro that replaces `#[tokio::test]`, injects a `TestServer`, and takes an optional `seed`.

**L1 pilots (5):** break a stone block and check the cobblestone drop, place a block, move the player, set a creative slot, and a boot + tick smoke test.

**L2 (protocol conformance):** the harness can also drive packets as real bytes. `send_serverbound` encodes a packet and runs it through the real `handle_play_packet` decode + dispatch; `assert_clientbound_sent` re-serialises the expected clientbound packet and matches it against the captured frames. Clientbound capture is deterministic (it doesn't rely on task scheduling). Three L2 pilots so far: a ping/pong round-trip, a position packet that moves the player, and a player-action packet that breaks a block. A deserialize-only serverbound packet is made drivable by gating a `Serialize` derive behind the `test-harness` feature.

**Data-driven:** a few tests assert known vanilla facts against Pumpkin's compiled data as pure static queries — a shapeless recipe (oak log to four planks), tag memberships (planks tag, mineable-with-pickaxe), and a block loot table (stone to one cobblestone). The expected values are hand-written vanilla truth, so they act as an independent oracle rather than re-checking the data the tables were generated from. A fully independent runtime oracle (a real vanilla server's reports) is left for the L4 differential work.

**L3 (end-to-end):** a headless `TestBot` connects to a real server over a loopback TCP socket, reuses `pumpkin-protocol`'s framing, and completes the server-list status exchange (handshake → status → ping/pong). Unlike L1/L2, which stay in memory, this drives the full socket + protocol stack. The full login → play join is the next L3 step.

Everything the server exposes for this sits behind a `test-harness` cargo feature and doesn't change production behaviour. The harness crate pulls the feature in transitively, so the tests run under a plain `cargo nextest run` with no extra flags.

## A bug this shook out

While writing the block-break test I hit an infinite loop in the dynamic light engine: sky-light propagation into an unloaded chunk never converges (writes to an unloaded chunk are dropped, reads come back as 0, so the "brighter than the neighbour" check stays true forever). It's fixed on this branch and also sent as a focused standalone PR in #2334.

## Roadmap

- [x] L1 in-memory harness (`pumpkin-test`) with real-handler player actions
- [x] `#[pumpkin_test]` macro + `TestServer` fixture injection
- [x] Fast empty-chunk fixtures + deterministic ticking
- [x] Five L1 pilot tests, green in CI
- [x] Fix the sky-light propagation loop on unloaded chunks (also #2334)
- [x] L2 protocol conformance (serverbound bytes in, assert clientbound out)
- [x] Data-driven checks against known vanilla facts (recipes / tags / block loot)
- [x] L3 status exchange over real TCP (headless bot reusing `pumpkin-protocol`)
- [ ] L3 full login → play join (offline, no encryption)
- [ ] L4 differential testing against a real vanilla server
- [ ] More test categories + a short "how to write a test" guide
- [ ] Dedicated CI jobs for the heavier levels (nightly / label-gated)

## Notes

- No production code paths change; all test hooks are `#[cfg(feature = "test-harness")]`.
- L1/L2 are meant to stay fast and hermetic (no Docker, no network); L3/L4 will be opt-in.

Feedback on the overall shape is very welcome while the rest gets built out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2832,10 +2832,19 @@ dependencies = [
  "pumpkin-config",
  "pumpkin-data",
  "pumpkin-protocol",
+ "pumpkin-test-macros",
  "pumpkin-util",
  "pumpkin-world",
  "tempfile",
  "tokio",
+]
+
+[[package]]
+name = "pumpkin-test-macros"
+version = "0.1.0-dev+26.2"
+dependencies = [
+ "quote",
+ "syn",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2833,6 +2833,7 @@ dependencies = [
  "pumpkin-data",
  "pumpkin-protocol",
  "pumpkin-util",
+ "pumpkin-world",
  "tempfile",
  "tokio",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2823,6 +2823,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "pumpkin-test"
+version = "0.1.0-dev+26.2"
+dependencies = [
+ "arc-swap",
+ "bytes",
+ "pumpkin",
+ "pumpkin-config",
+ "pumpkin-data",
+ "pumpkin-protocol",
+ "pumpkin-util",
+ "tempfile",
+ "tokio",
+]
+
+[[package]]
 name = "pumpkin-util"
 version = "0.1.0-dev+26.2"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,8 @@ members = [
     "pumpkin-data",
     "pumpkin-plugin-api",
     "pumpkin-codecs",
-    "pumpkin-test"
+    "pumpkin-test",
+    "pumpkin-test-macros"
 ]
 exclude = ["pumpkin-codegen"]
 
@@ -159,6 +160,7 @@ pumpkin-nbt = { path = "pumpkin-nbt", default-features = false }
 pumpkin-protocol = { path = "pumpkin-protocol", default-features = false }
 pumpkin-util = { path = "pumpkin-util", default-features = false }
 pumpkin-world = { path = "pumpkin-world", default-features = false }
+pumpkin-test-macros = { path = "pumpkin-test-macros" }
 quote = { version = "1.0", default-features = false, features = ["proc-macro"] }
 rand = { version = "0.10.1", default-features = false, features = ["std", "std_rng", "thread_rng"] }
 rsa = { version = "=0.10.0-rc.18", default-features = false, features = ["std", "encoding"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,8 @@ members = [
     "pumpkin/",
     "pumpkin-data",
     "pumpkin-plugin-api",
-    "pumpkin-codecs"
+    "pumpkin-codecs",
+    "pumpkin-test"
 ]
 exclude = ["pumpkin-codegen"]
 
@@ -148,6 +149,7 @@ num-traits = { version = "0.2", default-features = false, features = ["std"] }
 p384 = { version = "0.13.1", default-features = false, features = ["std", "arithmetic", "pkcs8"] }
 phf = { version = "0.14.0", default-features = false, features = ["std"] }
 proc-macro2 = { version = "1.0", default-features = false, features = ["proc-macro"] }
+pumpkin = { path = "pumpkin", default-features = false }
 pumpkin-codecs = { path = "pumpkin-codecs", default-features = false }
 pumpkin-config = { path = "pumpkin-config", default-features = false }
 pumpkin-data = { path = "pumpkin-data", default-features = false }
--- a/pumpkin-protocol/Cargo.toml
+++ b/pumpkin-protocol/Cargo.toml
@@ -10,6 +10,8 @@ default = ["query"]
 serverbound = []
 clientbound = []
 query = []
+# Adds encode derives to serverbound packets so tests can send them as bytes. Off in normal builds.
+test-harness = []
 
 [dependencies]
 pumpkin-nbt.workspace = true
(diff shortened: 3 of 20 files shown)


JunkyDeveloper:
How do you write tests?
we (a group of 3) started this kind of project long time ago, we have both a bot which joins and a direct lib which uses the rust server.
So would be nice, to share how you want to write tests, we currently started writing our tests in json. and how do you validate your tests to vanilla? We have for that our bot, which is usable for all servers also vanilla to validate all tests to vanilla

TheDarkSword:
> How do you write tests? we (a group of 3) started this kind of project long time ago, we have both a bot which joins and a direct lib which uses the rust server. So would be nice, to share how you want to write tests, we currently started writing our tests in json. and how do you validate your tests to vanilla? We have for that our bot, which is usable for all servers also vanilla to validate all tests to vanilla

I was thinking using yaml file to write tests but json is good too.
If you want we can work together for creating a good testing framework for Pumpkin (or in general for the community) to check Vanilla parity and avoid regressions

JunkyDeveloper:
> I was thinking using yaml file to write tests but json is good too. If you want we can work together for creating a good testing framework for Pumpkin (or in general for the community) to check Vanilla parity and avoid regressions

I hate yaml, so we settled with json, but one had the idea to move to groovy I think, but we are still with json. Our framework already works for testing most blocks and items. More will come later and we have already a tool which checks vanilla parity, still a bit flaky but 1-2 persons are working on that problem currently. The idea is to also test entities later and add all missing functionality to achieve 100% block testing (like shelf to allow to select each slot, etc).
We already have I think around 30 tests, a recorded, visual editor, cicd, vanilla parity checker. Even a pumpkin channel on your discord :D

I hope it is ok, the framework is called Flint: https://github.com/FlintTestMC/FlintBenchmark (our tests repo, if you want to see)

TheDarkSword:
> I hate yaml, so we settled with json, but one had the idea to move to groovy I think, but we are still with json. Our framework already works for testing most blocks and items. More will come later and we have already a tool which checks vanilla parity, still a bit flaky but 1-2 persons are working on that problem currently. The idea is to also test entities later and add all missing functionality to achieve 100% block testing (like shelf to allow to select each slot, etc). We already have I think around 30 tests, a recorded, visual editor, cicd, vanilla parity checker. Even a pumpkin channel on your discord :D
> 
> I hope it is ok, the framework is called Flint: https://github.com/FlintTestMC/FlintBenchmark (our tests repo, if you want to see)

Sure, I'd be happy to help! Is there anything specific I can work on? What should I focus on?
Should I just open PRs when I have something ready, or would you prefer that we coordinate on Discord first?

JunkyDeveloper:
> Sure, I'd be happy to help! Is there anything specific I can work on? What should I focus on? Should I just open PRs when I have something ready, or would you prefer that we coordinate on Discord first?

I would say join maybe the discord first, because I'm on exam phase and not much overview what coco and Dumbass are currently planing to do, but you can also do only PRs if you want :)
And the discord I think it's easier to ask questions, if you don't understand something. It's fully up to you, you can also first look at the code base and see is that what you wanted to do or not :)

JunkyDeveloper:
> Sure, I'd be happy to help! Is there anything specific I can work on? What should I focus on? Should I just open PRs when I have something ready, or would you prefer that we coordinate on Discord first?

Forgot to answer your first question, I think the first thing is you need to decide if you want to use FlintCLI or do something like flint-steel for pumpkin aka flint-pumpkin (I designed flint-steel first for pumpkin, because steel was not advanced enough). Theoretically it should be not to hard possible to do that.

For you FlintCLI is a pure bot (using azalea), which joins a minecraft server and executes tests in parallel. flint-steel, sits on top of steel-core (library and uses internal functions of the player) to be faster and allow to debug, your block impl (that was my feature which I wanted to have). flint-steel uses an empty world gen, to have a pure empty world in ram storage to quickly execute all tests. Currently serial but parallel is planned (currently not needed 25k tests are done in 1min).

Means FlintCLI should work out of the box for pumpkin, means you only need to add all missing commands in pumpkin. if you want to do flint-pumpkin, more coding is needed

Q2297045667:
Since Pumpkin supports server-side functionality for both Java and Bedrock platforms, our goal is to align as closely as possible with the vanilla features of the Java Edition. In the event of any feature conflicts between the two platforms, Java Edition will take precedence. Therefore, your testing framework needs to account for the specific conditions of both platforms.

JunkyDeveloper:
> Since Pumpkin supports server-side functionality for both Java and Bedrock platforms, our goal is to align as closely as possible with the vanilla features of the Java Edition. In the event of any feature conflicts between the two platforms, Java Edition will take precedence. Therefore, your testing framework needs to account for the specific conditions of both platforms.

I don't really get what you mean, the goal of the framework is to make 100% vanilla parity and not only claim it but also proof it. So which specific conditions are you talking? The tests are written in json to allow to create a bedrock bot and make pure bedrock tests if needed.

PR #2338, opened 2026-07-02
perf(pumpkin-world): smaller pump chunk saves and lighter in-memory chunks

TheDarkSword:
Two related changes to how the default `pump` world format handles chunks — one cuts what we write to disk, the other cuts what we hold in memory. Neither changes a single byte of what a client or an existing save sees.

## 1. Compress chunk blobs with zlib instead of zstd

`pump` compresses every chunk blob with `ruzstd`, whose pure-Rust encoder only supports its weakest "Fastest" level. On chunk NBT that lands about 55% larger than plain zlib at level 6 (measured ~744 vs ~476 bytes on a freshly generated, untouched chunk, where the payload is mostly light data — roughly -36%).

Bumping the zstd level would mean pulling in a native encoder, so the blobs now go through zlib via `flate2`, already a workspace dependency and pure Rust through `miniz_oxide`. Level 6 is the sweet spot: level 9 saved next to nothing here for more CPU.

Old worlds keep loading — a blob is sniffed for the zstd frame magic on read and decoded with `ruzstd` when it's an old one; everything written from now on is zlib, so re-saving migrates a world over.

## 2. Keep uniform sky light compact in memory

Generation fills a full-bright 2048-byte sky-light array for every section above the terrain, and those arrays sit in memory for as long as the chunk is loaded — about 30 KB of identical bytes per chunk. Once lighting is computed (after generation and after a chunk is read from disk) those uniform sections collapse back into the existing `Empty(level)` representation.

This is lossless, run-length style: a uniform array carries a single value, which is exactly what `Empty(level)` already stores. Only genuinely uniform, non-zero sections are collapsed; a non-uniform section (e.g. the gradient right at the surface) stays a full array and is written verbatim. `Empty(0)` is left alone because it's the "no data" sentinel that both disk and the light packets intentionally drop.

### Why this changes nothing a client or a save sees

Every place that turns light back into raw arrays now goes through `LightContainer::to_full_bytes` / `has_light`, which re-expand a compacted section on demand. That covers all three consumers:

- the on-disk chunk format,
- the clientbound chunk-data packet (`CChunkData`), and
- the clientbound light-update packet (`CLightUpdate`), sent when a block change relights a chunk.

I went through every reader of `LightContainer` for this. The light engine, mob spawning, the daylight sensor and the plugin light getters all read via `get()`, which already returns the right level for a compacted section, so they're untouched. The only readers that inspected the `Full`/`Empty` variant directly were those two packets, and they now key off `has_light()` + `to_full_bytes()`.

### How it's verified

- `compaction_shrinks_memory_but_keeps_disk_bytes` — serialises a chunk with and without compaction, asserts the on-disk bytes are identical, then round-trips the light values.
- `compacted_section_reads_identically_to_full` — a compacted section and the full array it replaced report the same `has_light()` and the same `to_full_bytes()`.
- `compaction_keeps_chunk_data_packet_identical` / `compaction_keeps_light_update_packet_identical` — build a lit chunk both ways and assert the actual `CChunkData` / `CLightUpdate` bytes a client receives are byte-for-byte identical.

## Tests

- pump: `test_pump_file_roundtrip` (zlib path), `test_reads_legacy_zstd_blob` (old zstd still decodes).
- light: the four tests above.

## Scope

Only the `pump` format (the default) is touched for compression. `linear` uses the same `ruzstd`-Fastest encoder and could get the same change in a follow-up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2816,6 +2816,7 @@ dependencies = [
  "pumpkin-nbt",
  "pumpkin-util",
  "pumpkin-world",
+ "rustc-hash",
  "serde",
  "thiserror 2.0.18",
  "tokio",
--- a/pumpkin-protocol/Cargo.toml
+++ b/pumpkin-protocol/Cargo.toml
@@ -37,6 +37,7 @@ bitflags.workspace = true
 
 [dev-dependencies]
 # Validate correctness
+rustc-hash.workspace = true
 
 [lints]
 workspace = true
--- a/pumpkin-protocol/src/java/client/play/chunk_data.rs
+++ b/pumpkin-protocol/src/java/client/play/chunk_data.rs
@@ -245,13 +245,16 @@ impl ClientPacket for CChunkData<'_> {
             for section_index in 0..num_sections {
                 let bit_index = section_index + 1; // Offset by 1 for the below-world section
 
-                if let LightContainer::Full(_) = &light_engine.sky_light[section_index] {
+                // A compacted uniform section (`Empty(level != 0)`) still carries
+                // light and must be sent as an array, so key off `has_light()`
+                // rather than only matching `Full`.
+                if light_engine.sky_light[section_index].has_light() {
                     sky_light_mask |= 1 << bit_index;
                 } else {
                     sky_light_empty_mask |= 1 << bit_index;
                 }
 
-                if let LightContainer::Full(_) = &light_engine.block_light[section_index] {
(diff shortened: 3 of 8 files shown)


Sawlyer:
Two notes: please rebase, the branch conflicts; this makes re-saved worlds unreadable by older Pumpkin builds, worth a changelog line. Optionally a shared full-bright buffer would avoid a 2048-byte alloc per compacted section on every save.

PR #2339, opened 2026-07-02
Harden the chunk scheduler and fix its stranded-node leak

TheDarkSword:
Keeps long-running worlds from slowly leaking memory (and, in debug builds, crashing) by making the single chunk-generation scheduler thread crash-safe and stopping a generation-graph node leak.

## Background

A real crash report from a long session pointed straight here: the `Schedule` thread panicked in a debug node-count assertion (`nodes count error`) after accumulating dozens of "unreleased" generation graph nodes.

## 1. Don't let a scheduler panic silently freeze the world

Chunk loading, generation and unloading all run on one dedicated `Schedule` thread, spawned with no panic handling. If its work loop panicked the thread just vanished: chunk loads would await forever and — worse — unloads would stop, so `loaded_chunks` and the entity/block-entity maps keyed off it would grow without bound until the process ran out of memory.

The thread body is now wrapped in `catch_unwind`. On a panic it logs the cause loudly, flags the chunk system as shut down and wakes anyone blocked on it, so the failure is visible and the server stops waiting on a thread that is never coming back instead of leaking until it OOMs. (This is what turned the crash report above into a clean, logged shutdown.)

## 2. Don't leak stranded generation nodes

When a chunk is downgraded, or its generation fails, while one of its tasks is in flight, some graph nodes can be left dangling — never advanced, never released. Once the scheduler next goes fully idle they are provably stranded: nothing queued, in flight or waiting can ever touch them again.

- **Debug** builds still abort on this (it is a real desync worth catching), but now first dump every stranded node together with its holder's target/current/dependency stages so the crash report is actionable.
- **Release** builds no longer leak or go down: the stranded nodes are dropped and the holder references cleared, and chunks that are still wanted are rescheduled on the next ticket level change.

I could not pin down the exact interleaving that strands the nodes from static reading alone, so rather than risk a wrong change to core generation this fixes the symptom safely and adds the diagnostics above (plus a debug log when a downgrade keeps an in-flight node) to pinpoint the root cause from the next report.

## Not included here

The block-entity per-chunk ticking perf change that was originally on this branch has moved to its own PR: Pumpkin-MC/Pumpkin#2340. This branch stays focused on scheduler stability.

## Checks

`cargo fmt --check` and `cargo clippy --all-targets --all-features` (debug and release) are clean; the existing `pumpkin` / `pumpkin-world` suites pass.

### DIFF
--- a/pumpkin-world/src/chunk_system/schedule.rs
+++ b/pumpkin-world/src/chunk_system/schedule.rs
@@ -24,6 +24,16 @@ use std::thread;
 use std::time::Duration;
 use tracing::{debug, error, info, trace, warn};
 
+/// Best-effort extraction of a human-readable message from a caught panic
+/// payload (`catch_unwind` hands back a `Box<dyn Any>`).
+fn panic_reason(payload: &(dyn std::any::Any + Send)) -> String {
+    payload
+        .downcast_ref::<&str>()
+        .map(|s| (*s).to_string())
+        .or_else(|| payload.downcast_ref::<String>().cloned())
+        .unwrap_or_else(|| "unknown cause".to_string())
+}
+
 pub(crate) struct TaskHeapNode(i8, NodeKey);
 impl PartialEq for TaskHeapNode {
     fn eq(&self, other: &Self) -> bool {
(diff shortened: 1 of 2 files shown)


DanMcgraw:
[crash-2026-07-02_20.07.02.txt](https://github.com/user-attachments/files/29612415/crash-2026-07-02_20.07.02.txt)
This new error seems to be related to this PR, just tried playing on the pr for a little bit.

TheDarkSword:
> [crash-2026-07-02_20.07.02.txt](https://github.com/user-attachments/files/29612415/crash-2026-07-02_20.07.02.txt) This new error seems to be related to this PR, just tried playing on the pr for a little bit.

Thank you very much! This PR listen for panic of the Scheduler thread, this log will let me to fix the real issue!

TheDarkSword:
This PR is not completed yet, I added some logs to find the real cause of the problem. Once it is found i fix and the PR will be ready to merge

Snowiiii:
@TheDarkSword so should this be draft?

PR #2340, opened 2026-07-02
perf(pumpkin): tick block entities per active chunk instead of scanning all

TheDarkSword:
Block entities were stored in one flat `DashMap<BlockPos, _>`, and every world tick rebuilt the tick list by iterating the entire map and filtering it down to the active chunks — O(all loaded block entities). The more of the world that has been explored, the more chests/signs/hoppers/... are loaded, and the longer every tick spends walking block entities it will not tick, so TPS slowly bleeds away on long-running worlds.

The map is now keyed by chunk (`chunk -> its block entities`). Ticking walks only the active chunks and their entries; inserts, removals and lookups compute the chunk key up front; unloading a chunk drops its whole bucket. The same block entities get ticked — it just stops scanning the ones that are loaded but nowhere near a player.

### DIFF
--- a/pumpkin/src/world/mod.rs
+++ b/pumpkin/src/world/mod.rs
@@ -162,7 +162,7 @@ use weather::Weather;
 
 type FlowingFluidProperties = pumpkin_data::fluid::FlowingWaterLikeFluidProperties;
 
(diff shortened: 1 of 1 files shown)


Snowiiii:

PR #2342, opened 2026-07-02
fix(pumpkin): stop entities from duplicating on chunk reload

TheDarkSword:
### The bug

Every reconnect (more generally, every entity-chunk load→unload cycle) duplicated the world's entities — exponentially: 1 → 2 → 4 → 8.

### Root cause

An entity existed in two places at once: the live `World::entities` list and the serialized NBT in the entity chunk's `data`. On load the saved NBT was turned into live entities but the list was **never cleared**, and on unload each live entity was **appended** back onto that still-populated list, so the persisted count doubled every cycle. Freshly spawned entities hit the same trap — `add_entity_silent` serialized them into the chunk immediately, so they were live *and* serialized and doubled on the first unload too.

### Fix — make the live entity the single source of truth (as in vanilla)

- On load, the chunk's serialized entities are **taken** (cleared) as they become live, and their **persisted UUID is restored** so an entity keeps its identity across reloads (it was previously regenerated with `Uuid::new_v4()`, which is not how vanilla behaves).
- A second watcher of an already-loaded chunk is sent spawn packets built from the **live** entities, not the stale NBT.
- Entities are serialized — fresh, from their current live state — only when their chunk **unloads** (`save_entity`).
- `add_entity_silent` no longer serializes on spawn.

An entity is now live **xor** serialized at any moment, matching vanilla.

### Side benefit

Because the live entity is serialized fresh on unload, any change made to it while it is loaded (health, effects, item contents, …) is persisted automatically — there is no longer a stale NBT copy that has to be rewritten by hand or the change is lost.

### Out of scope (pre-existing, to be tracked separately)

Passenger/vehicle NBT nesting and periodic autosave of loaded-chunk entities remain non-vanilla and are left for a follow-up; some per-type `write_nbt` fields are still `TODO` (e.g. mob equipment). None of these are affected by this change.

### Checks

`cargo fmt --check`, `cargo clippy --all-targets --all-features` (debug and release), and the `pumpkin` test suite pass; a unit test covers the UUID read matching the write format.


pumpkin/src/world/mod.rs
```diff
@@ -5383,3 +5333,44 @@ impl WorldPortalExt for WorldPortal {
         natural_spawner::spawn_mobs_for_chunk_generation(&self.0, cache, biome, chunk_x, chunk_z);
     }
 }
+
+/// Reads an entity's persistent UUID from its saved NBT (the `UUID` int-array
+/// written by `Entity::write_nbt`), so a reloaded entity keeps its identity.
+fn read_entity_uuid(nbt: &NbtCompound) -> Option<Uuid> {
```
Snowiiii: This should be not here, rather in nbt crate

TheDarkSword: Done

TheDarkSword:
I think CI fail is github problem

Snowiiii:

PR #2348, opened 2026-07-04
fix: prevent capacity overflow crash

### LINKED ISSUE #2345: Server crashes when flying with elytra and in creative

### Operating System

Windows 11

0.1.0-dev+26.2 (Commit: 5d117b3/release)

### What happened?

I was flying using the elytra to test the speed of the generation when it crashed. As I flew into new chunks, the server encountered a capacity overflow panic during the "Features" generation stage and crashed. I also tried flying in creative with no elytra the same issue.

I checked the existing issue tracker and found only one other ticket mentioning a "Panic while flying fast" crash, but it involved a completely different error message, so this appears to be a unique issue.

```
[ERROR] Chunk generation FAILED at Vector2 { x: -44, y: -6 } (Features): capacity overflow
[ERROR] Pumpkin has encountered a panic!
[ERROR] Received generation failure notification for chunk Vector2 { x: -44, y: -6 } at stage Features: capacity overflow
[ERROR] 
[WARN] Chunk Vector2 { x: -44, y: -6 } reset to None and re-queued for regeneration (target: Features)
[ERROR] Thread 'Gen-Pool-0' with ThreadId(18) panicked at /rustc/31fca3adb283cc9dfd56b49cdee9a96eb9c96ffd/library\alloc\src\raw_vec\mod.rs:28:5:
[ERROR] capacity overflow
[ERROR] 
[ERROR] The full backtrace will be printed to the crash report.
[ERROR] Encountered panic while shutting down: capacity overflow
[ERROR] Chunk generation FAILED at Vector2 { x: -44, y: 1 } (Features): capacity overflow
[ERROR] Received generation failure notification for chunk Vector2 { x: -44, y: 1 } at stage Features: capacity overflow
```

here is the full crash report 
```

Time: 2026-07-03 22:24:04.0072092 +00:00:00
Message: capacity overflow
Panic Location: /rustc/31fca3adb283cc9dfd56b49cdee9a96eb9c96ffd/library\alloc\src\raw_vec\mod.rs:28:5

--- Panicking Thread ---
ID: ThreadId(18)
Name: Gen-Pool-0
Backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: <unknown>
  10: <unknown>
  11: <unknown>
  12: <unknown>
  13: <unknown>
  14: <unknown>
  15: <unknown>
  16: <unknown>
  17: <unknown>
  18: BaseThreadInitThunk
  19: RtlUserThreadStart

--- Server Details ---
Pumpkin Version: 0.1.0-dev+26.2 (Commit: 5d117b3/release)
Minecraft Version: 26.2
Server compiled with Rust 1.96.1

--- System Details ---
Operating System: Windows 11 Pro
Kernel: 26200
Physical Memory: 16087 MiB/31621 MiB used, 15534 MiB free
Swap Memory: 0 MiB/1984 MiB used, 1984 MiB free

Total cores: 16

| CPU #1     Cores: 16
|            Names: CPU 1, CPU 2, CPU 3, CPU 4, CPU 5, CPU 6, CPU 7, CPU 8, CPU 9, CPU 10, CPU 11, CPU 12, CPU 13, CPU 14, CPU 15, CPU 16
|            Brand: AMD Ryzen 7 7700X 8-Core Processor             
|            Average Frequency: 4501 MHz
|            Vendor ID: AuthenticAMD

```

### To Reproduce

I tested this multiple times to confirm the trigger. The crash happens consistently in a brand-new world with no pre-generated chunks. 

You can reproduce it either by flying in Creative mode while holding Shift to go faster, or by flying in Survival mode with an Elytra and spamming firework rockets to maximize your speed.

### Expected behavior

Flying at high speeds should not cause a server crash. The generation speed itself was fast enough, but it should not result in a panic.

like this:
Image

CompileRider:
This fixes a crash in chunk generation caused by a negative feature count being converted into an enormous `usize`, which then overflows allocation in `Vec` construction.

### Root cause

`NoiseBasedCountPlacementModifier::get_count()` can return a negative `i32` after noise sampling and scaling. That value was then used as a count without clamping, so when it reached allocation or iteration code it became an invalidly large size request.

In Java, the equivalent pattern is safe because a `for` loop whose condition is false from the start executes zero iterations. For example, `for (int i = 0; i < -5; i++)` does nothing because the condition is false immediately.

### Fix

Clamp negative counts to zero before using them:

```rust
// Before
let count = self.get_count(random, pos);

// After
let count = self.get_count(random, pos).max(0);
```

This preserves the intended behavior: when noise indicates no feature placements, the loop performs zero iterations instead of attempting an impossible allocation.

### Why this is correct

The Java version relies on control-flow semantics to produce zero iterations when the bound is negative or otherwise unreachable. In Rust, an unchecked cast from a negative integer to `usize` does not panic by itself, but it does produce a very large unsigned value, which is unsafe to use as a count.

Clamping to zero makes the behavior explicit and matches the vanilla outcome more safely.

### Verification

- Before: chunk generation could panic with `capacity overflow` during feature placement.

- After: negative counts are treated as zero, chunk generation completes normally

Closes #2345

### DIFF
--- a/pumpkin-world/src/generation/feature/placed_features.rs
+++ b/pumpkin-world/src/generation/feature/placed_features.rs
@@ -516,7 +516,7 @@ pub trait CountPlacementModifierBase {
         random: &mut RandomGenerator,
         pos: BlockPos,
     ) -> Box<dyn Iterator<Item = BlockPos>> {
-        let count = self.get_count(random, pos);
+        let count = self.get_count(random, pos).max(0);
         Box::new(std::iter::repeat_n(pos, count as usize))
     }
 
--- a/pumpkin-world/src/generation/feature/placed_features.rs
+++ b/pumpkin-world/src/generation/feature/placed_features.rs
@@ -516,7 +516,7 @@ pub trait CountPlacementModifierBase {
         random: &mut RandomGenerator,
         pos: BlockPos,
     ) -> Box<dyn Iterator<Item = BlockPos>> {
-        let count = self.get_count(random, pos).max(0);
+        let count = self.get_count(random, pos).max(0); // Vanilla treats negative counts as 0, but `i32 as usize` in Rust would wrap.
         Box::new(std::iter::repeat_n(pos, count as usize))
     }
 


TheAwiteb (commented):
I don't think `CountPlacementModifierBase::get_count` should return a negative number in the first place. The function's return type should be changed from `i32` to `usize`.

yunuservices:
LGTM. This should resolve #2350. Also as a minor suggestion, consider adding a short inline comment explaining that vanilla treats negative counts as zero iterations. Since rust "i32 as usize" behaviour differs from java...

CompileRider:
> I don't think `CountPlacementModifierBase::get_count` should return a negative number in the first place. The function's return type should be changed from `i32` to `usize`.

 Hi, I've seen your point on using usize. My concern is that a refactor that large would bring more problems than solutions right now, especially for a targeted bugfix. I'm choosing to stick with the minimal fix just to get the crash resolved quickly.

lineality:
Confirming that this fixes the panic-crash for issue 2350:  https://github.com/Pumpkin-MC/Pumpkin/issues/2350

Necko1:
+1

Snowiiii:
Mhh, yeah saw vanilla does not do max(0), well i guess let keep that until we found a better way, Thank you @CompileRider :D

PR #2351, opened 2026-07-06
Implement chunk serialization API for plugins

vyPal:
## Description
Implementation of https://github.com/Pumpkin-MC/pumpkin-plugin-wit/pull/14, adds a new API to plugins which allows plugins to request a specific chunk by calling `world.get_serialized_chunk(x,z)`, which gets serialized to a custom flat format and is returned to the plugin. The plugin can then read and write to this copy of the chunk without needing to make any calls to the host. Once the plugin is done, it can send the modified serialized data back using `world.set_serialized_chunk(chunk_data)`, which the server will deserialize and create a new chunk from, this chunk then gets placed into the world and clients should receive a packet about the chunk being updated.

This API is only faster when the plugin is reading or modifying large amounts of blocks in a chunk (like setting the whole block to air, or copying a structure into several chunks), that's why it is not replacing the current chunk API. I am adding a screenshot of a small benchmark I did to show how fast this new API can be:
image

## Testing

### DIFF
--- a/pumpkin-plugin-api/src/chunk.rs
+++ b/pumpkin-plugin-api/src/chunk.rs
@@ -1,0 +1,205 @@
+use crate::world;
+
+/// A block entity stored as raw NBT bytes together with its absolute position.
+pub struct BlockEntityData {
+    pub x: i32,
+    pub y: i32,
+    pub z: i32,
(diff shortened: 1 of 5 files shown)


vyPal:
Closing in favor of #2919

PR #2353, opened 2026-07-07
feat(entity): pillagers and piglins fire crossbows 🤖🤖🤖

dongzh1:
## What was changed?

Adds `CrossbowAttackGoal`, a ranged crossbow behavior, and gives it to **pillagers and piglins** so they now **keep their distance and shoot arrows** instead of registering a `MeleeAttackGoal` (both had a `// for now we give them melee` note).

- **New `pumpkin/src/entity/ai/goal/crossbow_attack.rs`** — loosely mirrors vanilla `CrossbowAttackGoal`:
  - closes in until the target is within range,
  - holds position, backing off directly away from the target when it gets too close (< ¼ range),
  - charges the crossbow for a fixed time, then fires an arrow and reloads (20–40 tick delay),
  - vanilla ballistics `velocity(dx, dy + horizontal * 0.2, dz, 1.6, 14 - difficulty * 4)`, aiming ~⅓ up the target's body,
  - plays the crossbow load-start and shoot sounds.
- **Pillager** and **Piglin** swap their `MeleeAttackGoal` for `CrossbowAttackGoal`.

Pillagers and piglins are crossbow users, but they registered a `MeleeAttackGoal`, so they ran up to the player and hit in melee. This restores ranged crossbow behavior.

## Impact

Pillagers and piglins now fire arrows at their targets from range, close in when too far and back off when the player gets too close. No other mob changes.

## Known limitations

- Kiting uses navigation only (approach / hold / retreat) rather than lateral strafing, so this stays independent of any other AI change and touches no shared movement code.
- Rendering a held crossbow and the charged-crossbow pose depend on mob equipment, which no mob currently equips — orthogonal and left as a follow-up. The AI (approach/retreat, charging, firing, difficulty scaling, sounds) is complete.
- Part of a small series filling in ranged-attack AI (skeleton bows landed separately; snow golem snowballs and witch potions can follow).

## Testing

`cargo clippy -p pumpkin --all-targets` passes with no warnings. The change is additive and isolated to pillager/piglin AI; in-game verification of the shooting behavior is recommended.

### DIFF
--- a/pumpkin/src/entity/ai/goal/crossbow_attack.rs
+++ b/pumpkin/src/entity/ai/goal/crossbow_attack.rs
@@ -1,0 +1,209 @@
+use super::{Controls, Goal, GoalFuture};
+use crate::entity::ai::pathfinder::NavigatorGoal;
+use crate::entity::mob::Mob;
+use crate::entity::projectile::arrow::{ArrowEntity, ArrowPickup};
+use crate::entity::{Entity, EntityBase};
+use pumpkin_data::entity::EntityType;
+use pumpkin_data::sound::{Sound, SoundCategory};
+use pumpkin_util::math::vector3::Vector3;
+use rand::RngExt;
+use std::sync::Arc;
+
+/// Ranged crossbow-attack behavior for pillagers and piglins.
+///
+/// Loosely mirrors vanilla `CrossbowAttackGoal`: the mob closes in until the
+/// target is in range, holds position (backing off when the target gets too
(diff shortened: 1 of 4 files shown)


Sawlyer:
One note: this makes every piglin a crossbow user while vanilla piglins can be sword-wielders, acceptable interim, worth a TODO. See the parallel RangedCrossbowAttackGoal work with the vanilla CrossbowState machine to avoid duplicated effort.

RoosterBooster007:

PR #2354, opened 2026-07-07
feat(entity): snow golems throw snowballs at their target 🤖🤖🤖

dongzh1:
## What was changed?

Adds `SnowballAttackGoal` and gives it to the **snow golem** so it actually defends itself by throwing snowballs, instead of only wandering (it had a `// TODO: SnowballAttackGoal` and no attack goal at all).

- **New `pumpkin/src/entity/ai/goal/snowball_attack.rs`** — vanilla `ProjectileAttackGoal` style:
  - approaches until the target is within range, then holds position,
  - lobs a snowball every 20 ticks with vanilla ballistics: aim at `eyeY - 1.1`, `velocity(dx, dy + horizontal * 0.2, dz, 1.6, 12.0)`,
  - plays the snow golem shoot sound.
- **Snow golem** registers the goal at priority 1.

Snow golems are meant to pelt hostile mobs with snowballs (knocking them back, and damaging blazes). Without an attack goal they were passive wanderers.

## Impact

Snow golems now throw snowballs at their target from range and close the distance when too far. No other mob changes.

## Known limitations

- Part of a small series filling in ranged-attack AI (skeleton bows and pillager/piglin crossbows landed separately; witch potions can follow).

## Testing

`cargo clippy -p pumpkin --all-targets` passes with no warnings. The change is additive and isolated to the snow golem AI; in-game verification of the throwing behavior is recommended.

### DIFF
--- a/pumpkin/src/entity/ai/goal/mod.rs
+++ b/pumpkin/src/entity/ai/goal/mod.rs
@@ -24,6 +24,7 @@ pub mod owner_hurt_target;
 pub mod pick_up_block;
 pub mod place_block;
 pub mod revenge;
+pub mod snowball_attack;
 pub mod step_and_destroy_block;
 pub mod swim;
 pub mod teleport_towards_player;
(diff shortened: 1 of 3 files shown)


Sawlyer:
Only nit: navigator.set_progress every tick while approaching, consider throttling.

RoosterBooster007:

PR #2355, opened 2026-07-07
feat(entity): witches throw splash potions of Harming 🤖🤖🤖

dongzh1:
## What was changed?

Adds `WitchAttackGoal` and gives it to the **witch** so it actually attacks by throwing splash potions, instead of only wandering (it had a `// TODO: WitchAttackGoal (potions)` and no attack goal).

- **New `pumpkin/src/entity/ai/goal/witch_attack.rs`** — mirrors vanilla `WitchEntity::shootAt`:
  - approaches until the target is within range, then holds position,
  - throws a **splash potion of Harming** every 60 ticks, leading the target by its velocity, aiming at `eyeY - 1.1`, with vanilla ballistics `velocity(dx, dy + horizontal * 0.2, dz, 0.75, 8.0)`,
  - plays the witch throw sound. The existing `SplashPotionEntity` applies the potion effects (an AoE splash) on impact.
- **Witch** registers the goal at priority 2; the wander goal moves to priority 3 to match vanilla and avoid a `MOVE`-control conflict with the attack goal.

Witches attack by lobbing harmful splash potions, but the witch registered no attack goal, so it just wandered and never fought back.

## Impact

Witches now throw splash potions of Harming at their target from range and close the distance when too far. No other mob changes.

## Known limitations

- Only Harming is thrown for now. Vanilla also picks Slowness (target far), Poison (healthy target) or Weakness (target close); dynamic selection is left as a follow-up.
- Self-buff drinking (`isDrinking`) and witch potion immunity aren't modelled here.
- Final of a small series filling in ranged-attack AI (skeleton bows, pillager/piglin crossbows and snow golem snowballs landed separately).

## Testing

`cargo clippy -p pumpkin --all-targets` passes with no warnings. The change is additive and isolated to the witch AI; in-game verification of the throwing behavior is recommended.

### DIFF
--- a/pumpkin/src/entity/ai/goal/mod.rs
+++ b/pumpkin/src/entity/ai/goal/mod.rs
@@ -30,6 +30,7 @@ pub mod teleport_towards_player;
 pub mod tempt;
 pub(crate) mod track_target;
 pub mod wander_around;
+pub mod witch_attack;
 pub mod zombie_attack;
 
 #[must_use]
--- a/pumpkin/src/entity/ai/goal/witch_attack.rs
+++ b/pumpkin/src/entity/ai/goal/witch_attack.rs
@@ -1,0 +1,184 @@
+use super::{Controls, Goal, GoalFuture};
+use crate::entity::ai::pathfinder::NavigatorGoal;
+use crate::entity::mob::Mob;
+use crate::entity::projectile::splash_potion::SplashPotionEntity;
+use crate::entity::{Entity, EntityBase};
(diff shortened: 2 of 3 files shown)


Sawlyer:
Main gap is the fixed Harming payload, vanilla's Slowness/Poison/Weakness selection actually makes witches less lethal in several cases, so a selection table would be a good follow-up. Also WanderAroundGoal and LookAtEntityGoal now share priority 3; harmless, but worth tidying.

RoosterBooster007:

PR #2356, opened 2026-07-07
feat(plugin-api): expose missing plugin events 🤖🤖🤖

JinxedUp:
## Description
This PR exposes several existing server events to the plugin API and WASM plugin runtime, making them available for plugin developers without changing the underlying server behavior.

Newly exposed events
InventoryCloseEvent
InventoryClickEvent
PlayerInteractEntityEvent
ChunkSendEvent
BedrockFormResponseEvent
CustomClickActionEvent
Additional changes
Added the missing guest API event definitions and exports.
Registered the events with the WASM host event router.
Added the required WIT event definitions and Rust/WIT conversions.
Fixed the WASM packet event cancellation bridge so packet events marked as cancellable correctly propagate the cancellation result instead of panicking.

## Testing
cargo fmt
cargo check -p pumpkin-plugin-api
cargo check -p pumpkin
cargo test -p pumpkin-plugin-api
cargo test -p pumpkin topological_sort

All of the above completed successfully.

## Depends on: Pumpkin-MC/pumpkin-plugin-wit#15


Sawlyer:
I verified all six internal events already exist on master and that the panicking packet from_wasm stubs really do crash any blocking WASM packet handler today. Two things to document: a plugin-modified payload is accepted by the bridge but the network layer still dispatches on the original bytes, so only cancellation takes effect, the WIT surface implies otherwise; from_wasm for PacketSentEvent swaps the packet field for a private RawPacket unit, so any later native handler that downcasts packet silently fails. Both fine as documented limitations. Needs a rebase; expect conflicts with #2403/#2375/#2346. Please drop the emoji from the merge title.

Snowiiii:
Hey, Most of these were recently implemented

JinxedUp:
yeah because the PR was here for more than a month, smh im done with this project

PR #2358, opened 2026-07-07
fix: make SNBT operation `bool()` return `true` for all non-zero values

Laptop59:
## Description

Fixes the `bool` SNBT operation to return `true` for all non-zero values after Mojang has fixed the bug described in [MC-307774](https://bugs.mojang.com/browse/MC/issues/MC-307774).

## Testing

Edited a test to make it succeed again for the one test case affected by the bug, which confirms that it works as intended.


Snowiiii:
LGTM, Thank you @Laptop59 :D

PR #2360, opened 2026-07-07
perf(entity): swap std hashmap for fxhashmap in pathfind

nobuildersnotools:
## Description

Swaps out the usage of std::collections::HashMap to rustc_hash::Fx* in the pathfind module for faster non-cryptographic hashing.

## Testing

I measured a speedup of ~2.8x on average for 64 entries, inserting and looking up Vector3<i32> keys. Existing tests pass. Additionally, no deviation from the preexisting entity behavior was observed ingame.


Snowiiii:
LGTM, Thank you @nobuildersnotools :D

PR #2369, opened 2026-07-09
fix(data): remove redundant references flagged by newer clippy in panic! args 🤖🤖🤖

dongzh1:
## Summary

A newer stable Rust/clippy release (1.97) newly flags `clippy::useless_borrows_in_formatting` at 123 call sites in
`pumpkin-data/src/generated/block.rs` and `fluid.rs`: `panic!("...", &block.name)` / `panic!("...", &fluid.name)`
borrow a field that's already passed by reference to the formatting machinery, so the `&` is redundant.

This isn't specific to any one PR — because CI's `rust-toolchain.toml` tracks `channel = "stable"` and the runner
always installs the current latest stable, **every open PR that touches `pumpkin-data` (directly or transitively)
is currently failing `Run lints (debug)` / `Run lints (release)`** with the exact same 123 errors, regardless of
what the PR itself changes. I found this while chasing down what looked like CI failures on my own PRs and traced
it to this pre-existing latent issue in checked-in generated code that only a newer toolchain surfaces.

## What changed

Purely mechanical: for each of the 123 sites clippy flagged, removed exactly the `&` it pointed at. Nothing else
touched. No behavior change — `panic!`/`format!` already format the value through a reference internally, so
`&block.name` and `block.name` produce identical output.

## Testing

Applied precisely at the (file, line, column) clippy itself reported, with a script-level assertion that the
character at each position was actually `&` before removing it (verified no mismatches — all 123 applied cleanly).
Diff reviewed by hand: every changed line is a `panic!` argument list, nothing outside that context was touched,
and brace/paren balance across both files is unchanged. I was not able to run a local `cargo test`/`clippy` pass
against Rust 1.97 due to a local toolchain install issue on my end (unrelated to this repo) — CI will be the
first real verification of this diff; happy to iterate if it surfaces anything.

### DIFF
--- a/pumpkin-data/src/generated/fluid.rs
+++ b/pumpkin-data/src/generated/fluid.rs
@@ -902,7 +902,7 @@ impl FluidProperties for WaterLikeFluidProperties {
         if !["water", "lava"].contains(&fluid.name) {
             panic!(
                 "{} is not a valid fluid for {}",
-                &fluid.name, "WaterLikeFluidProperties"
+                fluid.name, "WaterLikeFluidProperties"
             );
         }
         let prop_index = self.to_index();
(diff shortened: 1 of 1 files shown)


dongzh1:
Rebased onto current master and found this branch now has zero diff against master — an equivalent fix for these panic! format args already landed upstream separately. Closing since there's nothing left to merge.

PR #2346, opened 2026-07-04
feat(plugin): add player drop item event

NicDevTV:
## Summary
- add `PlayerDropItemEvent` to the Rust plugin API
- wire the new event through the WASM event registration and conversion layer
- fire the cancellable event before held-item drop inventory mutation

## Depends on
- Pumpkin-MC/pumpkin-plugin-wit#13

## Testing
- `cargo fmt --check`
- `cargo check -p pumpkin-plugin-api -p pumpkin`
- `cargo clippy -p pumpkin-plugin-api -p pumpkin --all-targets -- -D warnings` fails on existing `pumpkin-data/src/data_component_impl.rs` `HashMap` disallowed-type warnings

PR #2375, opened 2026-07-10
feat(event): implement PluginLoadEvent

yunuservices:
Implements PluginLoadEvent, which fires after a plugin has been successfully loaded and initialized.
# Changes
- Added internal PluginLoadEvent in pumpkin/src/plugin/api/events/server/.
- Added WASM plugin API binding in pumpkin-plugin-api/src/events/server/.
- Added WASM host conversion in pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/events/server.rs.
- Added event registration routing in pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/context.rs.
- Fired the event in PluginManager::spawn_plugin_initialization after on_load succeeds and the plugin is marked active.
# Notes
- Non-cancellable: This is a post-load notification event. The plugin is already active when it fires.
- Metadata: Exposes the loaded plugin's name and version.
- Timing: Fired after on_load completes and is_active is set to true.

Pumpkin-MC/pumpkin-plugin-wit#16 have to merged first to merging this PR.


Sawlyer:
Three points: permission-gating event registration is a new pattern (no other EventType requires one, and a missing permission traps the plugin rather than failing gracefully), I'd like maintainer consensus before it spreads; get_loaded_plugins/PluginInfo/plugin.info are beyond the title, split or retitle; get_active_plugins is added but never exposed to WASM. Needs a rebase; the register-event match will conflict with #2356 and #2346.

crates/pumpkin/src/plugin/permissions.rs
```diff
@@ -111,6 +117,8 @@ pub fn get_permission_description(permission: &str) -> Option<&'static str> {
         SYS_INFO_CPU => Some("Allows the plugin to read CPU information."),
         SYS_INFO_RAM => Some("Allows the plugin to read RAM information."),
         SYS_INFO_OS => Some("Allows the plugin to read OS information."),
+        EVENTS_PLUGIN_LOAD => Some("Allows the plugin to receive plugin load events."),
```
QuantumSegfault, line 113: Not sure we need two separate permissions. I think one that allows both would be good enough.

There seems little value in allowing plugins to be able to inspect plugins that loaded before them, but not after, or vice versa; just allow viewing loaded plugins, period.

pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/context.rs
```diff
@@ -314,6 +321,15 @@ impl pumpkin::plugin::context::HostContext for PluginHostState {
         event_priority: EventPriority,
         blocking: bool,
     ) -> wasmtime::Result<()> {
+        let has_perm = |p: &str| self.permissions.iter().any(|perm| perm == p);
+
+        // Guard sensitive lifecycle events behind explicit permissions.
+        if event_type == EventType::PluginLoadEvent && !has_perm(permissions::EVENTS_PLUGIN_LOAD) {
+            return Err(wasmtime::Error::msg(
```
QuantumSegfault: This error here will crash the plugin. Seems like it'd be more appropriate to change the signature of `register_event` to return Result so plugins can recover from the error, now that it is fallible.

pumpkin/src/plugin/api/context.rs
```diff
@@ -74,6 +74,18 @@ impl Context {
...
 
+    /// Returns metadata for every plugin that is currently loaded.
+    #[must_use]
+    pub async fn get_loaded_plugins(&self) -> Vec<PluginMetadata> {
+        self.plugin_manager.loaded_plugins().await
+    }
+
+    /// Returns metadata for every plugin that has been activated.
+    #[must_use]
+    pub async fn get_active_plugins(&self) -> Vec<PluginMetadata> {
```
QuantumSegfault: Why does the native API get access to this but not the Wasm API

Also, what's the difference between the two?

QuantumSegfault (changes requested):

crates/pumpkin/src/plugin/permissions.rs
yunuservices: Hmmm, i will look on this. thanks for reminding

QuantumSegfault (approved):
Looks good now.

Though now there's no more ability to see which plugins have been loaded up this point? I think that's important in order for this feature as a whole to be complete (otherwise you can't know what plugins have loaded before you, only after).

Perfectly fine as a separate PR (and probably belongs as one). The permission will probably be renamed at that point, but it's accurate enough for now.

vyPal:
I actually don't think that the change to make the `register_event_handler()` method return a result is necessary, since if the permission is denied at startup, the plugin should probably just not load like when any other permission is denied

Snowiiii:
@yunuservices 
We already have Bukkit's `PluginEnableEvent`, is this different?

yunuservices:
important bit is the trigger: "PluginEnableEvent" is opt-in/manual "enable_plugin()", while "PluginLoadEvent" is automatic (post-"on_load"). Since there's no "on_enable" hook, "load" is the phase that actually happens automatically, so that's the event plugins can rely on.

yunuservices:
Commited stuffs inside PR already added into Pumpkin, after fc063c4dff40f581eec331253f87e72eb4710eb6 and fc063c4dff40f581eec331253f87e72eb4710eb6. Closing the PR, its been a while.

PR #2380, opened 2026-07-10
Fix shulker box losing contents when broken and not dropping in creative

### LINKED ISSUE #1569: Shulker Box doesn't retain contents when broken.

### Operating System

macOS Sonoma

0.1.0-dev+1.21.11

### What happened?

Shulker boxes doesn't retain it's content after being broken. It also doesn't drop in creative mode like it does in vanilla.

### To Reproduce

- Get a shulker box
- Open it
- Put some things into it
- Close it
- Break in survival and collect drop
- Place down again then open to see nothing inside

### Expected behavior

Shulker box drops in creative, and retains contents after being broken then placed.

dongzh1:
## Description

Shulker boxes lost their inventory when broken and placed again, and unlike vanilla they didn't drop at all when broken in creative mode with items still inside.

**What was changed**
- Implemented `write_data`/`get_hash` for `ContainerImpl` (the `minecraft:container` data component) so it actually round-trips through NBT instead of silently falling back to the trait defaults (`NbtTag::End` / `todo!()`).
- Added network (de)serialization for `DataComponent::Container` in `pumpkin-protocol`, mirroring vanilla's format: a positional item list padded with empty slots up to the highest occupied slot.
- `World::break_block` now snapshots a shulker box's inventory (tagged via `minecraft:shulker_boxes`) before its block entity is torn down, and forces a drop in creative mode when that snapshot is non-empty (vanilla always drops a shulker box that has items, even in creative).
- Implemented the previously unhandled `CopyComponents { source: "block_entity", include: [...] }` loot function for the `minecraft:container` include, attaching the snapshotted items onto the dropped item stack as its `Container` component. Other `block_entity` includes still log a warning and are left as follow-up work, same as before.
- Threaded the placed `ItemStack` through `BlockRegistry::place_block` / `player_placed` (both Java and Bedrock placement paths) so `ShulkerBoxBlock::player_placed` can read the `Container` component off the just-placed item and restore its contents into the freshly created block entity's inventory.

**Why these changes were necessary**

The `minecraft:container` component existed in name (struct + NBT read) but was never written back out and had no packet codec, so any contents attached to a dropped shulker box item were silently dropped on every serialization boundary (saving to disk, sending to a client). Separately, nothing in the break/place path ever populated or consumed that component in the first place, and creative-mode breaking used the same no-drop path as every other block instead of vanilla's shulker-box-specific override.

**Impact**

Breaking a shulker box with items now keeps those items on the dropped item (visible in the item's tooltip/model like vanilla) and restores them to the block entity's inventory when the item is placed down again, for both Java and Bedrock clients. Breaking a non-empty shulker box in creative mode now drops it instead of destroying its contents.

**Known limitations**
- Only the `minecraft:container` include of `CopyComponents { source: "block_entity" }` is implemented; any other include from that source still just logs a warning (pre-existing behavior, unchanged for those cases).
- This only covers the shulker box family (blocks tagged `minecraft:shulker_boxes`); no other block currently relies on this data path in vanilla.

## Testing

- `cargo clippy --all-targets --all-features --jobs 2 -- -D warnings` — clean, no warnings.
- `cargo test --jobs 2` — full suite passes (0 failures).
- Manually reasoned through the break → drop → pick up → place → open flow for both the survival and creative-mode cases against vanilla's documented behavior; I don't have a way to attach an in-game recording in this environment, so I'd appreciate an in-game sanity check from a reviewer before merge.

Fixes #1569

### DIFF
--- a/pumpkin-data/src/data_component_impl.rs
+++ b/pumpkin-data/src/data_component_impl.rs
@@ -2513,6 +2513,30 @@ impl ContainerImpl {
     }
 }
 impl DataComponentImpl for ContainerImpl {
+    fn write_data(&self) -> NbtTag {
+        let mut list = Vec::with_capacity(self.items.len());
+        for (slot, stack) in &self.items {
+            let mut compound = NbtCompound::new();
+            compound.put_int("slot", i32::from(*slot));
+            let mut item_compound = NbtCompound::new();
+            stack.write_item_stack(&mut item_compound);
+            compound.put_compound("item", item_compound);
+            list.push(NbtTag::Compound(compound));
+        }
+        NbtTag::List(list)
+    }
+
+    fn get_hash(&self) -> i32 {
+        let mut digest = Digest::new(Crc32Iscsi);
+        digest.update(&[Container.to_id()]);
+        for (slot, stack) in &self.items {
+            digest.update(&[*slot]);
+            digest.update(&get_str_hash(stack.item.registry_key).to_le_bytes());
+            digest.update(&get_i32_hash(i32::from(stack.item_count)).to_le_bytes());
+        }
+        digest.finalize() as i32
+    }
+
     default_impl!(Container);
 }
 #[derive(Clone, Debug)]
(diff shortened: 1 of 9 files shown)


RoosterBooster007:
I'm closing this PR for now because it appears to be largely AI-generated with no in-game testing. Please test and review your code again before reopening.

Issue #2372, opened 2026-07-10
Fall damage occurs when switching between Survival and Spectator


impresson:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

MacOS Tahoe 26.5

### Server Software Version/Commit

7ccb0c2

### What happened?

When you fall from any height and if you switch from survival to spectator anywhere before you land causes you to take the full fall damage when switching back to survival and land on the ground.

### To Reproduce

1. Find a height to fall from.
2. Switch to spectator before you land.
3. Switch back to survival.
4. You take fall damage.

### Expected behavior

The fall damage should be cancelled as in vanilla minecraft.

PR #2382, opened 2026-07-10
fix(entity): reset fall distance on Spectator/Creative gamemode switch

CompileRider:
## Summary
Fixes fall damage when switching between Survival and Spectator modes during a fall (#2372).

## Problem
When a player falls from a height, switches to Spectator mode before landing, and then switches back to Survival mode, they take full fall damage and die. Original Minecraft does not apply fall damage in this case.

## Root Cause
Pumpkin lacks the multiple layers of defense that the original version uses to prevent fall damage for Spectator/Creative players:
1. **No Physics**: Spectators ignore physics in `Entity.move()`.

2. **Reset on Tick**: `Player.aiStep()` resets `fallDistance` on every tick when `abilities.flying=true`.

3. **Force on Ground**: `Player.tick()` forces `onGround=false` for Spectators. 

4. **Mayfly Verification**: `Player.causeFallDamage()` returns false when `abilities.mayfly=true`.

## Workaround
Reset `fall_distance` to 0 when switching to Spectator or Creative mode. This is a temporary solution to compensate for the missing layers until the full original version is implemented. 

## Testing

**Before (bug):**

https://github.com/user-attachments/assets/c3da30ae-ff0a-44d7-8e51-dbec2a06c706

**After (fix):**

https://github.com/user-attachments/assets/4a435b7b-f4cc-465b-871c-48ada8d67b42

## Notes
This fix addresses the immediate issue. The pending tasks in the code indicate remaining deficiencies for a full and faithful implementation of the original fall damage prevention system.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -2935,6 +2935,26 @@ impl Player {
                     }
                 }
 
+                // Reset fall distance when switching to Spectator or Creative to prevent
+                // fall damage upon switching back to Survival.
+                //
+                // This is a pragmatic workaround: vanilla does NOT reset fallDistance in
+                // setGameMode(). Instead, vanilla relies on multiple defense layers:
+                //   1. noPhysics=true for Spectator — Entity.move() skips checkFallDamage()
+                //   2. abilities.flying=true → Player.aiStep() resets fallDistance every tick
+                //   3. setOnGround(false) forced for Spectators — prevents fallOn() invocation
+                //   4. abilities.mayfly=true → Player.causeFallDamage() returns false
+                //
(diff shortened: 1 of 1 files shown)


Sawlyer:
Good. One note: vanilla keeps fallDistance for a non-flying Creative faller who switches back to Survival mid-fall, so this over-forgives that edge, fine given the TODOs. Suggest filing them as a tracking issue.

RoosterBooster007:
Looks good, but maybe cut down on the length of the comment?

CompileRider:
> Looks good, but maybe cut down on the length of the comment?

That seems fine to me

RoosterBooster007:
Thanks, @CompileRider!

PR #2383, opened 2026-07-11
fix: Ender Dragon collision knockback/damage on every tick of overlap

### LINKED ISSUE #2220: Ender Dragon collision bug

### Operating System

Windows 11

0.1.0-dev+26.1/463c25d

### What happened?

When the player is located "inside" the Ender Dragon the player gets both kicked around and movement stuck while the Ender Dragon takes damage.

### To Reproduce

1. Summon and Ender Dragon
2. Fly "inside" the Ender Dragon in creative mode or jump at him in survival (preferredly with a high level of regeneration)
3. Watch the show

### Expected behavior

When invulnerable (creative mode): Do not take collision knockback
When vulnerable (survival & adventure mode): Take collision knockback once on contact

dongzh1:
## What

Fixes the Ender Dragon's player-collision handling in `EnderDragonEntity::handle_player_collisions`:

- Players who are invulnerable to `DamageType::MOB_ATTACK` (e.g. creative/spectator mode) are no longer knocked back when standing inside the dragon's hitbox. Vanilla only ever applies the collision knockback together with a successful hit, and an invulnerable target can never be hit, so it should never be pushed around either.
- For vulnerable players, the knockback + damage is now applied once per contact (on the tick the player's hitbox starts intersecting the dragon's), instead of being re-applied every single tick the player remains overlapping the dragon. A new `colliding_players: Mutex<HashSet<Uuid>>` field on `EnderDragonEntity` tracks who was already colliding on the previous tick so repeat overlap in the following ticks is skipped until the player separates and re-enters.

## Why

Fixes #2220. As reported: standing "inside" the Ender Dragon (flying into it in creative, or jumping into it in survival) caused the player to be flung around and damaged every tick for as long as the overlap lasted, and this happened even in creative mode where the player can't actually take damage. Expected behavior (from the issue):

- Invulnerable (creative): no collision knockback at all.
- Vulnerable (survival/adventure): knockback once on contact, not continuously while overlapping.

## Impact

- Behavior change is scoped to `EnderDragonEntity::handle_player_collisions`; no protocol, API, or config changes.
- Adds one small `Mutex<HashSet<Uuid>>` field to `EnderDragonEntity`, cleared/rebuilt each tick from the current set of colliding players, so it can't grow unbounded or leak entries for players who log off/despawn.
- Sitting-phase behavior (no damage while sitting) is unchanged; only the repeated-hit and invulnerable-knockback behavior is fixed.

## Limitations

- This only addresses the dragon's own body-collision handling. It does not touch dragon breath, other AoE damage sources, or any other mob's collision logic.
- "Once per contact" is tracked per-tick via hitbox intersection; a player who is knocked back but immediately re-enters the hitbox within the same continuous overlap will not be hit again until they fully separate first (matching the vanilla-like "hit on contact" behavior described in the issue).

### DIFF
--- a/pumpkin/src/entity/boss/ender_dragon.rs
+++ b/pumpkin/src/entity/boss/ender_dragon.rs
@@ -233,6 +233,12 @@ pub struct EnderDragonEntity {
     pub ticks_sitting: Mutex<i32>,
     pub sit_attack_timer: Mutex<i32>,
     pub breathing_timer: Mutex<i32>,
+
+    /// Players currently intersecting the dragon's hitbox that have already
+    /// received a collision knockback/hit for the current contact. Used so
+    /// `handle_player_collisions` only hits a player once per contact instead
+    /// of every tick they remain inside the dragon.
+    pub colliding_players: Mutex<std::collections::HashSet<uuid::Uuid>>,
 }
 
 impl EnderDragonEntity {
@@ -284,6 +290,7 @@ impl EnderDragonEntity {
             ticks_sitting: Mutex::new(0),
             sit_attack_timer: Mutex::new(0),
             breathing_timer: Mutex::new(0),
+            colliding_players: Mutex::new(std::collections::HashSet::new()),
         })
     }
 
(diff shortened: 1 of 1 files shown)


Sawlyer:
One design question: vanilla applies the dragon's body knockback every tick of overlap and relies on hurt-invuln frames to limit repeat damage, so 'once per continuous contact' under-hits. Since LivingEntity has hurt_cooldown, gating on that would match vanilla and remove the need for the colliding_players set. If the team prefers the issue's literal 'once per contact' reading, this is fine as-is.

PR #2384, opened 2026-07-11
Fix incorrect world spawn defaulting to (0, 0) in oceans

### LINKED ISSUE #1303: Incorrect world spawn

### Operating System

Arch Linux

0.1.0/578bab25

### What happened?

When creating a world, the world spawn location is centered around (0,0), even if it is an 'invalid' spawn location such as an ocean.

### To Reproduce

1. Delete current world and player data files
2. Generate a seed where (0,0) is in the ocean, such 118823198. 
3. Observe that you spawn at (0,0)

### Expected behavior

If you generate this seed in vanilla, you spawn at (-127, -63)

dongzh1:
## Description

**What was changed**

New Overworld levels now get a real spawn point instead of a hardcoded `(0, 0)`:

- Added `OVERWORLD_SPAWN_TARGET` (`pumpkin-world/src/biome/position_finder.rs`): vanilla's world-spawn climate target (`OverworldBiomeBuilder#spawnTarget()`), expressed as parameter ranges on the same scale used elsewhere in the biome parameter tree.
- Added `find_overworld_spawn_position()` (`pumpkin-world/src/generation/generator/mod.rs`), which reuses the existing `FittestPositionFinder` (already used for structures) against that climate target to locate the nearest position whose climate resembles ordinary dry land.
- Wired this into `Server` (`pumpkin/src/server/mod.rs`): when a world's `level.dat` is missing and a default `LevelData` has to be created, its spawn position is now set from the search above instead of staying at `(0, 0)`.
- Added `ParameterRange::new()` (codegen + generated output) since hand-written code building a standalone climate target needs a constructor; the biome codegen itself still only ever builds `ParameterRange`s as struct literals.

**Why**

Reported in #1303: a brand-new world's spawn is always `(0, 0)`, even when that point happens to be open ocean (e.g. seed `118823198`). Vanilla instead searches outward from the origin for the closest point that looks like normal land (checking temperature/humidity/continentalness/erosion/weirdness), which is what this PR reproduces.

**Impact**

Only affects world *creation*, i.e. the path taken when `level.dat` doesn't exist yet and a default `LevelData` is generated. Existing worlds with an existing `level.dat` are untouched. New Overworld levels get a spawn point chosen the same way vanilla picks one, instead of unconditionally `(0, 0)`.

**Known limitations**

- This reproduces vanilla's noise/climate-based search only. Vanilla also does a follow-up per-chunk scan for an actual non-waterlogged surface block, which is not implemented here, so on rare seeds a small lake or river could still be picked.
- The Y coordinate is intentionally not resolved by this search; that is left to the existing code path that re-derives it from generated terrain at the chosen `(x, z)` once chunks are loaded.
- The climate target values aren't derived from any registry/datapack asset generated by this repo, so they were cross-checked against the independently-maintained, vanilla-accurate `Cubitect/cubiomes` reimplementation (`spawn_np` in `finders.c`) rather than against an in-repo source of truth.

Fixes #1303

## Testing

- `cargo clippy --all-targets --all-features -- -D warnings` passes.
- `cargo test` passes (all suites green, no failures).
- Not verified: manually booting a server on the exact seed from the issue to confirm the resulting spawn coordinates client-side; the fix was validated via the checks above plus reasoning about parity with vanilla's algorithm and climate target values.

### DIFF
--- a/pumpkin-codegen/src/biome.rs
+++ b/pumpkin-codegen/src/biome.rs
@@ -501,6 +501,16 @@ pub fn build() -> TokenStream {
         }
 
         impl ParameterRange {
+            /// Constructs a parameter range directly. Not used by the biome
+            /// codegen output itself (which builds ranges as struct literals),
+            /// but needed by hand-written code that builds standalone climate
+            /// targets outside of the generated biome tree, such as the
+            /// world-spawn search.
+            #[must_use]
+            pub const fn new(min: i64, max: i64) -> Self {
+                Self { min, max }
+            }
+
             pub fn calc_distance(&self, noise: i64) -> i64 {
                 if noise > self.max {
                     noise - self.max
--- a/pumpkin-data/src/generated/biome.rs
+++ b/pumpkin-data/src/generated/biome.rs
@@ -9032,6 +9032,15 @@ pub struct ParameterRange {
     max: i64,
 }
 impl ParameterRange {
+    /// Constructs a parameter range directly. Not used by the biome codegen
+    /// output itself (which builds ranges as struct literals), but needed by
+    /// hand-written code that builds standalone climate targets outside of
+    /// the generated biome tree, such as the world-spawn search.
+    #[must_use]
+    pub const fn new(min: i64, max: i64) -> Self {
+        Self { min, max }
+    }
+
     pub fn calc_distance(&self, noise: i64) -> i64 {
         if noise > self.max {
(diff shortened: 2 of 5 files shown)


RoosterBooster007:
I'm closing this PR for now because it appears to be largely AI-generated with no in-game testing. Please test and review your code before reopening.

PR #2385, opened 2026-07-11
fix: mobs no longer clip into blocks after taking knockback

### LINKED ISSUE #2232: Mobs ignore hitbox after getting hit by player + ignore bamboo

### Operating System

Fedora 44

Pumpkin 0.1.0-dev+26.1 (Commit: 463c25d/release) - 224 Contributors

### What happened?

Mobs can get into block after getting hit by a player. Tested with small slime.

### To Reproduce

Hit a slime, nearby any solid block, so that it lands in it.

### Expected behavior

The slime should hit the block and stay in front.

dongzh1:
## Description

**What**

`BoundingBox`/`BoundingPlane` collision resolution (`pumpkin-util/src/math/boundingbox.rs`) now tolerates a tiny floating-point epsilon (`BoundingBox::COLLISION_EPSILON = 1.0E-7`) when a per-axis swept collision time comes out slightly negative:

- `BoundingBox::calculate_collision_time` clamps a collision time in `(-COLLISION_EPSILON, 0.0)` to `0.0` instead of discarding the collision (previously *any* negative time, including a few ULPs of rounding noise, made it return `None`).
- `BoundingPlane::intersects` grows each side of the comparison by the same epsilon, so a swept box that lands a hair short of/past the boundary on the perpendicular plane still counts as overlapping.

**Why**

Movement is clipped one axis at a time by solving `(other_plane_const - self_plane_const) / movement_on_axis` for the exact time an entity's leading face reaches an obstacle. Once an entity is resting exactly on that boundary (e.g. after a previous tick's clip), re-deriving the same boundary on a later tick goes through a fresh subtraction/division and does not always round back to the identical `f64` — it can land a few ULPs to either side.

When it lands a hair on the "past" side, the computed collision time is slightly negative. The old code treated *any* negative time as "already fully past this obstacle, nothing to block" and skipped clipping movement on that axis entirely, letting the entity's full, unclipped movement through and visibly into the block on the very next tick. This matches the report: hitting a mob (slime) that is resting against a solid block gives it enough knockback velocity that the rounding noise on the following tick skips the block on that side.

**Impact**

- An entity resting against a block stays correctly blocked on that axis instead of occasionally tunneling into it after receiving knockback (or any other movement) while resting.
- The change only widens an existing boundary check by `1e-7`, far smaller than any real per-tick movement, so normal (clearly-separated) collision behavior is unaffected.

**Limitations**

- This targets the floating-point rounding root cause of mobs clipping into blocks after knockback described in the issue body. I could not independently confirm the "ignore bamboo" wording from the issue title against a separate repro in the issue itself; if that turns out to be a distinct bug (e.g. related to bamboo's partial hitbox) it may need its own follow-up.

## Testing

- `cargo clippy --all-targets --all-features -- -D warnings` passes.
- `cargo test` passes (added targeted unit tests for `BoundingBox::calculate_collision_time` covering the rounding-noise regression and the epsilon boundary).

Fixes #2232

### DIFF
--- a/pumpkin-util/src/math/boundingbox.rs
+++ b/pumpkin-util/src/math/boundingbox.rs
@@ -23,13 +23,21 @@ struct BoundingPlane {
 impl BoundingPlane {
     /// Checks whether this plane intersects another plane.
     ///
+    /// A small epsilon is used so that two planes which are merely touching
+    /// (as happens when an entity is already resting flush against a block
+    /// from a previous tick's collision resolution) still count as
+    /// intersecting instead of being treated as fully separated. Without this
+    /// tolerance, floating-point rounding in [`BoundingBox::calculate_collision_time`]
+    /// can leave a resting entity a few ULPs to either side of the boundary,
+    /// which would otherwise let it slip through on the next tick.
+    ///
     /// # Arguments
     /// * `other` – The other bounding plane to check against.
     pub fn intersects(&self, other: &Self) -> bool {
-        self.min.x < other.max.x
-            && self.max.x > other.min.x
-            && self.min.y < other.max.y
-            && self.max.y > other.min.y
+        self.min.x < other.max.x + BoundingBox::COLLISION_EPSILON
+            && self.max.x > other.min.x - BoundingBox::COLLISION_EPSILON
+            && self.min.y < other.max.y + BoundingBox::COLLISION_EPSILON
+            && self.max.y > other.min.y - BoundingBox::COLLISION_EPSILON
     }
 
     /// Projects a 3D bounding box onto a 2D plane by excluding one axis.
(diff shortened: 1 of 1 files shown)


CompileRider:
Bro, take it easy or you'll leave us with no issues."

Sawlyer:
The intersects expansion is a problem: candidates come from stretch(movement) and gravity always stretches downward, so a grounded entity's X-sweep now intersects the next floor block on the YZ plane and horizontal movement gets clipped at every seam. Vanilla requires overlap greater than 1e-7, never expands. Your own regression test passes without the intersects change. Suggest reverting that hunk, keeping the clamp, and adding a walking-on-flat-floor test.

RoosterBooster007:
@Sawlyer is right. The clamp in `calculate_collision_time` is good, but the intersects epsilon expansion is wrong. Vanilla never expands overlap (it actually contracts it by 1e-7 in `VoxelShape.collideX`), and expanding it makes a grounded entity's X-sweep "overlap" the next floor block at every seam, so horizontal movement gets clipped, and entities stick at block boundaries. Revert that hunk, keep the clamp, and add a walking-on-flat-floor test. Your regression test already passes without the intersects change.

I'm closing this PR for now, since it appears to be largely AI-generated and lacks in-game testing. Feel free to reopen once the issues are resolved.

Issue #2366, opened 2026-07-09
Animals drop cooked food when killed


hydroko13:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11

### Server Software Version/Commit

0.1.0-dev+26.2 (commit 1196d92/release )

### What happened?

When I kill farm animals like pigs and cows they drop cooked meat instead of raw meat

### To Reproduce

1. Spawn a pig or cow
2. kill it without fire
3. it drops cooked meat instead of raw

### Expected behavior

Animals killed without fire or fire aspect should drop raw meat only

lineality:
Looked fixed: Tested this after the new commits (with is-on-fire handling) and cows and pigs now drop raw items (not cooked)

Nice. I was Bug Hunting through this yesterday but the new commits for this look better than what I cobbled together.

PR #2391, opened 2026-07-11
fix(loot): animals drop cooked food without fire or Fire Aspect

CompileRider:
## Summary

Fixes animals dropping cooked food when killed without fire or a Fire Aspect enchantment (#2366).

When a player kills a farm animal (pig, cow, chicken, sheep, rabbit, cod, salmon, etc.) without the entity being on fire and without using a weapon enchanted with Fire Aspect, the animal incorrectly drops cooked meat instead of raw meat.

This bug affects **16 entity types** across the game.

The fix adds proper predicate evaluation to the loot system, matching vanilla Minecraft behavior where the `FurnaceSmelt` loot function is only applied when **either** of the following conditions is true:

- The entity is on fire.
- The direct attacker is holding a weapon with an enchantment in the `#minecraft:smelts_loot` tag (Fire Aspect).

---

## Problem

In vanilla Minecraft, the `FurnaceSmelt` loot function is guarded by an `AnyOf` loot predicate with two subconditions:

1. The killed entity is currently on fire (`minecraft:flags.is_on_fire`).
2. The direct attacker's mainhand weapon has an enchantment in the `#minecraft:smelts_loot` tag (`minecraft:equipment.mainhand.predicates.minecraft:enchantments`).

Pumpkin's implementation of `LootCondition::EntityProperties` lacked the data model, code generation support, and runtime evaluation required to process these predicates. As a result, `FurnaceSmelt` was effectively applied **unconditionally**.

Consequently, every farm animal dropped cooked meat regardless of how it was killed:

- Bare hand
- Non-enchanted weapon
- No fire source present

---

## Root Cause

The loot system contained three independent issues that together prevented proper predicate evaluation.

### 1. Missing Data Model

`LootCondition::EntityProperties` in `pumpkin-util/src/loot_table.rs` only stored:

- `entity: &'static str`
- `expected_type: Option<&'static str>`

It had no representation for:

- Entity state flags (`is_on_fire`)
- Equipment predicates
- Mainhand enchantment tag checks

Without these fields, the condition could not express the predicates required by vanilla's `FurnaceSmelt` implementation.

---

### 2. Code Generation Discards Predicate Data

`EntityPredicateStruct` in `pumpkin-codegen/src/loot.rs` only deserializes the `type` field from vanilla's entity predicate JSON.

Its `ToTokens` implementation emits `()`, meaning all predicate information is discarded during code generation, including:

- `minecraft:flags`
- `minecraft:equipment`
- `minecraft:enchantments`

Additionally, the serde rename used `"type"` instead of `"minecraft:entity_type"`, causing every entity type check to deserialize as `None`.

---

### 3. Runtime Evaluator Bugs

The evaluator in `pumpkin/src/world/loot.rs` contained several logic errors:

- `"direct_attacker"` was missing from the entity name match and fell through to `_ => None`, always evaluating to `false`.
- `"direct_killer"` incorrectly resolved to `params.killer_entity` instead of `params.direct_killer_entity`.
- When `expected_type` was `None`, the evaluator returned `true` whenever the entity existed.

The combined result was:

```text
AnyOf(true, false) == true
```

which caused `FurnaceSmelt` to execute for every entity death.

---

### Before (bug)

https://github.com/user-attachments/assets/150b617d-5442-4e75-ad3d-9e2b47fca31c

### After (fix)

https://github.com/user-attachments/assets/f478d04e-bcb5-464d-af92-0d30fd67279c

### Validation

Added **18 unit tests** covering:

- Entity resolution
- Fire state predicates
- Enchantment tag predicates
- `AnyOf` evaluation logic

All checks pass:

```bash
cargo fmt
cargo clippy
cargo test
```

### DIFF
--- a/pumpkin-codegen/src/loot.rs
+++ b/pumpkin-codegen/src/loot.rs
@@ -312,15 +312,74 @@ pub struct ItemPredicateStruct {
     pub items: Option<StringOrVec>,
 }
 
+/// Entity flags predicate from loot table JSON.
+/// Mirrors vanilla `EntityFlagsPredicate`. Only `is_on_fire` is extracted;
+/// other flags (is_on_ground, is_baby, etc.) are ignored until needed.
+#[derive(Deserialize, Clone, Debug)]
+pub struct EntityFlagsPredicateStruct {
+    /// Whether the entity must be on fire. Mirrors vanilla `isOnFire: Optional<Boolean>`.
+    #[serde(default)]
+    pub is_on_fire: Option<bool>,
+}
+
+/// Entity equipment predicate from loot table JSON.
+/// Mirrors vanilla `EntityEquipmentPredicate`. Only mainhand enchantment tag is extracted;
+/// other slots (head, chest, legs, feet, offhand) and full item predicates are ignored.
+#[derive(Deserialize, Clone, Debug)]
+pub struct EntityEquipmentPredicatesStruct {
+    /// Mainhand slot predicate. Only enchantment tag is forwarded to runtime.
+    /// Mirrors vanilla `EntityEquipmentPredicate.mainhand`.
+    #[serde(default)]
+    pub mainhand: Option<EntityMainhandPredicateStruct>,
+}
+
+/// Mainhand item predicate for equipment slot checks.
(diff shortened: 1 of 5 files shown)


Snowiiii:
LGTM, Thank you @CompileRider :D

PR #2393, opened 2026-07-12
feat(plugin): add block type registry API for WASM plugins

### LINKED ISSUE #2211: Block IDs table for plugin developers

I can use world.get_block_state to get info about a specific block. However, the ID is provided numerically. It's hard to check for example if a block is any kind of planks or if it's an oak wood slab, as each variant (up slab, down slab, waterlogged slab, double slab...) has a different ID. We ought to have a way to check block type more easily like on Bukkit.

A Rust crate that contains all kind of info about blocks, and groupings, etc.

CompileRider:
# Summary

- Adds a new `blocks` WIT interface that exposes Minecraft's block type system to WASM plugin developers.
- Uses a **rich record pattern** (matching the existing `block-state` in `world.wit`) so plugins get all block metadata in a single call.

## What This Adds

The `blocks` WIT interface introduces a `record block-type` with **12 inline fields** and **6 functions**.

### Rich Record

```wit
record block-type {
    id: u16,
    name: string,
    display-name: string,
    hardness: f32,
    blast-resistance: f32,
    map-color: u8,
    is-solid: bool,
    is-air: bool,
    is-liquid: bool,
    burnable: bool,
    light-emission: u8,
    default-state-id: u16,
}
```

All properties reflect the block's default state values. Plugins receive the full record in **one call** instead of making 14 individual getter round-trips.

#### Block Type Lookup

- `get-block-type` — Look up a block type by registry name (e.g. `"minecraft:oak_slab"` or `"oak_slab"`). Returns the full `block-type` record.
- `block-type-from-state-id` — Convert a block state ID to its parent block type record.

#### Tag Queries

- `block-type-has-tag` — Check whether a block belongs to a tag (e.g. `"minecraft:slabs"`).
- `is-valid-block-tag` — Validate that a tag exists.
- `get-block-tag-members` — Return all block types in a tag (capped at **256** entries).

#### State Bridge

- `block-type-all-state-ids` — Return all possible state IDs for a block type (capped at **256**).

## Why

Previously, WASM plugins could only query block properties at a specific world position through `world.get-block-state()`. This interface enables type-level queries—including registry lookups, tag membership, physical property metadata, and state ID conversions—without requiring world access.

The design follows the same rich-record pattern as `block-state` in `world.wit`, providing a consistent API surface.

## Security

- Validates input strings (maximum length of **64** characters and rejects empty strings).
- Caps unbounded `Vec` allocations at **256** entries for both tag members and state ID lists.
- Prevents returning `NaN` or `Infinity` for `f32` values.
- Uses `ok_or_else` instead of `unwrap()` to avoid host crashes when invalid IDs are supplied.

## Closes

Closes #2211.

## Depends On

- Pumpkin-MC/pumpkin-plugin-wit#18


Snowiiii:
Instead of this why not just have a resource block with all these values and a get-block function, like we already have for block states

Snowiiii:
this is a bit sloppy, it would be muche better to have that in a world and some functions are just unnecessary

PR #2394, opened 2026-07-13
feat(pumpkin): add /save-all, /save-off and /save-on commands

Mcxiaocaibug:
Implements the three vanilla save-control commands, which were unchecked in the command tracking issue #15.

## What

- **`/save-all`** — Saves all online players' data and advancements, then requests a chunk save from every world's chunk scheduler. It works even while autosaving is disabled (matching Vanilla), and reports `commands.save.saving` / `commands.save.success`.
- **`/save-off`** — Disables periodic autosaving via a new `save_enabled` flag on `Level`. Running it while already disabled fails with `commands.save.alreadyOff`.
- **`/save-on`** — Re-enables autosaving. Running it while already enabled fails with `commands.save.alreadyOn`.

## How

- Added a `save_enabled: AtomicBool` (default `true`) to `Level` in `pumpkin-world`.
- The autosave trigger in `World::tick_environment` now checks `save_enabled` before flagging `should_save`, so `/save-off` cleanly pauses periodic saves while shutdown saves and manual `/save-all` are unaffected.
- Each command follows the existing command style (permission registration at OP level 4, `translate_cross` feedback, `CommandErrorType` for failures).

All three commands use the existing `commands.save.*` translation keys already present in `pumpkin-data`.

## Testing

- `cargo check` / `cargo clippy` (pedantic + nursery, as enforced by the workspace) pass on `pumpkin` and `pumpkin-world`.
- `cargo fmt --all` applied.
- Ran the server against a console and verified the full behavior:
  - `save-off` -> "Automatic saving is now disabled"
  - `save-off` again -> `commands.save.alreadyOff`
  - `save-on` -> "Automatic saving is now enabled"
  - `save-on` again -> `commands.save.alreadyOn`
  - `save-all` -> "Saving the game..." then "Saved the game"


Snowiiii:
LGTM, Thank you @Mcxiaocaibug

PR #2395, opened 2026-07-13
feat(pumpkin): add /tag command with entity scoreboard tags

Mcxiaocaibug:
Implements the vanilla `/tag` command, which was unchecked in the command tracking issue #15, together with the entity-side tag storage it requires.

## What

- **`/tag <targets> add <name>`** — adds a scoreboard tag; fails with `commands.tag.add.failed` when no target changed (already tagged or at the 1024-tag cap).
- **`/tag <targets> remove <name>`** — removes a tag; fails with `commands.tag.remove.failed` when no target had it.
- **`/tag <targets> list`** — lists the union of tags across targets, with the vanilla single/multiple/empty message variants.

## How

- New `scoreboard_tags: Mutex<HashSet<String>>` on `Entity`, plus `add_scoreboard_tag` / `remove_scoreboard_tag` helpers enforcing the vanilla 1024-tag cap.
- Tags serialize to the vanilla `Tags` NBT string list in `Entity::write_nbt` and load back in `read_nbt_non_mut`, so they round-trip through entity chunk saves (and player data, which delegates to the same entity NBT path).
- Command uses the current-style builder (`EntityArgumentType::Entities` + `StringArgumentType::SingleWord`), OP level 2, and the `commands.tag.*` translation keys already present in `pumpkin-data` for both Java and Bedrock.

Note: the `tag=` / `type=` entity-selector options are still unimplemented in the selector parser (they currently fall through to "unknown option") — that is selector-side work, intentionally out of scope here.

## Testing

- `cargo check` / `cargo clippy` (workspace pedantic + nursery) / `cargo fmt` all clean.
- Verified against a running server from console:
  - `summon minecraft:armor_stand 0 -60 0`, then `tag @e add testtag` -> "Added tag 'testtag' to Armor Stand"
  - duplicate add -> `commands.tag.add.failed`
  - `tag @e list` -> "Armor Stand has 2 tags: secondtag, testtag"
  - `tag @e remove secondtag` -> success; removing a missing tag -> `commands.tag.remove.failed`
- Confirmed persistence by inspecting the saved entity region file after `stop`: the `Tags` list with `testtag` is present on disk.

### DIFF
--- a/pumpkin/src/command/commands/mod.rs
+++ b/pumpkin/src/command/commands/mod.rs
@@ -48,6 +48,7 @@ mod spawnpoint;
 mod stop;
 mod stopsound;
 mod summon;
+mod tag;
 mod teleport;
 mod tellraw;
 mod tick;
@@ -155,6 +156,7 @@ pub async fn default_dispatcher(
     seed::register(&mut dispatcher, registry);
     setidletimeout::register(&mut dispatcher, registry);
     stop::register(&mut dispatcher, registry);
+    tag::register(&mut dispatcher, registry);
     tick::register(&mut dispatcher, registry);
     advancement::register(&mut dispatcher, registry);
     dispatcher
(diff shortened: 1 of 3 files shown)


Snowiiii:
LGTM, Thank you @Mcxiaocaibug

PR #2396, opened 2026-07-13
feat(pumpkin): add /random command

Mcxiaocaibug:
Implements the vanilla `/random` command, which was unchecked in the command tracking issue #15.

## What

- **`/random value <range>`** — draws a random value from an inclusive int range (e.g. `1..10`) and shows it to the command source only (`commands.random.sample.success`).
- **`/random roll <range>`** — draws a value and announces it to every online player (`commands.random.roll`, with the source display name and bounds), plus the console when it is the source.

## Behavior notes

- Uses the existing `IntRangeArgumentType`; open-ended ranges default to `i32::MIN`/`i32::MAX`, matching Vanilla.
- Validation matches Vanilla: a span of 0 (e.g. `5..5`) fails with `commands.random.error.range_too_small`; spans `>= i32::MAX - 1` (including open-ended ranges) fail with `range_too_large`.
- Permission `minecraft:command.random` defaults to Allow, like Vanilla (both subcommands are level 0).
- `/random reset` and the optional `[sequence]` argument are intentionally out of scope: named random-sequence storage does not exist in Pumpkin yet. This covers the two subcommands players actually use.

## Testing

- `cargo check` / `cargo clippy` (workspace pedantic + nursery) / `cargo fmt` clean.
- Verified on a running server from console:
  - `random value 1..10` -> "Randomized value: 2" / "Randomized value: 10" (inclusive bounds reached)
  - `random roll 1..6` -> "Server rolled 4 (from 1 to 6)"
  - `random value 5..5` -> range-too-small error
  - `random value 1..` and `..10` -> range-too-large error

### DIFF
--- a/pumpkin/src/command/commands/mod.rs
+++ b/pumpkin/src/command/commands/mod.rs
@@ -38,6 +38,7 @@ mod playsound;
 mod plugin;
 mod plugins;
 mod pumpkin;
+mod random;
 mod rotate;
 mod say;
 mod seed;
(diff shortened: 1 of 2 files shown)


Snowiiii:
LGTM, Thank you @Mcxiaocaibug

PR #2397, opened 2026-07-13
feat(pumpkin): add /spreadplayers command

Mcxiaocaibug:
Implements the vanilla `/spreadplayers` command, which was unchecked in the command tracking issue #15.

## What

`/spreadplayers <center> <spreadDistance> <maxRange> <respectTeams> <targets>` — teleports targets to random surface locations inside the `maxRange` square around the center column, keeping every pair at least `spreadDistance` apart.

## How

- Ports Vanilla's algorithm: random initial piles, an iterative force-based relaxation (up to 10000 iterations) that repels piles closer than `spreadDistance` and clamps them into the area, then a surface-grounding pass.
- Surface lookup uses `get_heightmap_height_async` (world-surface heightmap, loading chunks on demand); piles whose top block is a liquid are considered unsafe and re-randomized, matching Vanilla's refusal to drop entities onto water.
- Failure raises `commands.spreadplayers.failed.{entities,teams}` with Vanilla's suggested max spread (`maxRange / sqrt(count)`); success reports `commands.spreadplayers.success.*` with the average nearest-neighbor distance.
- `respectTeams=true` collapses all targets onto one pile: teams are not implemented in Pumpkin yet, and Vanilla groups all teamless entities into a single (null-team) pile, so behavior matches Vanilla for a server without teams. The center argument uses `ColumnPosArgumentType` (no float `vec2` argument type exists yet) with Vanilla's block-center correction; the `under <maxHeight>` variant is left as a follow-up.

## Testing

- `cargo check` / `cargo clippy` (workspace pedantic + nursery) / `cargo fmt` clean.
- Verified on a running server from console (ocean-spawn seed, which exercised both paths):
  - `spreadplayers 8 8 4 3000 false @e` teleported a summoned armor stand from (8, 100, 8) to (2984.5, 73, -2557.5) — on land, at surface height.
  - 3 armor stands: "Spread 3 entity/entities around 8.50, 8.50 with an average distance of 4018.15 block(s) apart".
  - Small ranges over open ocean correctly fail with "Could not spread ... (too many entities for space - try using spread of at most 11.55)" (suggestion = maxRange/sqrt(n)), for both the entities and teams wording.

### DIFF
--- a/pumpkin/src/command/commands/mod.rs
+++ b/pumpkin/src/command/commands/mod.rs
@@ -45,6 +45,7 @@ mod setblock;
 mod setidletimeout;
 mod setworldspawn;
 mod spawnpoint;
+mod spreadplayers;
 mod stop;
 mod stopsound;
 mod summon;
@@ -154,6 +155,7 @@ pub async fn default_dispatcher(
     list::register(&mut dispatcher, registry);
     seed::register(&mut dispatcher, registry);
     setidletimeout::register(&mut dispatcher, registry);
+    spreadplayers::register(&mut dispatcher, registry);
     stop::register(&mut dispatcher, registry);
     tick::register(&mut dispatcher, registry);
     advancement::register(&mut dispatcher, registry);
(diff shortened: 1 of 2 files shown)


Snowiiii:
LGTM, Thank you @Mcxiaocaibug

PR #2403, opened 2026-07-15
feat(plugin-api): block resolution + block entity NBT host functions

IndrajeethY:
## Description

Implements host-side logic for four new plugin API functions defined in pumpkin-plugin-wit (Pumpkin-MC/pumpkin-plugin-wit#19):

**`resolve_block_state(name, properties) -> Option<u16>`**

Resolves a block name and property map to a numeric block state ID.
- Returns the default state when no properties are specified
- Looks up specific property combinations via `Block::from_properties`
- Catches panics from unknown property values (e.g. schematics from newer/older MC versions) and returns `None` instead of crashing

**`set_block_entity_nbt(pos, nbt_data) -> Result<(), String>`**

Restores a block entity from raw NBT at a given world position.
- Overrides the x/y/z fields in the NBT with the caller-provided position so tile entities land at the correct coordinates during schematic pastes
- Uses `add_block_entity_nbt` for lazy loading to avoid broadcasting a packet per tile entity during bulk operations

**`get_block_entity_nbt(pos) -> Option<Vec<u8>>`**

Serializes the block entity at a position to raw NBT bytes via `write_internal`. Returns `None` if no block entity exists there. Useful for copying tile entities before a schematic save.

**`block_state_to_info(state_id) -> Option<BlockStateInfo>`**

Resolves a block state ID back to its block name and property key-value pairs. Panic-guarded so an out-of-range ID returns `None` instead of crashing.

Immediate use case is a litematica schematic paster / copier plugin, but all four are general-purpose additions to the plugin API.

## Testing

Tested end-to-end with a schematic paster plugin:
- Loaded multiple `.litematica` files with varying block types and tile entities
- Verified blocks resolve correctly including complex property states (stairs, slabs, fences)
- Confirmed unknown blocks from unsupported versions return `None` without crashing
- Verified tile entities (chests, signs, etc.) restore at correct positions
- Round-tripped block states via `block_state_to_info` -> `resolve_block_state`


IndrajeethY:
CI is expected to fail until the WIT changes land — this PR depends on Pumpkin-MC/pumpkin-plugin-wit#19 which adds `resolve-block-state` and `set-block-entity-nbt` to the world interface. Once that merges and the submodule is updated, this will build cleanly.

pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/events/mod.rs
vyPal, line 1: I'm guessing these changes were made for debugging? If so, it would probably be best to remove them

pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
```diff
@@ -192,7 +192,58 @@ impl PluginHostState {
...
+        // from_properties/from_value panics on unknown property values,
+        // so catch panics to avoid crashing on schematics from other MC versions.
+        let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let block_props = block.from_properties(&props);
+            block_props.to_state_id(block).as_u16()
+        }));
+        match result {
+            Ok(id) => Ok(Some(id)),
+            Err(_) => Ok(None),
+        }
```
vyPal: I think `Ok(result.ok())` could be used here as well to match the function below

vyPal (changes requested):
You need to fix merge conflicts, but otherwise LGTM

pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
IndrajeethY: Oh sure, will rebase and fix conflicts along with it :D

vyPal (approved):
LGTM, please update wit submodule reference to point to latest master

IndrajeethY:
Bumped submodule to wit master (97c0281c). Thanks for the review!

IndrajeethY:
@vyPal

vyPal:
Fixed one minor merge conflict, I'll wait for the checks to succeed and will merge then

PR #2404, opened 2026-07-15
User networking.java.online_mode for online mode detection

vyPal:
## Description
I recently found that the `server.is_online_mode()` in the plugin API always returned `true`, so I looked into the server side code and found that the wit implementation uses `server.advanced_config.networking.java.authentication.enabled`. While this is technically not wrong, however most users when setting up a offline mode server, will just set `online_mode = false`. This changes a different field, specifically `server.advanced_config.networking.java.online_mode` without affecting the authentication status. For this reason I decided to modify the wit implementation to use the online_mode field.

## Testing


Sawlyer:
+1, looks good to me.

Snowiiii:
LGTM, Thank you @vyPal :D

PR #2406, opened 2026-07-15
feat: adding Item predicate with the component ones

aure31:
## Description
start to implement the logic from the vanilla predicate which are used in advancement, lootable, function and lot of places.
## Overview:
 - add the predicate logic with the `ItemsPredicate` and his component ones
 - add a `Predicate` Trait which has a a test function
 - add Component Predicate trait and all the dependent struct
 - update `Bounds` to implement `RangeBounds` and impl From `RangeBounds` for the different bounds.
## Testing
- [x] test suite


Sawlyer:
Yeah quite good but three things need fixing before merge: bound_double_to_option/bound_float_to_option convert exclusive bounds by + or - 1.0, so DoubleBounds::from(1.0..3.0) becomes [1.0, 2.0] and rejects 2.5; DataComponentPredicate::Potions has inverted logic (!v.potion_id.is_some_and(...)), it matches items whose potion is NOT in the list; the WrittenBookContentImpl::read_data rewrite makes title/author/generation/resolved hard-required, and since read_item_stack bails on any unparseable component, a vanilla book omitting a default field would delete the whole item on load, #2422 already models books leniently, so rebase on it and drop the book.rs changes here. Minor: module-wide #![allow(unused)] hides a stray use std::path::Component;, and Trim silently returns false.

PR #2407, opened 2026-07-16
fix: use wolf-specific entity metadata tracker id

Matodor:
## What changed

Adds an entity-scoped `WOLF_VARIANT_ID` tracker constant and uses it when Pumpkin sends wolf variant
metadata.

Legacy protocol versions preserve their existing wolf `VARIANT` IDs. Minecraft 26.1 and 26.2 use tracker
index 23.

## Why

The flattened tracked-data asset loses duplicate entity field names. In 26.1/26.2, the generic
`TrackedData::VARIANT` resolves to index 18, but vanilla Minecraft expects the wolf variant at index 23.

entity data item type`.

## Impact

Wolf metadata is sent at the correct tracker index for supported protocol versions. Other entity metadata
and protocol behavior are unchanged.

## Tests

- `cargo test --manifest-path pumpkin-codegen/Cargo.toml`
- `cargo test --manifest-path Cargo.toml -p pumpkin --lib`
- `cargo clippy --all-targets -p pumpkin -- -D warnings`
- Rustfmt checks for all changed Rust files

The codegen tests also verify that the checked-in generated tracker table matches generator output.

## Limitations

This patch fixes the known wolf metadata collision. Additional entity-specific aliases may be required if
future mappings contain more duplicate tracked-data names.

### DIFF
--- a/pumpkin-codegen/src/tracked_data.rs
+++ b/pumpkin-codegen/src/tracked_data.rs
@@ -28,8 +28,21 @@ pub(crate) fn build() -> TokenStream {
 
         let content = fs::read_to_string(&path)
             .unwrap_or_else(|e| panic!("Failed to read JSON file: {path} {e}"));
-        let parsed: BTreeMap<String, u8> = serde_json::from_str(&content)
+        let mut parsed: BTreeMap<String, u8> = serde_json::from_str(&content)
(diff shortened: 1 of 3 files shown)


RoosterBooster007:
LGTM, @Matodor!

PR #2418, opened 2026-07-19
fix(entity): keep adding XP levels past the i32 total experience limit

### LINKED ISSUE #2094: XP points seemingly stop adding after the 32-bit signed integer limit

### Operating System

Arch Linux

0.1.0-dev+26.1 (Commit: ecc0af8/debug)

### What happened?

After the total XP points for the player reach the 32-bit signed integer limit, it seems like the XP just stops adding beyond the point, as the level and points don't change, even after the command to add XP is run many times.

### To Reproduce

1. Join the server.
2. Set your current gamemode to `survival` or `adventure`, so that the experience bar can be seen.
3. Clear all your pre-existing experience by running the commands:
    - `/xp set @s 0 levels`
    - `/xp set @s 0 points`
4. Now keep running the command `/xp add @s 300000000 points` until it is seen that experience level is stuck  at `21863`. Even after trying some different incremental values, the same problem is still persistent: The level doesn't increase past `21863`, and when that happens, the XP bar is also stuck at the same place.

Image

### Expected behavior

The experience level should increase further as more points are added and the experience progress should change in some way as well.

Mcxiaocaibug:
## Description

Adding experience points previously converted the player's level and points into a single accumulated total clamped to `i32::MAX`, so once a player reached that total, further `/xp add` commands (and any other XP gain) silently did nothing.

This PR carries points across level boundaries incrementally instead, mirroring vanilla's `giveExperiencePoints`, so levels keep increasing past the 32-bit total. The carry logic lives in `pumpkin-util` as a pure function (`experience::add_points_to_level`) with unit tests, including a regression test for adding `i32::MAX` points twice.

Fixes #2094

## Testing

- `cargo test -p pumpkin-util` — 5 new unit tests covering in-level adds, upward/downward carries, total consistency, and totals beyond `i32::MAX`.

### DIFF
--- a/pumpkin-util/src/math/experience.rs
+++ b/pumpkin-util/src/math/experience.rs
@@ -89,3 +89,96 @@ pub fn progress_in_level(points: i32, level: i32) -> f32 {
 
     progress.clamp(0.0, 1.0)
 }
+
+/// The highest level for which `points_in_level` still fits in an `i32`
+/// (`9 * level - 158 <= i32::MAX`).
+pub const MAX_LEVEL: i32 = i32::MAX / 9;
+
+/// Applies a point delta to a `(level, points_into_level)` pair, carrying
+/// across level boundaries like vanilla's `giveExperiencePoints`.
+///
+/// Unlike converting through total accumulated points, this never saturates
+/// once the total exceeds `i32::MAX`, so levels keep increasing.
+///
+/// # Returns
+/// The new `(level, points_into_level)` pair, with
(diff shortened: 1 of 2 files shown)


Sawlyer:
One caveat worth a follow-up: persistence still goes through XpTotal as an i32, so past the old cap the in-memory levels survive but the save wraps and the player resets on restart, vanilla stores XpLevel/XpP separately. Adjacent: read_nbt calls progress_in_level(level, points) with the arguments swapped versus its (points, level) signature.

PR #2419, opened 2026-07-19
fix(block): melt snow layers under high block light

### LINKED ISSUE #2179: Torches dont melt snow layer

Mcxiaocaibug:
## Description

Snow layers never melted because the block had no `random_tick` handler. This PR mirrors vanilla `SnowLayerBlock.randomTick`: on a random tick, the layer is removed when the block light level at its position exceeds 11 (e.g. from a nearby torch).

The snow layer's block states are already flagged as randomly ticking in the generated vanilla data, and the server-side lighting engine already exposes `World::get_block_light_level`, so this only adds the missing handler.

Fixes #2179

## Testing

- Verified via a temporary data test that `Block::SNOW` states have `has_random_ticks` set, so the new handler is actually dispatched.
- `cargo check` / `clippy` clean.

### DIFF
--- a/pumpkin/src/block/blocks/snow.rs
+++ b/pumpkin/src/block/blocks/snow.rs
@@ -10,7 +10,7 @@ use pumpkin_world::{
 
 use crate::block::{
     BlockBehaviour, BlockFuture, GetStateForNeighborUpdateArgs, OnPlaceArgs, OnScheduledTickArgs,
-    UseWithItemArgs, registry::BlockActionResult,
+    RandomTickArgs, UseWithItemArgs, registry::BlockActionResult,
 };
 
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thank you @Mcxiaocaibug :D

PR #2420, opened 2026-07-19
feat(command): add /locate structure

### LINKED ISSUE #2230: /locate command is missing

The /locate command is missing for finding structures. This could help compare pumpkin to vanilla.

Add the command like in vanilla.

Mcxiaocaibug:
## Description

Implements `/locate structure <name>` on top of the existing `find_nearest_structure` infrastructure (already used by eyes of ender), with:

- tab-completion for the generator's 20 known structure sets (`villages`, `desert_pyramids`, ..., optionally prefixed with `minecraft:`),
- vanilla-style clickable green `[x, ~, z]` coordinates that suggest a `/tp` command, plus the horizontal distance in the success message,
- vanilla `commands.locate.structure.invalid` / `.not_found` errors.

Closes #2230

## Notes / limitations

- Structure names refer to the generator's structure sets rather than individual structure keys, since that is what the placement data currently models.
- Like the existing ender-eye path, stronghold (concentric-rings) lookups depend on the global structure cache being populated.

## Testing

Smoke-tested on a live server from the console:

- `locate structure villages` → found a village at (88, ~, 152), 175 blocks away.
- `locate structure nonsense` → correct "no structure with type" error.

### DIFF
--- a/pumpkin/src/command/commands/locate.rs
+++ b/pumpkin/src/command/commands/locate.rs
@@ -1,0 +1,190 @@
+use std::borrow::Cow;
+use std::pin::Pin;
+
+use pumpkin_data::structures::StructureSet;
+use pumpkin_data::translation;
+use pumpkin_util::PermissionLvl;
+use pumpkin_util::math::position::BlockPos;
+use pumpkin_util::permission::{Permission, PermissionDefault, PermissionRegistry};
+use pumpkin_util::text::click::ClickEvent;
+use pumpkin_util::text::hover::HoverEvent;
+use pumpkin_util::text::{TextComponent, color::NamedColor};
+use pumpkin_world::generation::generator::structure_finder::find_nearest_structure;
+
+use crate::command::argument_builder::{ArgumentBuilder, argument, command, literal};
+use crate::command::argument_types::core::string::StringArgumentType;
+use crate::command::context::command_context::CommandContext;
+use crate::command::errors::error_types::CommandErrorType;
+use crate::command::node::dispatcher::CommandDispatcher;
+use crate::command::node::{CommandExecutor, CommandExecutorResult};
+use crate::command::suggestion::provider::SuggestionProvider;
+use crate::command::suggestion::suggestions::{Suggestions, SuggestionsBuilder};
+
+const DESCRIPTION: &str = "Locates the closest structure.";
+
+const PERMISSION: &str = "minecraft:command.locate";
+
+const ARG_STRUCTURE: &str = "structure";
+
+/// The maximum search radius in chunk regions, matching vanilla's
+/// `findNearestMapStructure` call in `LocateCommand`.
+const MAX_SEARCH_RADIUS: i32 = 100;
+
+/// The structure set names known to the generator, offered as suggestions.
+const STRUCTURE_SET_NAMES: &[&str] = &[
+    "ancient_cities",
+    "buried_treasures",
+    "desert_pyramids",
+    "end_cities",
+    "igloos",
+    "jungle_temples",
+    "mineshafts",
+    "nether_complexes",
+    "nether_fossils",
+    "ocean_monuments",
+    "ocean_ruins",
+    "pillager_outposts",
+    "ruined_portals",
+    "shipwrecks",
+    "strongholds",
+    "swamp_huts",
+    "trail_ruins",
+    "trial_chambers",
+    "villages",
+    "woodland_mansions",
+];
+
+const STRUCTURE_INVALID_ERROR_TYPE: CommandErrorType<1> = CommandErrorType::new(
+    translation::java::COMMANDS_LOCATE_STRUCTURE_INVALID,
+    translation::java::COMMANDS_LOCATE_STRUCTURE_INVALID,
+);
+
+const STRUCTURE_NOT_FOUND_ERROR_TYPE: CommandErrorType<1> = CommandErrorType::new(
+    translation::java::COMMANDS_LOCATE_STRUCTURE_NOT_FOUND,
+    translation::bedrock::COMMANDS_LOCATE_STRUCTURE_FAIL_NOSTRUCTUREFOUND,
+);
+
+struct StructureSuggestionProvider;
+
+impl SuggestionProvider for StructureSuggestionProvider {
+    fn suggest(
+        &self,
+        _context: &CommandContext,
+        mut builder: SuggestionsBuilder,
+    ) -> Pin<Box<dyn Future<Output = Suggestions> + Send>> {
+        Box::pin(async move {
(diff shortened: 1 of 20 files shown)


pumpkin/src/command/commands/locate.rs
```diff
@@ -0,0 +1,190 @@
...
+const PERMISSION: &str = "minecraft:command.locate";
+
+const ARG_STRUCTURE: &str = "structure";
+
+/// The maximum search radius in chunk regions, matching vanilla's
+/// `findNearestMapStructure` call in `LocateCommand`.
+const MAX_SEARCH_RADIUS: i32 = 100;
+
+/// The structure set names known to the generator, offered as suggestions.
+const STRUCTURE_SET_NAMES: &[&str] = &[
```
Snowiiii: these should not be hard coded

Snowiiii:
biome and poi subcommands are missing, also don't use StringArgument use ResourceOrTagKeyArgument like in vanilla

Mcxiaocaibug:
Thanks for the review! Addressed in 3f44e36:

- **`ResourceOrTagKeyArgument` / `ResourceOrTagArgument`**: added both argument types mirroring vanilla's pair (`id` or `#tag` syntax, proper `minecraft:resource_or_tag_key` / `minecraft:resource_or_tag` client-side parsers). `/locate structure` now uses the key variant instead of `StringArgument`; the validated variant raises vanilla's `argument.resource.not_found` / `argument.resource_tag.not_found` at parse time where generated registry/tag data exists.
- **`/locate biome`**: vanilla-parity spiral search (6400 block radius, 32 horizontal / 64 vertical steps) over the dimension's noise biome source via a new `find_closest_biome_3d` in pumpkin-world, including the `possibleBiomes` short-circuit (collected from the `BiomeTree`) and superflat support. Runs on a blocking thread; reports absolute-y clickable coordinates and the 3D distance, and tag searches append the concrete biome found, e.g. `#minecraft:is_forest (minecraft:forest)`.
- **`/locate poi`**: searches the world's POI storage within 256 blocks via a new `PoiStorage::find_closest_matching` (chebyshev gather, closest by 3D distance, like `PoiManager.findClosestWithType`), with `point_of_interest_type` tag support.

Current limitations (called out in the commit message): there is no generated structure tag data or POI type registry yet, so structure `#tags` resolve to the invalid-structure error and unknown POI ids fall through to not-found; the server currently only records nether portal POIs.

Smoke-tested all three subcommands live (bare/namespaced ids, tags, invalid inputs, cross-dimension biome short-circuit) plus unit tests against known seed data.

Sawlyer:
Two minor points: the structure search still runs on the async executor while the biome search moved to spawn_blocking, a 100-region scan probably deserves the same; and the .expect(\"locate biome search panicked\") could be mapped to a command error instead of panicking the task.

Mcxiaocaibug:
Thanks! Both addressed in 13f6bef:

- `/locate structure` now runs its search on `spawn_blocking` as well — the 100-region scan is just as CPU-bound as the biome spiral.
- The `.expect` on the join handle is gone: join errors (panic/cancellation) from both blocking searches are now mapped to a "The locate search failed unexpectedly" command error instead of panicking the task.

Mcxiaocaibug:
@Snowiiii

Mcxiaocaibug:
Friendly ping @Snowiiii — this is still ready for review: Sawlyer's feedback was addressed in 13f6bef, CI is green on all platforms, and the branch still merges cleanly into current master. Happy to rebase or adjust if anything else is needed!

Snowiiii:
Hey, Sorry it took so long. @Mcxiaocaibug Could you please provide in game footage this actually works and also fix conflicts

Mcxiaocaibug:
Thanks for getting back to me! Conflicts are fixed — the branch is rebased on current master (399bf14).

Most of the conflict was the `crates/` move; on top of that three upstream signatures had changed since this branch was written, which I adapted in a separate commit so the rebase itself stays readable:

- `WorldGenerator` gained a `Custom` variant for plugin generators. Those only expose biomes by generating a whole chunk, so there is no sampler to spiral over — `/locate biome` reports no match instead of pretending to search.
- `BIOME_REGISTRY` became a reference.
- `Level::world_gen` is now an `ArcSwap`, so the search tasks take a `load_full()` snapshot.

`cargo fmt --check`, `cargo clippy --all-targets --all-features` and the test suites for both `pumpkin` and `pumpkin-world` (257 and 158 passing) are clean on the rebased branch.

I'll follow up with the in-game footage.

Mcxiaocaibug:
CI is green again — the last failure was the `register` signature going from `&mut PermissionRegistry` to `&PermissionRegistry` in c196aa5, which I folded into the API-adaptation commit. All 16 checks pass.

While testing this against a running server I found and fixed a real bug in the command, so it is worth re-reading `locate.rs`: `/locate structure` was reporting the **placement grid's candidate chunk** without checking whether a structure grows there. Candidates are only *possible* sites — the biome at one can reject every structure in the set — so the command pointed at empty terrain. Teleporting to the reported village on a test world found nothing there.

It now resolves the structure start for random-spread sets, across every structure the set contains, which is what explorer maps already do via `find_nearest_structure_start`. Verified against a server on seed `1786557535426458755`:

| Command | Result |
|---|---|
| `/locate structure minecraft:villages` | 611, ~, -233 (653 blocks) |
| `/locate structure minecraft:desert_pyramids` | -1894, ~, 5898 (6194 blocks) |
| `/locate structure minecraft:shipwrecks` | 296, ~, 264 (396 blocks) |
| `/locate biome minecraft:desert` | -5024, 200, -2880 (5790 blocks) |
| `/locate biome minecraft:jungle` | -320, 200, -192 (373 blocks) |
| `/locate structure minecraft:village` (not a set) | clean "no structure with type" error |
| `/locate structure #minecraft:village` (tag) | clean error, see limitations below |

Resolving real starts costs time — these take a few seconds of CPU, which is why both searches run on `spawn_blocking`.

### Two limitations worth knowing before review

**Strongholds return nothing on a freshly generated world.** They use the concentric-rings path, which reads a ring cache that is only populated as a side effect of chunk generation. I looked at computing it from the command instead and deliberately did not: the computation takes a `ProtoChunk` as its biome source, and `ProtoChunk::get_biome_id` is chunk-local, so triggering it from `/locate` would make the world's stronghold layout depend on where the player was standing — and it is a `OnceLock`, so that would stick. Filed as #3005 since it is a worldgen issue rather than a command one.

**Structure tags are rejected**, as noted originally — there is no structure tag registry data yet, so `#minecraft:village` cannot resolve. The argument type already accepts tags for biomes and POIs, so this starts working once that data exists.

Footage is coming separately.

Snowiiii:
LGTM, Thank you @Mcxiaocaibug :D

PR #2421, opened 2026-07-19
fix: persist item data components dropped on save

TheDarkSword:
### Problem

`DataComponentImpl::write_data` has a default implementation that returns `NbtTag::End`, i.e. it serializes to nothing. `ItemStack::write_item_stack` walks the component patch and calls `write_data` on each entry, so every component that didn't override it was silently dropped whenever an item stack got written to NBT — chests, player inventories, chunk saves, etc.

The read side already has deserializers for most components, so this showed up as data quietly disappearing across a save/load cycle: an item with a modified `max_damage`, a custom `food` value, tool rules or block-entity data would come back stripped after a restart.

### Fix

Implemented `write_data` for the data-carrying components that were missing it and can round-trip cleanly:

- `MaxDamage`, `Enchantable`, `Food`, `Tool` — added `write_data` plus a matching `read_data`, and registered them in the `read_data` dispatcher.
- `BlockEntityData`, `Container`, `BundleContents` — already had a deserializer and a dispatcher entry, they were just missing `write_data`.

Each writer follows the vanilla NBT shape for the component and mirrors the existing `read_data` where one already existed.

Added round-trip tests (`write_data` → `read_data` → `equal`) for the scalar and compound cases.

### Still open (out of scope here)

A number of components are still `write_data = End`, but on purpose: they're fieldless placeholder impls that don't model any data yet (`Lore`, `Rarity`, `RepairCost`, `BannerPatterns`, …), so there's nothing to serialize until they get a real data model.


Sawlyer:
+1, looks good to me.

Snowiiii:
Hey @TheDarkSword, Sorry for the late response, I'm not sure how much of your changes got already implemented in master now, Do you plan to fix conflicts?

TheDarkSword:
Hey @Snowiiii no worries. Conflicts fixed

Snowiiii:
LGTM, Thank you, @TheDarkSword :D

PR #2422, opened 2026-07-19
feat: model book contents instead of dropping their text

TheDarkSword:
### Problem

`WritableBookContentImpl` and `WrittenBookContentImpl` didn't actually model book text. Their `read_data` just walked the `pages` list and pushed an empty `String` per element, and neither implemented `write_data`. So a book's title, author and page contents were thrown away — reading a book stack produced blank pages, and writing it back to NBT stored nothing.

### Change

Reworked both components to mirror the vanilla NBT layout (`WritableBookContent` / `WrittenBookContent` codecs):

- **writable_book_content**: `pages` is now a list of `Filterable<String>` — each page a `raw` string plus an optional server-side `filtered` variant, matching vanilla's `{raw, filtered?}` (bare strings are still accepted on read).
- **written_book_content**: now carries `title` (filterable), `author`, `generation`, `pages` and `resolved`. Page components are kept as their raw `NbtTag`, so any text-component payload (plain string or full compound) round-trips verbatim without a lossy conversion — the same approach already used by `custom_data` / `block_entity_data`.

Both now implement `write_data` and a faithful `read_data`, so book contents survive `write_item_stack` / `read_item_stack` (chests, inventories, chunk saves). Reads are lenient (missing optional fields fall back to vanilla defaults) so a slightly malformed book never drops the whole item.

Added round-trip tests for both, including a written page stored as a compound component.

### Notes

- `DataComponentImpl::equal` for `WrittenBookContentImpl` now relies on `PartialEq` over the raw page NBT; `Eq`/`Hash` were dropped from that struct since `NbtTag` isn't `Eq`/`Hash` (nothing depended on them — the only consumer, the lectern comparator output, just reads `pages.len()`).
- The one generated `WritableBookContentImpl { pages: Vec::new() }` default constructor still compiles unchanged.


Sawlyer:
+1, looks good to me.

PR #2425, opened 2026-07-20
Fix crash when serializing advancement item icons

notrootshark:
## Description

Fixes a server crash when sending advancement packets with item icons.

`update_advancement` did not handle item serialization failures. When serialization failed, the error propagated and caused the server to panic. This change falls back to an empty item instead of crashing the server.

## Testing

- Tested advancements triggered by Mace interaction
- Confirmed the server no longer crashes when advancements are triggered

### DIFF
--- a/pumpkin-protocol/src/java/client/play/update_advancement.rs
+++ b/pumpkin-protocol/src/java/client/play/update_advancement.rs
@@ -66,8 +66,15 @@ impl ClientPacket for CUpdateAdvancements {
                 write.write_slice(&display.get_description().encode())?;
 
                 // Item icon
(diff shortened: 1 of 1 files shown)


notrootshark:
Reopened because the referenced PR does not fix this issue. The issue is still reproducible. (This PR fix this)

aure31:
how do you make the server crash ? maybe here is the issue

notrootshark:
> how do you make the server crash ? maybe here is the issue

Hit an entity with a Mace. The server crashes while sending the advancement packet. (This PR fix this)

aure31:
I have tried in the latest version and it doesn't crash for me. did you tried in the latest commit ?

PR #2426, opened 2026-07-21
feat: persist item components that were only set at runtime

TheDarkSword:
### Problem

A bunch of data components still used the default `write_data` (`NbtTag::End`), so their contents were quietly thrown away whenever an item stack got written to NBT (chests, inventories, chunk saves). None of these show up as item defaults in the generated item table, they only ever get set at runtime, so I could give them a real data model without having to touch any generated code.

### Change

Added `write_data` plus a matching `read_data` (and the `read_data` dispatcher entry) for:

- Scalars and simple compounds, following the vanilla shape: `dyed_color` (rgb int), `base_color`, `note_block_sound`, `tooltip_style`, `container_loot` (loot table + seed), `custom_model_data` (floats/flags/strings/colors) and `lodestone_tracker` (target dimension+pos and the tracked flag).
- Unit markers: `glider` and `intangible_projectile` now serialize to an empty compound `{}`, same as vanilla's `Unit` codec (that's what `unbreakable` already does).
- The predicate-ish ones I kept as raw NBT for now so they round-trip without losing anything: `lock`, `trim`, `can_place_on` and `can_break`. Each has a `TODO` to move to a proper typed model later.

All of these survive `write_item_stack` / `read_item_stack` now. Added round-trip tests too.

### Notes

- Network-only components (`creative_slot_lock`, `additional_trade_cost`, `map_post_processing`) are left as they are on purpose, vanilla never writes those to disk so `End` is already correct there.
- A proper block/item predicate subsystem, to replace the raw NBT on `lock`, `can_place_on` and `can_break`, is planned as its own PR.

### DIFF
--- a/pumpkin-data/src/data_component_impl/basic.rs
+++ b/pumpkin-data/src/data_component_impl/basic.rs
@@ -247,63 +247,61 @@ impl DataComponentImpl for EnchantmentGlintOverrideImpl {
 
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub struct TooltipStyleImpl {
-    pub id: Cow<'static, str>,
+    pub id: String,
 }
 impl TooltipStyleImpl {
     pub fn read_data(data: &NbtTag) -> Option<Self> {
-        data.extract_string().map(|id| Self {
-            id: Cow::Owned(id.to_string()),
-        })
+        data.extract_string().map(|id| Self { id: id.to_string() })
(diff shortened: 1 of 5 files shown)


TheDarkSword:
Sorry, I closed it by mistake

Sawlyer:
+1, looks good to me.

Snowiiii:

Issue #2429, opened 2026-07-21
Incorrect mushrooms


Alfiestuff:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Arch Linux

### Server Software Version/Commit

Pumpkin 0.1.0-dev+26.2 (Commit: f4815f5/release) - 238 Contributors

### What happened?

The mushrooms look strange compared its vanilla version.

<img width="1520" height="688" alt="Image" src="https://github.com/user-attachments/assets/50fe611b-a810-4410-859e-2fac52d785f0" />

<img width="1456" height="704" alt="Image" src="https://github.com/user-attachments/assets/4639fc9a-8d0b-4b2c-9009-1b906ad91fcb" />

### To Reproduce

Go to a dark forest in pumpkin hosted server.

### Expected behavior

the mushroom should look like the first attached image

RoosterBooster007:
This should be fixed as of #2430. Please reopen if not.

PR #2430, opened 2026-07-21
fix: match vanilla generation for huge mushrooms in dark oak forests

oswaldsch:
## Description
Fixed the generation of huge mushrooms (red and brown) in Dark Oak Forests to mirror Vanillas generation.
Fixes Issue #2429
I have not written any unit tests as performance/gameplay impact is not to be expected by this change.

## Testing
All Unit Tests pass and there are no new Clippy Warnings.
I have manually tested the feature and the fixed mushrooms now exactly match the vanilla behavior (See before and after below)

## Before (image taken from issue #2429)
image

## After
image


Sawlyer:
Two optional notes: mushroom_tree_height could live in a shared parent module rather than red importing from brown, and vanilla additionally gates each block on replaceability (pre-existing gap here). LGTM otherwise.

RoosterBooster007:
I agree with @Sawlyer on moving `mushroom_tree_height` to a shared parent module. Also, for exact vanilla parity, the cap blocks should use `down: true`. Vanilla leaves the default all-faces-`true` state and only overrides the horizontal faces, so `down: false` culls the visible underside of the cap overhang. Minor/cosmetic: the stem's `up`/`down` are hidden by the cap and ground, but vanilla's default stem state is all-`true` too. Otherwise, this looks great!

oswaldsch:
Thanks for the feedback! I'm away for the next ~2 weeks, but I'll add the changes as soon as I get back.

Snowiiii:
Hey @oswaldsch, Sorry for the late response, did you look at decompiled Minecraft code for this?

Snowiiii:
LGTM, Thank you @oswaldsch :D

PR #2436, opened 2026-07-21
feat(command): add /reload command

JamieLittle16:
## Description

Implements the `/reload` command, which reloads data packs. In
vanilla Minecraft this triggers the `PreparableReloadListener`
pipeline (advancements, recipes, loot tables, tags, functions).
Since Pumpkin does not yet have a data pack system, the command
currently sends the standard "Reloading!" success message and is
wired up so it can be extended when the data pack system arrives.

- Adds `pumpkin/src/command/commands/reload.rs`
- Registers command with `minecraft:command.reload` permission (OP level 2)
- Uses the `CommandContext`-based pattern
- Sends translated feedback for both Java and Bedrock clients

## Testing

- `cargo fmt --all --check` clean.
- `cargo check -p pumpkin` clean.
- `cargo clippy --all-targets` / `cargo test` pending CI.

### DIFF
--- a/pumpkin/src/command/commands/mod.rs
+++ b/pumpkin/src/command/commands/mod.rs
@@ -48,6 +48,7 @@ mod plugins;
 mod pumpkin;
 mod random;
 mod recipe;
+mod reload;
 mod ride;
 mod rotate;
 mod saveall;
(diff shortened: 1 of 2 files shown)


JamieLittle16:
Had bee planning on working on the datapack pipeline so this would be more than a stub. But ive ended up focusing on different issues, so to avoid merging a stub that looks like an implementation I'm closing the request. 

I may circle back to this eventually, but no plans right now.

PR #2437, opened 2026-07-21
fix(world): create directory before saving level.dat

### LINKED ISSUE #2423: ERROR Failed to save level.dat: Info not found! on first startup — level.dat is never written while running

### Operating System

Windows 11

master @ e0117ab5 (built 2026-07-20)

### What happened?

I set up a test server and found this ERROR on the very first start (in an empty directory):

```
WARN  pumpkin::server: Failed to get level_info, using default instead: Info not found!
ERROR pumpkin::server: Failed to save level.dat: Info not found!
```

While the server is running, world/level.dat is never created (other folders like region/ are). It only appears after a graceful stop — I confirmed this on my machine. If the process is killed, no level.dat exists at all.

Cause: Server::new calls AnvilLevelInfo.write_world_info (pumpkin/src/server/mod.rs:193) before the world/ directory exists. write_world_info (pumpkin-world/src/world_info/anvil.rs:156) calls File::create, which fails with ErrorKind::NotFound when the parent directory doesn't exist. The world folder is only created later during world loading.

Fix plan: call fs::create_dir_all before File::create in write_world_info.

Note: I see open PR #2384 touches the same unwrap_or_else block — if it conflicts, I'll rebase.
If it's OK, I'd like to work on the fix myself.

### To Reproduce

1. Build latest master
2. Run the server in an empty directory
3. Observe the ERROR at startup, and check that world/level.dat is missing while running
4. Type stop, then check again — level.dat now exists

### Expected behavior

No ERROR on first start. level.dat should be written immediately at startup (the world directory should be created before writing).

MegalithOfficial:
## Description
This PR fixes https://github.com/Pumpkin-MC/Pumpkin/issues/2423

On a fresh server, Pumpkin tried to create `level.dat` before the world directory existed, so the initial save failed with `Info not found!`.

This creates the target directory before writing `level.dat`. I also added a regression test that writes to a missing nested world directory and reads the saved metadata back.

## Testing
I tested the fix by creating a clean Pumpkin instance and confirmed that `level.dat` was created successfully on first startup. I joined the server, waited for chunks to generate, and then logged off. I then restarted the server using the generated world to verify that the saved metadata could be read back correctly. Both startup and shutdown completed without the previous save error.

### DIFF
--- a/crates/pumpkin/src/command/argument_types/entity_selector/option.rs
+++ b/crates/pumpkin/src/command/argument_types/entity_selector/option.rs
@@ -10,6 +10,7 @@ use crate::command::suggestion::suggestions::SuggestionsBuilder;
 use pumpkin_data::entity::EntityType;
 use pumpkin_data::translation;
 use pumpkin_util::GameMode;
+use pumpkin_util::identifier::{Identifier, VANILLA_NAMESPACE};
 use pumpkin_util::math::bounds::{DoubleBounds, FloatDegreeBounds, IntBounds};
 use pumpkin_util::text::TextComponent;
 use std::str::FromStr;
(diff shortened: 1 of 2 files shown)


NetroAki:
This PR fixes the directory creation. I'll add the atomic write fix in a combined PR.

Sawlyer:
+1 looks good for me.

PR #2442, opened 2026-07-21
fix: preserve navigator movement while move control is idle

RedstoneGithub:
## What changed

Prevented `MoveControl` from clearing movement input while it is in the `Wait` state.

## Why

`Navigator` runs first and writes forward movement for path-following. `MoveControl` then ran in its normal idle state and overwrote that input with zero, so mobs could find paths but could not follow them.

## Impact

Mobs can now follow existing navigator paths.

This PR does not change pathfinding, goals, strafing, or movement speed.

## Verification

- `cargo fmt --check`
- `cargo test`
- Playtested, and mobs can move now. Before, they could not move.

AI disclaimer: Codex was used to help with this PR.


Snowiiii:
LGTM

Issue #2226, opened 2026-06-04
Wierd chunk trail


Barni-Mezei:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Server: Alpine Linux v3.23 (aarch64) running on a Raspberry Pi 4B

### Server Software Version/Commit

a2f0122/release

### What happened?

A line of stone is in the middle of a `minecraft:birch_forest`

<img width="1919" height="1080" alt="Image" src="https://github.com/user-attachments/assets/7dee2a1f-6e7a-418d-b728-15f991181c80" />

### To Reproduce

Start a world with the seed: 1779920288596261407
Teleport to: /execute in minecraft:overworld run tp @s 902.56 177.46 1022.64 -2340.60 90.00

### Expected behavior

No random stone line in a forest

RoosterBooster007:
This should now be resolved thanks to #2448

PR #2448, opened 2026-07-22
fix: stone-capped chunks when generation resumes from a saved proto chunk

oswaldsch:
## Description

Image of the Bug (taken from Issue #2226)
image

ProtoChunk::from_chunk_data restored the flat heightmaps using index z * 16 + x, while every reader of those maps (local_position_to_height_map_index) uses x * 16 + z. Any chunk saved to disk mid-generation and later reloaded resumed with transposed heightmaps. 

### Why
This is the cause of the stone trail bug mentioned in #2226.

1. Chunks at the generation frontier are (correctly) saved at an intermediate stage, e.g. Noise, when they get flushed (autosave/unload/shutdown)
2. When the player later moves toward them, theyre reloaded via from_chunk_data and generation resumes, now with transposed heightmaps
3. The surface stage iterates each column down from top_block_height_exclusive. On sloped terrain the transposed height lies below the real column top for part of the chunk, so the real top blocks are never touched by the surface rules and stay default stone
4. The chunk completes to Full and is saved, so the damage is permanent

Because the trigger is which chunks happened to be mid-generation at save time, the artifact looked non-deterministic and couldn't be reproduced from the seed (recreating the world with the same seed gives different results each time). The transpose is also invisible on flat terrain, which is why it only showed on slopes and "stopped" at rivers.

Reproduced deterministically by loading a status=Noise chunk from a real world save and resuming it in-process: 104/256 columns came out stone-capped vs. 0 for uninterrupted generation, and the heightmap error grid was antisymmetric across the chunk diagonal, the transpose signature.

### Impact
- Chunks resumed from disk now generate identical surfaces to uninterrupted generation (verified: resumed vs. fresh = 0 diffs on the repro chunk)
- Also fixes the motion_blocking / motion_blocking_no_leaves maps being transposed on resumed chunks, which affected feature placement
- Existing worlds: frontier chunks not yet finalized will now resume correctly; chunks already finalized with a stone surface are baked in and won't be repaired

### Limitations
While investigating I noticed from_chunk_data also doesn't restore flat_ocean_floor_height_map or structure starts/references. Neither is implicated in this bug, so they're left untouched here.

## Testing

- New regression test heightmap_roundtrip_through_chunk_data_resume: generates a sloped chunk to the Noise stage, round-trips it through ChunkData (the actual save/load path), and asserts heightmaps and the resumed surface match uninterrupted generation. Fails on master (208/256 columns transposed), passes with the fix
- cargo test -p pumpkin-world: 91 passed, 0 failed
- In-game: flew ~3000 blocks over fresh terrain crossing old generation frontiers, no stone caps

> [!NOTE]
> I did use AI for the debugging process because this shit needed sifting through the whole world gen code only to find out it was completly unrelated to that. All the code changes were applied by me, and I tested it myself.

### DIFF
--- a/pumpkin-world/src/generation/proto_chunk.rs
+++ b/pumpkin-world/src/generation/proto_chunk.rs
@@ -292,7 +292,7 @@ impl ProtoChunk {
 
         for z in 0..16 {
             for x in 0..16 {
-                let index = ((z << 4) + x) as usize;
+                let index = Self::local_position_to_height_map_index(x, z);
 
                 proto_chunk.flat_motion_blocking_height_map[index] = heightmap_data.get(
                     ChunkHeightmapType::MotionBlocking,
--- a/pumpkin-world/src/generation/proto_chunk_test.rs
+++ b/pumpkin-world/src/generation/proto_chunk_test.rs
@@ -5,6 +5,93 @@ mod test {
     use pumpkin_data::dimension::Dimension;
     use pumpkin_util::world_seed::Seed;
 
+    // Regression test for transposed heightmaps during Noise-stage chunk resume.
+    // Flat terrain cannot expose this bug, so use a sloped chunk.
+    #[test]
+    fn heightmap_roundtrip_through_chunk_data_resume() {
+        use crate::chunk_system::chunk_state::Chunk;
+        use pumpkin_config::lighting::LightingEngineConfig;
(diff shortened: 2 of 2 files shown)


Sawlyer:
+1, looks good to me.

RoosterBooster007:
LGTM! Thanks, @oswaldsch.

PR #2450, opened 2026-07-22
chore: Bedrock Raknet Handshake / update to bedrock 26.30

ZlordHUN:
## Description
Fixes the Bedrock RakNet handshake and chunk delivery so modern Bedrock clients can connect and remain in the world reliably. (By remain in world reliably I mean they wont get kicked seconds after joining, or suffocate in walls after walking up to them, getting kicked and crashing the client - in that order)

What
- Corrects RakNet handshake field encoding and IPv4 address serialization.
- Adds reliable-packet tracking, acknowledgements, retransmission, timeouts, and backpressure.
- Prevents chunk delivery from flooding the Bedrock connection.
- Corrects Bedrock chunk biome-palette encoding.
- Uses only palette bit widths supported by the Bedrock protocol.

## How
- Encodes RakNet GUIDs, timestamps, MTU values, and system indices using the correct byte order.
- Runs connection maintenance independently from packet reception so missing acknowledgements can be retransmitted without blocking incoming traffic.
- Limits Bedrock chunk delivery to one batch per tick and waits when the reliable-packet window is full.
- Removes invalid subchunk headers from biome palette data.
- Rounds palette storage widths to Bedrock-supported values.

## Testing
- `cargo test -p pumpkin-world bedrock_palette`
- Connected using Bedrock 1.26.30 and verified that the player could authenticate, enter the world, receive chunks, and remain connected and not suffocate after walking up to walls.

### DIFF
--- a/pumpkin-protocol/src/bedrock/client/level_chunk.rs
+++ b/pumpkin-protocol/src/bedrock/client/level_chunk.rs
@@ -77,32 +77,26 @@ impl PacketWrite for CLevelChunk<'_> {
             .read()
             .map_err(|_| Error::other("biome_sections read lock poisoned"))?;
 
-        for (i, biome_palette) in biome_sections.iter().enumerate() {
-            let num_storages = 1;
-            let y = (i as i8) + min_y_section;
-            data_write.write_all(&[VERSION, num_storages, y as u8])?;
+        for biome_palette in biome_sections.iter() {
+            let network_repr = biome_palette.convert_be_network();
 
-            for _ in 0..num_storages {
-                let network_repr = biome_palette.convert_be_network();
+            (network_repr.bits_per_entry << 1 | 1).write(data_write)?;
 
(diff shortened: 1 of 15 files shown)


Sawlyer:
Two things before merge: the new reliable-frame semaphore introduces a stall vector, a client that keeps the connection alive but withholds ACKs blocks send_frame_set forever, fills the bounded outgoing channel, and broadcast_packet_except_editioned then stalls world broadcasts for everyone on a single bad client; please bound the permit wait (disconnect when no ACK arrives within a deadline). The serializer/deserializer and client raknet changes are byte-identical to #2400, which also ships tests, suggest rebasing on it. Also consider splitting the unrelated if let refactors into their own PR.

Snowiiii:
Didn't you just update to 26.30?, when yes please make it more clear and title and desc

ZlordHUN:
@Snowiiii done but if you merge the java player visibility PR this can be discarded cause that PR has the same code as this.

PR #2451, opened 2026-07-22
fix(protocol): remap particle IDs for older clients

MegalithOfficial:
## Description

This PR fixes particle IDs being sent incorrectly to older clients.
I first noticed this with TNT on 1.21.11. The explosion happened normally on the server, but the client received the raw 26.2 particle ID and disconnected with:
```text
Internal Exception: io.netty.handler.codec.DecoderException: Failed to decode packet 'clientbound/minecraft:explode'
```
After fixing that packet, I found that normal particle packets had the same issue. They disconnected the client with:
```text
Internal Exception: io.netty.handler.codec.DecoderException: Failed to decode packet 'clientbound/minecraft:level_particles'
```
This was not only affecting TNT. Normal particle packets and area effect cloud metadata were also using raw 26.2 IDs. So this PR adds particle ID mappings for every supported version and applies them when encoding explosion packets, normal particle packets and particle metadata.

On 1.21.11, 121 out of 125 particle IDs are different from 26.2. The only particles that keep the same ID are `angry_villager`, `block`, `block_marker` and `bubble`.

The particles currently used directly by Pumpkin that were affected are `cloud`, `crit`, `entity_effect`, `explosion`, `explosion_emitter`, `heart, poof`, `portal`, `smoke` and `sweep_attack`. The particle command can use the rest of the registry, so those shifted IDs were affected as well.
## Testing

I tested this using a 1.21.11 Fabric client and reproduced both disconnects.

I also added regression tests for explosion packets, normal particle packets and particle metadata. The tests verify that older clients receive the remapped ID, 26.2 clients keep the original ID and the remaining particle data is kept after remapping.


Sawlyer:
Yeah, one follow-up: particle payloads embedding block-state IDs (block, block_marker, falling_dust, dust_pillar) still carry 26.2 state IDs to older clients; block_state_remap could be applied there later.

Snowiiii:

PR #2452, opened 2026-07-22
fix: persist block entity data on chunk save

JamieLittle16:
## Summary

  - Snapshot live block-entity NBT before autosave, unload processing, `/save-all`, and shutdown.
  - Use each block entity’s async canonical NBT writer for disk persistence, so contended inventory slots are awaited rather than skipped.
  - Remove pending NBT when a block entity is removed, preventing stale data from being restored.
  - Add NBT round-trip coverage for generic inventories, furnaces, hoppers, brewing stands, crafters, and chiseled bookshelves.

  ## Testing

  - `cargo fmt --check`
  - `cargo check -p pumpkin`
  - `cargo test -p pumpkin --lib nbt_round_trip`
  - `cargo test -p pumpkin --lib inventory_serialization_waits_for_a_contended_slot`

## Split
Campfire cooking was moved to #2488 so these persistence changes can be reviewed and merged independently.


JamieLittle16:
The `auto_pop_off`  bool flag is probably not specific enough. I should probably add some reference to campfire in this. 
in hindsight it probably would have worked better as two PRs (one with the feature another with the fix), apologies if this makes it difficult to follow


Since been addressed

Sawlyer:
The persistence half is high-value. Two blockers: drop the unrelated /reload stub that reports success without reloading anything; the try_lock pattern can silently persist wrong data, sync_write_items_to_nbt skips contended slots and the try_lock().ok()? style aborts the whole compound on contention; at minimum log, ideally snapshot under one lock. +1 to splitting persistence and campfire. Round-trip NBT tests for the inventory entities would guard the vanilla key names.

JamieLittle16:
Split update: campfire cooking now lives in #2488. This PR is persistence-only, so either change can be reviewed and merged independently.

JamieLittle16:
> The persistence half is high-value. Two blockers: drop the unrelated /reload stub that reports success without reloading anything; the try_lock pattern can silently persist wrong data, sync_write_items_to_nbt skips contended slots and the try_lock().ok()? style aborts the whole compound on contention; at minimum log, ideally snapshot under one lock. +1 to splitting persistence and campfire. Round-trip NBT tests for the inventory entities would guard the vanilla key names.

Thanks - agreed on the split; campfire is now in #2488 and this PR is persistence-only. I’ve removed the no-op /reload stub (this was accidentally included in the first place). For contended inventory locks, serialization now skips the affected slot and emits a warning rather than aborting the whole block-entity compound silently. That makes contention visible while keeping the save path non-blocking.

I agree that snapshotting an inventory under one lock would be the stronger long-term solution. I’ll add round-trip NBT coverage for the inventory entities to validate the vanilla key names and prevent regressions.

JamieLittle16:
> > The persistence half is high-value. Two blockers: drop the unrelated /reload stub that reports success without reloading anything; the try_lock pattern can silently persist wrong data, sync_write_items_to_nbt skips contended slots and the try_lock().ok()? style aborts the whole compound on contention; at minimum log, ideally snapshot under one lock. +1 to splitting persistence and campfire. Round-trip NBT tests for the inventory entities would guard the vanilla key names.
> 
> Thanks - agreed on the split; campfire is now in #2488 and this PR is persistence-only. I’ve removed the no-op /reload stub (this was accidentally included in the first place). For contended inventory locks, serialization now skips the affected slot and emits a warning rather than aborting the whole block-entity compound silently. That makes contention visible while keeping the save path non-blocking.
> 
> I agree that snapshotting an inventory under one lock would be the stronger long-term solution. I’ll add round-trip NBT coverage for the inventory entities to validate the vanilla key names and prevent regressions.

The solution is lossless per slot. A single-lock whole-inventory snapshot would still be a stronger future atomicity improvement for multi-slot transfers, but the save path no longer silently skips or aborts inventory data under contention

Snowiiii:
LGTM

PR #2456, opened 2026-07-22
Fix egg: initialize git submodules during install

MythicSorcerer:
The Pumpkin repo uses `pumpkin-plugin-wit` as a git submodule, but the install script only does `git clone --depth 1` without `--recurse-submodules`. This causes `bindgen!` / `wit_bindgen::generate` to fail with:

```
error: failed to read path for WIT [../pumpkin-plugin-wit/v0.1]
       Caused by: No such file or directory (os error 2)
```

This adds `git submodule update --init --recursive --depth 1` after `cd Pumpkin` so the WIT files are available during compilation.


Snowiiii:
LGTM, Thanks

PR #2460, opened 2026-07-22
fix: /execute as @e now runs commands for every entity, not just the player

### LINKED ISSUE #2447: 2 more bugs

### Operating System

Win 11

f4815f5

### What happened?

- Killing @a stops the server

- execute command

### To Reproduce

- /kill @a

- /execute as @e at @p run say 1 -- Nothing happens

### Expected behavior

- Stops the server
- Nothing Happens

JamieLittle16:
Fixes #2447: `/execute as @e at @p run say 1` produced no output because commands like `say` were only registered in the old command dispatcher and the inner executor used `execute_input` which has no fallback.

## Two bugs found and fixed

**1. `ExecuteRunExecutor` used `execute_input` (new dispatcher only) instead of `handle_command`**
- `handle_command` falls back to the old command dispatcher for commands like `say`, `item`, `particle`, `setblock`, and `summon` that aren't registered in the new tree yet.

**2. `try_flatten` skipped the root context's modifier**
- When building the execution chain, the root context's `RedirectModifier` was never included, so `execute_as_modifier` was never called and entity identity was never propagated to `source.output`. Now the root is included as the first modifier when it has a non-`KeepSource` redirect.

## Additional changes
- Added `CommandSender::Entity(String)` variant so entity-sourced commands display the correct name in output
- Full exhaustiveness updates across the old command system (`item`, `particle`, `setblock`, `summon`)
- Added regression test `root_with_modifier_is_included_in_chain` to ensure `try_flatten` includes the root modifier
- All tests pass (150), clippy clean, fmt clean

### DIFF
--- a/pumpkin/src/command/commands/execute.rs
+++ b/pumpkin/src/command/commands/execute.rs
@@ -1,6 +1,5 @@
 use std::sync::Arc;
 
-use crate::command::CommandSender;
 use crate::command::argument_builder::{ArgumentBuilder, argument, command, literal};
 use crate::command::argument_types::block::BlockArgumentType;
 use crate::command::argument_types::coordinates::block_pos::BlockPosArgumentType;
@@ -35,10 +34,7 @@ impl CommandExecutor for ExecuteRunExecutor {
         Box::pin(async move {
             let command_str = StringArgumentType::get(context, "command")?;
             let dispatcher = context.server().command_dispatcher.read().await;
-            dispatcher
-                .handle_command(&context.source, command_str)
-                .await;
-            Ok(1)
+            dispatcher.execute_input(command_str, &context.source).await
         })
     }
 }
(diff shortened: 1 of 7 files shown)


Sawlyer:
Three follow-ups: CommandSender::Entity hardcodes PermissionLvl::Two, if minecraft:command.execute is ever granted to a non-op, the inner command runs with level 2 instead of the invoker's level; vanilla preserves the source's level. ExecuteRunExecutor always returns Ok, dropping the inner result, this will bite /execute store later. Entity(String) has no position/world, so at/positioned don't reach fallback commands like setblock ~ ~ ~; fine for #2447 but worth a TODO.

JamieLittle16:
> Three follow-ups: CommandSender::Entity hardcodes PermissionLvl::Two, if minecraft:command.execute is ever granted to a non-op, the inner command runs with level 2 instead of the invoker's level; vanilla preserves the source's level. ExecuteRunExecutor always returns Ok, dropping the inner result, this will bite /execute store later. Entity(String) has no position/world, so at/positioned don't reach fallback commands like setblock ~ ~ ~; fine for #2447 but worth a TODO.

Thanks, all three were addressed in 781981d2.

execute as now retains the original CommandSource.output, so nested commands keep the invoker’s actual permission level and explicit grants instead of falling back to CommandSender::Entity’s hard-coded level 2. ExecuteRunExecutor now returns the nested command’s result directly as well.

I also added a TODO documenting that fallback commands still do not consume the source’s entity world/position context, so at/positioned are not yet propagated to commands such as setblock.

JamieLittle16:
Closing as superseded by the subsequent command-system rewrite.

The current execute implementation now applies the root as modifier across the generated command sources, so the underlying bug this PR was addressing is no longer present in the current architecture.

PR #2461, opened 2026-07-22
feat: add item frame entity and real comparator analog output

darkstarworks:
## What

The comparator's item frame reading was stubbed: `get_attached_itemframe_level` returned a hardcoded `Some(1)` regardless of the frame's rotation (the "Vanilla Parity TODO" in `comparator.rs`).

The underlying problem turned out to be bigger than the stub: item frames had no entity implementation at all. Frames loaded from imported vanilla worlds fell into the generic entity fallback in `from_type`, which silently drops their `Item`, `ItemRotation` and `Facing` NBT on the next save. So this PR adds the entity first and fixes the comparator on top of it.

- New `ItemFrameEntity` (decoration) holding the displayed item, rotation, facing, drop chance, and the invisible/fixed flags. NBT round-trips the vanilla keys (`Item`, `ItemRotation`, `ItemDropChance`, `Facing`, `Invisible`, `Fixed`). The facing is mirrored into the spawn packet data field, which is how vanilla clients orient frames.
- `from_type` maps `minecraft:item_frame` and `minecraft:glow_item_frame` to the new entity.
- `get_attached_itemframe_level` now returns the vanilla signal: 0 when the frame is empty, otherwise `rotation % 8 + 1`. This matches vanilla `ItemFrame.getAnalogOutput`.
- Two side fixes in the same function: glow item frames are now included (vanilla checks `ItemFrame.class`, which covers both), and the facing filter uses the frame's stored facing instead of the yaw-derived horizontal facing, so frames hanging above or below a block no longer produce false matches.

## Why

Comparator plus item frame is a common contraption (shop selectors, combination locks, and so on). On imported worlds the comparator read a constant 1 no matter the rotation, and worse, the frames lost their contents after a save cycle.

## Known limitations

- No placement or right-click rotation yet, so this currently serves worlds imported from vanilla. The entity is the groundwork for interaction support later.
- Punching a frame removes it, same as paintings currently do; popping the displayed item first is left as a TODO.
- The displayed item is not yet synced to clients (frames have no metadata sync yet); the frame itself spawns with the correct orientation.

## Verification

`cargo check`, `cargo clippy --all-targets --all-features`, `cargo fmt --check` and `cargo test` (250 passed, 0 failed) are all clean on Linux with stable 1.97.1.


darkstarworks:
Hi, by the way. 
This is my first contribution to Pumpkin
Have a nice day.

Sawlyer:
Two minor points: please open PRs from a feature branch rather than your fork's master; ItemFrameEntity::new sets facing to South but leaves entity.data at 0 (Down), so a frame created without NBT spawns facing the wrong way, mirror the default into entity.data in new. Otherwise LGTM.

darkstarworks:
Good catch on the facing default, fixed in 8f8c4eb: the default now goes into both the facing field and the entity data field, so the two can't disagree.

I also rebased on latest master to clear the conflict in `decoration/mod.rs`, which came in with the Leash PR. Both module lines are kept.

On the feature branch: sorry about that. GitHub won't let me change the source branch of an open PR without closing this one and opening a new one, which would throw away your review here. I'll open from a feature branch for anything after this.

RoosterBooster007:
Looks great! Thanks, @darkstarworks.

PR #2465, opened 2026-07-22
fix: nametags apply custom names to mobs and persist across reloads

JamieLittle16:
Three fixes that make the nametag item work end-to-end — renaming a nametag in an anvil and right-clicking a mob now correctly displays the custom name, and the name survives chunk unloads and player relogs.

### 1. Show the name (`name_tag.rs`)
The `use_on_entity` handler had a `// TODO` and was only calling `set_custom_name()` — which stores the name but does **not** make it visible. Added `set_custom_name_visible(true)` so the name label actually appears above the mob.

### 2. Reach all client versions (`entity/mod.rs`)
`set_custom_name()` sent `CUSTOM_NAME` metadata via `MetaDataType::OPTIONAL_TEXT_COMPONENT`, which has type ID **-1** for Minecraft v26+ clients. The entry was silently skipped — the client received `CUSTOM_NAME_VISIBLE = true` (showing the label) but never got the actual name text, so it defaulted to the entity type name ("Cow").

Now sends **both** `OPTIONAL_TEXT_COMPONENT` (1.21.x) and `OPTIONAL_COMPONENT` (v26+). The version-specific type ID check in `Metadata::write()` filters each per-client, so every version picks the valid one.

### 3. Persist across reloads (`world/mod.rs`)
The NBT save/load already handled storing and restoring `custom_name` in memory. But on chunk load, nobody sent the `CUSTOM_NAME` / `CUSTOM_NAME_VISIBLE` metadata to clients. Added metadata re-send in both code paths:
- **First load** (new chunk loaded from disk): sends metadata after the spawn packet
- **Already-live** (player joins a chunk another player already loaded): same, after spawn

Metadata must arrive **after** the spawn packet — clients silently ignore `CSetEntityMetadata` for entity IDs they haven't seen yet.

## Verification
- `cargo clippy --all-targets` — clean
- `cargo fmt --check --all` — clean
- `cargo test -p pumpkin` — all 149 tests pass

### DIFF
--- a/pumpkin/src/entity/mod.rs
+++ b/pumpkin/src/entity/mod.rs
@@ -1059,13 +1059,88 @@ impl Entity {
             visible,
         );
         self.send_meta_data(
-            &[Metadata::new(
-                TrackedData::CUSTOM_NAME,
-                MetaDataType::OPTIONAL_TEXT_COMPONENT,
-                Some(name),
-            )],
+            &[
+                Metadata::new(
+                    TrackedData::CUSTOM_NAME,
+                    MetaDataType::OPTIONAL_TEXT_COMPONENT,
+                    Some(name.clone()),
+                ),
+                Metadata::new(
+                    TrackedData::CUSTOM_NAME,
+                    MetaDataType::OPTIONAL_COMPONENT,
+                    Some(name),
+                ),
+            ],
             Some(&bedrock_meta),
         );
+    }
+
+    /// Sends the current custom-name metadata to one player after that player has
+    /// received this entity's spawn packet.
+    pub fn send_custom_name_to_player(&self, player: &Player) {
+        let Some(name) = &**self.custom_name.load() else {
+            return;
+        };
+        let name = name.clone();
+        let visible = self.custom_name_visible.load(Ordering::Relaxed);
+
+        match player.client.as_ref() {
+            ClientPlatform::Java(client) => {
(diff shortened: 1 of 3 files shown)


prajapatyogesh710-byte (changes requested):
Instagram bio change
[https://www.instagram.com/prajapat_palak_18741?igsh=MXYwcXM4ZTlzdmUyNQ==]
Change `ID`name
@prajapat_palak_18741 
`change`bio ```
Hey yogesh prajapat 
```

Sawlyer:
Two small points : the chunk-load path re-broadcasts to every watcher rather than the loading player; and CUSTOM_NAME_VISIBLE isn't re-sent on reload, so a saved always-visible name loses that flag.  Was this playtested on a v26+ client?

JamieLittle16:
> Two small points : the chunk-load path re-broadcasts to every watcher rather than the loading player; and CUSTOM_NAME_VISIBLE isn't re-sent on reload, so a saved always-visible name loses that flag. Was this playtested on a v26+ client?

Thanks, addressed in `99795960`.

The chunk-load path now sends the entity’s nametag metadata directly to the player loading the chunk instead of rebroadcasting it to every watcher. The reload metadata includes both the custom name and `CUSTOM_NAME_VISIBLE`, so saved always-visible names retain that flag.

I playtested this in latest release

PR #2470, opened 2026-07-23
feat: Implement pillager outpost structures

ZlordHUN:
## Description
- This PR adds Pillager Outpost generation, matching vanilla - flaws included like floating scarecrows over ditches.
- Another item off of this list: https://github.com/Pumpkin-MC/Pumpkin/issues/36
- Also includes a scheduler fix without which the server panics and crashes when the player nears a Pillager Outpost.

## Testing
cargo test -p pumpkin-protocol
cargo test -p pumpkin-world bedrock_palette
cargo fmt --all -- --check
- Inspected Pillager outposts in game for the same seed in both vanilla and pumpkin.

### DIFF
--- a/pumpkin-world/src/chunk_system/generation.rs
+++ b/pumpkin-world/src/chunk_system/generation.rs
@@ -183,4 +183,99 @@ mod tests {
         );
         assert_eq!(jigsaw_blocks, 0, "jigsaw blocks were not replaced");
     }
+
+    #[test]
+    fn seed_zero_generates_the_vanilla_pillager_outpost_chunk() {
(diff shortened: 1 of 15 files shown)


Sawlyer:
Solid vanilla-parity work, and the drop_satisfied_tasks fix looks like the actual root cause of the stranded-node panics #2339 mitigates, worth cross-referencing. Two minor points: pending_structure_entities isn't persisted, so a chunk saved between Features and Spawn loses its caged mobs on resume; and mutually-exclusive structure sets would recurse forever in should_generate_structure, safe with vanilla data, but a depth guard would be cheap. Neither blocks merging.

PR #2474, opened 2026-07-23
fix: restore terminal state when exiting while console input is active

### LINKED ISSUE #2441: When the server crashes, I can't type in terminal

### Operating System

Linux Mint Cinnamon edit: X11 Intel CPU + Nvidia GPU

0.1.0-dev26.2 (Commit f4815f5/release)

### What happened?

The server has crashed many times as seen in bug report https://github.com/Pumpkin-MC/Pumpkin/issues/2438#issue-4943375042 but whenever I try to restart the server via terminal, I can't type and need to open a new terminal window. Note that when running the "$stop" command this doesn't happen. I'm guessing when the server crashes unexpectedly on Linux or specific distros, it keeps trying to take user input, even after closing, so you can't type in normal terminal; but I'm not that smart so idk.

### To Reproduce

1. Launch the server via Linux terminal
2. Logon as a player on version 26.2
3. Get any advancement to crash the server(presumably other crashes work too)
4. Try to type in the terminal window or do an action like pressing one of the arrow keys, nothing will happen other than the cursor stops flashing as you type

### Expected behavior

Server crashes should completely release your keyboard so that you can type as normal.

eshankiyer:
## Description

Fixes #2441.

`rustyline` switches the terminal into raw mode while a `readline()` call is active and only restores it when that call returns. If the process exits while the console thread is still blocked inside `readline()` — e.g. after a panic in another thread, where the panic hook ends in `std::process::exit(1)` — the terminal is left in raw mode. Typing in the shell afterwards is invisible until the user runs `reset`, which matches what everyone in #2441 reported (and why running the server again "fixes" it: rustyline saves the already-broken state and restores that).

This PR adds a small `terminal` module (unix-only, no-op on other platforms) that:

- saves the original termios of stdin in `init_logger`, right before the rustyline editor is created (only when stdin is a TTY and `use_tty` is enabled)
- registers a `libc::atexit` handler that restores those attributes on any exit path — `exit(1)` from the panic hook, `exit(code)` from `stop_or_exit_server`, and normal returns. On the graceful `$stop` path rustyline has already restored the state, so the handler is a harmless no-op there.

`libc` is added as a unix-only dependency of the `pumpkin` crate (it is already in the dependency tree transitively).

## Testing

- `cargo fmt --check` passes
- The restore runs on every process exit via `atexit`, so the panic-exit path from #2441 is covered by construction; I have not reproduced the original advancement crash end-to-end, so a confirmation from someone who can trigger it would be welcome.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2653,6 +2653,7 @@ dependencies = [
  "hyper",
  "image",
  "indexmap",
+ "libc",
  "libloading",
  "notify",
  "num-bigint",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,6 +140,7 @@ heck = { version = "0.5", default-features = false }
 hmac = { version = "=0.13.0", default-features = false }
 indexmap = { version = "2.14", default-features = false, features = ["std"] }
 itertools = { version = "0.15.0", default-features = false, features = ["use_std"] }
+libc = { version = "0.2", default-features = false, features = ["std"] }
 libloading = { version = "0.9", default-features = false, features = ["std"] }
 lru = { version = "0.18.1", default-features = false }
 lz4-java-wrc = { version = "0.2.0", default-features = false, features = ["lz4_flex"] }
--- a/pumpkin/Cargo.toml
+++ b/pumpkin/Cargo.toml
@@ -113,6 +113,10 @@ image = { version = "0.25.10", default-features = false, features = ["png"] }
 
 indexmap.workspace = true
 
+# Restoring the terminal state on exit (see `src/terminal.rs`)
+[target.'cfg(unix)'.dependencies]
+libc.workspace = true
+
 [dev-dependencies]
 tempfile.workspace = true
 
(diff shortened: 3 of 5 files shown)


Sawlyer:
The unix-only dependency scoping is clean. One optional follow-up: rustyline on Windows manipulates console modes via SetConsoleMode and could leak similarly on abnormal exit, worth a tracking note, though all #2441 reports are Unix.

eshankiyer:
Good catch on Windows — you're right that `rustyline` uses `SetConsoleMode` there and can leak the same way on an abnormal exit, so the non-unix stub isn't a "nothing to do here" case, it's an unimplemented one. I've reworded the stub's comment to say that explicitly rather than implying Unix is the only platform that can leak.

I've deliberately left the Windows implementation out of this PR: it needs `GetConsoleMode`/`SetConsoleMode` on the stdin handle plus an exit hook, and I have no Windows machine to verify it on. Since every report in #2441 is Unix, I'd rather ship the verified half than guess at the other. Happy to open a tracking issue for the Windows side, or to attempt it if someone can test the result.

PR #2475, opened 2026-07-23
fix: reject native plugins with implausible metadata instead of aborting

### LINKED ISSUE #2434: Crash — memory allocation of ~94 trillion bytes failed in PluginManager::spawn_plugin_initialization when loading PatchBukkit

### Operating System

Arch Linux, x86_64

0.1.0-dev+26.2 (Minecraft protocol 776)

### What happened?

# ⚠️ most of this text used in this issue was made by ai

Server crashes with a Rust allocation-failure panic during plugin initialization, only when the PatchBukkit plugin (libpatchbukkit-X64-Linux.so) is present in ./plugins/. Removing the plugin allows the server to start normally. The allocation size requested is enormous and changes slightly between runs (e.g. 94287725934367, 94855951694367, 93868530125215), which looks like an address or other non-length value being misused as an allocation size somewhere in the formatting/logging path — not a genuine memory shortage.

### To Reproduce

1. Place libpatchbukkit-X64-Linux.so in ./plugins/
2. Run ./pumpkin-X64-Linux
3. Server crashes with SIGABRT shortly after "Found cached permission decision for plugin "" (approved: true)"

Backtrace (RUST_BACKTRACE=full):

memory allocation of 94398717781991 bytes failed
stack backtrace:
...
  10: <alloc::raw_vec::RawVecInner<_>>::reserve::do_reserve_and_handle
  11: <alloc::string::String as core::fmt::Write>::write_str
  12: core::fmt::write
  13: alloc::fmt::format::format_inner
  14: <pumpkin::plugin::PluginManager>::spawn_plugin_initialization::{closure#0}::{closure#0}
  ...
Notes:
- Frame 14 is inside Pumpkin's own PluginManager, in a format!-style call during plugin init — not inside the plugin itself or j4rs/JVM code.
- Suspect this is triggered by a field in PatchBukkit's plugin metadata (name/description/version or similar) being used unsafely in a formatting call — e.g. a width/precision specifier or length value taken from plugin-supplied data without validation.
- Ruled out: JVM/JDK version and path issues (confirmed JAVA_HOME/libjvm.so correctly resolved), plugin .so architecture/dlopen issues, and cached-resource corruption (crash persists after deleting and re-extracting all j4rs resources).
- PatchBukkit repo: https://github.com/Pumpkin-MC/PatchBukkit

### Expected behavior

# ⚠️ this text was not made with ai

the server crashes (note: i did change the server port but it worked without the plugin)

eshankiyer:
## Description

Fixes #2434.

The `METADATA` symbol in a native plugin is plain Rust data (`String`/`Vec` fields) whose layout is not stable across toolchains. When a plugin was built with a different Rust version or against a different Pumpkin build, the field lengths read from it can be garbage even though the `PUMPKIN_API_VERSION` check passes (the constant only captures the plugin API revision, not the toolchain). Cloning or logging such metadata then aborts the whole server with an allocation failure in the trillions of bytes, which is exactly the backtrace in #2434: the abort happens inside `format!` in `spawn_plugin_initialization`, after the "cached permission decision for plugin \"\"" line shows an empty misread name.

This PR validates the metadata before it is ever cloned or formatted:

- new `is_metadata_plausible` check in the native loader that bounds every text field to 4 KiB, every list to 1024 entries, requires a non-empty name, and requires `len <= capacity` for each field
- only lengths and capacities are inspected; the potentially invalid data pointers are never dereferenced
- implausible metadata fails the load with a new `LoaderError::MetadataCorrupt` that tells the user to rebuild the plugin against this Pumpkin build

This is deliberately a best-effort mitigation: any ABI-mismatched dylib is undefined behavior territory, and the durable fix would be passing metadata across the boundary in a stable representation (C strings or serialized bytes) plus embedding the rustc version in the exported symbols. That is an API design decision I did not want to make unilaterally, but I am happy to follow up on it if there is interest.

## Testing

- `cargo fmt --check` passes
- The reported failure mode (a field length in the trillions) is caught by the length bound before the `metadata.clone()` in the loader, so the PatchBukkit case from the issue reports gets a clean "metadata appears corrupt" load error instead of an abort. I do not have a PatchBukkit binary to verify end to end.

### DIFF
--- a/pumpkin/src/plugin/loader/native.rs
+++ b/pumpkin/src/plugin/loader/native.rs
@@ -23,8 +23,20 @@ const MAX_METADATA_LIST_LEN: usize = 1024;
 /// against a different Pumpkin build, the field lengths read here can be
 /// garbage even when the API version check passes, and cloning or logging
 /// such metadata aborts the server with an enormous allocation (see issue
-/// #2434). Only lengths and capacities are inspected here; the potentially
-/// invalid data pointers are never dereferenced.
+/// #2434).
+///
+/// This is a mitigation, not a sound fix: reading a `#[repr(Rust)]` struct
+/// written by a different toolchain is already undefined behaviour, and
+/// checking the strings inside the lists follows the lists' data pointers.
+/// It does not dereference anything that the `metadata.clone()` below would
+/// not dereference anyway, and it bails out earlier: the scalar fields are
+/// checked first and short-circuit, so mismatched metadata is almost always
+/// rejected before any pointer is followed.
+///
+/// The sound fix is to give the metadata a stable representation across the
+/// boundary (C strings or serialized bytes) and to export a build
+/// fingerprint alongside `PUMPKIN_API_VERSION` so a toolchain mismatch is
+/// rejected deterministically.
 fn is_metadata_plausible(metadata: &PluginMetadata) -> bool {
     let text_plausible =
         |text: &String| text.len() <= MAX_METADATA_TEXT_LEN && text.len() <= text.capacity();
@@ -122,6 +134,11 @@ impl PluginLoader for NativePluginLoader {
     }
 }
 
+// These cover only the pure predicate on well-formed values. The actual
+// hazard (a struct whose layout came from another toolchain) cannot be
+// constructed in safe Rust, so the `len <= capacity` conjunct is untestable
+// here. Adding a field to `PluginMetadata` deliberately breaks `valid()`, so
+// that the new field has to be considered in `is_metadata_plausible`.
 #[cfg(test)]
 mod tests {
     use super::{MAX_METADATA_LIST_LEN, MAX_METADATA_TEXT_LEN, is_metadata_plausible};
--- a/pumpkin/src/plugin/loader/native.rs
+++ b/pumpkin/src/plugin/loader/native.rs
@@ -28,8 +28,13 @@ const MAX_METADATA_LIST_LEN: usize = 1024;
 fn is_metadata_plausible(metadata: &PluginMetadata) -> bool {
     let text_plausible =
         |text: &String| text.len() <= MAX_METADATA_TEXT_LEN && text.len() <= text.capacity();
-    let list_plausible =
-        |list: &Vec<String>| list.len() <= MAX_METADATA_LIST_LEN && list.len() <= list.capacity();
+    // The inner strings must be checked too: a list header can look sane
+    // while its elements are garbage, and those elements are cloned as well.
+    let list_plausible = |list: &Vec<String>| {
+        list.len() <= MAX_METADATA_LIST_LEN
+            && list.len() <= list.capacity()
+            && list.iter().all(text_plausible)
+    };
 
     !metadata.name.is_empty()
         && text_plausible(&metadata.name)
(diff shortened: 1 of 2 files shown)


Sawlyer:
Good two optional notes: inner strings of the Vec<String> fields are still cloned unvalidated, so a partially-corrupt layout can still abort, worth a comment; and is_metadata_plausible is pure and easily unit-testable. +1 on the follow-up idea of a stable ABI representation.

eshankiyer:
Thanks — both notes addressed, and the first one turned out to matter more than it looked.

**Inner strings.** `list_plausible` now checks every `String` inside the lists against the same bound, so a sane-looking list header with garbage elements is rejected before the clone.

**Correction to my own doc comment.** While implementing that I realised the comment I originally wrote ("the potentially invalid data pointers are never dereferenced") became false: `list.iter().all(...)` goes through `Deref for Vec<T>`, which builds a slice from the data pointer and loads through it. I've rewritten the comment rather than leaving a claim in the tree that a future reader would rely on. It now says what is actually true:

- this is a mitigation, not a sound fix — reading a `#[repr(Rust)]` struct written by another toolchain is UB before we get here at all
- it dereferences no more than the `metadata.clone()` immediately below already does, and it bails out earlier: the scalar fields are checked first and `&&` short-circuits, so mismatched metadata is almost always rejected before any pointer is followed
- the sound fix is a stable representation across the boundary (C strings or serialized bytes) plus a build fingerprint exported next to `PUMPKIN_API_VERSION`

On that last point, I agree with your +1 on a stable ABI. The cheap deterministic half is a `build.rs`-generated `u32` fingerprint of the rustc version checked alongside `PUMPKIN_API_VERSION` — reading a `u32` is layout-independent, so it is genuinely safe, and it converts the exact failure mode in #2434 (right API version, wrong toolchain) from probabilistic to deterministic. Happy to do that as a follow-up PR, or fold it into this one if you'd rather it land together.

**Tests.** Six unit tests added for the predicate. I've noted in a comment what they can't cover: the real hazard can't be constructed in safe Rust, so the `len <= capacity` conjunct is untestable here and the tests shouldn't be read as covering the corruption path. The struct literal in `valid()` intentionally breaks the build if a field is added to `PluginMetadata`, so the new field has to be considered in the check.

pumpkin/src/plugin/loader/native.rs
```diff
@@ -11,6 +11,52 @@ use super::{LoaderError, Path, Plugin, PluginLoader, PluginMetadata};
...
+/// checking the strings inside the lists follows the lists' data pointers.
+/// It does not dereference anything that the `metadata.clone()` below would
+/// not dereference anyway, and it bails out earlier: the scalar fields are
+/// checked first and short-circuit, so mismatched metadata is almost always
+/// rejected before any pointer is followed.
+///
+/// The sound fix is to give the metadata a stable representation across the
+/// boundary (C strings or serialized bytes) and to export a build
+/// fingerprint alongside `PUMPKIN_API_VERSION` so a toolchain mismatch is
+/// rejected deterministically.
```
vyPal: No need for this many comments

vyPal (approved):
No need for all these comments, this is a very well known issue. Please fix merge conflicts, otherwise LGTM

eshankiyer:
Conflicts are resolved. Linux checks are green and the remaining release builds are still running.

Issue #2453, opened 2026-07-22
Cows = Crash


Aarinator2010:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

windows

### Server Software Version/Commit

f4815f5

### What happened?

If you kill a cow, the server crashes, and if you load the area, the server crashes again.

### To Reproduce

kill a cow

### Expected behavior

The cow drops its items.

Alfiestuff:
can confirm

Alfiestuff:
its not just cows its any mob in general please fix

andro-32:
Update Pumpkin.

andro-32:
I just realised that the latest download isn't currently the latest version so you have to get it from the actions.

MythicSorcerer:
I can reproduce this on latest download from the website 0.1.0-dev+26.2 on MacOS

Issue #2424, opened 2026-07-20
Server crash (panic) when unlocking any advancement: "Item serialization is not yet implemented without serde"


soichiro0520:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows 11 

### Server Software Version/Commit

master @ e0117ab5 (release build)

### What happened?

The server panics and shuts down as soon as a player unlocks an advancement that has a display (icon).

Steps I took: I joined my test server, crafted planks from logs, then crafted a crafting table. The moment the "Minecraft" advancement (crafting table) unlocked, the server crashed:

Thread 'tokio-rt-worker' panicked at pumpkin-protocol\src\java\client\play\update_advancement.rs:69:17:
not implemented: Item serialization is not yet implemented without serde

Looking at the code, the display branch of update_advancement.rs hits `unimplemented!` when serializing the item icon. This seems related to the recent serde removal (e0117ab5 "perf: ditch serde because its slower") — any advancement with a display will crash the server, so normal survival gameplay is currently not possible.

Full crash report attached below (Panic Location: update_advancement.rs:69:17, commit e0117ab/release).

Fix ideas (for discussion): implement ItemStack serialization for this packet without serde, or temporarily skip sending the display/icon instead of panicking.

[crash-2026-07-20_00.26.54.txt](https://github.com/user-attachments/files/30172675/crash-2026-07-20_00.26.54.txt)

### To Reproduce

1. Build and run latest master (e0117ab5)
2. Join with a client
3. Craft a crafting table (or unlock any advancement with a display/icon)
4. The server panics and shuts down

### Expected behavior

Unlocking an advancement should not crash the server. The update_advancement packet should be sent (or the display skipped gracefully).

PR #2476, opened 2026-07-23
fix: do not panic on unknown or unprefixed items in loot table entries

eshankiyer:
## Description

The `Item` entry arm of `LootPoolEntryTypes::get_stacks` unwraps both the `minecraft:` prefix strip and the item registry lookup:

```rust
let key = &item_entry.name.strip_prefix("minecraft:").unwrap();
vec![ItemStack::new(1, Item::from_registry_key(key).unwrap())]
```

A loot table entry naming an item without the `minecraft:` prefix, or one that is missing from the item registry, panics the server in the middle of loot generation (for example on a mob kill). The neighboring `LootTable` and `Tag` arms already handle both cases defensively with `unwrap_or` and `filter_map`, so this arm was the odd one out.

This PR makes the `Item` arm match: the prefix is treated as optional, and an unknown item logs a warning and drops the entry instead of panicking, the same way the `Tag` arm skips unknown items.

I looked at this while investigating #2453; that crash turned out to be the already-fixed advancement panic (#2424), so this PR does not claim to fix it, but the panic path here is real and reachable from data.

## Testing

- `cargo fmt --check` passes
- Behavior for well-formed entries is unchanged: prefixed names resolve to the same registry key as before

### DIFF
--- a/pumpkin/src/world/loot.rs
+++ b/pumpkin/src/world/loot.rs
@@ -385,8 +385,17 @@ impl LootPoolEntryTypesExt for LootPoolEntryTypes {
                     })
             }
             Self::Item(item_entry) => {
-                let key = &item_entry.name.strip_prefix("minecraft:").unwrap();
-                vec![ItemStack::new(1, Item::from_registry_key(key).unwrap())]
+                let key = item_entry
+                    .name
+                    .strip_prefix("minecraft:")
+                    .unwrap_or(item_entry.name);
+                Item::from_registry_key(key).map_or_else(
+                    || {
+                        tracing::warn!("Unknown item {:?} in loot table entry", item_entry.name);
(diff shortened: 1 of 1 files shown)


Sawlyer:
Good, one optional simplification: Item::from_registry_key already strips the minecraft: prefix internally, so the manual strip_prefix could be dropped.

eshankiyer:
Thanks for the review. Pushed that simplification — confirmed `Item::from_registry_key` strips the `minecraft:` prefix itself (`pumpkin-data/src/generated/item.rs`, first line of the function), so the manual `strip_prefix` was dead weight. Dropped it and left a short comment noting that prefixed and unprefixed names both resolve, so nobody re-adds it later.

PR #2477, opened 2026-07-23
fix: bedrock java player visibility

ZlordHUN:
## Description
- Built on top of this PR https://github.com/Pumpkin-MC/Pumpkin/pull/2450 cause without it players cant even join. Anyway, this PR fixes the visibility issues between bedrock and java players. With this they can both see each other.

## Testing
- cargo test -p pumpkin-protocol
- cargo test -p pumpkin-world bedrock_palette
- cargo fmt --all -- --check
- Logged onto pumpkin with both a java and bedrock player and looked at each other.

### DIFF
--- a/crates/pumpkin/src/net/bedrock/mod.rs
+++ b/crates/pumpkin/src/net/bedrock/mod.rs
@@ -1270,7 +1270,8 @@ impl BedrockClient {
                 self.handle_set_local_player_as_initialized(
                     player,
                     &SSetLocalPlayerAsInitialized::read(reader)?,
-                );
+                )
+                .await;
             }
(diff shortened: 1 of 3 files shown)


Sawlyer:
Two requests: this inherits the reliable-frame-window stall concern from #2450, and it matters more here because spawn packets now flow through the blocking broadcast, please address it there first; the client_loaded guard races with itself since Bedrock packets are handled in independent tasks, a compare_exchange would make the double-send impossible. Suggest rebasing to the pure delta once #2450 and #2400 land.

Snowiiii:
fix conflicts please

ZlordHUN:
@Snowiiii done!

ZlordHUN:
Already fixed.

PR #2408, opened 2026-07-17
feat: Add villager trading and professions

ZlordHUN:
## Description
Implements villager workstation detection, profession assignment, and the normal villager trading lifecycle.

## What
- Adult unemployed villagers locate and claim Vanilla workstation blocks within the 48-block POI search radius.
- Villagers path to potential job sites, acquire the corresponding profession on arrival, and release unreachable or invalid sites.
- Claimed workstations are persisted through Pumpkin’s existing POI region storage.
- Merchant interactions consume payment, deliver results, increment offer uses, lock exhausted offers, award villager/player XP, and support normal and shift-click trading.
- Villagers level up after Vanilla’s 40-tick delay, unlock the next trade tier, update their visible badge metadata, receive regeneration, and display the associated particles.
- Villagers restock at their workstation during working hours, including Vanilla’s demand updates, two-restocks-per-day limit, and delay between restocks.
- Trading includes reputation and Hero of the Village price adjustments.
- Villagers provide Vanilla-style interaction feedback: trade sounds, throttled valid/invalid input sounds, unhappy head movement, job-site particles, and trade particles.
- Villagers face and follow their current customer instead of continuing to wander while a merchant screen is open.

Supported workstations include lecterns, composters, blast furnaces, smokers, cartography tables, brewing stands, barrels, fletching tables, cauldrons, stonecutters, looms, smithing tables, and grindstones.

## How
- Ports the relevant behavior from vanilla, including merchant offer matching, adjusted prices, offer usage, XP thresholds, delayed level-up, restocking, job-site validation, and merchant feedback.
- Adds a compact POI ownership layer on top of Pumpkin’s existing persisted POI storage. Job-site acquisition checks the five nearest candidates for path reachability before claiming one.
- Separates potential job sites from established job sites. Villagers only change profession and emit happy particles after reaching the workstation.
- Resets a profession after losing its workstation only for an untraded novice, matching Vanilla’s XP and level restrictions.
- Uses Vanilla’s generated trade-set data for every profession and level, including secondary costs and dynamic modifiers:
  - Tradeable enchanted books with Vanilla’s level and price formula.
  - Enchanted equipment with additional price components.
  - Explorer maps with actual structure-start lookup and translated item names.
  - Random dyes, potions, and suspicious stew effects.
- Adds the required item-component network and NBT codecs so dynamic trade results survive packet encoding and world saves.
- Synchronizes villager data and active-effect particle metadata so profession badges and level-up particles update without reconnecting.

## Testing
[...]

PR #2376, opened 2026-07-10
feat: add chest, furnace, hopper, and TNT minecarts

ZlordHUN:
## Description
Adds functional chest, furnace, hopper, and TNT minecarts on top of Pumpkin’s existing rideable minecart implementation.

## What
- Adds chest minecarts with persistent 27-slot inventories and deferred loot-table support.
- Adds furnace minecarts with fuel consumption, player-directed propulsion, visual fueled state, and persistent fuel/push data.
- Adds hopper minecarts with five inventory slots, item and container collection, activator-rail disabling, and persistent state.
- Adds TNT minecarts with activator-rail priming, fuse persistence, smoke, collision and damage handling, speed-scaled explosions, and Vanilla rail protection.
- Persists minecart-specific inventories and state across chunk unloads and server restarts.
- Restores players to their minecart after reconnecting or restarting the server using Vanilla’s `RootVehicle` player data.
- Fixes minecart damage-wobble metadata for Java and Bedrock clients.
- Drops the appropriate minecart item and stored inventory contents when broken.

## How
- Splits each minecart variant into a dedicated module while keeping shared rail movement and collision behavior in the base minecart entity.
- Uses Pumpkin’s existing inventory, screen-handler, NBT, loot-table, entity, and explosion systems.
- Dispatches variant-specific interaction, ticking, persistence, drops, and metadata through a compact internal minecart-kind enum.
- Preserves the off-rail gravity implementation already merged into `master`.

## Testing
- `cargo check -p pumpkin`
- `cargo test -p pumpkin --lib`
- `cargo test -p pumpkin-world --lib`
- `cargo clippy -p pumpkin -p pumpkin-world --all-targets -- -D warnings`
- `cargo fmt --all -- --check`

### In-game testing
- Placed and interacted with chest, furnace, hopper, TNT, and rideable minecarts.
- Verified chest and hopper inventories persist after reconnecting and restarting.
- Verified breaking container minecarts drops the minecart and stored contents.
- Verified furnace minecart fueling and propulsion.
- Verified hopper collection and activator-rail disabling.
- Verified TNT priming and explosions.
- Verified players reconnect inside their previously occupied minecart.

PR #2479, opened 2026-07-23
feat(combat): implement spears and sync equipment attributes

MegalithOfficial:
## Description

This PR implements spear combat. Spears already existed as items, but none of their attacks worked.

I added the normal jab and kinetic attack behavior using the same calculations and conditions as the official server. This includes the different values for each spear material, damage, knockback, dismounting, movement checks, contact cooldowns, durability damage, sounds and attack ranges.

Also, I noticed that I could spam the jab animation even though most of the attacks did not register. My server-side cooldown calculation is correct, but equipped item modifiers are not currently applied to the player's runtime attributes or synced to the client. The client still thinks the player has the default attack speed, so it keeps sending attacks before the server is ready for them.

That synchronization issue is not fixed yet. I am currently working on applying equipment modifiers to the runtime attribute map, syncing changed attributes to the client and remapping attribute registry IDs for older client versions. I am doing this as a general equipment attribute fix because the same problem can affect swords, axes, armor and other equipment too.

This is a draft because the equipment attribute synchronization is still being implemented and tested.


MegalithOfficial:
Update: I finished the equipment attribute synchronization.

Initially, while testing the normal jab, I found that my target ray did not match vanilla. Large or close hitboxes could intersect the ray before the spear's minimum range, which caused valid hits to be rejected. I changed the ray to start at the minimum attack range and added the vanilla forward movement extension.

I then fixed the synchronization issue by tracking equipment changes, removing old modifiers and applying new modifiers to the player's runtime attributes. Changed attributes are now synced to the client, and normal melee and spear attacks use the same effective attack damage and attack speed values.

I also added attribute registry ID remapping for older client versions and skip attributes that are not supported by the connected client. The client and server now agree on the spear's attack cooldown instead of playing jab animations for attacks that the server rejects

Sawlyer:
Two things before undrafting: since update_equipment_attributes runs every tick, the explicit calls in attack()/sync_hand_slot/swap_item are redundant, one strategy is enough; the kinetic path uses get_attribute_base while the jab uses get_attribute_value, if that's vanilla parity, a comment would help. Also note conflicts with #2408 and #2376 in player.rs/play.rs.

Snowiiii:
@MegalithOfficial please split this into multiple PRs

PR #2482, opened 2026-07-23
perf(protocol): buffer encrypted writes

JamieLittle16:
## Description

Replace the byte-at-a-time Java encryption write path with a bounded, backpressure-safe ciphertext buffer.

Previously, encrypted writes followed this path:

Packet encoder -> bytewise StreamEncryptor -> 8 KiB BufWriter -> TCP

Each plaintext byte caused a separate encryption adapter write and downstream dispatch. The old implementation also retained only the last unwritten byte, relying on the caller to continue writing the same logical byte stream across poll_write calls. That assumption is not guaranteed by the AsyncWrite contract.

The new encrypted path is:

Packet encoder -> coalescing StreamEncryptor -> TCP

StreamEncryptor now:

- Encrypts up to 16 KiB at a time.
- Owns ciphertext until the downstream writer accepts it.
- Preserves pending ciphertext across partial writes and Poll::Pending.
- Drains pending ciphertext before accepting more plaintext.
- Drains pending ciphertext during flush and shutdown.
- Coalesces small application writes until the buffer is full or the caller flushes.
- Reports WriteZero and downstream I/O errors correctly.

The unencrypted path keeps its existing BufWriter. The BufWriter is removed only after encryption is enabled, avoiding a redundant buffering layer for encrypted traffic.

## Performance

Criterion benchmarks compare the old bytewise encryption plus 8 KiB BufWriter path against the new 16 KiB coalescing encryption path writing directly to the downstream writer.

Results from the local benchmark:

| Workload | Old | New | Change |
| --- | ---: | ---: | ---: |
| 64 B contiguous | 1.149 us | 1.132 us | 1.5% faster |
| 1 KiB contiguous | 16.545 us | 16.021 us | 3.2% faster |
| 16 KiB contiguous | 268.01 us | 258.93 us | 3.4% faster |
| 64 KiB contiguous | 1.078 ms | 1.033 ms | 4.1% faster |
| 1 MiB contiguous | 16.711 ms | 16.474 ms | 1.4% faster |
| 64 KiB split into 64 B writes | 1.167 ms | 1.102 ms | 5.6% faster |
| 64 KiB split into 1 KiB writes | 1.129 ms | 1.037 ms | 8.2% faster |
| 64 KiB split into 16 KiB writes | 1.081 ms | 1.080 ms | No meaningful change |

For payloads larger than 8 KiB, non-empty downstream write attempts are approximately halved because the old 8 KiB buffering boundary is replaced by the 16 KiB encryption buffer.

These are encryption-path microbenchmark results, not total server CPU measurements. The largest benefit is expected for fragmented writes and under real downstream backpressure.

## Memory

Each encrypted connection owns a reusable 16 KiB ciphertext buffer.

The previous one-byte pending state has been removed. The additional memory is approximately 16 KiB per encrypted connection, or approximately 1.6 MiB for 100 encrypted connections.

## Tests

Added coverage for:

- Ciphertext equality against a reference AES-CFB8 implementation.
- Fragmented application writes.
- Large writes spanning multiple encryption buffers.
- Partial downstream writes.
- Alternating Pending and Ready behaviour.
- Ciphertext retained after Poll::Pending.
- Small writes coalesced until flush.
- WriteZero propagation.
- Flush with pending ciphertext.
- Shutdown with pending ciphertext.
- Encryption combined with compression.
- Rejecting encryption activation while unencrypted data remains buffered.

## Passed locally

- cargo fmt --check
- cargo clippy --all-targets --all-features
- cargo clippy --release --all-targets --all-features
- cargo test -p pumpkin-protocol --lib
- Protocol benchmark compilation and execution


RoosterBooster007:
@JamieLittle16 It looks like the bytewise encrypted-write path is still present after some changes. Are you still interested in getting this PR through (and/or #2489)? If so, please rebase and rerun the benchmark, and we'll review it.

RoosterBooster007:
CI is red on every platform because `encode_with_encryption` and `encode_with_compression_and_encryption` now read an empty buffer (the encryptor holds ciphertext until flush, and `build_packet_with_encoder` never flushes), and the description says the 8 KiB `BufWriter` goes away once encryption is on but the diff doesn't touch `pending.rs`, so we'd actually be double-buffering.

PR #2484, opened 2026-07-23
fix: combat between bedrock and java players

ZlordHUN:
## Description
- Uses the following PR as its foundation: https://github.com/Pumpkin-MC/Pumpkin/pull/2477 cause without it bedrock players cannot join pumpkin or see java players. With this PR bedrock players can hit java players.

- Issue was:
- Bedrock encodes the attack action type as a signed VarInt.
- Attack value `1` is ZigZag-encoded as wire value `2`.
- Pumpkin decoded it as an unsigned VarInt, producing action `2`.
- The combat handler only recognizes action `1`, so it discarded the attack.

## Testing
- cargo test -p pumpkin-protocol
- cargo test -p pumpkin-world bedrock_palette
- cargo fmt --all -- --check
 - Logged onto pumpkin with both a java and bedrock player and had the bedrock player kill the java player.

### DIFF
--- a/crates/pumpkin-protocol/src/bedrock/server/inventory_transaction.rs
+++ b/crates/pumpkin-protocol/src/bedrock/server/inventory_transaction.rs
@@ -270,3 +270,50 @@ impl PacketRead for SInventoryTransaction {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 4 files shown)


ZlordHUN:
Already fixed.

PR #2486, opened 2026-07-24
fix(pumpkin): restore terminal state when exiting during a blocked console read

shiminshen:
## Description

The console thread spends nearly all its time blocked inside rustyline's `readline()`, which puts the terminal into raw mode for the duration of the call. Every exit path (a graceful stop, a forced re-exit when already stopping, and a main-thread panic) calls `std::process::exit`, which skips destructors — so exiting while a read is in flight leaves the launching shell without local echo. This matches the reports in #2441: typing becomes invisible after a crash, and — as one commenter confirmed — it also persists after a normal `/stop`, since that path ends in the same unconditional `exit()` call at the end of `main()`.

This adds a small `restore_terminal()` helper that runs `stty sane` (guarded on stdin actually being a terminal, so headless/piped/Docker deployments are unaffected) and calls it at all three `exit()` call sites: the normal end-of-`main` exit, the forced re-exit in `stop_or_exit_server`, and the main-thread panic path in `handle_panic`.

## Testing

- [x] `cargo check -p pumpkin`
- [x] `cargo clippy -p pumpkin --all-targets --all-features`
- [x] `cargo fmt --check -p pumpkin`

This is fundamentally a terminal-state side effect that's only observable in a real interactive TTY session (launch the server from a terminal, trigger a crash or run `/stop`, and check whether the shell still echoes typed input afterward), so I wasn't able to exercise it end-to-end in a sandboxed environment. Reasoning through the fix: `stty sane` is a no-op whenever stdin isn't a terminal (guarded by `is_terminal()`), so it can't affect non-interactive/Docker deployments, and it directly reverses the raw-mode side effect `rustyline` leaves behind when a blocking `readline()` call is interrupted by `exit()`.

### DIFF
--- a/crates/pumpkin/src/lib.rs
+++ b/crates/pumpkin/src/lib.rs
@@ -31,7 +31,7 @@ use rustyline::history::FileHistory;
 use rustyline::{Config, error::ReadlineError};
 use std::collections::HashMap;
 use std::io::{ErrorKind, IsTerminal, stdin};
-use std::process::exit;
+use std::process::{Command, Stdio, exit};
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
 use std::sync::{Arc, OnceLock};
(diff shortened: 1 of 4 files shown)


DaRealNoteed:
> it also persists after a normal `/stop`

just want to note that it only persists after stop if the bug has already happened and the server is relaunched in same terminal. running /stop normally restores the terminal if it wasn't already in a bugged state

PR #2491, opened 2026-07-24
perf(world): preserve tick headroom during generation

JamieLittle16:
## Summary

  Bounds background chunk-generation admission so sustained exploration is less likely to starve latency-sensitive tick and network work.

  This does not reserve CPU cores. It reduces the number of simultaneously runnable generation workers and limits how much generation work can be queued ahead.

  ## Changes

  - Adds configurable generation concurrency:

  [world.generation]
  max_threads = 6

  When omitted, Pumpkin selects:

  | Logical CPUs | Generation workers|
  |---|---|
  |1|1|
  |2–4| N - 1|
  |Above 4| N - 2|

  - Uses one generation runtime shared by all dimensions and dynamically created worlds.
  - Globally bounds admitted generation work to worker_count + 2.
  - Holds admission permits until results are consumed or dropped, including failure and shutdown paths.
  - Separates generation and disk-I/O in-flight accounting.
  - Allows eligible reads to continue when generation admission is saturated.
  - Preserves scheduler priority when temporarily deferring saturated work.
  - Adds structured generation and tick-health telemetry.

  ## Motivation

  Rayon already limits actively executing CPU work to its worker count. The scheduler’s previous 4 × available_parallelism limit primarily allowed significant work to
  accumulate ahead of execution.

  That could:

  - retain more chunk-generation caches;
  - increase memory pressure;
  - delay reprioritisation;
  - commit resources to chunks that are no longer urgent;
  - increase contention with ticking and networking.

  The new global admission limit stays close to actual generation capacity while retaining two queued jobs to keep workers supplied.

  Disk reads are accounted for separately, so limiting CPU-bound generation does not unnecessarily block I/O.

  ## Instrumentation

  Generation telemetry includes:

  - ready generation and I/O work;
  - dependency-waiting tasks;
  - local and global admitted work;
  - admission limit;
  - jobs admitted, completed and failed;
  - chunks reaching Full;
  - average duration per generation stage.

  Tick telemetry includes rolling:

  - p50, p95 and p99 MSPT;
  - maximum MSPT;
  - ticks exceeding the configured tick budget.

  Enable it with:

  RUST_LOG=pumpkin_world::generation_metrics=debug,pumpkin::tick_metrics=debug

  ## Benchmark

  Tested on an Intel Core Ultra 7 155H using the existing concurrent full-generation benchmark:

  16 chunks
  4 Rayon workers
  3 alternating runs per revision
  10 samples per run

 |Revision| Median runtime| Approx. throughput|
 |---|---|---|
 |Master| 159.38 ms| 100.39 chunks/s|
 |This PR|160.83 ms| 99.48 chunks/s|

  The approximately 0.9% difference is within observed run-to-run variation, indicating no material raw-generation throughput regression.

  This benchmark bypasses the live scheduler and tick loop, so it does not measure the intended tail-latency improvement. The added telemetry enables a follow-up scripted
  exploration comparison using p99 MSPT, ticks over budget, requested-chunk latency and connection responsiveness.

  ## Validation

  - cargo test -p pumpkin-world --lib - 101 passed
  - New configuration and admission-limit tests pass
  - Clippy passes with warnings denied across pumpkin, pumpkin-world and pumpkin-config
  - Formatting and git diff --check pass

  Adaptive generation concurrency based on tick health remains suitable for a follow-up. This PR establishes the configuration, accounting, global admission control and
  instrumentation needed to implement it safely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2602,7 +2602,6 @@ dependencies = [
  "pumpkin-util",
  "pumpkin-world",
  "rand",
- "rayon",
  "rsa",
  "rustc-hash",
  "rustc_version_runtime",
--- a/pumpkin-config/src/world.rs
+++ b/pumpkin-config/src/world.rs
@@ -1,6 +1,38 @@
 use serde::{Deserialize, Serialize};
+use std::num::NonZeroUsize;
 
 use crate::{chunk::ChunkConfig, lighting::LightingEngineConfig};
+
+/// Controls the CPU workers used for background chunk generation.
+#[derive(Deserialize, Serialize, Default, Clone)]
+#[serde(default)]
+pub struct GenerationConfig {
+    /// Maximum number of Rayon workers used for chunk generation.
+    ///
+    /// When omitted, Pumpkin chooses a conservative value from the logical CPU count.
+    pub max_threads: Option<NonZeroUsize>,
+}
+
+impl GenerationConfig {
+    /// Resolves the configured worker count for a host reporting `available_cpus`.
+    #[must_use]
+    pub fn resolve_threads(&self, available_cpus: usize) -> usize {
+        self.max_threads.map_or_else(
+            || default_generation_threads(available_cpus),
+            NonZeroUsize::get,
+        )
+    }
+}
+
+/// Chooses a low-core-aware default while leaving some scheduling headroom.
+#[must_use]
+pub const fn default_generation_threads(available_cpus: usize) -> usize {
+    match available_cpus {
+        0 | 1 => 1,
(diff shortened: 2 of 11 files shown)


JamieLittle16:
Closing rather than rebasing this onto the current architecture. Chunk generation, packet offload, and scheduling have moved substantially since this PR, with meaningful overlap in the original goal of preserving tick/runtime headroom.

At this point a rebase would effectively mean designing a second scheduling layer around newer infrastructure, which isn’t worthwhile. The ideas and measurements here can still be useful context for future scheduler work.

PR #2492, opened 2026-07-24
perf(world): bulk-build generated chunk palettes

JamieLittle16:
## Summary

Speed up the final conversion from a generated `ProtoChunk` into a loaded `ChunkData`.

The old path populated empty block and biome palettes through roughly 98,000 individual mutation calls for a normal Overworld chunk. Those mutations maintain bookkeeping intended for live block updates. The new path bulk-builds each palette in one pass, then constructs random-tick metadata once.

The conversion also reuses the heightmaps already calculated during generation instead of rescanning the final chunk.

## Performance

Release Criterion results on the same machine and workload:

| Benchmark | Before | After | Change |
| --- | ---: | ---: | ---: |
| Final proto-to-level conversion | 667.22 us | 289.45 us | 56.6% faster |
| Full generated chunk pipeline | 43.62 ms | 41.47 ms | 4.9% faster |

The conversion itself is now much cheaper, but generation stages such as noise, features, and lighting still dominate total fresh-chunk time. The full-pipeline figure is therefore the useful expectation for exploration.

Also adds an NBT deserialisation baseline for already-generated terrain: 188.84 us median for a representative generated chunk. This is a measurement baseline only, not a claimed improvement.

## Correctness

- Bulk palettes are checked against the individual-mutation implementation for homogeneous, indirect, direct, liquid, and randomly ticking sections.
- Generated heightmaps are checked against a complete recalculation.
- Existing Anvil and other world-format round-trip tests remain in the suite.

## Passed locally

- Strict clippy for pumpkin and pumpkin-world
- `cargo test -p pumpkin-world --lib`

### DIFF
--- a/pumpkin-world/Cargo.toml
+++ b/pumpkin-world/Cargo.toml
@@ -68,9 +68,9 @@ serde_json5.workspace = true
 name = "chunk"
 harness = false
 
-#[[bench]]
-#name = "chunk_io"
-#harness = false
+[[bench]]
+name = "chunk_io"
+harness = false
 
 [[bench]]
 name = "chunk_gen"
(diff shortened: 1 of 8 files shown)


Snowiiii:
LGTM, Thank you @JamieLittle16

PR #2494, opened 2026-07-24
fix(elytra): restore vanilla flight and firework behavior

MegalithOfficial:
## Description
This PR will fix Elytra flight. Elytra can already be activated, but most of its server-side behavior is currently missing or incomplete.

While testing it, I noticed that you can never actually land after a glide. When the player touches the ground, they remain stuck in the fall-flying state instead of returning to normal movement.

Landing also deals fall damage based on the full height descended from. The current code keeps accumulating normal fall distance while the player is gliding, so landing is handled like one long fall.

I also noticed that firework rockets boost the player, but the rocket itself detaches and moves around incorrectly. The attached-player metadata uses the wrong tracked data entry for 1.21.11 and is sent before the rocket is spawned. The rocket also runs normal projectile movement before being moved back to the player every tick, which causes the visible jittering.

After comparing this with the Mojang's official server code, I found that fall-flying is currently implemented mostly as an entity flag and pose. The glider validation, dedicated flight movement, fall-distance handling, collision damage, durability loss, landing behavior, and flight termination conditions are still missing.

Firework behavior is also incomplete. Rockets do not use their actual item data for flight duration and explosion effects, are not consumed correctly and do not implement vanilla explosion damage.

I will fix this as a complete Elytra implementation instead of only patching the landing damage and detached firework symptoms. This will include the full flight lifecycle, vanilla movement and collision handling, durability, correct firework attachment and the remaining firework behavior.


MegalithOfficial:
I Implemented and pushed the remaining fixes.

Elytra flight now follows the vanilla lifecycle, including activation validation, flight movement, landing, fall-distance handling, collision damage and durability loss. I also fixed the tracked metadata compatibility issue affecting 1.21.11.

Firework rockets now remain attached while boosting, use their actual flight duration, launch correctly when placed and are consumed after use.

Firework explosion effects and damage are still unimplemented. I originally included the remaining firework behavior in the planned scope, but I am leaving explosions for a separate PR since they are not required for the Elytra and boosting fixes here.

RoosterBooster007:
@MegalithOfficial there have been several elytra/flight fixes recently, but server-side elytra flight is still one of the big missing pieces afaik. Please fix conflicts.

MegalithOfficial:
> @MegalithOfficial there have been several elytra/flight fixes recently, but server-side elytra flight is still one of the big missing pieces afaik. Please fix conflicts.

I will redo the PR. The conflicts are way too big that it requires me to rewrite it from scratch. I will pull this PR into draft mode tomorrow.

PR #2496, opened 2026-07-24
perf(world): use dense chunk lighting visitation

JamieLittle16:
## Summary

Chunk lighting repeatedly tracks propagated block positions and looks up neighbouring surface heights. Both operations are constrained to the light cache, but the previous implementation used hash tables keyed by absolute positions. This replaces those hot-path hash table lookups with cache-relative, contiguous storage.

- Replace the per-propagator FxHashSet<BlockPos> with a reusable DenseVisited bitset. Each cache block has one bit, so visited and out-of-cache checks are handled by one bounds-checked index lookup and bit test.
- Reuse the bitset between propagation runs and only resize it when the cache layout changes.
- Replace the temporary 18 by 18 skylight surface-height FxHashMap with a fixed 324-element array.
- Keep propagation semantics unchanged: positions outside the cached lighting region are skipped, and a position is processed at most once per propagation pass.

This targets chunk loading directly. Lighting visits neighbours for every propagated block, so avoiding hashes, tuple construction, hasher work, and per-run map allocation reduces CPU time while chunks are generated or relit.

## Benchmark results

Benchmarks were run in release mode on the same machine, comparing the parent commit with this branch. The focused result measures lighting generation; the full-pipeline result includes the surrounding chunk-generation work and is the more useful expectation for normal chunk loading.

| Scenario | Before | After | Change |
| --- | ---: | ---: | ---: |
| Focused lighting generation | 19.575 ms | 12.464 ms | 36.3% faster |
| Full chunk-generation pipeline, 20 s sample | 48.368 ms | 44.520 ms | 8.0% faster |

The 36.3% figure is for the isolated lighting portion, not a claim that all chunk generation becomes 36.3% faster. In the end-to-end benchmark, work such as terrain generation, structure work, and chunk assembly remains unchanged, which is why the measured overall reduction is 8.0%. The exact benefit in a server will vary with world generation settings, cache size, terrain, and how much lighting work a player triggers.

## Correctness and safety

The dense map is indexed from the cache origin and validates x, y, and z before accessing the bitset. Out-of-range positions continue to be rejected. Its layout includes the cache origin and vertical range, so moving to another cache cannot reuse stale coordinate mappings. Clearing resets all bits before every independent propagation pass.

Unit coverage includes cache boundaries, negative coordinates, out-of-bounds rejection, clearing and reuse, and relayouts that would otherwise risk aliasing positions from a previous cache. Existing world tests cover the broader lighting and chunk paths.

## Passed locally

- cargo fmt --all -- --check
- git diff --check
- cargo clippy -p pumpkin-world --all-targets --all-features -- -D warnings
- cargo test -p pumpkin-world --lib

### DIFF
--- a/pumpkin-world/src/lighting/engine.rs
+++ b/pumpkin-world/src/lighting/engine.rs
@@ -11,11 +11,99 @@ use pumpkin_util::math::vector3::Vector3;
 use std::collections::VecDeque;
 //use std::time::Instant;
 
-// These are hit on every neighbour of every propagated block, so the hash
-// function dominates. Use rustc-hash's fast hasher instead of the std default
-// (SipHash), which is what "Fast" was meant to imply.
-type FastHashSet<K> = rustc_hash::FxHashSet<K>;
-type FastHashMap<K, V> = rustc_hash::FxHashMap<K, V>;
+const LIGHTING_REGION_WIDTH: usize = 18;
+
+/// Dense cache-relative visitation map used by light propagation.
+///
+/// Propagation never leaves the chunk cache, so hashing absolute block positions
+/// wastes work and memory. A bit per cache block also lets the hot path combine
+/// the cache-boundary and visited checks.
+struct DenseVisited {
+    bits: Vec<u64>,
+    origin_x: i32,
+    origin_z: i32,
+    min_y: i32,
+    width: usize,
+    height: usize,
+}
+
+impl DenseVisited {
+    const fn new() -> Self {
+        Self {
+            bits: Vec::new(),
+            origin_x: 0,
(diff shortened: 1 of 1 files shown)


RoosterBooster007 (approved):

RoosterBooster007:
@JamieLittle16, it looks like the lighting engine hasn't changed. Rebase this PR, and we'll get it reviewed.

JamieLittle16:
Closing as superseded. The dense lighting visitation/fixed surface-height optimisation this PR was targeting is now present on master through later work, so there’s no value in maintaining a parallel implementation here.

PR #2499, opened 2026-07-24
fix(auth): don't require signatureRequired in texture profiles

### LINKED ISSUE #301: Missing field 'signatureRequired' when using drasl auth server

### Operating System

Ubuntu 24.04

0.1.0/00af2b98683b11946cddb4f08de3b0ce04c1fddd

### What happened?

I've set up an drash auth server and changed the `auth_url` in `features.toml` to point to the auth server.
When I try to join the server, I get an error: 
```
[pumpkin::client@tokio-runtime-worker] Kicking Client id 0 for Texture Error Failed to parse JSON from player texture: missing field `signatureRequired` at line 1 column 279
```

The response returned by my auth server at `https://<hostname>/session/minecraft/hasJoined?username={username}&serverId={server_hash}` looks like this:
```
{
    "id": "<an-id>",
    "name": "stefan",
    "properties": [
        {
            "name": "textures",
            "value": "<value>",
            "signature": <signature>"
        }
    ]
}
```

there is indeed no `signatureRequired` property. I removed the `signature_required` property from the `ProfileTextures` struct in `src/client/authentication.rs` and I could join the world.

### To Reproduce

- Set up drasl auth server
- Create account
- Change `auth_url` in `features.toml`
- Run server
- Try to join server from client authenticated through drasl

### Expected behavior

The server should let you join

Sawlyer:
## Description

Third-party auth servers (drasl, Blessing Skin, littleskin.cn) don't send the `signatureRequired` field in the `hasJoined` texture property JSON. `ProfileTextures` declared it as a mandatory `bool`, so `serde_json` failed to deserialize the profile in `validate_textures` and the player was kicked.

The field is unused (the struct is `#[expect(dead_code)]`), so this just marks it `#[serde(default)]` to default it to `false` when absent. Mojang responses are unaffected.

Fixes #301

## Testing

Added unit tests deserializing a profile both without and with `signatureRequired`; both parse.

`cargo test -p pumpkin`, `cargo clippy -p pumpkin` and `cargo fmt --check` pass.

### DIFF
--- a/pumpkin/src/net/authentication.rs
+++ b/pumpkin/src/net/authentication.rs
@@ -19,6 +19,9 @@ pub struct ProfileTextures {
     timestamp: i64,
     profile_id: Uuid,
     profile_name: String,
+    // Mojang always sends this, but third-party auth servers (drasl, Blessing Skin, ...)
+    // omit it. It is unused here, so default it instead of failing to parse the profile.
+    #[serde(default)]
     signature_required: bool,
     textures: HashMap<String, Texture>,
 }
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thanks @Sawlyer

PR #2503, opened 2026-07-24
fix(net): don't kick players for attacking a removed entity

### LINKED ISSUE #1800: attemping to attack an invalid entity

### Operating System

Ubuntu

b6ecb9a

### What happened?

When attacking multiple mobs in survival the player sometimes randomly gets kicked with message: attemping to attack an invalid entity

Console log: Player id 216 interacted with entity id 255, which was not found.

### To Reproduce

Spawn multiple mobs like zombies, attack until error occurs

### Expected behavior

Player should not get kicked

Sawlyer:
## Description

When an attack/interact packet references an entity id the server can no longer find, `handle_attack` and the unknown-entity branch of `handle_interact` disconnected the player with `MULTIPLAYER_DISCONNECT_INVALID_ENTITY_ATTACKED` ("attempting to attack an invalid entity").

This is a normal race: the target can die or despawn between the client sending the packet and the server handling it. Vanillas `handleInteract` ignores a null target instead of kicking.

Both not-found branches now log at `debug` and return. The separate anti-cheat check for a client claiming to attack its own entity id is untouched and still kicks.

Fixes #1800

## Testing

`cargo test -p pumpkin`, `cargo clippy -p pumpkin` and `cargo fmt --check` pass. The change is control flow inside async packet handlers, which are not covered by the existing unit tests, so it is verified by matching vanilla behaviour rather than a new test.

### DIFF
--- a/crates/pumpkin/src/entity/mod.rs
+++ b/crates/pumpkin/src/entity/mod.rs
@@ -797,7 +797,7 @@ pub trait EntityBase: Send + Sync + std::any::Any {
     }
 }
 
-#[derive(Clone, Copy, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum RemovalReason {
     Killed,
     Discarded,
--- a/crates/pumpkin/src/net/java/play/attack.rs
+++ b/crates/pumpkin/src/net/java/play/attack.rs
@@ -32,12 +32,14 @@ impl JavaClient {
             .map(|p| Arc::clone(p) as Arc<dyn EntityBase>)
             .or_else(|| world.get_entity_by_id(entity_id.0));
         let Some(target) = target else {
-            self.kick(TextComponent::translate_cross(
-                translation::java::MULTIPLAYER_DISCONNECT_INVALID_ENTITY_ATTACKED,
-                translation::java::MULTIPLAYER_DISCONNECT_INVALID_ENTITY_ATTACKED,
-                [],
-            ))
-            .await;
+            // The target was removed between the client sending the packet and us
+            // handling it (e.g. it died or despawned). Vanilla ignores this rather
+            // than kicking the player.
+            debug!(
+                "Player id {} attacked entity id {}, which was not found.",
+                player.entity_id(),
+                entity_id.0
+            );
             return;
         };
         if let Some(player_victim) = &player_target {
--- a/crates/pumpkin/src/net/java/play/attack.rs
+++ b/crates/pumpkin/src/net/java/play/attack.rs
@@ -36,9 +36,10 @@ impl JavaClient {
             // handling it (e.g. it died or despawned). Vanilla ignores this rather
             // than kicking the player.
             debug!(
-                "Player id {} attacked entity id {}, which was not found.",
+                "Player id {} attacked entity id {}, which was not found. {}",
                 player.entity_id(),
-                entity_id.0
+                entity_id.0,
+                describe_missing_entity(&world, entity_id.0)
             );
             return;
         };
--- a/crates/pumpkin/src/net/java/play/interact.rs
+++ b/crates/pumpkin/src/net/java/play/interact.rs
@@ -131,9 +131,10 @@ impl JavaClient {
                         // and us handling it (e.g. it died or despawned). Vanilla ignores
                         // this rather than kicking the player.
                         debug!(
-                            "Player id {} attacked entity id {}, which was not found.",
+                            "Player id {} attacked entity id {}, which was not found. {}",
                             player.entity_id(),
-                            event.entity_id
+                            event.entity_id,
+                            describe_missing_entity(&world, event.entity_id)
                         );
                     }
                 }
--- a/crates/pumpkin/src/net/java/play/interact.rs
+++ b/crates/pumpkin/src/net/java/play/interact.rs
@@ -127,13 +127,14 @@ impl JavaClient {
 
                 'after: {
                     if event.action == ActionType::Attack {
-                        error!(
-                            "Player id {} interacted with entity id {}, which was not found.",
+                        // The target was removed between the client sending the packet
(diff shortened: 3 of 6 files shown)


MegalithOfficial:
Wouldn't it be better to figure out why removed entities remain present?

Sawlyer:
> Wouldn't it be better to figure out why removed entities remain present?

The race is inherent: the client picks its target from the world state it had a round-trip ago, so an attack packet can already be on the wire when the entity is removed. Perfect removal broadcasting doesn't close that window, which is why vanilla null-checks rather than reacting to it. Also worth pinning down what this disconnect means in vanilla. In handleAttack a null target is skipped silently (if (target != null && ...)), and multiplayer.disconnect.invalid_entity_attacked is only sent when the target exists but is never attackable: ItemEntity, ExperienceOrb, the player itself, or a non-attackable arrow. Pumpkin was reusing that message for "not found", which is a different condition. If entities do additionally linger client-side after removal, that's a real bug and worth its own issue, but kicking is the worst way to find it: the player loses their session and we get nothing to debug. The debug line here logs the entity id, which is what you'd actually grep for. I did check the removal paths, remove_entity and remove_player both broadcast CRemoveEntities, so I don't have a concrete ghost path to point at. 

Oh and also one gap I noticed while doing this: Pumpkin doesn't implement vanilla's real use of this disconnect (attacking item entities / XP orbs / non-attackable arrows). Can add it here or as a follow-up.

MegalithOfficial:
> > Wouldn't it be better to figure out why removed entities remain present?
> 
> The race is inherent: the client picks its target from the world state it had a round-trip ago, so an attack packet can already be on the wire when the entity is removed. Perfect removal broadcasting doesn't close that window, which is why vanilla null-checks rather than reacting to it. Also worth pinning down what this disconnect means in vanilla. In handleAttack a null target is skipped silently (if (target != null && ...)), and multiplayer.disconnect.invalid_entity_attacked is only sent when the target exists but is never attackable: ItemEntity, ExperienceOrb, the player itself, or a non-attackable arrow. Pumpkin was reusing that message for "not found", which is a different condition. If entities do additionally linger client-side after removal, that's a real bug and worth its own issue, but kicking is the worst way to find it: the player loses their session and we get nothing to debug. The debug line here logs the entity id, which is what you'd actually grep for. I did check the removal paths, remove_entity and remove_player both broadcast CRemoveEntities, so I don't have a concrete ghost path to point at.
> 
> Oh and also one gap I noticed while doing this: Pumpkin doesn't implement vanilla's real use of this disconnect (attacking item entities / XP orbs / non-attackable arrows). Can add it here or as a follow-up.

I understand that a short race is unavoidable, but we’re assuming that’s what caused #1800 without actually verifying it. Can we temporarily retain removal information and log the removal tick/reason, whether the player was tracking the entity, and when CRemoveEntities was broadcast? Then we can distinguish a packet arriving one or two ticks late from an entity genuinely lingering client-side.

Not kicking is reasonable, but I don’t think it should replace investigating why these unknown-entity attacks are occurring.

KamelotWalker:
We hit this exact bug on a live Pumpkin server (0.1.0-dev+26.2-26.40): a player was disconnected with `multiplayer.disconnect.invalid_entity_attacked` ("Attempting to attack an invalid entity") when attacking a mob right after a far teleport + death, reproducible twice in a row. The attacked entity simply was not in `world.entities` anymore when the packet was handled.

Before finding this PR we applied the same approach locally: both not-found branches in `handle_attack`/`handle_interact` now log at debug and return, keeping the self-attack anti-cheat kick intact. `cargo check --workspace` passes and the server boots and runs cleanly with it. (We deployed it, but no player sessions happened since, so live stress-testing is still pending on our side.)

The reasoning in this PR matches what we observed: the race window is inherent — the client targets an entity from a world state it had a round-trip ago — and vanilla null-checks instead of kicking. Kicking here costs the player the whole session for a normal combat race, so I'd love to see this merged.

Sawlyer:
@MegalithOfficial rebased on master, and added the removal record you asked for: the world keeps the last 256 removals (id, reason, and when the entity was dropped) and the debug line prints the matching one, so a packet arriving a tick or two late can be told apart from an entity that genuinely lingers client side.

The lint and flake failures are master, not this branch: since aee90883e, encryption_response.rs:46 has a bare `return;` in a function returning `Option<PacketHandlerResult>`.

MegalithOfficial:
> @MegalithOfficial rebased on master, and added the removal record you asked for: the world keeps the last 256 removals (id, reason, and when the entity was dropped) and the debug line prints the matching one, so a packet arriving a tick or two late can be told apart from an entity that genuinely lingers client side.
> 
> The lint and flake failures are master, not this branch: since aee90883e, encryption_response.rs:46 has a bare `return;` in a function returning `Option<PacketHandlerResult>`.
> 

If you want my honest opinion: you just invented unnecessary code and clearly don't know how Minecraft servers work. Plus, you replaced the kick message with a hardcoded string. That was working fine.

Im sorry If i sound harsh but thats what worth mentioning.

RoosterBooster007:
Hey @Sawlyer, thanks for digging into this. You're right that vanilla never kicks for a missing target, so the direction is solid. The pushback is mainly about the removal-history machinery riding along with it; that's a lot of added state in `World` just to decorate one debug line, and @Snowiiii quietly closed the same minimal fix in #2386 last month. We'll need his sign-off on whether the no-kick approach is wanted at all. If it is, the cleanest path is the two-line fix on its own, with the diagnostic bit either trimmed way down or as a separate PR. Also needs a rebase to the current `try_kick` handlers and a clean CI run either way.

Developers13:
Hit this exact disconnect in live play while investigating #3435 — in our case the "removed entity" was one of the ghost copies produced by the entity duplication on chunk unload/reload described there (client keeps one visible golem, server-side lookup can no longer resolve it, `handle_attack` kicks). The guard proposed here would turn those ghost mobs into harmless no-ops instead of disconnects.

PR #2504, opened 2026-07-24
fix(data): don't panic when the data directory can't be created

### LINKED ISSUE #2057: Panic when Docker start

### Operating System

Ubuntu 24

db885f4

### What happened?

When I started using Docker to run Pumpkin, the container stopped immediately. Here is the error message:
```
Backtrace:
   0: pumpkin::handle_panic
   1: std::panicking::panic_with_hook
   2: std::panicking::panic_handler::{closure#0}
   3: std::sys::backtrace::__rust_end_short_backtrace::<std::panicking::panic_handler::{closure#0}, !>
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: core::result::unwrap_failed
   7: pumpkin::main::{{closure}}
   8: pumpkin::main
   9: std::sys::backtrace::__rust_begin_short_backtrace
  10: main
  11: <unknown>
```

### To Reproduce

Here is my docker compose file:
```yaml
services:
  pumpkin:
    image: ghcr.io/pumpkin-mc/pumpkin:master
    container_name: minecraft
    ports:
      - 25565:25565
    volumes:
      - ./pumpkin:/pumpkin
    restart: unless-stopped
```

### Expected behavior

The container runs normally, just as it does when using `cargo run --release`.

Sawlyer:
## Description

`VanillaData` configs (bans, ops, whitelist, usercache) panicked with a raw backtrace when the `data/` directory could not be created. This commonly happens in Docker when the container runs as a non-root user without write permission on its working directory, so the server crashed on startup with a Rust panic instead of a usable message.

On load, log a clear error explaining the permission problem and exit cleanly, matching how other fatal startup errors are handled in `lib.rs`. On save, log a warning and skip the write instead of crashing a running server.

This does not make the server run without permissions, which is inherent to the setup. It replaces the panic with an actionable message.

Fixes #2057

## Testing

`cargo test -p pumpkin`, `cargo clippy -p pumpkin` and `cargo fmt --check` pass. The change is filesystem error handling on the startup path, which the existing unit tests do not cover.

### DIFF
--- a/crates/pumpkin/src/main.rs
+++ b/crates/pumpkin/src/main.rs
@@ -60,9 +60,11 @@ async fn main() {
 
     let config = PumpkinConfig::load(&exec_dir);
 
-    let vanilla_data = VanillaData::load();
-
+    // Install the logger before loading the vanilla data files, so their errors
+    // (for example an unwritable data directory) actually reach the console.
     pumpkin::init_logger(&config.advanced);
+
+    let vanilla_data = VanillaData::load();
 
     info!(
         "{}",
--- a/crates/pumpkin/src/data/mod.rs
+++ b/crates/pumpkin/src/data/mod.rs
@@ -46,7 +46,7 @@ pub trait LoadJSONConfiguration {
     {
         let exe_dir = env::current_dir().unwrap_or_else(|_| std::path::PathBuf::from("."));
         let data_dir = exe_dir.join(DATA_FOLDER);
-        debug!("creating new data root folder");
+        debug!("ensuring the data root folder exists");
         if let Err(err) = fs::create_dir_all(&data_dir) {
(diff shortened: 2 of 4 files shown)


Sawlyer:

crates/pumpkin/src/main.rs
```diff
@@ -60,10 +60,12 @@ async fn main() {
 
     let config = PumpkinConfig::load(&exec_dir);
 
-    let vanilla_data = VanillaData::load();
-
+    // Install the logger before loading the vanilla data files, so their errors
+    // (for example an unwritable data directory) actually reach the console.
```
QuantumSegfault:
```suggestion
```

PR #2505, opened 2026-07-24
fix(entity): don't panic reading a missing attribute base value

Sawlyer:
## Description

`get_attribute_base` unwrapped the result of looking up an entity types registry base value, so any attribute the type does not declare panicked the server. `/effect give <player> <attribute-backed effect>` could hit this through `send_attribute_updates_for_living` and crash.

`update_attribute` right next to it already handled the same lookup gracefully. Extract that logic into `registry_base_value`, which falls back to the attributes default value with a warning, and use it in both places.

Fixes #1985

## Testing

Added unit tests for `registry_base_value`: a type declaring no attributes (`EntityType::ITEM`) falls back to the attribute default instead of panicking, and a type that does declare it returns the registry value.

`cargo test -p pumpkin`, `cargo clippy -p pumpkin` and `cargo fmt --check` pass.

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -3341,18 +3341,15 @@ mod tests {
         );
     }
 
-    /// When the type does declare the attribute, its registry value is returned.
+    /// When the type does declare the attribute, its registry value is returned and
+    /// not the attribute default. Zombies walk at 0.23, while the default movement
+    /// speed is 0.7, so the two cases can't be confused.
     #[test]
     fn registry_base_value_returns_registry_value_when_present() {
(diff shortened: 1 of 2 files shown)


Sawlyer:
Rebased on master. get_attribute_base no longer unwraps upstream, so what is left here is the shared lookup, the warning and the tests.

RoosterBooster007:
Hey @Sawlyer, thanks for sticking with this one and rebasing. The panic itself is already handled on master (`get_attribute_base` uses `map_or` now, and the `/effect` path does a single `get_effect` lookup), so what's left here is really the dedupe plus tests. Maybe retitle so it isn't billed as a fix?

PR #2506, opened 2026-07-24
perf(world): optimize chunk noise generation and buffer allocations

JamieLittle16:
# perf(world): optimize chunk noise generation and buffer allocations

## Summary

This PR optimizes the noise block population stage (`populate_noise_stage`) and density function buffer allocations in `pumpkin-world`, achieving a **56.4% reduction in chunk noise generation time** (**2.30x speedup**, or **+130% chunk generation throughput**) while preserving **100% bit-for-bit vanilla terrain parity**.

---

## Benchmark Results

Evaluated using Criterion (`cargo bench --package pumpkin-world --bench noise_gen`) comparing `master` against this PR (`perf/noise-generation`):

| Metric | `master` (Baseline) | This PR (`perf/noise-generation`) | Improvement |
| :--- | :--- | :--- | :--- |
| **`populate_noise_stage` Time** | **43.14 ms / chunk** | **18.80 ms / chunk** | **−56.4% (−24.34 ms)** |
| **Generation Rate / Throughput** | Baseline (1.00x) | **2.30x faster** | **+130% chunk gen rate** |
| **Bit-for-Bit Terrain Parity** | Identical | Identical | **100% Verified** |

### Criterion Output

```text
populate_noise_stage    time:   [18.336 ms  18.802 ms  19.310 ms]
                        change: [−58.471%  −56.419%  −54.339%] (p = 0.00 < 0.05)
                        Performance has improved.
```

---

### 1. Direct Flat Array Indexing & Heightmap Pruning (`proto_chunk.rs`)
- **`set_block_state_direct`**: Introduced a direct block setter for local coordinates `(local_x, local_y, local_z)` during noise sampling. This bypasses `set_block_state`'s global-to-local coordinate conversions (`bottom_y`, `start_block_x/z` offsets) and vector destructuring overhead in the hot inner loop (evaluated 98,304 times per chunk).
- **Subterranean Heightmap Pruning**: Leveraged top-down Y-iteration in `populate_noise`. Heightmap scans (surface, ocean floor, motion blocking, leaves) are only evaluated when updated, eliminating ~90% of redundant heightmap branch checks for subterranean blocks.

### 2. Zero-Boxing Buffer Recycling (`chunk_density_function.rs`)
- **Refactored `F64_BUFFER_POOL`**: Modified `F64_BUFFER_POOL` to store and recycle `Vec<f64>` instances directly.
[...]

PR #2508, opened 2026-07-24
perf(world): add 2D flat-Y fast-path for surface noise sampling

JamieLittle16:
## Description

Introduces a 2D fast-path (`sample_flat_xz`) for `DoublePerlinNoiseSampler`, `OctavePerlinNoiseSampler`, and `PerlinNoiseSampler` that eliminates redundant Y-axis math when sampling surface noise at a fixed Y = 0.0.

Surfaces in Minecraft's terrain builder (`SurfaceTerrainBuilder`) and material condition rules consistently call noise samplers with a hard-coded `y = 0.0`. The standard `sample(x, 0.0, z)` path computes:
- A full Y fade curve (`6t⁵ − 15t⁴ + 10t³`)
- Y-axis gradient dot products at all 4 Y-faces of the unit cube
- Y-axis lerp interpolation across those results

When Y = 0.0, the Y fractional offset is a constant determined entirely by the Y-origin offset of the sampler, and no Y-scale clamping is applied. The new `sample_flat_xz` path short-circuits this by forwarding directly to the existing `sample_no_fade(x, 0.0, z, 0.0, 0.0)` path, which is guaranteed bit-for-bit identical to `sample(x, 0.0, z)`.

### `pumpkin-util`
- Added `PerlinNoiseSampler::sample_flat_xz(&self, x, z)` — convenience method forwarding to `sample_no_fade(x, 0.0, z, 0.0, 0.0)`.
- Added `OctavePerlinNoiseSampler::sample_flat_xz(&self, x, z)` — maps octave lacunarity/precision and calls `sample_flat_xz` on each `PerlinNoiseSampler`.
- Added bit-exact parity unit test (`sample_flat_xz_bit_exact_parity`): asserts `sample(x, 0.0, z) == sample_flat_xz(x, z)` for 400 coordinate pairs across multiple seeds.

### `pumpkin-world`
- Added `DoublePerlinNoiseSampler::sample_flat_xz(&self, x, z)` — samples both internal octave samplers with `sample_flat_xz`.
- Updated all `SurfaceTerrainBuilder` noise calls to use `sample_flat_xz`:
  - `place_badlands_pillar`: badlands surface noise, pillar noise, pillar roof noise
  - `place_iceberg`: iceberg surface noise, pillar noise, pillar roof noise
  - `get_terracotta_block`: terracotta band offset noise
- Updated `test_noise_threshold` in surface material rules to use `sample_flat_xz`.
- Added `surface_bench` Criterion benchmark suite with:
  - `build_surface_stage` — aggregate surface stage throughput
  - `double_perlin_sample_3d_y0` — baseline 3D sampling at Y=0
  - `double_perlin_sample_flat_xz` — optimized 2D sampling

## Benchmark Results

Per-call micro-benchmark (256 calls, `DoublePerlinNoiseSampler::BADLANDS_SURFACE`):

| Method | Time (256 calls) | Per-call |
|---|---|---|
| `sample(x, 0.0, z)` | 91.86 µs | ~359 ns |
| `sample_flat_xz(x, z)` | 82.67 µs | ~323 ns |
| **Improvement** | **−10.0%** | **−36 ns/call** |

> **Note on aggregate benchmark**: The `build_surface_stage` aggregate benchmark shows no change at chunk (0,0) with seed 42, because `place_badlands_pillar` and `place_iceberg` are biome-gated (only fire in `ERODED_BADLANDS` / frozen ocean biomes). The per-call improvement is real and measurable in isolation.

## Verification

- **Bit-exact parity**: `sample_flat_xz_bit_exact_parity` — 400 coordinate pairs, bit-for-bit equality verified.
- **Existing deterministic checksums**: All 4 proto-chunk generation tests pass unchanged.
- `cargo fmt --check` and `cargo clippy` pass with no warnings.

## Related

IMPORTANT: Follows on from #2506 (buffer recycling + `set_block_state_direct`). This branch is based on `perf/noise-generation`.
<details>

## Mathematical Equivalence
`sample_flat_xz(x, z)` is **bit-for-bit identical** to `sample(x, 0.0, z)` for all inputs. Both paths terminate at the same `sample_no_fade(x, 0.0, z, 0.0, 0.0)` call - the only eliminated work is the redundant `0.0 * lacunarity` multiply and `maintain_precision(0.0)` (which always returns `0.0`):

* sample(x, 0.0, z) - existing path for each octave: mapped_y = maintain_precision(0.0 * lacunarity)
* always 0.0 sample_no_fade(mapped_x, 0.0, mapped_z, 0.0, 0.0)
* sample_flat_xz(x, z) - new path for each octave: sample_no_fade(mapped_x, 0.0, mapped_z, 0.0, 0.0) 
* identical result

No approximation or rounding difference is introduced. The guarantee is enforced by `sample_flat_xz_bit_exact_parity`, which asserts `assert_eq!(sample(x, 0.0, z), sample_flat_xz(x, z))` using Rust's exact `f64` bit equality across 400 coordinate pairs.

### DIFF
--- a/pumpkin-world/Cargo.toml
+++ b/pumpkin-world/Cargo.toml
@@ -84,6 +84,10 @@ harness = false
 name = "noise_router"
 harness = false
 
+[[bench]]
+name = "noise_gen"
+harness = false
+
 
 [lints]
 workspace = true
--- a/pumpkin-world/benches/noise_gen.rs
+++ b/pumpkin-world/benches/noise_gen.rs
@@ -1,0 +1,117 @@
+use criterion::{BatchSize, Criterion, criterion_group, criterion_main};
+use pumpkin_config::lighting::LightingEngineConfig;
+use pumpkin_data::BlockStateId;
+use pumpkin_data::dimension::Dimension;
+use pumpkin_util::world_seed::Seed;
+use pumpkin_world::ProtoChunk;
+use pumpkin_world::chunk_system::{Cache, Chunk, StagedChunkEnum};
+use pumpkin_world::generation::generator::WorldGenerator;
+use pumpkin_world::generation::get_world_gen;
+use pumpkin_world::world::WorldPortalExt;
+use std::hint::black_box;
+
+const SEED: Seed = Seed(42);
+
+struct BlockRegistry;
+impl WorldPortalExt for BlockRegistry {
+    fn can_place_at(
+        &self,
+        _block: &pumpkin_data::Block,
+        _state: &pumpkin_data::BlockState,
+        _block_accessor: &dyn pumpkin_world::world::BlockAccessor,
+        _block_pos: &pumpkin_util::math::position::BlockPos,
+    ) -> bool {
+        true
+    }
+
+    fn mirror(
+        &self,
+        block: &pumpkin_data::Block,
+        state_id: BlockStateId,
+        mirror: pumpkin_data::Mirror,
+    ) -> &'static pumpkin_data::BlockState {
+        block.mirror(state_id, mirror)
+    }
+
+    fn rotate(
+        &self,
+        block: &pumpkin_data::Block,
+        state_id: BlockStateId,
+        rotation: pumpkin_data::Rotation,
+    ) -> &'static pumpkin_data::BlockState {
+        block.rotate(state_id, rotation)
+    }
+
+    fn spawn_mobs_for_chunk_generation(
+        &self,
+        _cache: &mut dyn pumpkin_world::generation::proto_chunk::GenerationCache,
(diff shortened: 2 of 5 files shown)


JamieLittle16:
Closing this for now because it is stacked on #2506 and the current version only demonstrates an isolated sampler microbenchmark improvement, not a measurable end-to-end surface-generation gain. Greater performance benefits were expected but it seems LLVM optimisations do most of the work. 

I’ll preserve the implementation and revisit it after #2506 is resolved, either as a clean standalone PR with representative biome benchmarks or as a more substantial fixed-Y specialisation.

PR #2510, opened 2026-07-24
security: return after failed Mojang auth; centralize join gates in add_player

kodareef5:
Part 2 of the hardening stack (stacked on #2509).

## Commits
- fix auth fallthrough and centralize join gates

## What and why
- On failed Mojang session auth the login handler kicked but did not return, and fell through to completing login with the unverified client-supplied profile/UUID. The handler now returns after the kick on every failure branch.
- A missing profile at config-acknowledge panicked the connection task; it now kicks cleanly.
- Duplicate UUID/name and ban/whitelist checks existed only on the encrypted Java login path. They are now centralized in Server::add_player, so every transport (Java offline, Velocity, BungeeCord, Bedrock) inherits them.
- The Mojang HTTPS calls are made with a real agent carrying connect/recv timeouts (the configured values were previously never applied; ureq defaults have no timeouts).
- The prevent_proxy_connections URL now actually substitutes the player IP ({ip}); it was a no-op.
- Player name validation tightened on Java, Velocity-forwarded, and Bedrock paths (empty/control/NUL/format-code characters rejected), which also closes a query-task panic on NUL names.

## Tests
cargo test -p pumpkin passes, including updated name-validation tests.

### DIFF
--- a/pumpkin-nbt/src/compound.rs
+++ b/pumpkin-nbt/src/compound.rs
@@ -44,6 +44,16 @@ impl NbtCompound {
     }
 
     pub fn skip_content<'a, R: NbtReadHelper<'a>>(reader: &mut R) -> Result<(), Error> {
+        Self::skip_content_depth(reader, 0)
+    }
+
+    pub(crate) fn skip_content_depth<'a, R: NbtReadHelper<'a>>(
+        reader: &mut R,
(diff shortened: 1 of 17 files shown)


Snowiiii:
use Atomic PRs

PR #2511, opened 2026-07-24
security: fail-closed RCON/proxy configuration; fix RCON connection limits

kodareef5:
Part 3 of the hardening stack (stacked on #2510).

## Commits
- fail-closed RCON/proxy config and fix RCON limits

## What and why
- RCON enabled with the default empty password previously authenticated an empty-body auth packet and granted full console (permission level 4). Startup now refuses RCON with an empty/whitespace password, with a runtime guard as a second layer.
- The RCON max_connections counter was decremented immediately after spawn, so the limit never applied; it now tracks live connections correctly. Adds a read timeout and a 64 KiB buffer bound.
- Velocity proxy mode with an empty forwarding secret allowed forged HMAC validation. Startup now rejects velocity.enabled with an empty/whitespace secret.
- BungeeCord mode now prints a loud startup warning that direct connections can spoof any UUID/name and that the backend must be firewalled.
- Velocity-forwarded player names are validated before use.

## Tests
cargo test -p pumpkin passes.

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -107,6 +107,32 @@ impl LoadConfiguration for PumpkinConfig {
                 "When allow_chat_reports is enabled, java.online_mode must be enabled"
             );
         }
+
+        // Fail-closed: RCON with an empty password is full unauthenticated console access.
+        if self.advanced.networking.rcon.enabled {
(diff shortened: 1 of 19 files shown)


Snowiiii:
use Atomic PRs

PR #2512, opened 2026-07-24
security: block inventory dupe paths; gate Bedrock item/flight authority

kodareef5:
Part 4 of the hardening stack (stacked on #2511).

## Commits
- stop inventory dupes and drag/swap panics
- gate Bedrock item/flight authority and harden handlers

## What and why
- Container drag with 256+ slots wrapped a u8 to zero and caused a division-by-zero panic. Bounded and made u32-wide.
- An operator-precedence bug let any click action with button 40 run the hotbar/offhand swap branch. Grouping fixed.
- Merchant and anvil result slots could be swept for free (first layer of defense; the architectural fix lands later in the stack).
- Bedrock: ItemStackRequest::CraftCreative and legacy inventory-transaction item writes are now gamemode-gated; StartFlying requires allow_flying; invalid block actions no longer unwrap; a potion-amplifier shift that could reach a heal() assert is capped; full-packet INFO dumps softened to debug.

## Tests
cargo test -p pumpkin-inventory and -p pumpkin pass.

PR #2513, opened 2026-07-24
security: sweep-proof result slots, verified-login gate, panic cleanup, Velocity secret

kodareef5:
Part 5 of the hardening stack (stacked on #2512).

## Commits
- make trade/craft outputs non-sweepable and refresh on take
- harden Bedrock inventory moves and slot mapping
- require finished login before Config and kick-old on dup
- cleanup players after play-task panics and supervise ticks
- cap Bedrock item wire buffers and require Velocity secret

## What and why
- Merchant/anvil result slots become take-only (no insertion), so PickupAll sweeps cannot steal them; craft results refresh per take. Fixes the trade/anvil dupe cycles architecturally instead of per-action.
- Bedrock slot mapping hardened; bounded container slot indices.
- A login-finished flag is now required before the Config state; the flag is only set after a verified login on every transport, and the profile is cleared on auth failure. Closes the residual auth-bypass race. Duplicate logins kick the old session (vanilla parity).
- Play-task panics now always run full player cleanup (remove, save), so contained panics cannot leave ghost players; the world ticker is supervised so a tick panic cannot silently freeze the world; crash reports are written only for the first panic.
- Bedrock item wire buffers (user_data/extra_data) are capped pre-allocation; Velocity secret required when velocity is enabled regardless of the proxy.master switch.

## Tests
cargo test passes for pumpkin-inventory, pumpkin-protocol, and pumpkin (including take-only slot tests).

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -108,20 +108,28 @@ impl LoadConfiguration for PumpkinConfig {
             );
         }
 
-        // Fail-closed: RCON with an empty password is full unauthenticated console access.
+        // Fail-closed: RCON with an empty/whitespace password is full console access.
         if self.advanced.networking.rcon.enabled {
             assert!(
-                !self.advanced.networking.rcon.password.is_empty(),
+                !self.advanced.networking.rcon.password.trim().is_empty(),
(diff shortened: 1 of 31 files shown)


Snowiiii:
use Atomic PRs

PR #2514, opened 2026-07-24
security: centralize Bedrock craft-output authority; bound request arrays

kodareef5:
Part 6 of the hardening stack (stacked on #2513).

## Commits
- centralize Bedrock craft-output authority
- bound Bedrock ItemStackRequest and auth-input arrays
- finish kick-old login path and close spawn/cleanup gaps

## What and why
- CraftRecipe/CraftRecipeAuto are gated to crafting screens with a take-only output slot, so a chest slot cannot be crafted from. Craft repetitions consume the grid per craft.
- Preview container names resolve to real result slots and takes are settled through the slot hooks instead of being served free (keeps vanilla Bedrock workstation flows working while charging them).
- ItemStackRequest request/action/filter/result-item counts and PlayerAuthInput block-action counts are capped pre-allocation.
- Login-stage duplicate checks removed in favor of the centralized kick-old path in add_player; player spawn now runs inside the supervised task so spawn panics clean up; contained worker panics no longer force exit code 1 on manual shutdown.

## Tests
cargo test passes for pumpkin and pumpkin-protocol (includes new cap tests).

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -108,20 +108,28 @@ impl LoadConfiguration for PumpkinConfig {
             );
         }
 
-        // Fail-closed: RCON with an empty password is full unauthenticated console access.
+        // Fail-closed: RCON with an empty/whitespace password is full console access.
         if self.advanced.networking.rcon.enabled {
             assert!(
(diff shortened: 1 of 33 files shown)


Snowiiii:
use Atomic PRs

PR #2515, opened 2026-07-24
security: charge trade/anvil output on verified take; settle Bedrock craft takes via slot hooks; throttle registry/decompress/panic reports

kodareef5:
Part 7 (final) of the hardening stack (stacked on #2514).

## Commits
- charge merchant/anvil output on verified take
- cap remaining Bedrock length-driven allocations
- settle Bedrock craft-output takes through slot hooks
- throttle registry dump, packet reserve, and panic reports
- fix Bedrock zigzag VarInt decode and cap remaining counts
- replace known-packs sentinel with a dedicated flag

## What and why
- Merchant/anvil charging moved into the result slot's take hook, so every verified take (pickup, shift-click, throw, swap) is charged exactly once and failed deliveries never charge; the Throw double-call is fixed. Completes the dupe fixes.
- The Bedrock created-item cursor mirror is single-authority (never a destination base; consumed as a source); all take paths settle through the slot's take hook exactly once; writes into take-only slots are rejected.
- Remaining Bedrock length-driven allocations capped (legacy transaction entries, inventory actions, emote list).
- known-packs registry dump answered once per connection (previously re-triggerable, a full deep-clone+dump per small packet); decompression buffer grows incrementally instead of trusting the client-declared length up front (pre-auth alloc/CPU amplification); crash reports only built for the first panic.
- Bedrock zigzag VarInt decode fixed for negative wire values (BlockPos, legacy ids); transaction/legacy counts, BitSet decode, and the shared Vec<u8> reader capped; known-packs guard uses a dedicated flag instead of a keep-alive sentinel.

## Tests
cargo test passes for all touched crates (35 pumpkin-nbt, 75 pumpkin-protocol, 21 pumpkin-inventory, 166 pumpkin), including exploit-sequence regression tests for the dupe and crash classes fixed in this stack.

## Notes
- This stack intentionally does not attempt full movement/reach anti-cheat, login/config timeouts, packet rate limits, Velocity message_id binding, query bounds, or the plugin permission model; those are documented as follow-ups.
- Detailed technical report was sent privately to the address in SECURITY.md.

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -108,20 +108,28 @@ impl LoadConfiguration for PumpkinConfig {
             );
         }
 
-        // Fail-closed: RCON with an empty password is full unauthenticated console access.
+        // Fail-closed: RCON with an empty/whitespace password is full console access.
         if self.advanced.networking.rcon.enabled {
             assert!(
-                !self.advanced.networking.rcon.password.is_empty(),
+                !self.advanced.networking.rcon.password.trim().is_empty(),
                 "RCON is enabled but password is empty. Set networking.rcon.password to a non-empty value, or disable RCON."
             );
         }
 
-        // Fail-closed: Velocity with empty secret makes HMAC forgeable by anyone.
-        if self.advanced.networking.proxy.enabled && self.advanced.networking.proxy.velocity.enabled
-        {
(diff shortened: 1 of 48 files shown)


Snowiiii:
use Atomic PRs

PR #2518, opened 2026-07-25
fix(world/generate/huge_fungus): :bug: Fixed something wrong about the huge fungus generation

### LINKED ISSUE #2466: Something wrong in the nether

### Operating System

Arch Linux

712bb96

### What happened?

The nether generation seems strange......
That's freaking strange so that i cant describe. Just give you screenshots:

Image

Image

Yes. The trees are mixed, some even in lava!! Gosh that makes me nervous. The fence's facing is incorrect as well.

### To Reproduce

Just create a nether portal and find out a crimson forest, and a stronghold...

### Expected behavior

 - The trees are not in lava and not mix together
 - The fence's direction should be correct

zhangxuan2011:
## Description
As the #2466 describes, the generation of the nether is a bit incorrect. This PR mainly fixed the fungus's generation logic, added some checks to align to vanilla's behavior.

Fixed #2466 

After-fixed screenshots:

### Changes
 - Added biome check instead of random to decide fungus type through biome
 - Added block check to avoid fungus generates on incorrect block
 - Aligned to vanilla generation logic

## Testing
 - `cargo clippy --all-targets` - No warnings
 - `cargo test` - All passed
 - In-game test - In expected behavior (see above)

### DIFF
--- a/crates/pumpkin-data/Cargo.toml
+++ b/crates/pumpkin-data/Cargo.toml
@@ -127,5 +127,5 @@ harness = false
 [package.metadata.cargo-machete]
 ignored = ["phf"]
 
-# [lints]
-# workspace = true
+[lints]
+workspace = true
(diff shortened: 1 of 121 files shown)


zhangxuan2011:
So, I'd like to add more verification before a fungus generations, and further fungus varients...

Q2297045667:
LGTM

PR #2521, opened 2026-07-25
fix(entity): implement per-player entity tracking

MegalithOfficial:
## Description

Entities are currently not properly removed from the client when their chunks are unloaded. This causes ghost entities to remain visible, allows mobs to appear before their terrain is loaded and can disconnect players when they attack an entity that no longer exists on the server. The purpose of these changes is to implement a proper entity tracking lifecycle and keep the client and server synchronized.

Pumpkin sends entity spawn packets based on view distance. When the player travels away, Pumpkin sends `CUnloadChunk`, but it does not send `CRemoveEntities`. If the chunk has no remaining watchers, Pumpkin removes the entity from the world, but the client continues rendering it because it never received a removal packet.

I checked Mojang's 26.2 implementation. Each entity keeps a set of players currently tracking it, pairing only happens after the player has actually received the entity's terrain chunk, and the tracking state is recalculated every tick. I used a similar approach, but adapted it to Pumpkin's async architecture so tracking is only recalculated when something relevant changes:

- A chunk is sent or unloaded
- A player crosses a chunk boundary
- An entity crosses a chunk boundary
- An entity is spawned or removed

When a pairing becomes invalid, `CRemoveEntities` is immediately sent to that player. Entities are also registered before their spawn packet is sent. This should give us the same observable behavior as the official server without recalculating every entity for every stationary player each tick.

While testing this, I also found that some entities were saved without their base NBT because their custom NBT implementation was empty. This caused `Entity has no ID` messages when those chunks were loaded again. Base entity data is now always saved and loaded together with any custom data. Fast-moving entities such as projectiles and minecarts also use their live position when deciding which chunk they should be saved in.

Attacking an entity ID that no longer exists is now ignored like vanilla instead of disconnecting the player. The separate check for attacking your own entity ID is unchanged.

## Testing
Besides the cargo/fmt/clippy checks I have done, I had worries about performance impact. So i did a small benchmark:

I tested both upstream and this branch using the same world, one connected player and 500 armor stands.

Upstream averaged 23.9% CPU, while this branch averaged 24.4% CPU. Median tick time stayed around 1.4 ms on both, and the P95/P99 results were also within normal run-to-run variance.

I could not find any measurable performance regression from the new tracking system. I did movement-heavy benchmarks too but the results were excluded because chunk generation made the runs too inconsistent for a fair comparison.

I have I7 10700K and 16 GB RAM DDR4. Though we need a larger test in the future.

### DIFF
--- a/crates/pumpkin-config/src/entity_tracking.rs
+++ b/crates/pumpkin-config/src/entity_tracking.rs
@@ -1,0 +1,137 @@
+use std::collections::BTreeMap;
+
+use serde::{Deserialize, Serialize};
+
+/// Configuration for the distance at which entities are sent to clients.
+#[derive(Deserialize, Serialize)]
+#[serde(default)]
+pub struct EntityTrackingConfig {
+    /// Tracking range, in chunks, for entity types without an explicit override.
+    pub default_range: u32,
+    /// Per-entity tracking range overrides, in chunks.
+    pub ranges: BTreeMap<String, u32>,
+}
+
+impl Default for EntityTrackingConfig {
+    fn default() -> Self {
+        let ranges = [
+            ("marker", 0),
+            ("arrow", 4),
+            ("breeze_wind_charge", 4),
+            ("cod", 4),
+            ("dragon_fireball", 4),
+            ("egg", 4),
+            ("ender_pearl", 4),
+            ("experience_bottle", 4),
+            ("eye_of_ender", 4),
+            ("fireball", 4),
+            ("firework_rocket", 4),
+            ("fishing_bobber", 4),
+            ("lingering_potion", 4),
+            ("llama_spit", 4),
+            ("pufferfish", 4),
+            ("salmon", 4),
+            ("small_fireball", 4),
+            ("snowball", 4),
+            ("spectral_arrow", 4),
+            ("splash_potion", 4),
+            ("trident", 4),
+            ("tropical_fish", 4),
+            ("wind_charge", 4),
+            ("wither_skull", 4),
+            ("bat", 5),
+            ("dolphin", 5),
+            ("evoker_fangs", 6),
+            ("experience_orb", 6),
+            ("item", 6),
+            ("allay", 8),
+            ("bee", 8),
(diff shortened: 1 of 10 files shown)


pumpkin/src/world/entity_tracker.rs
```diff
@@ -0,0 +1,206 @@
...
+    pub fn remove_player_from_entity_tracking(&self, player: &Player) {
+        let player_id = player.gameprofile.id;
+        for tracker in &self.entity_trackers {
+            tracker.viewers.lock().unwrap().remove(&player_id);
+        }
+    }
+}
+
+/// Vanilla 26.2 client tracking ranges, measured in chunks.
+fn client_tracking_range(entity_type: &pumpkin_data::entity::EntityType) -> i32 {
```
Snowiiii: Hell nah, we don't hardcode

MegalithOfficial: These values are provided by the official Minecraft server source code. There isn't a way to calculate them.

The only way to avoid hardcoding them is to put them inside a config file.

MegalithOfficial: Plus, I reviewed multiple versions of the source code; they all use the same values.

RoosterBooster007: I'd suggest moving them to a config section, then.

crates/pumpkin-config/src/entity_tracking.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Deserialize, Serialize)]
+#[serde(default)]
+pub struct EntityTrackingConfig {
+    /// Tracking range, in chunks, for entity types without an explicit override.
+    pub default_range: u32,
+    /// Per-entity tracking range overrides, in chunks.
+    pub ranges: BTreeMap<String, u32>,
+}
+
+impl Default for EntityTrackingConfig {
```
Snowiiii, line 15: This really should not be hardcoded, and i'm also not sure if we should have that in a config at all

PR #2522, opened 2026-07-25
fix(enderman): synchronize teleport landing

MegalithOfficial:
## Description

Endermen could teleport to a position in the air and then appear to walk there instead of landing on the terrain below.

There were two separate problems causing this. The destination search did not fully match the official server behavior, and successful teleports updated the position without updating Pumpkin’s movement cache. The next relative movement packet could therefore be calculated from the old position, making the client apply part of the teleport twice.

The destination search now scans downward for solid ground, rejects water and other liquids, checks collisions using the Enderman’s actual dimensions and stops the teleport when no valid landing position exists.

Successful teleports now use the shared entity teleport path. This updates both the server position and the last position sent to clients before any relative movement updates are generated. Absolute position synchronization also preserves the entity’s current velocity.

## Testing
I shot an Enderman with projectiles such as arrows for around 10 minutes to repeatedly trigger its teleportation.

It consistently landed on valid terrain and never teleported into the air, water or anywhere else it was not supposed to. It also no longer appeared to walk in the air after teleporting.

### DIFF
--- a/crates/pumpkin/src/entity/mob/enderman.rs
+++ b/crates/pumpkin/src/entity/mob/enderman.rs
@@ -21,10 +21,7 @@ use pumpkin_data::{
     tracked_data::TrackedData,
 };
 use pumpkin_nbt::compound::NbtCompound;
-use pumpkin_protocol::{
-    codec::var_int::VarInt,
-    java::client::play::{CEntityPositionSync, Metadata},
-};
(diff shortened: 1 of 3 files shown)


RoosterBooster007:
Looks good. Thanks, @MegalithOfficial!

PR #2526, opened 2026-07-25
feat(plugin-api): expose ItemStack, uuid, event and command arg types

dorochadev:
## What

Re-export the already-generated `item_stack`, `uuid`, and `event` WIT modules (plus a few convenience types) from `pumpkin-plugin-api`, so plugin authors can name them:

- `item_stack` + `ItemStack` (with its `new(registry_key, count)` constructor)
- `uuid`
- `event::{self as events_wit, EventType}`, plus `EventHandler` and `FromIntoEvent`
- command `Arg`, `ArgumentType`, `StringType`

## Why

These modules are generated by `wit_bindgen` and are already used internally (by `gui`, `player`, etc.), but they were never added to the public re-export list in `lib.rs`, so plugin authors can't name them:

- **`ItemStack` can't be constructed by a plugin** — it's only reachable by *inference* off methods like `player.get_inventory_item()`. That blocks building GUI icons (`gui.set_item(...)`), and commands like `/give`, `/i`, `/skull`.
- **`uuid` can't be named**, so plugins can't build one to pass to `server.get_player_by_uuid(...)` and have to hand-format the `{high, low}` fields.
- The command `Arg`/`ArgumentType`/`StringType` types and `EventHandler`/`FromIntoEvent` sit behind non-obvious paths (`command_wit`, `events::...`).

## Change

One file (`pumpkin-plugin-api/src/lib.rs`), purely additive re-exports. Compiles for `wasm32-wasip2`.

Revives #2325 (opened and self-closed unmerged by @dongzh1) — same intent, credit to them.

### DIFF
--- a/pumpkin-plugin-api/src/lib.rs
+++ b/pumpkin-plugin-api/src/lib.rs
@@ -44,7 +44,8 @@ pub mod scheduler;
 
 pub mod command {
     pub use crate::wit::pumpkin::plugin::command::{
-        Command, CommandError, CommandNode, CommandSender, ConsumedArgs,
+        Arg, ArgumentType, Command, CommandError, CommandNode, CommandSender, ConsumedArgs,
+        StringType,
     };
 }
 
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thank you @dorochadev :D

PR #2540, opened 2026-07-25
feat(bedrock): configurable username prefix and space replacement

### LINKED ISSUE #1706: Bedrock Username Prefix & Space

You know how in GeyserMC, Bedrock players have like a prefix in their username. Let's say for example, your Java & Bedrock username are the same. This will make sure you don't get conflicted usernames. It could also be customized or even removed by leaving it empty. Additionally, we should add a toggle that on by default to replace the space that Bedrock players get in the username with _ so that it's possible to type it in chat when mentioning the player (@player_name instead of @player name).

Add a prefix option to Bedrock players username that can be customized or removed (by leaving it blank).
Add a toggle to replace the space in Bedrock players username with _ ("player_name" instead of "player name")

Sawlyer:
## Description

Bedrock gamertags can contain spaces and can be spelled exactly like a Java account name, so on a cross-play server two players can end up indistinguishable and neither can be addressed by name in a command.

Adds two options under `networking.bedrock`:

- `username_prefix`, prepended to the Bedrock name. Empty by default, so existing setups keep the names they have.
- `replace_username_spaces`, on by default, which turns the existing unconditional space-to-underscore rewrite into something servers can opt out of.

The rewrite itself already existed in `handle_login`; it is now built by a small helper so both options go through one place.

Fixes #1706

## Testing

Added unit tests for the name transform: spaces become underscores by default, are kept when the toggle is off, the prefix is prepended, and an empty prefix leaves the name unchanged.

`cargo test -p pumpkin`, `cargo clippy -p pumpkin -p pumpkin-config` and `cargo fmt --check` pass.

### DIFF
--- a/crates/pumpkin-config/src/networking/bedrock.rs
+++ b/crates/pumpkin-config/src/networking/bedrock.rs
@@ -78,6 +78,12 @@ pub struct BedrockConfig {
     pub compression: CompressionConfig,
     /// Message of the Day; the server's description displayed on the status screen.
     pub motd: String,
+    /// Prefix prepended to Bedrock Edition player names, so they cannot collide with
+    /// Java Edition account names on a cross-play server. Empty means no prefix.
+    pub username_prefix: String,
+    /// Whether spaces in Bedrock Edition player names are replaced with underscores.
+    /// Names containing spaces cannot be typed as command arguments.
+    pub replace_username_spaces: bool,
     /// Bedrock Edition authentication settings.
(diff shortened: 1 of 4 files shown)


Snowiiii:
@Sawlyer do you plan to fix conflicts

Sawlyer:
@Snowiiii done, rebased on master.

Snowiiii:
LGTM, Thank you @Sawlyer :D

PR #2542, opened 2026-07-25
fix(entity): use block particles when an armor stand breaks

### LINKED ISSUE #1621: Armor stands produce death particles when destroyed

### Operating System

Arch Linux

8bd5009

### What happened?

When you destroy an armor stand in creative or survival mode, it produces death particles

### To Reproduce

1. Place an armor stand
2. Destroy said armor stand
3. Observe death particles

### Expected behavior

Armor stands shouldn't produce death particles

Sawlyer:
## Description

Breaking an armor stand spawned `Particle::Poof`, the puff a mob leaves when it dies, so destroying one read as a mob death rather than a block break. Vanilla `ArmorStand.showBreakingParticles` emits `block` particles textured with oak planks, at the same count, spread and speed already used here.

The `block` particle needs a block state id on top of the particle id, which nothing in the codebase was sending yet, so this adds `spawn_block_particle` to `Player` and `World`. The state id is remapped to the receiving client's protocol version before it is written, because `CParticle` carries extra particle data as already-encoded bytes and cannot remap it itself.

The vertical offset also used a truncated `0.6666` where vanilla uses two thirds of the entity height.

Fixes #1621

## Testing

`cargo test -p pumpkin`, `cargo clippy -p pumpkin` and `cargo fmt --check` pass. The change is packet construction, which the existing unit tests do not reach, so the payload was verified against vanilla `BlockParticleOption` (a single VarInt block state id after the particle id) rather than a new test.

### DIFF
--- a/crates/pumpkin/src/entity/decoration/armor_stand.rs
+++ b/crates/pumpkin/src/entity/decoration/armor_stand.rs
@@ -4,11 +4,11 @@ use crate::entity::{Entity, EntityBase, EntityBaseFuture, NbtFuture, living::Liv
 use crossbeam::atomic::AtomicCell;
 use pumpkin_data::item_stack::ItemStack;
 use pumpkin_data::{
+    Block,
     damage::DamageType,
     data_component_impl::{EquipmentSlot, EquipmentType},
     entity::EntityStatus,
     item::Item,
-    particle::Particle,
     sound::{Sound, SoundCategory},
 };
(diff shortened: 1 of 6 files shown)


Sawlyer:
Rebased on master, conflicts fixed. try_enqueue_packet takes Bytes now, so the packet goes through the Java client's serialize_packet first.

RoosterBooster007 (approved):

RoosterBooster007:

Issue #2531, opened 2026-07-25
Using bone meal to accelerate crop growth is not possible


bloodclover:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Gentoo Linux

### Server Software Version/Commit

Pumpkin 0.1.0-dev+26.2

### What happened?

Nothing, the item is not used and the crop is not growing.

### To Reproduce

Use (M2/ Right Click) bone meal on any crop.

### Expected behavior

The crop gets the green particle effect and has a chance to grow to one of its next stages.

DemonFiend:
Im gonna take a look at this

Edit: Looks like this mechanic was missing, This now works specifically on Crops given the below feat fix. just gotta wait for merge.
Due to more missing work  on any bonemealable item that has a block build (Saplings, Muchrooms, ect) require a separate PR to fix. as those require Tree Generation work, which is currently WIP and requires another PR to start.

PR #2544, opened 2026-07-25
feat: bone meal grows crops (#2531) 🤖🤖🤖

DemonFiend:
## Description

Implements bone meal fertilizing crops, closing #2531.

Adds a `BonemealableBlock`-style interface on `BlockBehaviour` — `is_bonemeal_target` / `is_bonemeal_success` (default `true`) / `perform_bonemeal` (default no-op) — and a `bone_meal.rs` item that dispatches through the block registry, with impls for wheat, carrots, potatoes, beetroot, and melon/pumpkin stems.

Behavior matches Java 26.2:
- Wheat/carrots/potatoes and melon/pumpkin stems advance a random 2–5 stages, clamped to max. Stems only advance age — bone meal never spawns the gourd (that's the separate mature-stem tick).
- Beetroot rolls 75% to advance +1 (max age 3).
- The stack is consumed and the growth effect (world event 1505: `happy_villager` particles + `bone_meal.use` sound) plays whenever the target is *valid*, even when the success roll fails — a mature crop is not a valid target, so nothing is consumed.
- Light level and farmland hydration are ignored; not consumed in creative.

Scope is crops only. Saplings, mushrooms/fungi, area spreaders (grass/nylium), the underwater and dispenser paths, and the remaining +1 plants (sweet berry, cocoa, torchflower, pitcher) are follow-ups on the same interface.

## Testing

Tested in-game on 26.2: 2–5 growth/clamping on wheat/carrots/potatoes, beetroot's ~75% + `+1`, melon/pumpkin stems advancing without spawning the gourd, mature-crop no-op (item not consumed), and creative non-consumption. Added a unit test locking the 2–5 crop increase and beetroot's `+1`. `cargo clippy --all-targets` (debug + release), `cargo fmt --check`, tests, and typos all pass.


Sawlyer:
`CWorldEvent` gets `0` as data. All four vanilla call sites pass 15 (`BoneMealItem` twice, `DispenseItemBehavior`, `UseBonemeal`). Did you check a client actually renders the particles with 0?

On the open question in the beetroot comment: vanilla has no `isBonemealSuccess` override there, it is `getBonemealAgeIncrease(level) { return super.getBonemealAgeIncrease(level) / 3; }`, so `randInt(2, 5) / 3` gives 0 at 1/4 and 1 at 3/4. Same distribution as what you have, so that note can go.

DemonFiend:
Good catch, reverted to 15 to match the vanilla call sites. That was a bad change on my part; I'd swapped it to 0 on an automated suggestion without re-verifying. 

Re-tested in-game: particles + sound still show.

Also, Dropped the note per suggestion.

PR #2545, opened 2026-07-25
fix(net): acknowledge block changes per packet, not on keep-alive 🤖🤖🤖

DemonFiend:
## Description

`CAcknowledgeBlockChange` — which resolves the client's block-interaction prediction — was only flushed inside the keep-alive branch (~every 15s). Interactions that change no block (e.g. using an item on a block where nothing happens) therefore left the client's prediction pending until the next keep-alive, blocking further use on that block until the player interacted elsewhere.

This flushes the pending sequence acknowledgement immediately after each play packet is handled. It only sends when a sequence was actually set, so non-interaction packets add no traffic.

## Testing

Reproduced with repeated no-op item uses (bone meal on a fully-grown crop) which previously froze further right-clicks until interacting with a different block; after the change each click registers immediately. `cargo clippy --all-targets` (debug + release), `cargo fmt --check`, tests, and typos all pass.

### DIFF
--- a/pumpkin/src/net/java/mod.rs
+++ b/pumpkin/src/net/java/mod.rs
@@ -275,13 +275,6 @@ impl JavaClient {
                     self.last_keep_alive_time.store(Instant::now());
                     let packet = pumpkin_protocol::java::client::play::CKeepAlive::new(keep_alive_id);
                     self.enqueue_packet(&packet).await;
-
-                    let seq = self.packet_sequence.swap(-1, Ordering::Relaxed);
-                    if seq != -1 {
-                        self
-                            .send_packet_now(&CAcknowledgeBlockChange::new(seq.into()))
-                            .await;
-                    }
                 }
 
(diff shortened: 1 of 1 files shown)


Sawlyer:
`send_packet_now` goes through `outgoing_packet_priority_send`, the corrective `CBlockUpdate`s through `outgoing_packet_queue_send`. Sending the ack here means it goes out while the block updates the same handler just enqueued are still sitting in the normal queue, so the client can accept its own prediction before the correction arrives. On the keep-alive tick the normal queue had time to drain first.

#2443 touches the same lines and moves the ack to `enqueue_packet` for that ordering. These two will conflict.

DemonFiend:
I missed that PR, apologies/ Closing in favor of #2443 

Reason: i independently hit this from the bone-meal side. This is a global networking fix, so nothing in the bone-meal PR #2544 depends on it, that one will just benefit from #2443 once it lands.

PR #2546, opened 2026-07-25
fix(loot): roll binomial_with_bonus_count without fortune 🤖🤖🤖

DemonFiend:
## Description

`apply_bonus` gated its entire body on `enchantment_level > 0`, so `binomial_with_bonus_count` never rolled its base `extra` trials when no fortune was present. Crops that drop via that formula (wheat/beetroot seeds) always yielded exactly 1 seed instead of `1 + binomial(extra, p)` (e.g. wheat = 1–4).

Removed the outer `enchantment_level > 0` guard so the formulas always apply. `uniform_bonus_count` computes `0..=(level * multiplier)`, a no-op at level 0, and `ore_drops` self-gates on its own match arm, so neither regresses.

## Testing

Broke ~10 mature wheat by hand: seed drops now vary (1–4) instead of always 1; carrots/potatoes (which use `SetCount`/uniform and were unaffected) still drop 2–5. `cargo clippy --all-targets` (debug + release), `cargo fmt --check`, tests, and typos all pass.

### DIFF
--- a/pumpkin/src/world/loot.rs
+++ b/pumpkin/src/world/loot.rs
@@ -115,40 +115,51 @@ fn apply_bonus(
         pumpkin_data::Enchantment::from_name(enchantment_name)
             .map_or(0, |enchantment| tool.get_enchantment_level(enchantment))
     });
-    if enchantment_level > 0 {
-        for stack in stacks {
-            match formula {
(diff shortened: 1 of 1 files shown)


Sawlyer:
+1 looks good for me.

PR #2548, opened 2026-07-25
fix: double-tall plant worldgen & consumable on-consume effects (milk/honey)

sirfusebox:
Double-tall plants

Fixed world generation so double-tall plants (rose bushes, sunflowers, lilacs, peonies, tall grass, large ferns, pitcher plants, small dripleaf) place both the bottom and top halves.
Fixed neighbor updates so tall plants only stay if the other half is the same plant, matching vanilla behavior.

Milk & consumable effects

Fixed consumable effects not being applied after finishing an item.
Added support for:
clear_all_effects (milk)
remove_effects (honey bottle)
apply_effects (suspicious stew)
play_sound
teleport_randomly (chorus fruit) is not implemented yet and currently logs a warning.

### DIFF
--- a/pumpkin-world/src/generation/feature/features/simple_block.rs
+++ b/pumpkin-world/src/generation/feature/features/simple_block.rs
@@ -1,4 +1,7 @@
-use pumpkin_data::Block;
+use pumpkin_data::block_properties::{
+    BlockProperties, DoubleBlockHalf, SmallDripleafLikeProperties, TallSeagrassLikeProperties,
+};
+use pumpkin_data::{Block, BlockState};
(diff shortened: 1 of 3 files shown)


Sawlyer:
`ConsumeEffect::RemoveEffects` only handles `IDSet::IDs` and silently drops `IDSet::Tag`. Nothing in vanilla data hits that today (honey bottle lists `minecraft:poison` directly), but a datapack using a tag would do nothing at all with no trace. Worth a `warn!` there like you did for `TeleportRandomly`.

These are also two unrelated fixes in one PR. The worldgen half is self-contained and could go in on its own.

RoosterBooster007:
Please split this into at least 2 PRs. Your code looks good, though.

PR #2551, opened 2026-07-25
fix: nether chunk height, portal travel stalls, block acks and log spam

sirfusebox:
ProtoChunk was sized from dimension.logical_height (128 in the Nether) while chunk sections use dimension.height (256), so every Nether chunk panicked with an index out of bounds at the Full stage.

find_safe_location did an async block fetch per block over a 65x65 area, stalling the tick for seconds. Chunks are now preloaded and the scan is synchronous, with vanilla's 16 block radius.

search_for_portal read POI candidates synchronously from unloaded chunks, so existing portals were never reused and a duplicate was built every trip.

CAcknowledgeBlockChange was only sent on the 15s keep-alive timer, leaving client predicted blocks (fire in portal frames) on screen for seconds.

Per-tick portal collision logs and repeated chunk generation failures no longer spam the console.

### DIFF
--- a/pumpkin-world/src/chunk_system/chunk_state.rs
+++ b/pumpkin-world/src/chunk_system/chunk_state.rs
@@ -205,6 +205,13 @@ impl Chunk {
     }
 
     fn build_level_sections(proto_chunk: &ProtoChunk, dimension: &Dimension) -> ChunkSections {
+        // The proto chunk must be allocated with the dimension's full storage height,
+        // otherwise the reads below run past the end of its block/biome buffers.
+        debug_assert_eq!(
+            proto_chunk.height() as i32,
+            dimension.height,
+            "proto chunk height does not match the dimension height"
+        );
         let total_sections = dimension.height as usize / BlockPalette::SIZE;
         let biome_min_y = biome_coords::from_block(dimension.min_y);
         let block_sections = (0..total_sections)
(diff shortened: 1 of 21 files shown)


sirfusebox:
AHH I just saw it failed to build imma fix it in 8 hours when I'm back home sorry :(

Sawlyer:
`cargo fmt --all -- --check` fails on `pumpkin/src/item/items/ignite/ignition.rs:55` and `pumpkin-world/src/generation/proto_chunk_test.rs:113`, so CI will reject it as is.

This is five unrelated fixes in one PR, and the respawn-point change in `player.rs` (validating the bed/anchor in the stored spawn dimension instead of the death world, plus loading that chunk first) isn't mentioned in the description at all. The block-ack part also overlaps #2545, which does the same thing in the packet loop instead of in `Player::tick`. The Nether height fix is a crash fix with a regression test and doesn't depend on any of the rest; splitting it out would get it merged a lot faster than reviewing all 21 files together.

RoosterBooster007:
Thanks for putting this together! I'm closing this since the main issue (the Nether chunk-height panic) was already fixed in #2803.

This PR also bundles several unrelated fixes across 21 files, which makes it tough to review... especially with the current fmt errors and merge conflicts.

If you're up for it, we'd love to see these split into separate, focused PRs against master:
- Log-spam rate-limiting: Still looks relevant.
- Block-ack timing: Worth re-verifying against the latest master.
- Portal travel preload: The portal code was recently rewritten, so this needs a fresh look.

PR #2554, opened 2026-07-26
feat(config): read the Velocity forwarding secret from a file 🤖🤖🤖

### LINKED ISSUE #2516: Add functionality to resolve file-based secrets for Velocity configuration. (Example @forwarding.secret)

Having to manually update secret keys, having to hide and constantly be making sure the .toml can't be viewed by anyone or any plugins. 

Add file based secrets to the velocity proxy config, this allows for secure secrets without having to worry about embedding the secret directly into the config. One thing this allows for is synced secrets across a network, rather than having to update each config file every time you change the key, you can use symlinks and update all servers by just changing one file, this is extremely important for security and proxy support.

Example:
```
# Use '@' to specify an external file (path is relative to the working directory)
[networking.proxy.velocity]
enabled = true
secret = "@forwarding.secret"
```

> [!NOTE] 
> I've opened a Discussion (#2485), and was just going to wait but decided to open an issue. (since that's what the contribute.md said to do.)

RicePollution:
Closes #2516.

## What changed

`networking.proxy.velocity.secret` may now point at a file instead of holding the secret inline. A value beginning with `@` is read from the file at that path:

```toml
[networking.proxy.velocity]
enabled = true
secret = "@forwarding.secret"
```

Relative paths resolve against the server's working directory, the same place `pumpkin.toml` is read from. Absolute paths work too, so the value can point straight at a Docker or Kubernetes secret (`@/run/secrets/velocity_forwarding_secret`). Surrounding whitespace is trimmed, since almost everything that writes a secret file leaves a trailing newline. Any value not starting with `@` is used literally, exactly as before.

## Why

`pumpkin.toml` gets committed to version control, baked into images, and pasted into support threads. The forwarding secret is the only thing standing between a server and anyone who can reach its port, so it is worth being able to keep it somewhere else. Secret mounts in Docker and Kubernetes are files, so this is the shape those tools already expect.

## Implementation note

There is one non-obvious hazard worth flagging for review. `LoadConfiguration::load` writes the deserialized config back to disk whenever it fills in missing defaults:

```rust
let (merged_config, changed) = Self::merge_with_default_toml(parsed_toml_value);
if changed {
    fs::write(&path, toml::to_string(&merged_config).unwrap())
}
```

So resolving `@forwarding.secret` into the `secret` field directly would make the server copy the real secret into `pumpkin.toml` the first time any default is filled in, which defeats the whole point of the feature.

To make that impossible rather than merely avoided, `secret` keeps the raw `@file` reference and the resolved value lives in a separate `#[serde(skip)]` `OnceLock`, read through `VelocityConfig::secret()`. The field that gets serialized never holds the plaintext. `tests/velocity_secret_file.rs` covers this against the real `load()` path with a deliberately incomplete config, so the write-back actually fires.

Resolution is also forced during `validate()`, for two reasons. It surfaces a broken path at startup rather than on a player's first connection, and it keeps the blocking `fs::read_to_string` off the async login path: `handle_plugin_response` reaches the secret whenever `velocity.enabled` is set, independently of the proxy master switch, so resolution is forced on that same condition. The emptiness assertion stays gated on both switches so that configs which boot today keep booting.

## Performance

No benchmark added, and I do not believe one is warranted here. The file is read at most once per process and cached in a `OnceLock`, so after startup reading the secret is an atomic load where it was previously a plain field access. It sits on the login path rather than in any hot loop. Happy to add a Criterion benchmark if you would rather have the number.

## Impact

- No behaviour change for existing configs. A secret without a leading `@` is used exactly as before.
- `VelocityConfig` gains a private field, which would otherwise make the struct impossible to build outside `pumpkin-config` (functional update syntax still requires every field to be visible). A `VelocityConfig::new(enabled, secret)` constructor is added so downstream code keeps a supported way to construct it. Nothing in the workspace built it by struct literal.
- The `validate()` check is gated on `proxy.enabled && proxy.velocity.enabled`, matching what actually turns on Velocity forwarding in `login.rs`, so no config that boots today starts failing.

## Testing

18 new tests.

Resolver units (`pumpkin-config/src/secret.rs`): literal value, file reference, whitespace trimming, `@@` escape, missing file, empty file, bare `@`.

Config integration (`pumpkin-config/tests/velocity_secret_file.rs`), driving the real `PumpkinConfig::load()`: the secret resolves from the referenced file; the resolved secret is **not** written back into `pumpkin.toml`; the reference still resolves with the proxy master switch off; single-quoted TOML paths work (the form Windows users need, since `"@C:\...\"` is an invalid TOML escape).

Forwarding integration (`pumpkin/src/net/proxy/velocity.rs`), driving the real `receive_velocity_plugin_response` with a signed payload: a payload signed with the secret *as stored in the file* is accepted; one signed with a different secret is rejected; and one signed with the raw `@path` string is rejected, which is what proves resolution actually happened rather than the literal reference being used as the HMAC key.

Also run: `cargo fmt --check`, `cargo clippy --all-targets --all-features` in both debug and release with `RUSTFLAGS=-Dwarnings`, `cargo test --workspace`, `cargo test --doc`, `cargo machete`, and a full `cargo build --release`.

## Limitations

- The HMAC verification path is covered by the tests above, but I did not run this against **actual Velocity software** end to end. The forwarding payload in those tests is constructed to match `receive_velocity_plugin_response`'s expectations, so it verifies Pumpkin's side rather than interop with a live proxy.
- `@@` escapes a literal leading `@` (`@@value` means `@value`). This is slightly beyond what the issue asked for, but without it a secret genuinely starting with `@` cannot be expressed at all. Happy to drop it if you would rather keep the surface minimal.
- `networking.rcon.password` has the same plaintext-in-config problem and the resolver is written to be reusable, but I kept this PR to the scope of the issue. Glad to follow up if wanted.
- Lightly overlaps #2513, which also touches the Velocity secret (making it required). Different change, adjacent lines.

Docs PR for the `@file` syntax to follow in Pumpkin-Docs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2652,6 +2652,7 @@ version = "0.1.0-dev+26.2"
 dependencies = [
  "pumpkin-util",
  "serde",
+ "tempfile",
  "toml 1.1.3+spec-1.1.0",
  "tracing",
  "uuid",
--- a/pumpkin-config/Cargo.toml
+++ b/pumpkin-config/Cargo.toml
@@ -13,6 +13,8 @@ uuid.workspace = true
 
 toml.workspace = true
 
+[dev-dependencies]
+tempfile.workspace = true
 
 [lints]
 workspace = true
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -38,6 +38,7 @@ pub mod op;
 mod advancement;
 mod player_data;
 mod pvp;
+mod secret;
 mod server_links;
 pub mod whitelist;
 pub mod world;
@@ -106,6 +107,24 @@ impl LoadConfiguration for PumpkinConfig {
                 self.advanced.networking.java.online_mode,
                 "When allow_chat_reports is enabled, java.online_mode must be enabled"
             );
+        }
+
+        // Resolve the forwarding secret up front. Reading the file here keeps the
+        // blocking read off the async login path, since `handle_plugin_response`
+        // reaches the secret whenever velocity is enabled, regardless of the proxy
+        // master switch. It also surfaces a broken `@file` reference at startup
+        // instead of on a player's first connection.
+        if self.advanced.networking.proxy.velocity.enabled {
+            let secret = self.advanced.networking.proxy.velocity.secret();
+
+            // Only required when the proxy is actually switched on, so that
+            // configs which boot today keep booting.
+            if self.advanced.networking.proxy.enabled {
+                assert!(
+                    !secret.is_empty(),
+                    "When velocity is enabled, a forwarding secret must be set"
+                );
+            }
         }
     }
 }
--- a/pumpkin-config/src/networking/proxy.rs
+++ b/pumpkin-config/src/networking/proxy.rs
@@ -1,4 +1,8 @@
+use std::sync::OnceLock;
+
 use serde::{Deserialize, Serialize};
+
+use crate::secret::resolve_file_reference;
 
 /// Configuration for proxy support.
 ///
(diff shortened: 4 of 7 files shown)


Sawlyer:
+1 looks good for me.

Rocraften:
Can't wait.

Rocraften:
This is taking a little longer than expected.

Rocraften:
If this has been completed and tested, what's taking so long?

xRookieFight (approved):

DerGrumpf:
> If this has been completed and tested, what's taking so long?

You should chill a bit, but the nixpkgs is also depending on it :)

06kellyjac:
> If this has been completed and tested, what's taking so long?

There are other PRs getting reviewed and merged

image

---

Also there are now conflicts blocking so you'll need to resolve those please:

image

DerGrumpf:
@Snowiiii why closed what is the Problem with it? Currently this PR dependency for https://github.com/NixOS/nixpkgs/pull/545183

RicePollution:
> @Snowiiii why closed what is the Problem with it? Currently this PR dependency for https://github.com/NixOS/nixpkgs/pull/545183

there are definitely conflicts now since this PR is like 3 weeks old. I could probably fix this sometime.

DerGrumpf:
> there are definitely conflicts now since this PR is like 3 weeks old. I could probably fix this sometime.

Pls

Rocraften:
> there are definitely conflicts now since this PR is like 3 weeks old. I could probably fix this sometime.

Please do.

PR #2562, opened 2026-07-26
Clamp proto_chunk out of bounds array access

### LINKED ISSUE #2560: Server crashes when a player teleports through a nether portal

### Operating System

Arch Linux

0.1.0-dev+26.2/faf8c50

### What happened?

I entered into a nether portal and when it was the time to teleport the server crashed.

https://github.com/user-attachments/assets/332a6372-4c7d-4ae9-8dad-a6c5b5ba6e6c

[crash-2026-07-26_08.33.24.txt](https://github.com/user-attachments/files/30384800/crash-2026-07-26_08.33.24.txt)

[latest.log](https://github.com/user-attachments/files/30384803/latest.log)

### To Reproduce

Try to enter the nether.

### Expected behavior

Should teleport to the nether.

Akorn27:
this may break worldgen but this is a quickfix rather than a in-depth fix.

## Description
Should fix #2560
Out of bounds access
## Testing
Testing done by hand on default cargo build settings.

### DIFF
--- a/pumpkin-world/src/generation/proto_chunk.rs
+++ b/pumpkin-world/src/generation/proto_chunk.rs
@@ -446,7 +446,9 @@ impl ProtoChunk {
     #[inline]
     #[must_use]
     pub fn get_block_state_raw(&self, x: i32, y: i32, z: i32) -> BlockStateId {
-        let index = self.local_pos_to_block_index(x, y, z);
+        let index = self
+            .local_pos_to_block_index(x, y, z)
+            .clamp(0, self.flat_block_map.len() - 1);
         self.flat_block_map[index]
     }
 
@@ -505,7 +507,7 @@ impl ProtoChunk {
             x & biome_coords::from_block(15),
             y - biome_coords::from_block(self.bottom_y() as i32),
             z & biome_coords::from_block(15),
-        );
+        ).clamp(0, self.flat_biome_map.len() - 1);
         self.flat_biome_map[index]
     }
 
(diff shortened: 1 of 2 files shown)


Sawlyer:
#2560 isn't really a bad coordinate to guard against, it's an undersized buffer. `ProtoChunk::new` sizes `flat_block_map` from `dimension.logical_height`, which is 128 in the Nether, while the level-chunk conversion walks `dimension.height` = 256.

That matters for what the clamp does. The index is `height * 16 * x + 16 * y + z` with `height` = 128, so a Nether read at y >= 128 mostly aliases into the next `x` column and only actually leaves the buffer up near x=15. Index (0, 130, 0) resolves to 2080, which is (1, 2, 0). So the panic is the tail of the problem, not the whole of it: clamping silences the crash and leaves every read above y=128 pointing at the wrong column, which is the "may break worldgen" you flagged.

#2551 fixes the allocation itself (`logical_height` -> `height`) and has a regression test for it, so I'd rather that one land for #2560.

If a guard is still wanted here, `get_block_state` a few lines below already has the right shape: bounds check, return `VOID_AIR`, rather than clamping onto a valid-looking index.

The `main.rs` panic-message change also looks unrelated to the rest, and it leaves the two branches formatting the payload differently.

Akorn27:
@Sawlyer well, in regards to the panic message, it was what I needed to do (because the panic message wasn't previously good enough because it didn't tell where, it just told what)
second, actually, as i tested it, (feel free to test it yourself on my branch) the nether looked okay.
that being said, you can make changes, and as i said, this was a hotfix to maybe introduce me to the codebase, so pardon me for my terrible-ness when it comes to making code here.

Akorn27:
And also in regards to panic formatting stuff diffrently, yes, kindof better to do it that way, if you fail twice you need more info to send the devs as its probably a bug

Akorn27:
I'd consider it uncontended that there is a better fix, one thing i would change overall is the panic formatting, but that is up to debate.

PR #2563, opened 2026-07-26
feat(plugin): add get-process-memory and cheapen get-sys-info

Ashura006:
## Description

Depends on Pumpkin-MC/pumpkin-plugin-wit#23.

Earlier draft extended `sys-info` with `process-memory`. That would break already-built v0.1 plugins, because changing a returned record changes the imported function type and fails `instantiate_pre` without record subtyping.

This PR now:

- leaves `sys-info` unchanged
- adds `get-process-memory` for process RSS (same RAM permissions)
- refreshes only the counters the caller is allowed to see in `get-sys-info`, instead of `System::new_all()` + `refresh_all()`

## Testing

- Confirmed the WIT change is additive (new method, no record shape change)
- Host `get-process-memory` uses a single-pid memory refresh
- Permission gating matches existing `sys.info` / `sys.info.ram` checks

### DIFF
--- a/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
+++ b/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
@@ -53,17 +53,14 @@ impl pumpkin::plugin::server::HostServer for PluginHostState {
         let (total_memory, used_memory, process_memory) = if need_ram {
             sys.refresh_memory();
 
-            let process_memory = match sysinfo::get_current_pid() {
-                Ok(pid) => {
-                    sys.refresh_processes_specifics(
-                        sysinfo::ProcessesToUpdate::Some(&[pid]),
-                        true,
-                        sysinfo::ProcessRefreshKind::nothing().with_memory(),
-                    );
-                    sys.process(pid).map(sysinfo::Process::memory)
-                }
-                Err(_) => None,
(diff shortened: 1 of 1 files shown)


julesdemangeot-ship-it (approved):

Sawlyer:
One question before this lands: does adding `process-memory` to the `sys-info` record inside v0.1 break plugins already built against the current WIT? `instantiate_pre` type-checks the component's imports against the linker, and as far as I know records aren't extensible there, so a plugin compiled against the 5-field record would fail to instantiate against a host offering 6. If that's right this wants a new interface version rather than an in-place field. Worth trying with a plugin binary built before the change.

Also Pumpkin-MC/pumpkin-plugin-wit#23 is still open, so the submodule bump can't be merged yet either way.

Ashura006:
@Sawlyer Good catch — you're right. Adding a field to `sys-info` in v0.1 would change the return type of `get-sys-info`, and without record subtyping that breaks already-built components at `instantiate_pre`.

I switched the approach:

- `sys-info` is unchanged
- process RSS is exposed via a new `get-process-memory: func() -> option<u64>` on `server` (same `sys.info` / `sys.info.ram` gating)
- the selective refresh change in `get-sys-info` stays

pumpkin-plugin-wit#23 is updated the same way. Once that lands, the submodule bump here can follow.

PR #2564, opened 2026-07-26
fix(entity): feeding a parrot a cookie poisons and kills it

### LINKED ISSUE #2534: Parrots can not eat cookies

### Operating System

Gentoo Linux

Pumpkin 0.1.0-dev+26.2

### What happened?

Nothing, the cookie stays in the inventory and the parrot is unaffected.

### To Reproduce

Feed a parrot a cookie (M1/Right Click)

### Expected behavior

The parrot consumes the cookie and gets poisoned.

Sawlyer:
## Description

Fixes #2534.

`ParrotEntity` had no interact handler at all, so a cookie used on a parrot did nothing: the item stayed in the inventory and the parrot was unaffected.

Vanilla `Parrot.mobInteract` treats the `parrot_poisonous_food` tag as its last branch: the item is consumed, poison is applied for 900 ticks, and the parrot is then killed with a player-attack damage source.

This adds a `mob_interact` override doing the same. Anything that isn't in `parrot_poisonous_food` falls through to `MobEntity::mob_interact`, so leashing and name tags keep working.

Two details worth noting:

- The branch is gated on the item tag rather than on `Item::COOKIE`, matching vanilla. Nothing in `parrot_food` (seeds, used for taming) is also in `parrot_poisonous_food`, so implementing this branch doesn't get in the way of taming later.
- Vanilla guards the kill with `player.isCreative() || !this.isInvulnerable()`. `hurt` re-checks invulnerability itself and `player_attack` doesn't bypass it, so that guard only skips a call that would already do nothing. The call here is unconditional.

## Testing

`cargo test -p pumpkin --lib` (160 pass), `cargo clippy -p pumpkin --lib --tests`, `cargo fmt --all -- --check`.

Three unit tests cover the tag gate: cookie is in `parrot_poisonous_food`, seeds and cooked chicken are not, and the poison duration is 900 ticks. The interaction itself needs a live client to verify, so the death and the poison particles aren't covered by an automated test.

### DIFF
--- a/pumpkin/src/entity/passive/parrot.rs
+++ b/pumpkin/src/entity/passive/parrot.rs
@@ -1,15 +1,24 @@
 use std::sync::{Arc, Weak};
 
+use pumpkin_data::damage::DamageType;
+use pumpkin_data::effect::StatusEffect;
 use pumpkin_data::entity::EntityType;
+use pumpkin_data::item_stack::ItemStack;
+use pumpkin_data::tag::{self, Taggable};
 
 use crate::entity::{
-    Entity, NBTStorage,
+    Entity, EntityBase, EntityBaseFuture, NBTStorage,
     ai::goal::{
         look_around::RandomLookAroundGoal, look_at_entity::LookAtEntityGoal, swim::SwimGoal,
         wander_around::WanderAroundGoal,
     },
     mob::{Mob, MobEntity},
+    player::Player,
 };
+
+/// Duration in ticks of the poison a parrot gets from eating a cookie, matching
+/// vanilla `Parrot.mobInteract`.
+const COOKIE_POISON_DURATION: i32 = 900;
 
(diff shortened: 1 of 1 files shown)


Thesupereevee1:
@Sawlyer I belive this will crash as entitys can not have effects yet. I will test for you

Sawlyer:
Effects on non-player entities already work. `wither_rose.rs` applies wither to any living entity through the same `LivingEntity::add_effect`. The `/effect` crash from #1985 is in `registry_base_value`, which `add_effect` only reaches when the effect has attribute modifiers, and `POISON` has none.

RoosterBooster007 (approved):
This pr seems fine. It uses the ``parrot_poisonous_food`` item tag (matching vanilla), applies the actual poison effect, and includes unit tests.

RoosterBooster007:
Thanks, @Sawlyer!

PR #2565, opened 2026-07-26
feat(entity): spawn a baby when using a matching spawn egg on an animal

### LINKED ISSUE #2536: Ricgt clicking a mob with a spawn egg doesnt spawn a baby

### Operating System

Gentoo Linux

Pumpkin 0.1.0-dev+26.2

### What happened?

Nothing, no baby was spawned and the mob wasn't affected.

### To Reproduce

Right click a mob that has a baby version with its spawn egg.

### Expected behavior

The baby version of the mob is spawned at the same position as its "parent" (the mob you used the egg on).

Sawlyer:
## Description

Fixes #2536.

Vanilla handles this in `Mob.checkAndHandleImportantInteractions`, before any food interaction: a spawn egg used on a mob of its own type spawns a baby at that mob's position and consumes one egg, rather than placing a new adult next to it. `SpawnEggItem` in Pumpkin only implements `use_on_block`, so right-clicking a mob with an egg did nothing.

This adds the check to `Animal::animal_interact`, mirroring `SpawnEggItem.spawnOffspringFromSpawnEgg`: same-type egg only, baby spawned at the parent's position with yaw/pitch 0, variant components applied from the egg, one egg consumed.

Scope worth being explicit about: vanilla does this for every mob, this does it for the four Pumpkin currently models as breedable (chicken, cow, pig, sheep). Those are the mobs that already produce babies through `BreedGoal`, so no mob gains a baby state it couldn't already have. Extending it further is a matter of more mobs implementing `Animal`, not of changing this code.

The baby is flagged by its negative age, the same way `BreedGoal::breed` does it. `init_data_tracker` reads that age during `spawn_entity` and sends the client-side baby flag, which is why the age is set before the entity is spawned.

## Testing

`cargo test -p pumpkin --lib` (160 pass), `cargo clippy -p pumpkin --lib --tests`, `cargo fmt --all -- --check`.

Three unit tests cover the type guard, which is the part that decides between spawning a baby and falling through: a cow egg matches a cow, doesn't match a pig or chicken, and a non-egg item doesn't match at all. Spawning itself needs a world, so it isn't covered by an automated test.

### DIFF
--- a/crates/pumpkin/src/entity/passive/animal.rs
+++ b/crates/pumpkin/src/entity/passive/animal.rs
@@ -1,12 +1,26 @@
 use std::sync::Arc;
 
+use pumpkin_data::entity::{EntityType, entity_from_egg};
 use pumpkin_data::item_stack::ItemStack;
 use pumpkin_data::particle::Particle;
 use pumpkin_data::sound::{Sound, SoundCategory};
+use uuid::Uuid;
 
-use crate::entity::{EntityBaseFuture, mob::Mob, player::Player};
+use crate::entity::{
+    EntityBaseFuture, ageable::BABY_START_AGE, mob::Mob, player::Player, r#type::from_type,
+};
+use crate::item::items::spawn_egg::apply_entity_variant;
 use pumpkin_protocol::bedrock::server::actor_event::ActorEventType;
 use pumpkin_util::math::vector3::Vector3;
+
+/// Whether `item_stack` is the spawn egg for `entity_type`.
+///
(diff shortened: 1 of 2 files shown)


Sawlyer:

RoosterBooster007 (approved):

RoosterBooster007:
Looks good. Please fix conflicts (again).

zinedinp:
@RoosterBooster007 this seems already implemented

PR #2567, opened 2026-07-26
fix: require light for crops, stems and berry bushes to grow

eshankiyer:
## Description

**What vanilla does (1.21):** `CropBlock#randomTick` and `StemBlock#randomTick` both begin with `if (level.getRawBrightness(pos, 0) >= 9)`. `SweetBerryBushBlock#randomTick` uses the same threshold but samples `pos.above()` rather than the plant's own position. The `0` matters: it is the ambient darkness argument, so sky light is *not* reduced by the time of day and crops keep growing at night. `AttachedStemBlock` does not tick randomly at all, and nether wart, sugar cane and cactus are deliberately ungated in vanilla.

**What Pumpkin did:** all three growth paths advanced a stage on every random tick with no light check, with `//TODO add impl for light level` markers in the crop and stem paths. A sealed, unlit room grew wheat, carrots, potatoes, beetroot and gourd stems at full speed.

**Change:**

- New `World::get_raw_brightness(pos, ambient_darkness)`, which is vanilla's `getRawBrightness`: the greater of the block light and the sky light reduced by the given ambient darkness. The growth checks pass `0`.
- Crops, stems and berry bushes return early below brightness 9. Berry bushes check the block above, as vanilla does, so they gate separately from the shared crop tick.
- `get_max_local_raw_brightness` now delegates to the new function and keeps its existing `TODO: getSkyDarken`.

**Why not just call the existing `get_max_local_raw_brightness`?** It is named after vanilla's skyDarken-aware accessor and carries a TODO to become that. It happens to return `getRawBrightness(pos, 0)` today *only because that TODO is unimplemented*. Its other callers (bat and slime spawn checks) genuinely want the skyDarken behaviour, so whoever implements that TODO would silently stop crops growing outdoors at night, and the breakage would look like their fault. Splitting the two makes each call site state what it actually wants.

**Impact / known limitations:**

- A server running the non-default `lighting = "dark"` mode fills all light with zero, so this stops growth entirely there. Previously that mode had no gameplay effect.
- Bone meal is correctly unaffected: vanilla's `isBonemealSuccess` returns `true` regardless of light, and Pumpkin has no bone meal implementation yet.
- Saplings, bamboo and mushrooms have related light rules that are still missing; they are out of scope here and left with their existing TODOs.

## Testing

- `cargo clippy -p pumpkin --all-targets` and `cargo fmt --check` pass with the workspace deny-level lints, and with `RUSTFLAGS="-D warnings"`.
- Verified the vanilla thresholds and the `pos` vs `pos.above()` distinction against the 1.21 sources rather than trusting the TODO comments; the crop check moved from `pos.up()` to `pos` between 1.12 and 1.14, so older references disagree.

Suggested manual check on a server: place farmland, water and `minecraft:wheat[age=0]` inside a sealed dark box, set `gamerule random_tick_speed 1000`, and confirm it no longer advances; repeat with a torch and confirm it does.

### DIFF
--- a/pumpkin/src/block/blocks/plant/crop/gourds/stem.rs
+++ b/pumpkin/src/block/blocks/plant/crop/gourds/stem.rs
@@ -1,7 +1,10 @@
 use crate::block::{
     BlockBehaviour, BlockFuture, BlockMetadata, CanPlaceAtArgs, GetStateForNeighborUpdateArgs,
     RandomTickArgs,
-    blocks::plant::{PlantBlockBase, crop::get_available_moisture},
+    blocks::plant::{
+        PlantBlockBase,
+        crop::{MIN_GROWTH_LIGHT, get_available_moisture},
+    },
 };
 use pumpkin_data::{
     Block, BlockDirection, BlockId, BlockStateId,
@@ -77,7 +80,9 @@ impl BlockBehaviour for StemBlock {
 
     fn random_tick<'a>(&'a self, args: RandomTickArgs<'a>) -> BlockFuture<'a, ()> {
         Box::pin(async move {
-            // TODO add light level check
+            if args.world.get_raw_brightness(args.position, 0) < MIN_GROWTH_LIGHT {
+                return;
+            }
             let f: f32 = get_available_moisture(args.world, args.position, args.block).await;
             if rand::rng().random_range(0..=(25.0 / f).floor() as i32) == 0 {
                 let (block, state) = args.world.get_block_and_state_id(args.position);
--- a/pumpkin/src/block/blocks/plant/crop/mod.rs
+++ b/pumpkin/src/block/blocks/plant/crop/mod.rs
@@ -49,6 +49,9 @@ trait CropBlockBase: PlantBlockBase {
     }
 
     async fn random_tick(&self, world: &Arc<World>, pos: &BlockPos) {
+        if world.get_raw_brightness(pos, 0) < MIN_GROWTH_LIGHT {
+            return;
+        }
         let (block, state) = world.get_block_and_state_id(pos);
         let age = self.get_age(state, block);
         if age < self.max_age() {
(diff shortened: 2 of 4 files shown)


Sawlyer:
The sweet berry bush now needs light >= 9 at two positions: the new check on `pos.up()`, then the one added to `CropBlockBase::random_tick`, which runs on `pos`. Vanilla only tests the one above:

```java
if (age < 3 && random.nextInt(5) == 0 && level.getRawBrightness(pos.above(), 0) >= 9) {
```

eshankiyer:
Thanks for checking this. I don't think the second check fires on this path, though I had to prove it to myself before saying so.

`SweetBerryBushBlock` has its own `impl CropBlockBase` containing a full `random_tick` override (`crop/sweet_berry_bush.rs:190`). It's pre-existing on master and untouched by this PR, it never calls the trait default, and it has no light check. Since the call site is `<Self as CropBlockBase>::random_tick` with a concrete `Self`, it resolves to that override rather than the default in `crop/mod.rs`. So on a berry bush the only brightness test is the `pos.up()` one, which is the vanilla check.

The dispatch is genuinely easy to miss, so I've added a comment at the call site pointing at the override.

Separately, while verifying this I noticed something pre-existing worth its own issue: that override gates growth on `random_range(0..=25)` on top of the outer `0..5`, so a berry bush advances with probability ~1/130 per random tick where vanilla is 1/5. Nether wart has a smaller version of the same problem, `random_range(0..=10)` against vanilla's `nextInt(10)`. Neither is caused by this PR; I'll open an issue rather than widen this one.

eshankiyer:
Flagging a gap in my own verification of this PR, since it is the one of my light-related changes that carries real risk.

Unlike the others, this one **removes** behaviour rather than adding it. Before this change crops grew unconditionally; after it, anything that makes the light read come back low stops crop growth entirely. So "I could not observe a failure" is not good enough here, and I do not currently have a positive result showing crops still grow.

I did test on a live server, but the test was invalid: I had raised `random_tick_speed` to make growth observable, and `random_tick_speed` turns out to be a no-op (`Level::get_tick_data` samples a hardcoded 3 per section and never reads the game rule, which I have opened as #2587). Everything ran at the default rate, where a specific block ticks about once every 68 seconds, so my observation window was far too short to conclude anything.

What I have actually verified is static: the condition matches vanilla `CropBlock.randomTick`, which is `getRawBrightness(pos, 0) >= 9`, and I confirmed the lighting engine is real and is invoked on `set_block_state`, so the light values should be populated. What I have not done is plant wheat under open sky and watch it advance.

I would rather not have this merged on my say-so until I can show that positive result. Once #2587 lands I can raise the tick rate and confirm growth directly, and I will report back here either way. If you would prefer to hold this until then, that seems right to me.

One related thing worth knowing while this is open: `get_max_local_raw_brightness` currently always passes `ambient_darkness = 0`, with a TODO noting the sky darken value is not tracked. That does not affect this PR, which passes 0 deliberately to match vanilla's crop check, but it does mean anything else relying on time-of-day light is reading daylight values at night.

PR #2569, opened 2026-07-26
fix(block): make rain extinguish fire and hydrate farmland

eshankiyer:
## Description

Rain currently has no effect on two blocks that vanilla says it should. This adds the shared primitive both need and wires up the two consumers.

### `World::is_raining_at`

Mirrors vanilla `Level#isRainingAt`: it must be raining, the topmost motion blocking block must be at or below the position, the position must see the sky, and the biome's precipitation there must be rain rather than snow or none.

One detail worth review: Pumpkin's `get_heightmap_height` returns the Y of the topmost matching block, whereas vanilla's `Heightmap#getFirstAvailable` returns that Y **plus one**. The repo states this itself in `ChunkHeightmaps::update`, which does `self.get(...) + 1`. So the comparison here carries a `+ 1`. Fire at y=64 on grass at y=63 gives `63 + 1 > 64` = false, i.e. not rejected, matching vanilla. Fire is not motion blocking (`state_flags` has neither the solid nor liquid bit) so it does not raise its own heightmap and cannot mask itself.

`can_see_sky` mirrors `BlockAndTintGetter#canSeeSky`, a raw sky light comparison against the maximum light level, and rejects positions outside the world because an out of range lookup reports full sky light.

### Fire

`FireBlock::is_near_rain` was a hardcoded `false` behind a "when weather is implemented" TODO. Weather has since been implemented, so the two call sites were correctly placed but permanently dead: fire never extinguished in rain and spread freely through storms.

Note the two call sites are deliberately different, matching vanilla: `checkBurnOut` tests **only its own position** (`!level.isRainingAt(pos)`), while the spread loop uses the five position `isNearRain`.

### Farmland

Vanilla dries farmland only when `!isNearWater(pos) && !isRainingAt(pos.above())`, and `isNearWater` tests the **fluid** state so waterlogged blocks count. Pumpkin had neither: it compared block identity, so water hidden under a slab never hydrated, and it had no rain check at all. Both TODOs in that file asked for exactly this.

## Testing

- `RUSTFLAGS="-D warnings" cargo clippy -p pumpkin -p pumpkin-util --all-targets` is clean, as is `cargo fmt --check`.
- Four unit tests added for the biome precipitation helpers (no precipitation never rains, warm biomes rain, cold biomes snow, temperature falls with height).
- Verified the fluid tag genuinely covers both `water` and `flowing_water` and does not match lava, and that `get_fluid` maps waterlogged blocks to `FLOWING_WATER`.

## Known limitations, called out deliberately

- **Rain does not start on its own.** `Weather::tick_weather` advances the cycle when `weather_cycle_enabled` is **false**, and that field defaults to `true` and is never wired to the `doWeatherCycle` gamerule. So this is only observable after `/weather rain`. That inversion looks like a separate pre-existing bug; happy to fix it in its own PR rather than smuggle it in here.
- **Worlds imported from vanilla.** Sections saved without a `SkyLight` array load as sky light 0, and `needs_relighting` returns false because `isLightOn` is set, so `can_see_sky` is false in open air and this silently does nothing. That is a pre-existing light round trip gap, not introduced here, but it does bound where this fix takes effect. The heightmap test runs first partly for this reason. If maintainers would rather drop the sky light clause entirely and rely on the heightmap, that is a one line change and arguably more robust than vanilla parity here.
- `is_water_nearby` scans 162 positions and `get_fluid` allocates for blocks with properties. That is pre-existing, but this change makes it run more often; a non-allocating waterlogged fast path would be a sensible follow up.
- Two behaviours in one PR because they share the new primitive. Happy to split if preferred.

### DIFF
--- a/pumpkin-util/src/biome.rs
+++ b/pumpkin-util/src/biome.rs
@@ -67,7 +67,7 @@ impl TemperatureModifier {
 /// Represents weather information for a biome, including temperature and precipitation.
 #[derive(Clone, Debug)]
 pub struct Weather {
-    #[expect(dead_code)]
+    /// Whether this biome has any precipitation at all (rain or snow).
     has_precipitation: bool,
     /// Base temperature of the biome.
     temperature: f32,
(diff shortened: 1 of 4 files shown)


Sawlyer:
+1 looks good for me.

PR #2570, opened 2026-07-26
fix: correct cross-dimension respawn logic, End exit portal, and void damage

Necko1:
## Description
This PR addresses several critical bugs related to cross-dimension respawning, the End exit portal, and out-of-world (void) damage.

**What was changed?**
1. Throttled void damage application to once every 10 ticks instead of every tick.
2. Fixed respawn point validation to query the correct dimension and load the chunk before checking for a bed/anchor block.
3. Updated death respawn fallback to explicitly route players to the Overworld when dying without a bed, instead of reusing the death dimension.
4. Modified the End exit portal teleportation logic to respect the player's personal respawn point (bed/anchor). If no spawn point exists, it now calculates the top block at the world spawn instead of using a hardcoded `Y=200`.

- Void damage was being applied on every tick, resulting in instant deaths (80 damage/sec instead of the vanilla 8 damage/sec).
- Players dying in the End were getting stuck there because the game attempted to find their bed in the End and subsequently fell back to the End's world spawn coordinates.
- The End exit portal ignored beds completely and dropped players from the sky at `Y=200`, ignoring `/setworldspawn` top block logic.

- Void damage behaves exactly like vanilla Minecraft (4.0 damage every 10 ticks).
- Cross-dimension respawning works as intended.
- Exiting the End functions correctly, prioritizing player beds and safely placing players on the ground at world spawn if no bed is found.

- None introduced by this PR.

## Testing
- [x] Code passes `cargo clippy --all-targets` with no warnings.
- [x] Code passes `cargo test` successfully.
- [x] Tested in-game: jumping into the void, dying in the End with/without a bed, and entering the End exit portal.


RoosterBooster007:
Hey @Necko1, thanks for this. Two of the four fixes already landed in other PRs, though (the cross-dimension respawn lookup and the Overworld fallback on death), and the branch is still on the old pre-crates paths, so it won't merge as-is. It looks like the end portal and void damage fixes are still needed, so please open separate PRs if you'd like to get them in (smaller, per-fix PRs are often better for review). I'll close this PR out now.

PR #2575, opened 2026-07-26
fix: wire up gamerules that were never read

eshankiyer:
## Description

Six gamerules exist in `pumpkin-data` but nothing in `pumpkin/src` ever read them, so setting any of them had no effect. I found these by grepping every `GameRuleRegistry` field for read sites.

### `block_drops`

`World::break_block` gated drops only on `BlockFlags::SKIP_DROPS`, and `drop_loot` fed the loot table straight into `drop_stack`. `/gamerule block_drops false` did nothing.

Now gated in `drop_loot`, which has exactly two callers (block break and explosions) and so matches vanilla's scope. Vanilla checks this rule in `Block#popResource` but deliberately **not** in `Block#popExperience`, so ore experience still drops with the rule off; the gate is placed accordingly. It is deliberately not applied in `World::drop_stack`, which also serves `/loot`, chicken eggs, player death and vehicles.

### `fall_damage` and `freeze_damage`

Neither was read; `handle_fall_damage` checked only the immunity flags and the safe-fall attribute. These are now checked in the damage path directly alongside the `fire_damage` gate that was already there, which keeps all four player-damage rules in one place. Like the existing `fire_damage` and `drowning_damage` handling, they apply to players only, so mobs still take fall and freeze damage.

### `reduced_debug_info`, `immediate_respawn`, `limited_crafting`

These were passed to `CLogin` as hardcoded `false, true, false`. They now come from the gamerules. Note `immediate_respawn` is inverted at the wire because the protocol field is `enable_respawn_screen`.

## Testing

Suggested checks on a server:

```
gamerule block_drops false
# break a stone block, then:
kill @e[type=item]          # echoes a count; expect 0, was 1

gamerule fall_damage false
tp <player> ~ ~60 ~         # expect no damage on landing

gamerule immediate_respawn true
# reconnect, then die: expect no death screen
```

## Known limitations

- The three login fields are only sent at join, so changing those rules mid-session still needs a reconnect. Vanilla re-sends a `ClientboundGameEvent` on change; that felt like a separate change rather than something to fold in here, but happy to add it if preferred.
- I have not exercised these on a live server yet. The reads are one-liners against `level_info.game_rules` following the existing `fire_damage` pattern, but flagging it rather than claiming verification I did not do.

### DIFF
--- a/pumpkin/src/block/mod.rs
+++ b/pumpkin/src/block/mod.rs
@@ -444,18 +444,21 @@ pub async fn drop_loot(
     experience: bool,
     params: LootContextParameters,
 ) {
-    // Vanilla gates item drops on this rule in `Block#popResource`, but
-    // deliberately does not gate experience in `Block#popExperience`, so ore
-    // experience still drops with the rule off.
-    if world.level_info.load().game_rules.block_drops
-        && let Some(loot_table) = &block.loot_table
-    {
+    // Vanilla gates both item drops and experience on this rule, in
+    // `Block#popResource` and `Block#popExperience` respectively, so with the
+    // rule off an ore drops neither its item nor its experience orbs.
+    let block_drops = world.level_info.load().game_rules.block_drops;
+
+    if block_drops && let Some(loot_table) = &block.loot_table {
         for stack in loot_table.get_loot(params) {
             world.drop_stack(pos, stack).await;
         }
     }
 
-    if experience && let Some(experience) = &block.experience {
+    if block_drops
+        && experience
+        && let Some(experience) = &block.experience
+    {
         let mut random = RandomGenerator::Xoroshiro(Xoroshiro::from_seed(get_seed()));
         let amount = experience.experience.get(&mut random);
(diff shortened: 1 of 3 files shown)


Sawlyer:
`Block#popExperience` does check the rule in 26.2:

```java
public void popExperience(final ServerLevel level, final BlockPos pos, final int amount) {
    if (level.getGameRules().get(GameRules.BLOCK_DROPS)) {
```

So with `blockDrops` off, ore XP orbs still spawn. The `experience` branch in `drop_loot` needs the same gate as the loot table.

eshankiyer:
You're right, and my comment asserting the opposite was wrong. Pushed the fix: `block_drops` is now read once and gates both branches, so with the rule off an ore drops neither its item nor its experience orbs.

I've also corrected the comment above it. It previously claimed vanilla "deliberately does not gate experience in `Block#popExperience`", which is exactly backwards from the 26.2 source you quoted, and that's the kind of claim someone would later rely on rather than re-check.

PR #2578, opened 2026-07-26
fix(player): cost hunger for attacking and mining

eshankiyer:
## Description

**What vanilla does (1.21):** `Player.attack` calls `causeFoodExhaustion(0.1F)` on a hit that lands, and `Block.playerDestroy` calls `causeFoodExhaustion(0.005F)` per block mined.

**What Pumpkin did:** `add_exhaustion` was only ever called from the Hunger status effect and from jumping and sprinting. Attacking and mining cost nothing, so a player could fight and mine indefinitely without the hunger bar moving, and hunger was effectively cosmetic.

**Change:** adds the attack exhaustion at the end of `Player::attack`, and the mining exhaustion at the four player block-break call sites (two Java, two Bedrock).

Everything downstream was already correct and did not need touching: `hunger.rs` matches vanilla `FoodData` (4.0 exhaustion converts to one saturation or food point, the two-tier regen timers, the peaceful check, starvation).

Notes on gating, which is already handled and was not added here:

- `Player::add_exhaustion` early-returns on `abilities.invulnerable`, which `set_for_gamemode` sets for Creative and Spectator. That mirrors vanilla `causeFoodExhaustion`'s `if (!this.abilities.invulnerable)`.
- Peaceful is handled in the hunger tick, as in vanilla `FoodData.tick`.
- The attack path early-returns when `damage_with_context` reports no damage, so a miss costs nothing and sweep secondary victims correctly do not each add exhaustion.

## Testing

Suggested check: `gamerule natural_health_regeneration false` to isolate, top the player up, then break a few hundred blocks and watch the hunger bar move where it previously did not.

## Known limitations

- **Swimming exhaustion is not included.** `progress_motion` only implements the `onGround` branch of the movement statistics, and adding swimming needs the full else-if chain plus 3D distance, on top of `Player::is_swimming` which is itself an approximation carrying a TODO. Better as its own change than bolted on here.
- **Bare-handed mining costs no exhaustion**, because the call sites sit inside the existing `can_harvest` branch. I believe that matches vanilla, where `destroyBlock` only reaches `playerDestroy` with the correct tool, but I am flagging it as the least certain part of this change rather than asserting it.
- The three constants are from vanilla rather than derived from anything in this repo. Worth a maintainer's eye. Note that the `exhaustion: 0.1` field on `player_attack` in `damage_type.rs` looks like corroboration but is not: that is the victim-side `DamageSource` exhaustion, a separate mechanic which is generated but never read anywhere.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -122,6 +122,14 @@ const MAX_PREVIOUS_MESSAGES: u8 = 20; // Vanilla: 20
 
 pub const DATA_VERSION: i32 = 4903; // 26.2
 
+/// Food exhaustion applied for every block a player mines.
+///
+/// Vanilla: `Block#playerDestroy` calls `player.causeFoodExhaustion(0.005F)`.
+/// `ServerPlayerGameMode#destroyBlock` only reaches `playerDestroy` for
+/// non-creative players holding a tool that can harvest the block, so callers
+/// must apply the same gating.
+pub const MINE_BLOCK_EXHAUSTION: f32 = 0.005; // Vanilla: 0.005F
+
 struct HeapNode(i32, Vector2<i32>, Weak<ChunkData>);
 
 impl Eq for HeapNode {}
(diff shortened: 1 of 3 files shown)


Sawlyer:
+1 looks good for me.

PR #2579, opened 2026-07-26
feat(block): melt ice in bright light

eshankiyer:
## Description

**What vanilla does (1.21):** `IceBlock.randomTick` melts ice when the block light exceeds `11 - state.getLightBlock()`. `melt()` sets water in normal dimensions and removes the block in ultrawarm ones, then notifies neighbours. Only `ice` melts; packed ice and blue ice never do.

**What Pumpkin did:** there was no `ice.rs` and no behaviour registered for `Block::ICE`, so the random tick was delivered and discarded. Ice next to a torch never turned to water, and ice placed in the Nether persisted.

**Change:** adds `IceBlock` with a `random_tick`, mirroring the existing `snow.rs` melt structure, and registers it.

**On the light threshold.** This is `11 - opacity` computed from the live state rather than hardcoded. `BlockState::opacity` is Pumpkin's generated equivalent of vanilla `getLightBlock` (its own doc says "how much light is subtracted as it passes through this block"), and the lighting engine uses that same field as the per-block attenuation. `minecraft:ice` is generated with `opacity: 1`, so the effective threshold is block light `> 10`, one lower than snow's flat `> 11`. Spot checks line up with vanilla: stone 15, glass 0, water 1, packed ice and blue ice 15.

`state_flags` confirms the scope: `ice` carries the random-tick bit while `packed_ice` and `blue_ice` do not, matching vanilla.

This uses `set_block_state` rather than `break_block`, because vanilla's `melt()` drops nothing; `break_block` would have wrongly dropped an ice item.

## Testing

Suggested check:

```
setblock 100 70 100 ice
setblock 101 70 100 torch
# after a random tick
execute if block 100 70 100 water run say melted
```

## Known limitations

- **Ultrawarm is detected as "is the Nether"**, because `Dimension` has no `ultrawarm` field. That follows the existing convention in `block/fluid/lava.rs`, which does the same with the same comment. A custom datapack dimension declaring `ultrawarm: true` would melt to water rather than evaporating.
- `IceBlock.playerDestroy`, where mining ice without Silk Touch leaves water, is still unimplemented and out of scope here.
- The melt formula's shape is recalled from vanilla rather than read from a source in this repo; the constant itself is derived from the generated data as described above.

### DIFF
--- a/pumpkin/src/block/blocks/ice.rs
+++ b/pumpkin/src/block/blocks/ice.rs
@@ -1,0 +1,52 @@
+use pumpkin_data::{Block, dimension::Dimension};
+use pumpkin_macros::pumpkin_block;
+use pumpkin_world::world::BlockFlags;
+
+use crate::block::{BlockBehaviour, BlockFuture, RandomTickArgs};
+
+#[pumpkin_block("minecraft:ice")]
+pub struct IceBlock;
+
+impl BlockBehaviour for IceBlock {
+    fn random_tick<'a>(&'a self, args: RandomTickArgs<'a>) -> BlockFuture<'a, ()> {
+        Box::pin(async move {
+            // `IceBlock.randomTick` melts once the block light exceeds
+            // `11 - state.getLightBlock()`. Java's `getLightBlock()` is `BlockState::opacity`
+            // here, and ice has an opacity of 1, so the effective threshold is a block light
+            // level above 10 (one lower than the flat `> 11` used for snow layers).
+            let opacity = args.world.get_block_state(args.position).opacity;
+            if args.world.get_block_light_level(args.position).unwrap_or(0)
+                <= 11u8.saturating_sub(opacity)
+            {
+                return;
+            }
+
+            // `IceBlock.melt`: ultrawarm dimensions evaporate the ice, everything else leaves
+            // a water source behind. Melting never drops an ice item, so the state is
+            // overwritten instead of broken.
+            if args.world.dimension == Dimension::THE_NETHER {
+                args.world
+                    .set_block_state(
+                        args.position,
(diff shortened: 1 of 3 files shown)


Sawlyer:
+1 looks good for me.

eshankiyer:
Correcting myself on this one.

I tested this on a live server and reported that the behaviour did not fire. That result was wrong, and I want to retract it before anyone tries to reproduce it.

I had set `random_tick_speed` high to make the behaviour observable. `random_tick_speed` turns out to be a no-op: `Level::get_tick_data` samples a hardcoded 3 positions per chunk section per tick and never reads the game rule. So every one of my tests ran at the default rate, where a *specific* block is random-ticked roughly once every 68 seconds. My checks ran for 20 to 60 seconds and had somewhere between a 7 and 25 percent chance of observing the behaviour even if the code were perfect. The negative results were sampling noise, not evidence.

Two of the premises I reasoned from were also wrong, and I would rather say so than leave them in the thread:

- I used snow failing to melt in daylight as corroboration. Vanilla `SnowLayerBlock.randomTick` uses `getBrightness(LightLayer.BLOCK, pos)`, block light only, so snow in open daylight with no torches should never melt. It was working correctly.
- I used fire coral bleaching as a control proving random ticks were firing. Pumpkin implements coral death as a *scheduled* tick queued from `placed`, not a random tick, so it proved nothing about the random tick path.

I have opened #2587 for the `random_tick_speed` bug. Once that lands I can raise the rate for real, and the same checks resolve in seconds instead of being unobservable. I will re-test this PR then and post an actual result.

Nothing about the code in this PR has changed. What changed is that I no longer have evidence either way, and I would rather say that plainly than leave an unsupported claim standing.

PR #2580, opened 2026-07-26
feat(block): spread and kill grass blocks on random tick

eshankiyer:
## Description

**What vanilla does (1.21):** `SpreadingSnowyDirtBlock.randomTick` reverts grass to dirt when it can no longer survive, and otherwise, when `getMaxLocalRawBrightness(pos.above()) >= 9`, makes four spread attempts at random offsets `(nextInt(3)-1, nextInt(5)-3, nextInt(3)-1)`, converting a target `dirt` to grass when that target could itself support grass and has no water above it.

**What Pumpkin did:** `grass_block.rs` implemented only the `snowy` property on neighbour updates. With no `random_tick`, the tick fell through to the no-op default, so dirt exposed by mining never regrew and grass under a placed block never died.

**Change:** implements `random_tick` for `grass_block` only. Mycelium and podzol have no behaviour registered at all and are left for a separate change.

Note the survival and spread checks use different rules in vanilla, and that distinction is preserved: survival is occlusion-based with no light check, while the `>= 9` brightness gate applies only to spreading.

## Testing

Suggested check: `fill` a grass platform, `setblock` one block to dirt, and confirm it regrows. Note `random_tick_speed` is currently hardcoded to 3 and ignores the gamerule, so this cannot be sped up and needs wall-clock patience.

## Known limitations, stated plainly

- **The survival check is not fully faithful.** Vanilla runs the block above through `LightEngine#getLightBlockInto`, which reports full occlusion when the touching faces occlude. Pumpkin has no face-occlusion lookup, so only the raw opacity is available. Blocks that occlude downwards while carrying a low opacity, notably carpets, bottom slabs, and snow layers of two or more, let grass survive where vanilla would kill it. Snow layers are handled explicitly since vanilla calls that case out by name; the rest remain divergent. Grass under a full opaque block does die, which is the common case. Closing this properly needs a face-occlusion flag from the data generator, which felt out of scope. `BlockState::is_side_solid` is not a usable substitute: glass reports the same `side_flags` as a full opaque cube.
- **Grass spreads at night and during storms**, because `get_max_local_raw_brightness` does not yet subtract sky darkening (it carries its own TODO for this). Pre-existing, but this change is what makes it observable, so it roughly doubles the real spread rate.
- The fluid check matches on `block_state_id` rather than using `Fluid::get_level`, because those helpers index with `state_id % states.len()`, which is wrong for a 16-state fluid (`get_level(94)` yields 7 rather than 8). That looks like a separate pre-existing bug worth its own issue; I worked around it rather than touching it here.
- No plugin event is fired for spread or reversion. The repo has `BlockGrowEvent` but reusing it here would be semantically wrong.

### DIFF
--- a/pumpkin/src/block/blocks/grass_block.rs
+++ b/pumpkin/src/block/blocks/grass_block.rs
@@ -2,7 +2,7 @@ use pumpkin_data::BlockStateId;
 use pumpkin_data::fluid::Fluid;
 use pumpkin_data::{
     Block,
-    block_properties::{BlockProperties, GrassBlockLikeProperties},
+    block_properties::{BlockProperties, GrassBlockLikeProperties, SnowLikeProperties},
     tag::{self, Taggable},
 };
 use pumpkin_macros::pumpkin_block;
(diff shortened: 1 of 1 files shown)


Sawlyer:
+1 looks good for me.

eshankiyer:
Correcting myself on this one.

I tested this on a live server and reported that the behaviour did not fire. That result was wrong, and I want to retract it before anyone tries to reproduce it.

I had set `random_tick_speed` high to make the behaviour observable. `random_tick_speed` turns out to be a no-op: `Level::get_tick_data` samples a hardcoded 3 positions per chunk section per tick and never reads the game rule. So every one of my tests ran at the default rate, where a *specific* block is random-ticked roughly once every 68 seconds. My checks ran for 20 to 60 seconds and had somewhere between a 7 and 25 percent chance of observing the behaviour even if the code were perfect. The negative results were sampling noise, not evidence.

Two of the premises I reasoned from were also wrong, and I would rather say so than leave them in the thread:

- I used snow failing to melt in daylight as corroboration. Vanilla `SnowLayerBlock.randomTick` uses `getBrightness(LightLayer.BLOCK, pos)`, block light only, so snow in open daylight with no torches should never melt. It was working correctly.
- I used fire coral bleaching as a control proving random ticks were firing. Pumpkin implements coral death as a *scheduled* tick queued from `placed`, not a random tick, so it proved nothing about the random tick path.

I have opened #2587 for the `random_tick_speed` bug. Once that lands I can raise the rate for real, and the same checks resolve in seconds instead of being unobservable. I will re-test this PR then and post an actual result.

Nothing about the code in this PR has changed. What changed is that I no longer have evidence either way, and I would rather say that plainly than leave an unsupported claim standing.

PR #2586, opened 2026-07-26
feat: add multi-instance hosting core API

nglmercer:
## Summary
This PR implements the P0 core infrastructure for using Pumpkin as a multi-instance library:
- Adds an InstanceManager lifecycle API for create/start/stop/delete operations.
- Allocates TCP and UDP ports per instance with conflict detection and fallback.
- Adds per-instance world, data, plugin, and plugin-data directories.
-  Adds per-instance configuration roots and data persistence paths.
- Uses per-instance cancellation and graceful shutdown instead of process-wide stop state.
- Returns bind failures through the library constructor path instead of exiting the process.
- Preserves the existing single-instance entry points for backward compatibility.
## Verification
running 8 tests
test instance::port_allocator::tests::preferred_outside_range_is_kept_when_available ... ok
test instance::port_allocator::tests::tcp_and_udp_can_share_a_port ... ok
test instance::port_allocator::tests::test_allocate_or_any_fallback ... ok
test instance::port_allocator::tests::test_out_of_range ... ok
test instance::config::tests::explicit_config_root_is_used_for_default_paths ... ok
test instance::config::tests::reserve_ports_updates_network_configuration ... ok
test instance::port_allocator::tests::test_allocate_any ... ok
test instance::port_allocator::tests::test_allocate_and_free ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 157 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s (8 passed)
## Scope
This PR covers the P0 core needed for orchestration. REST/gRPC management, resource enforcement, metrics, snapshots, hot-swapping, and auto-scaling remain follow-up work rather than being represented as completed here.

### DIFF
--- a/pumpkin-config/src/advancement.rs
+++ b/pumpkin-config/src/advancement.rs
@@ -2,7 +2,7 @@ use serde::{Deserialize, Serialize};
 /// Configuration for advancements
 ///
 /// Controls whether the advancements should be saved and loaded
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize, Serialize, Clone)]
 #[serde(default)]
 pub struct AdvancementConfig {
     /// Whether saving advancements is enabled.
--- a/pumpkin-config/src/chat.rs
+++ b/pumpkin-config/src/chat.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize};
 /// Configuration for in-game chat behaviour.
 ///
 /// Controls chat formatting and display.
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize, Serialize, Clone)]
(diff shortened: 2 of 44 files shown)


nglmercer:
***note: exist performance issue not ready for merge***

nglmercer:
There’s an issue where if the first player takes too long to join or times out, subsequent players can still connect fine. I suspect a CPU spike during chunk loading might be the culprit, though I’m not 100% sure. It feels similar to the behavior we had before these changes, but it could just be me.

PR #2587, opened 2026-07-26
fix: honour the randomTickSpeed game rule

eshankiyer:
## What

`/gamerule random_tick_speed` does nothing. Setting it to `0` does not stop random ticks, and raising it does not speed them up.

## Why

`Level::get_tick_data` samples a hardcoded 3 positions per chunk section per tick:

```rust
for _ in 0..3 {
    let r = rand::random::<u32>();
```

`random_tick_speed` exists in the generated game rule data and round-trips through `level.dat`, but nothing in the tick path ever reads it.

## How

The value is threaded in from `World::tick_chunks`, which reads it from the level info on each tick so `/gamerule` applies immediately rather than at restart. A value of `0` skips the sampling entirely, matching vanilla's `if (randomTickSpeed > 0)` guard. Scheduled block and fluid ticks stay outside that guard and keep running at `0`, which is also what vanilla does.

`pumpkin-world` has no access to the game rules, so this is a parameter rather than a new dependency edge. `get_tick_data` has exactly one caller.

## The cap

The value is clamped to `MAX_RANDOM_TICK_SPEED = 4096`, the block count of a chunk section, past which every block in a section already gets roughly one expected random tick per game tick.

This is not cosmetic. The game rule argument is an unbounded `i64`, and the old capacity hint was:

```rust
random_ticks: Vec::with_capacity(active_chunks.len() * 3),
```

With the rule wired up and no cap, `active_chunks` is 289 entries per player and `RandomTickSample` is 16 bytes, so `random_tick_speed = i64::MAX` asks for about 19.8 TB every tick. That is below `isize::MAX`, so it does not raise the catchable capacity-overflow panic; it reaches the allocator, fails, and `handle_alloc_error` aborts the process. Even `100000` would allocate and free roughly 462 MB per tick per world. Since the rule was previously inert, wiring it up without a cap would have introduced a one-command server kill.

I dropped the capacity hint rather than scaling it. It multiplied by chunk *columns* while the loop pushes per *section* and then filters on `has_random_ticks`, so it both under- and over-counted; plain `Vec::new` growth is bounded by the samples that actually survive the filter.

The command argument itself is left unbounded. `init_command_tree` builds one `BoundedNumArgumentConsumer::<i64>` shared by every integer game rule, so adding min/max there would apply to all of them. Vanilla also accepts arbitrary values and only bounds the interpretation.

## Why this one matters beyond itself

At the hardcoded rate a *specific* block is random-ticked about once every 68 seconds. That makes gameplay behaviour effectively untestable: I spent a while chasing what looked like broken ice melting, snow melting and grass spreading on a live server, and all three were sampling noise from tests that had a 7 to 25 percent chance of observing the behaviour at all. With this fix and `random_tick_speed` raised, the same checks resolve in seconds.

## Verification

`cargo fmt --check` and `RUSTFLAGS="-D warnings" cargo clippy -p pumpkin -p pumpkin-world --all-targets` clean.

Not yet confirmed in game, for the circular reason above: confirming it requires the fix. I will report back once I have it running on my test server.

## Note for existing worlds

A world whose `level.dat` already carries a non-default `random_tick_speed` will start honouring it. Worlds that never set it default to 3 and are unaffected.

### DIFF
--- a/pumpkin-world/src/level.rs
+++ b/pumpkin-world/src/level.rs
@@ -47,6 +47,15 @@ use tokio_util::task::TaskTracker;
 pub type SyncChunk = Arc<ChunkData>;
 pub type SyncEntityChunk = Arc<ChunkEntityData>;
 
+/// Upper bound applied to the `randomTickSpeed` game rule before it is used for sampling.
+///
+/// A chunk section holds 16 * 16 * 16 = 4096 blocks, so sampling 4096 positions per section
+/// already gives every block roughly one expected random tick per game tick. Anything beyond
+/// that adds cost without adding behaviour, and the game rule itself is stored as an
+/// unbounded signed integer, so it has to be capped somewhere to keep the per-tick work
+/// (and the allocations it drives) bounded.
+pub const MAX_RANDOM_TICK_SPEED: u32 = 4096;
+
 /// The `Level` module provides functionality for working with chunks within or outside a Minecraft world.
 ///
 /// Key features include:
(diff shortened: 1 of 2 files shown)


eshankiyer:
No problem with the close. The rule is still unwired on current master and #2575 does not cover it; would you take a rebased minimal version, or is this planned differently?

PR #2593, opened 2026-07-27
fix(entity): separate the entity tick counter from the ageable age 🤖🤖🤖

RicePollution:
## What changed

`Entity::age` was serving two different vanilla concepts at once:

- **`Entity#tickCount`** — ticks since spawn, only ever counting up.
- **`AgeableMob#age`** — negative for a baby counting up to zero, zero for an adult, positive for the breeding cooldown left behind by breeding and counting back *down* to zero.

The world entity tick did `age.fetch_add(1)` once per tick, which made the field usable as a tick counter but meant the ageable meaning was never actually implemented — a positive age grew instead of shrinking, and nothing ever consulted the age lock.

This PR adds `Entity::tick_count` for the tick-counter meaning and leaves `age` to mean only the ageable age, progressed per vanilla `AgeableMob#aiStep` by the new `Entity::tick_age`. Every "time since I spawned" reader moves to `tick_count`; the `age` readers that remain are all baby/adult checks (`< 0` / `>= 0`), which is what they always wanted.

Both fields are documented so the next reader does not have to rediscover the distinction.

## Why — four defects fell out of the overload

**1. Baby mobs re-planned their AI twice as often as vanilla.** The gate in `Mob`'s tick was `if (age + entity_id) % 2 != 0 && age > 1`, mirroring vanilla's

```java
int i = this.level().getServer().getTickCount() + this.getId();
if (i % 2 != 0 && this.tickCount > 1) { ...tickRunningGoals(false) } else { ...tick() }
```

but with `age` standing in for both `getTickCount()` and `tickCount`. A baby's age is negative for its whole 20 minutes of childhood, so `age > 1` never held, the light pass was never taken, and the full goal-selector pass — which re-evaluates `can_start`/`should_continue` for every goal — ran on *every* tick. That doubles the goal-selection work for every baby mob, and because ~14 goals size their intervals with `to_goal_ticks` on the assumption of being re-evaluated every other tick, it also halves all of those intervals. The gate is now `runs_full_goal_pass`, keyed off the server tick count and the entity's own tick count, matching vanilla.

**2. Infinite status effects never ticked on baby mobs.** `tick_effects` uses the entity's tick counter as the clock for effects with no duration of their own (`duration == -1`). A baby's negative age meant every `duration <= 0` guard in `should_apply_effect_tick` bailed out, so infinite Regeneration, Poison and Wither did nothing at all on a baby animal.

**3. Adult cows, pigs and chickens persisted a meaningless `Age` NBT tag.** Their age accumulated ticks-alive without bound, and `write_ageable_nbt` saves it as `Age`. An adult alive for an hour wrote `Age: 72000`, which vanilla reads back as a 72000-tick breeding cooldown. Adults now correctly sit at zero.

**4. `AgeLocked` and `ForcedAge` were inert.** Both are read and written in `read_ageable_nbt`/`write_ageable_nbt`, but the only code that consulted the lock was `AgeableMob::ageable_ai_step`, which was never called from anywhere. That dead function is removed, and `age_locked` moves next to `age` on `Entity` so the central progression can honour it without needing to know the entity is ageable.

Also repoints the plugin API's `get-ticks-lived`/`set-ticks-lived`. The WIT contract (`v0.1/world.wit`) declares those separately from `get-age`/`set-age`, but the host implementation backed both with the same field, so a plugin asking a baby cow how long it had lived got a negative number.

## Impact

Baby mobs now plan on vanilla's cadence rather than double-rate, which is also strictly less per-tick work. Infinite effects work on babies. Adult animals stop writing a bogus `Age` to disk, so worlds round-trip correctly and stay readable by vanilla. `AgeLocked` starts working.

Behaviour for players and for non-ageable mobs is unchanged: their age was always non-negative, so every reader that moved to `tick_count` sees the same values it saw before.

## Tests

11 new unit tests, all pure-function so they need no running server:

- `Entity::stepped_age` — babies count up, breeding cooldowns count down, adults stay at zero, no overflow at `i32::MIN`/`i32::MAX`, and a baby reaches adulthood in exactly 24000 ticks (vanilla's 20 minutes).
- `runs_full_goal_pass` — alternates every other tick, is staggered by entity id so neighbouring mobs do not re-plan on the same tick, always runs during the spawn warm-up, skips half the passes once settled (the regression in defect 1), and survives tick-counter wraparound.

Full local CI is green: `typos`, `cargo fmt --check`, `cargo clippy --all-targets --all-features`, the same in `--release`, `cargo nextest run` (449 passed), `cargo test --doc`, `cargo machete`.

## Known limitations

- `Player::tick_counter` now duplicates `Entity::tick_count`. I left it alone to keep this diff focused; unifying them would be a reasonable follow-up.
- `AgeableMob` is currently only implemented by cow, pig and chicken, so defects 3 and 4 were only observable on those three. The fix is in `Entity`, so it applies to whatever else adopts the trait.
- `ForcedAge` is now honoured only through `age_up`, which is the sole caller; it is still not applied when a forced-age baby reaches adulthood. Out of scope here.
- `stepped_age` applies to every entity, not just ageable ones. For anything that never sets an age this is a no-op, exactly as the previous unconditional increment was for non-ageable entities.

### DIFF
--- a/pumpkin/src/entity/ageable.rs
+++ b/pumpkin/src/entity/ageable.rs
@@ -1,7 +1,7 @@
 use pumpkin_data::meta_data_type::MetaDataType;
 use pumpkin_data::tracked_data::TrackedData;
 use pumpkin_protocol::java::client::play::Metadata;
-use std::sync::atomic::{AtomicBool, AtomicI32, Ordering::Relaxed};
+use std::sync::atomic::{AtomicI32, Ordering::Relaxed};
 
 use crate::entity::mob::Mob;
 
@@ -11,7 +11,6 @@ pub const FORCED_AGE_PARTICLE_TICKS: i32 = 40;
 pub struct AgeableData {
     pub forced_age: AtomicI32,
     pub forced_age_timer: AtomicI32,
-    pub age_locked: AtomicBool,
     pub age_lock_particle_timer: AtomicI32,
 }
 
@@ -20,7 +19,6 @@ impl Default for AgeableData {
         Self {
             forced_age: AtomicI32::new(0),
             forced_age_timer: AtomicI32::new(0),
-            age_locked: AtomicBool::new(false),
             age_lock_particle_timer: AtomicI32::new(0),
         }
     }
@@ -64,11 +62,22 @@ pub trait AgeableMob: Mob {
     }
 
     fn is_age_locked(&self) -> bool {
-        self.get_ageable_data().age_locked.load(Relaxed)
+        self.get_mob_entity()
+            .living_entity
+            .entity
+            .age_locked
+            .load(Relaxed)
     }
 
+    /// The lock lives next to the age on [`crate::entity::Entity`] so that the
+    /// central age progression in [`crate::entity::Entity::tick_age`] can honour
+    /// it without having to know that this entity is ageable.
     fn set_age_locked(&self, locked: bool) {
-        self.get_ageable_data().age_locked.store(locked, Relaxed);
+        self.get_mob_entity()
+            .living_entity
+            .entity
+            .age_locked
+            .store(locked, Relaxed);
     }
 
     fn can_age_up(&self) -> bool {
@@ -127,15 +136,5 @@ pub trait AgeableMob: Mob {
 
     fn can_be_a_baby(&self) -> bool {
         true
-    }
-
-    fn ageable_ai_step(&self) {
-        if self.can_age_up() {
-            let age = self.get_age() + 1;
-            self.set_age(age);
-        } else if self.get_age() > 0 {
-            let age = self.get_age() - 1;
-            self.set_age(age);
-        }
     }
 }
--- a/pumpkin/src/entity/ai/goal/escape_danger.rs
+++ b/pumpkin/src/entity/ai/goal/escape_danger.rs
@@ -36,8 +36,8 @@ impl EscapeDangerGoal {
         if last_attacked == 0 {
             return false;
         }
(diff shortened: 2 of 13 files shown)


Sawlyer:
Vanilla is `int idBasedTickCount = this.tickCount + this.getId();` in `Mob#serverAiStep`, so the parity comes off the entity's own tick count. `runs_full_goal_pass` takes `entity_tick_count` but only uses it for the warm-up and feeds the server tick into the parity instead. Any reason not to use the entity count for both?

RicePollution:
Good catch, and you're right — fixed in the latest commit.

`Mob#serverAiStep` derives both terms from one field:

```java
int idBasedTickCount = this.tickCount + this.getId();
if (idBasedTickCount % 2 != 0 && this.tickCount > 1) { ...reduced pass... }
```

`this.tickCount` is `Entity#tickCount`, so the alternation belongs to the entity and the id is only a stagger. I had corrected the field for the warm-up and then left the server tick in the parity, which splits what vanilla keeps as one value. `runs_full_goal_pass` now takes `(entity_id, entity_tick_count)` and the `server_tick` parameter is gone.

It's worth being precise about why this isn't cosmetic, since the two differ only by a per-entity constant and so look interchangeable: an entity that misses server ticks keeps its own counter contiguous while the server's runs on, so keying off the server would shift that entity's parity for reasons unrelated to how often it has actually ticked.

Also confirmed while checking: `serverAiStep` is declared `protected final`, so no subclass can alter the parity — this holds for every mob. Added a test pinning the alternation to the entity counter.

PR #2594, opened 2026-07-27
fix(entity): spawn one child per breeding pair, not one per parent 🤖🤖🤖

### LINKED ISSUE #2532: Breeding results in 2 babies instead of 1

### Operating System

Gentoo Linux

Pumpkin 0.1.0-dev+26.2

### What happened?

When breeding two cows it always produces 2 babies but when breeding pigs it often produces two babies but sometimes only one.

### To Reproduce

Breed an animal.

### Expected behavior

Breeding an animal only results in one baby.

RicePollution:
Fixes #2532.

## What changed

Breeding two animals produced **two** babies. Both partners run their own `BreedGoal`, both find each other via `find_mate`, and both count `timer` up in lockstep — so both reach `timer >= 60` on the same tick, and `breed()` spawned a child for each of them.

Two changes:

1. **Decide which partner spawns the child, from the two entity ids.** `BreedGoal::spawns_the_child(entity_id, mate_entity_id)` is `entity_id < mate_entity_id`, so exactly one side of any pair says yes.
2. **Guard `breed()` against re-entry** on `!is_in_love() || !is_breeding_ready()`.

## Why it has to be an id comparison rather than a state re-check

The obvious fix is to have `breed()` re-read the partner's in-love state and bail if it's already been consumed. That is not sufficient here. `World::tick` spawns each entity's tick into a `tokio::task::JoinSet`, so entity ticks run **concurrently** — both partners can execute `BreedGoal::tick` at the same time and both observe the not-yet-bred state before either one writes. Any read-then-write guard has a window.

Comparing the two entity ids has no window: both sides compute the same answer from the same two inputs, with no shared state to race on. The re-entry guard is still worth having, but for a different reason — see below.

## Why the second guard is needed too

`BreedGoal::should_run_every_tick()` is `true`, so `tick` runs on every tick, but `should_continue` — which contains the `self.timer < 60` stop condition — is only re-evaluated on the full goal-selector pass, i.e. every *other* tick. So on the tick after the pair breeds, `tick` can run again with `timer == 61` and still satisfy `timer >= 60 && dist_sq < 9.0`, spawning a second child before the goal is stopped. Breeding consumes the in-love state and puts both parents on a 6000-tick cooldown, so refusing when that state is gone closes the window.

## This explains the reported asymmetry

The issue notes cows *always* producing two babies but pigs only sometimes. That follows from the above: when the two partners' goals were a tick out of phase, whether the second one got stopped by `should_continue` before its own `tick` reached 60 depended on the order the two entity tick tasks happened to run in that server tick. In phase → both always bred. Out of phase → it depended on scheduling.

## Impact

One child per successful breeding, as in vanilla. Both parents still get their in-love state cleared and a 6000-tick breeding cooldown, and the breeder still gets the `AnimalsBred` stat and the `BredAnimal` advancement exactly once, so those are no longer double-counted either.

## Tests

4 unit tests, pure-function so no server is needed. The important one asserts the actual invariant rather than the implementation: for a pair, asking from both sides yields exactly one yes — two yeses is this bug, two noes would mean breeding silently stopped working. Also covers negative entity ids (Bedrock paths can hand those out) and the equal-id case.

Full local CI green: `typos`, `cargo fmt --check`, `cargo clippy --all-targets --all-features`, the same in `--release`, `cargo nextest run` (443 passed), `cargo test --doc`, `cargo machete`.

## Known limitations

- The lower-id partner is the one that spawns, so if *its* goal is not running (preempted by a higher-priority goal holding `MOVE`) while the higher-id one's is, no child is spawned on that attempt. Both animals stay in love, so the next attempt succeeds once the lower-id goal runs again. In practice both goals run, since breeding already requires mutual love and a mutual `find_mate`.
- Unlike vanilla, `breed()` still does not set the parents' `Age` to 6000; Pumpkin tracks the post-breeding cooldown in a separate `breeding_cooldown` field, which has the same effect. Left alone deliberately to keep this change scoped.
- The baby spawns at the initiating parent's position rather than between the two parents. Pre-existing, not touched here.

### DIFF
--- a/pumpkin/src/entity/ai/goal/breed.rs
+++ b/pumpkin/src/entity/ai/goal/breed.rs
@@ -60,8 +60,33 @@ impl BreedGoal {
         closest.map(|(_, e)| e)
     }
 
+    /// Which of the two partners is the one that spawns the child.
+    ///
+    /// Both partners run their own [`BreedGoal`], find each other as a mate, and
+    /// count their timers up in lockstep, so both reach the point of breeding on
+    /// the same tick and each would spawn a child of its own. Having one side
+    /// re-read the other's in-love state is not enough to stop that: the world
+    /// ticks entities concurrently on a `JoinSet`, so both sides can observe the
+    /// not-yet-bred state before either of them writes.
+    ///
+    /// Deciding it from the entity ids instead needs no shared state and cannot
+    /// race, because both sides compute the same answer from the same two ids.
+    #[must_use]
+    pub const fn spawns_the_child(entity_id: i32, mate_entity_id: i32) -> bool {
+        entity_id < mate_entity_id
+    }
+
     async fn breed(mob: &dyn Mob, mate: &dyn EntityBase) {
         let mob_entity = mob.get_mob_entity();
+
+        // `tick` keeps running on the ticks in between goal re-evaluations, so
+        // this can be reached again after the pair has already bred. Breeding
+        // consumes the in-love state and puts both parents on cooldown, so once
+        // that is gone there is nothing left to breed.
+        if !mob_entity.is_in_love() || !mob_entity.is_breeding_ready() {
+            return;
+        }
+
         let entity = mob.get_entity();
         let world = entity.world.load();
 
@@ -168,7 +193,10 @@ impl Goal for BreedGoal {
 
             self.timer += 1;
 
-            if self.timer >= 60 && dist_sq < 9.0 {
+            if self.timer >= 60
+                && dist_sq < 9.0
+                && Self::spawns_the_child(mob.get_entity().entity_id, mate.get_entity().entity_id)
+            {
                 Self::breed(mob, mate.as_ref()).await;
             }
         })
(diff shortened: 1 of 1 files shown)


Sawlyer:
`find_mate` takes the closest in-love candidate, so the pairing is not necessarily mutual. Three animals in love, A---B-C with B nearer to C: A picks B, B picks C, C picks B. With ids 1/2/3, `spawns_the_child(1, 2)` and `spawns_the_child(2, 3)` are both true, so B is a parent in two breed calls, and the only thing left stopping the second one is the `is_in_love` re-check, which is the read-then-write window the id comparison exists to avoid.

Vanilla gets one baby out of three in-love animals. In scope here, or a follow-up?

RicePollution:
You're right that the pairing isn't mutual and that the chain is reachable. `find_mate` takes the nearest in-love candidate with no reciprocity check, so for A---B-C with B nearer C you get A→B, B→C, C→B, and with ids 1/2/3 both `spawns_the_child(1, 2)` and `spawns_the_child(2, 3)` are true. B is a parent in two breed calls and the only thing stopping the second is the `is_in_love()` re-check at the top of `breed`, which is exactly the read-then-write window the id comparison exists to close. No argument there.

Where I'd push back is the vanilla baseline. Vanilla doesn't give one baby from three in-love animals either:

- `BreedGoal#canUse` tests `this.animal.isInLove()` and `getFreePartner()` → `Animal#canMate`, which requires both sides in love. But `getFreePartner` is nearest-wins, not mutual, so vanilla forms the same A→B, B→C, C→B chain.
- `BreedGoal#canContinueToUse` is `this.partner.isAlive() && this.partner.isInLove() && this.loveTime < 60`. It validates the **partner**, never the goal owner. And `GoalSelector`'s update loop skips goals where `isRunning()`, so a running `BreedGoal` never re-consults `canUse`/`canMate` — those constrain acquisition only, never continuation.
- `finalizeSpawnChildFromBreeding` resets love on exactly the two participants.

So if the tail animal breeds first, A and B are reset but C isn't; B's goal still passes `canContinueToUse` because C is still in love, and B breeds a second time. Two babies. If B or C goes first, both get reset and you get one. It's order-dependent — reported as MC-201000, still open.

That matters for scoping. The invariant vanilla actually holds is one baby per **mutual pair**, which is #2532 and what this PR fixes; sequential ticking is what makes the check-then-act atomic there, and the concurrent `JoinSet` is what destroys it — hence the id tiebreak. The three-animal chain is a separate, pre-existing non-mutual-pairing bug that master has too, and "fixing" it here would make Pumpkin stricter than vanilla, which seems like the wrong thing to do quietly inside a bug fix.

Happy to open a follow-up for the chain case if you'd like it made deterministic — the honest fix is reciprocity in `find_mate` rather than anything in `spawns_the_child` — but I'd rather not widen this PR into deliberately diverging from vanilla without that being the stated intent.

One thing worth noting for the record: Pumpkin already has a guard vanilla lacks. `breed()` re-checks the caller's own `is_in_love()`/`is_breeding_ready()` before doing anything, where vanilla's `breed()` has no self-check at all. That's what makes the chain case merely racy here instead of reliably producing two babies.

PR #2595, opened 2026-07-27
fix(entity): restore the state a status effect implies when loading it 🤖🤖🤖

### LINKED ISSUE #2520: /Effect command desynchronizing, When Leaving / Joining

### Operating System

Lunix Mint 22.3

Pumpkin 0.1.0-dev+26.2 (Commit: 0d44394/release)

### What happened?

When I Joined The Server On minecraft java And Applied the speed effect to Myself By Running
`/effect give <target> minecraft:speed infinite 7 false`

Image

 The Effect Fully Works. But After Leaving and rejoining the Server, the effect Isnt Applied, But on the top Right hand side of the screen, It still says the Effect is applied. But when i try moving, it isnt.

Image

### To Reproduce

1. Start PumpkinMC and join the server.
2. Run `/effect give <target> minecraft:speed infinite 7 false`.
3. Confirm that the Speed effect is active.
4. Leave the server
5. Rejoin the same server
6. Observe that Speed is displayed in the effect HUD, but the players movement speed appears normal.
7. Run `/effect clear <target>`.
8. Observe that the HUD entry disappears, But the speed behavior is then activated
9. Run `/effect clear <target>` again.
10. Pumpkin responds with "Target has no effects to remove", Even though the effect is enabled.

### Expected behavior

When i Ran the command, Left the Server, then rejoined. I Expected the effect To be Active When i rejoined the server because before hand i Enabled the effect. That wasnt the case.

RicePollution:
Fixes #2520.

## What changed

Status effects already survive a rejoin — `LivingEntity::write_nbt` saves them, `read_nbt_non_mut` reads them back, and `Player::send_active_effects` re-sends `CUpdateMobEffect` to the client on join. So the effect icon reappears in the HUD and looks fine.

What was missing is everything an effect implies **outside** the `active_effects` map. `add_effect` applies that on the live path; the load path only refilled the map. The two silently disagreed:

- **Attribute modifiers** — a restored Speed II showed on the client while the entity kept moving at its base speed, because no modifier was ever put back into the `AttributeInstance`. This is the desync in the issue.
- **The invisibility flag** — a restored Invisibility left the entity plainly visible.
- **The glowing flag** — likewise.

This PR pulls that work out of `add_effect` into a shared `LivingEntity::apply_effect_state`, and calls it from both paths, so an effect read back from disk lands in the same server-side state as one that was just applied — and the two cannot drift apart again.

## Where it is called

- `read_nbt_non_mut` re-applies it for every restored effect, via `reapply_active_effect_state`. This covers mobs loaded from chunk NBT as well as players.
- There is no connected client at that point, so `Player::send_active_effects` — which already runs during the join sequence right after the effects are re-sent — now also sends the attribute updates the restored effects imply. Without that the client keeps predicting movement from its base speed and a restored Speed or Slowness still looks ignored, even though the server side is now correct.

## Absorption is deliberately excluded

`add_effect` grants +4 absorption per level, but the absorption *amount* is already persisted separately as `AbsorptionAmount` (and clamped to `MAX_ABSORPTION` on read). Re-granting it on load would stack another +4 per level onto the saved value on every single load, so `apply_effect_state` does not touch it. This is called out in the doc comment so it does not get "fixed" later by accident.

## Why calling it twice is safe

The restore path and the join path both re-apply, and that is fine:

- `AttributeInstance::add_or_replace_modifier` removes any existing modifier with the same id before pushing, so applying the same effect's modifier repeatedly leaves exactly one.
- `set_invisible(true)` / `set_glowing(true)` are idempotent flag writes.

The fix leans on that first property, so the tests pin it explicitly.

## Tests

6 unit tests on `AttributeInstance`, pure-function so no server is needed:

- re-applying the same modifier does not stack (the property this fix depends on — if it appended instead, a reloaded Speed II would come back at double strength and grow on every load)
- modifiers with distinct ids still stack, so the idempotence is keyed on the id rather than collapsing everything
- replacing a modifier uses the new amount, so a restored effect at a different amplifier overwrites rather than merges
- the three operations compose in vanilla order: `(base + add) * (1 + multiply_base) * prod(1 + multiply_total)`
- removing a modifier restores the base value
- the memoised `cached_value` is invalidated on every change

Full local CI green: `typos`, `cargo fmt --check`, `cargo clippy --all-targets --all-features`, the same in `--release`, `cargo nextest run` (445 passed), `cargo test --doc`, `cargo machete`.

## Known limitations

- Nothing syncs a living entity's attributes to a Java client on join in general, so attributes changed by other means (e.g. `/attribute`) are still not re-sent on rejoin. This PR only covers the attributes implied by active effects, which is what the issue is about; a general join-time attribute sync would be a broader change.
- Invisibility and glowing are set to `true` when restored but nothing recomputes them to `false` from the full effect set, so an entity made invisible by another source and then losing the effect can still end up out of sync. That path is unchanged by this PR — `remove_effect` has the same shape today.
- Instant effects (`INSTANT_HEALTH` / `INSTANT_DAMAGE`) are never stored in `active_effects`, so they are correctly not re-triggered on load.

### DIFF
--- a/pumpkin/src/entity/attributes.rs
+++ b/pumpkin/src/entity/attributes.rs
@@ -260,7 +260,7 @@ impl AttributeRegistry {
 
 #[cfg(test)]
 mod tests {
-    use super::{AttributeInstance, Modifier, ModifierOperation};
+    use super::{AttributeInstance, Attributes, Modifier, ModifierOperation};
 
     fn modifier(id: &str, amount: f64, operation: ModifierOperation) -> Modifier {
         Modifier {
@@ -346,5 +346,45 @@ mod tests {
 
         inst.remove_modifier("x");
         assert!((inst.value() - 1.0).abs() < f64::EPSILON);
+    }
+
+    #[test]
+    fn max_absorption_is_zero_until_the_absorption_effect_is_applied() {
+        use pumpkin_data::effect::StatusEffect;
+
+        // The order the NBT restore path has to preserve. `AbsorptionAmount` is
+        // clamped against MAX_ABSORPTION, every entity type has that attribute at
+        // base 0, and the only thing that lifts it is the Absorption effect's own
+        // modifier. So clamping before the effect's state is re-applied computes a
+        // ceiling of 0 and throws the saved amount away — log out with Absorption,
+        // come back with none.
+        let saved_amount = 8.0f64;
+        let mut max_absorption = AttributeInstance::new(0.0);
+
+        assert!(max_absorption.value().abs() < f64::EPSILON);
+        assert!(
+            saved_amount.max(0.0).min(max_absorption.value()).abs() < f64::EPSILON,
+            "clamping before the effect returns must be what zeroes it"
+        );
+
+        // Absorption II, as `apply_effect_state` builds it.
+        let amplifier = 1.0;
+        let m = &StatusEffect::ABSORPTION.attribute_modifiers[0];
+        assert_eq!(
+            m.attribute.id,
+            Attributes::MAX_ABSORPTION.id,
+            "absorption is expected to modify MAX_ABSORPTION"
+        );
+        max_absorption.add_or_replace_modifier(modifier(
+            m.id,
+            m.base_value * (amplifier + 1.0),
+            ModifierOperation::Add,
+        ));
+
+        assert!((max_absorption.value() - 8.0).abs() < 1e-9);
+        assert!(
+            (saved_amount.max(0.0).min(max_absorption.value()) - 8.0).abs() < 1e-9,
+            "clamping after the effect returns must preserve the saved amount"
+        );
     }
 }
--- a/pumpkin/src/entity/attributes.rs
+++ b/pumpkin/src/entity/attributes.rs
@@ -257,3 +257,94 @@ impl AttributeRegistry {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{AttributeInstance, Modifier, ModifierOperation};
+
+    fn modifier(id: &str, amount: f64, operation: ModifierOperation) -> Modifier {
+        Modifier {
+            id: id.to_string(),
+            amount,
+            operation,
+        }
+    }
+
+    #[test]
+    fn reapplying_the_same_modifier_does_not_stack() {
+        // Restoring a status effect re-applies its attribute modifiers, and that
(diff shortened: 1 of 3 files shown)


Sawlyer:
The absorption note does not hold on the NBT path. `read_nbt_non_mut` clamps the saved value at the top of the same function:

```rust
let raw_abs = nbt.get_float("AbsorptionAmount").unwrap_or(0.0);
let max_abs = self.get_attribute_value(&Attributes::MAX_ABSORPTION) as f32;
let clamped_abs = raw_abs.max(0.0).min(max_abs);
```

`StatusEffect::ABSORPTION` carries the `minecraft:effect.absorption` modifier on `MAX_ABSORPTION` (+4, AddValue), and every entity type has that attribute at base 0. At that point the effects map is still empty, and `reapply_active_effect_state` only runs at the end of the function, so `max_abs` is 0 and the saved amount is clamped to 0. Log out with Absorption, come back with none.

The tag this defers to is already zeroed by the time the effect comes back.

RicePollution:
You're right, and the doc comment you're quoting was asserting something false. Fixed.

I confirmed both halves of your reasoning: `assets/effect.json` gives absorption a single `max_absorption` modifier at `+4.0 ADD_VALUE`, and every entity type in the generated data carries `(Attributes::MAX_ABSORPTION, 0f64)` as its base. So at the point the clamp ran, `max_abs` was 0 and any saved amount went to zero — and the `AbsorptionAmount` tag that `apply_effect_state` deliberately defers to had already been emptied by the time the effect came back.

The clamp now runs *after* `reapply_active_effect_state()`, which is what installs the modifier that makes the saved amount survivable. I also rewrote the `apply_effect_state` doc comment: it still doesn't grant absorption (re-granting would stack +4 per level on every load), but it now says what it *does* restore and that the restore path depends on the ordering, rather than implying the tag alone is sufficient.

Worth flagging that the clamp itself is pre-existing on master, not something this PR introduced — but the PR is about restoring the state an effect implies on load, and absorption is exactly that, so it belongs here.

Test added in `attributes.rs` rather than against a live `LivingEntity`: it drives the real `StatusEffect::ABSORPTION` modifier data against a base-0 `MAX_ABSORPTION` instance and asserts that clamping before the effect returns zeroes the saved amount while clamping after preserves it. It pins the reason rather than the statement order, which is the part that would silently rot.

PR #1693, opened 2026-02-22
Add Crimson and Warped Fungus blocks

GreenedDev:
## Description
added placement rules and neighbor updates

PR #2604, opened 2026-07-27
fix: twisting vines checked the wrong support face

eshankiyer:
## Twisting vines check the wrong support face

`twisting_vines.rs` tested `is_side_solid(BlockDirection::Down)` on the block below. Twisting vines grow upward, so the sturdy face required of the block below is its **top** face. This looks like a copy from `weeping_vines.rs`, where `Down` is correct because weeping vines grow downward and hang from the block above. `kelp.rs`, the other upward grower, already uses `Up`.

Vanilla `GrowingPlantBlock.canSurvive` tests `isFaceSturdy` against `growthDirection`:

```java
BlockPos blockpos = pos.relative(this.growthDirection.getOpposite());
BlockState blockstate = level.getBlockState(blockpos);
return !this.canAttachTo(blockstate) ? false
    : blockstate.is(getHeadBlock()) || blockstate.is(getBodyBlock())
   || blockstate.isFaceSturdy(level, blockpos, this.growthDirection);
```

`TwistingVinesBlock` passes `Direction.UP` to super, `WeepingVinesBlock` passes `Direction.DOWN`, and neither overrides `canAttachTo`.

The difference is only visible on partial blocks, since on nylium both faces are sturdy. A top slab or upside-down stair has a sturdy top face but not a sturdy bottom face: vanilla allows twisting vines there and this refused them. A bottom slab is the mirror case, where vanilla refuses and this allowed them.

## Fungus and roots dispatch on the wrong block

`fungus.rs` and `roots.rs` read the **ground** block and then compared it against `Block::WARPED_FUNGUS` / `Block::WARPED_ROOTS` to choose which support tag to check. Those are the plants, not the ground, so the warped branch was unreachable and everything fell through to the crimson tag.

This is currently a no-op, and I want to be upfront about that: I checked the generated tag data and in 26.2 `supports_crimson_fungus`, `supports_warped_fungus`, `supports_crimson_roots` and `supports_warped_roots` are byte-identical lists (the dirt family, mycelium, farmland, both nyliums, soul soil). So the wrong branch produced the right answer by accident. I fixed it because the comparison is unambiguously wrong and those tags are Mojang-owned data that can diverge in any version. Say the word if you would rather not carry a no-op change.

It was introduced in #1693, which added these blocks.

## Deliberately not touched

- **Bone meal on fungus is entirely unimplemented.** There is no `BonemealableBlock` equivalent in the codebase; `BONE_MEAL` appears only in composter, bamboo, bamboo sapling, sweet berry bush and sea pickles. Growing huge fungi from bone meal means building that infrastructure plus a server-side path into the worldgen feature, which is a port rather than a bug fix.
- **Huge fungus worldgen** is a stub (random 50/50 warped/crimson, no biome check, fixed cap geometry), but #2518 is already fixing exactly that against #2466, so I stayed out of the file.
- **Nylium has no block behaviour registered at all**, so no spread and no bone meal vegetation. Absent rather than wrong.
- **No random tick growth for twisting or weeping vines.** Also absent rather than wrong.
- The extra `&& support_block.is_solid()` gate in twisting, weeping and kelp has no counterpart in vanilla's `canSurvive`, but it is applied consistently across all three and I could not construct a concrete divergent case, so changing it would be speculative.

## Verification

Not verified in game. The slab case above is easy to check by hand if you want it confirmed.

### DIFF
--- a/pumpkin/src/block/blocks/plant/fungus.rs
+++ b/pumpkin/src/block/blocks/plant/fungus.rs
@@ -1,5 +1,6 @@
-use crate::block::{BlockBehaviour, BlockFuture, BlockMetadata, CanPlaceAtArgs};
-use crate::block::{GetStateForNeighborUpdateArgs, blocks::plant::PlantBlockBase};
+use crate::block::{
+    BlockBehaviour, BlockFuture, BlockMetadata, CanPlaceAtArgs, GetStateForNeighborUpdateArgs,
+};
 use pumpkin_data::BlockStateId;
 use pumpkin_data::tag::Taggable;
 use pumpkin_data::{Block, BlockId, tag};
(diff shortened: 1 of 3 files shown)


zhangxuan2011:
> * **Huge fungus worldgen** is a stub (random 50/50 warped/crimson, no biome check, fixed cap geometry), but [fix(world/generate/huge_fungus): 🐛 Fixed something wrong about th… #2518](https://github.com/Pumpkin-MC/Pumpkin/pull/2518) is already fixing exactly that against [Something wrong in the nether #2466](https://github.com/Pumpkin-MC/Pumpkin/issues/2466), so I stayed out of the file.

The fixing process of this issue has done, and I have pushed to my forked repository. thx for leaving a chance to let me contribute :)

PR #2607, opened 2026-07-27
Fix spectator players being visible to normal players

sirfusebox:
Spectators were fully visible to everyone else, both on join and when switching to spectator mid-game with /gamemode. Now they're only visible to other spectators, matching vanilla.

Fixes the issue where logging in with two accounts (one spectator, one not) let the normal account see the spectator like any other player.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -2945,17 +2945,29 @@ impl Player {
                     matches!(gamemode, GameMode::Creative | GameMode::Spectator),
                     Ordering::Relaxed,
                 );
-                self.living_entity
-                    .entity
-                    .world
-                    .load()
-                    .broadcast_packet_all(&CPlayerInfoUpdate::new(
-                        PlayerInfoFlags::UPDATE_GAME_MODE.bits(),
-                        &[pumpkin_protocol::java::client::play::Player {
-                            uuid: self.gameprofile.id,
-                            actions: &[PlayerAction::UpdateGameMode((gamemode as i32).into())],
-                        }],
-                    ));
+
+                let world = self.living_entity.entity.world.load();
+
+                // Spectators must be fully invisible to non-spectators. Since this player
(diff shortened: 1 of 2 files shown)


RoosterBooster007:
Hey @sirfusebox, thanks for your work here! There are a few things to sort out before we can land this:

* **Needs a rebase:** It's using pre-crate paths, and player spawning has shifted quite a bit since July (`broadcast_skin_parts` landed, and `pair_new_player_with_tracked_entities` now runs right after `CLogin`).
* **Duplicate visibility checks:** The entity tracker already handles this rule via `TrackedEntity::broadcast_to_player` (`viewer is spectator || target is not`, matching vanilla's `ServerPlayer.broadcastToPlayer`). Since `add_pairing` already handles spawn, metadata, and equipment packets, we should route through the tracker rather than maintaining a parallel spawn path with `except` lists in `world/mod.rs`.
* **Tracker state desync:** The manual remove/respawn bypasses the tracker's `seen_by` set. Because the tracker still thinks non-spectators have the entity, it'll leak metadata/equipment deltas to them. Then, when leaving spectator, `seen_by.insert` returns `false`, preventing `add_pairing` from firing (which is why `show_former_spectator_to_non_spectators` was needed). Keeping `seen_by` in sync or triggering a tracker re-check would let us drop both manual helpers entirely.
* **Gamemode switch timing:** Vanilla doesn't immediately resync tracking during `/gamemode`. `ServerPlayerGameMode` just updates abilities and broadcasts player info, leaving the visibility update to `ChunkMap.tick` (which gates on section changes). Master matches that gate in `update_all`. Instantly sending `CRemoveEntities` is a bit more aggressive than vanilla, but if we want immediate flipping, triggering a single tracker re-eval on gamemode change would be the cleanest way to do it.

Btw, there's an active tracker rework in #3397 / #3401 (both green) that moves these manual spawn broadcasts into the tracker, which might resolve the join-time half on its own.

PR #2608, opened 2026-07-27
Build- and run-time optimizations in debug mode

PilkeySEK:
## Description
Building pumpkin, even in debug mode, can take a long time and produce a very resource-intensive binary (my experience was that it used close to 100% CPU while being much slower than the release mode binary).

**The fix for a resource-intensive debug binary:** All dependencies (which does not include workspace members[^1]) are compiled at `opt-level=3` (the maximum, esentially release mode) while workspace members are compiled at `opt-level=0`. On my machine this made the server use almost no CPU.

**Attempts to make the build times faster:** I was not really able to improve the build times of debug mode through the `Cargo.toml` (`debug=1` might improve it too, while hopefully not removing too much debug information). Instead, I made rustc use the `lld` linker instead of the default (which seems to be cc[^2]). This had the following impact:
- default (cc): 1m 11s
- lld: 40s

Additionally, `cargo +nightly check -Zcargo-lints --workspace --all-targets`[^3] showed the following unused dependencies:
- `cipher`
- `crossbeam-utils`
- `lru`
- `signature`
- `xxhash-rust`
As the cargo docs mention too, it could show false positives in specific cases which is why I decided to not remove these dependencies, but it might be something to consider.

Here's a flamegraph for building `pumpkin` only (after changing one file). This uses the default linker (the linking step takes the most time):
image
I don't have a flamegraph for the lld linker, but might be able to generate one later.

## Testing
I did not observe an impact on the functionality of Pumpkin.

I'm not sure whether lld is pre-installed on basically every system. If not, might want to switch back to the default linker.

[^1]: https://doc.rust-lang.org/cargo/reference/profiles.html#overrides
[^2]: https://stackoverflow.com/questions/57812916/how-do-i-change-the-default-rustc-cargo-linker
[^3]: https://doc.rust-lang.org/cargo/guide/build-performance.html#removing-unused-dependencies

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.x86_64-unknown-linux-gnu]
+# Use LLVM's lld instead of the default for (likely) faster linking
+rustflags = [
+    "-C", "link-arg=-fuse-ld=lld",
+]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,0 @@
-[target.x86_64-unknown-linux-gnu]
-# Use LLVM's lld instead of the default for (likely) faster linking
-rustflags = [
-    "-C", "link-arg=-fuse-ld=lld",
-]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,11 @@ debug = 1
 [profile.dev.package."*"]
 opt-level = 3
 
+[profile.dev.package.wasmtime]
+opt-level = 1
+[profile.dev.package.cranelift-codegen]
+opt-level = 1
+
 [profile.release]
 lto = true
 strip = "debuginfo"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,15 @@ edition = "2024"
 rust-version = "1.95"
 license = "GPL-3.0"
 
+[profile.dev]
+lto = false
+opt-level = 0
+incremental = true
+debug = 1
+
+[profile.dev.package."*"]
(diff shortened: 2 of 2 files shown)


PigTurtleDev:
`lld` has been the default linker for `x86_64-unknown-linux-gnu` for about a year now (Rust 1.90). Pumpkin already specifies 1.95 as it's rust version, are you sure your system was defaulting to `cc`?
I'm against including a `.cargo/config.toml` if it isn't necessary, since doing so creates a hassle for anyone using a linker like mold or wild locally (or even just has project-specific cargo aliases).

I haven't tested and therefore can't comment on the effectiveness of the opt-level overrides yet, but if they work I don't see much reason not to use them. They should only need to be compiled once unless someone manually clears their build cache. I would nevertheless look into whether excluding some of our slower-to-compile dependencies (wasmtime, cranelift-codegen) from the opt-level=3 might be justifiable for "cold" compilations, that way new people won't quite have to wait eons the first time they compile the project using what should be the fast build.

PilkeySEK:
> using a linker like mold or wild locally (or even just has project-specific cargo aliases).

That's something I totally didn't think of...

So, I removed the .cargo/config.toml and made wasmtime and cranelift-codegen compile at opt-level=1 instead of 3

Snowiiii:
Most of the stuff in the debug profile is already the default, also it does not make sense to let packages optimize in debug, i understand performance might be lower, but that can be potentially fixed by code changes instead forcing even higher compile times

Issue #2592, opened 2026-07-27
execute if block ignores its predicate; blockstate args, forceload and CLightUpdate also broken


eshankiyer:
While building an RCON-driven test harness against a live server I hit four command and protocol bugs. Filing them together since they were all found the same way and none seemed worth a separate issue on its own, but happy to split if you would prefer.

Server was built from master plus some of my own open PRs, none of which touch command dispatch. Tested over RCON with a player connected so chunks stayed loaded.

## 1. `execute if block` ignores its predicate

```
setblock 8 101 8 minecraft:stone
execute if block 8 101 8 minecraft:stone run seed   ->  Seed: [1784236824834363978]
execute if block 8 101 8 minecraft:air   run seed   ->  Seed: [1784236824834363978]
execute unless block 8 101 8 minecraft:air run seed ->  Seed: [1784236824834363978]
```

The block at that position is stone, so the second one should not run. `if` and `unless` behave identically, and the condition appears not to gate the command at all.

This is the one with the widest blast radius of the four: any datapack or command block relying on `execute if block` will run its command unconditionally.

## 2. Block states are not accepted in `setblock` or as a `clone` filter

```
setblock 21 100 20 minecraft:wheat[age=0]
  ->  Unknown block type 'minecraft:wheat[age=0]'

clone 20 100 20 21 100 20 40 122 40 filtered minecraft:wheat[age=0]
  ->  Expected whitespace to end one argument, but found trailing data
```

Plain `minecraft:wheat` works in both. Vanilla accepts the bracketed state form everywhere a block argument or block predicate is taken, so anything that needs to place or match a specific state cannot be expressed.

## 3. `forceload` reports success but does not load the chunk

```
forceload add -32 -32 32 32   ->  Marked 25 chunks in -2 from -2 to 2 to be force loaded
forceload query               ->  Chunk at 0 in 0 is marked for force loading
setblock 0 64 0 minecraft:stone
  ->  That position is not loaded
```

The mark is recorded and queried back correctly, but nothing loads, so the command has no effect. With no player nearby the chunk stays unloaded indefinitely. I worked around it by keeping a client connected purely as a chunk loader.

## 4. `CLightUpdate` is never sent, so runtime light changes do not reach clients

`pumpkin-protocol/src/java/client/play/light_update.rs` defines the packet, but grepping for its send sites returns none. Light is only ever carried by `CChunkData` at chunk send time.

[...]

PR #2609, opened 2026-07-27
fix: execute subcommands never ran their modifiers

eshankiyer:
Fixes the first item in #2592.

## What

`execute if block` runs its command whether or not the block matches, and `unless` behaves identically. Reproduced on a live server over RCON:

```
setblock 8 101 8 minecraft:stone
execute if block 8 101 8 minecraft:stone run seed   ->  Seed: [...]
execute if block 8 101 8 minecraft:air   run seed   ->  Seed: [...]   <- should not run
execute unless block 8 101 8 minecraft:air run seed ->  Seed: [...]
```

## Why

It turned out not to be in the condition functions at all. `ContextChain::try_flatten` walks the context list and pushes each **child** onto the modifier list, where Brigadier pushes the **current** context:

```rust
loop {
    if let Some(child) = current.get_child() {
        modifiers.push(child.clone());   // pushes the child
        current = child;
    } else {
        ...
    }
}
```

A context carries the redirect modifier of the node that redirected *away* from it. For `execute if block <pos> <block> run seed` the modifier lives on the **root** context, whose last parsed node is the `block` argument holding `Redirection` plus `execute_if_block_modifier`. The child context is `run <command>`, whose modifier is the default `KeepSource`.

So the chain was built as `modifiers = [child]` with `execute = child`, the same context twice. `run_modifier` short-circuits on `KeepSource` before ever calling `sources()`, so `execute_if_block_modifier` and `execute_unless_block_modifier` were **never invoked**. Not computed and discarded, simply never reached.

The fix is to push `current` rather than the child.

## Scope is wider than the reported symptom

The same defect silently broke every other subcommand in the tree, since they all route through this path. Fixed by the same change:

- `if` / `unless` for `entity`, `block`, `loaded`, `dimension`
- `as`, `at`, `in`, `positioned`, `rotated`, `align`, `anchored`, `facing`

So `execute as @a run ...` was running once as the original source rather than once per target. The condition functions in `execute.rs` were already correct and are untouched.

## Test

Added `conditional_redirect_gates_execution` to the existing test module in `dispatcher.rs`, following the surrounding `#[tokio::test]` / `execute_input` convention. It registers a command whose two literals redirect to root with a satisfied and an unsatisfied modifier, and asserts the command runs in one case and not the other.

I checked it actually catches the bug rather than just passing: with `try_flatten` reverted to the old form it fails with `left: Ok(1), right: Ok(0)`, which is exactly the reported symptom.

## The other half of #2592

Block state syntax like `minecraft:wheat[age=0]` is a separate root cause and I did not fold it in. There are two independent block argument parsers, neither of which handles brackets: the newer one stops consuming at `[` because it is not in the allowed character class, and the legacy one hands the whole token to `Block::from_name`. Both have `Item = &'static Block`, and consumers use `block.default_state.id`, so real support means changing the argument's item type to a resolved block state and threading it through `setblock`, `fill`, `clone` and `execute if block`. That is a materially different change.

## Verification

`cargo fmt --check` and `RUSTFLAGS="-D warnings" cargo clippy -p pumpkin --all-targets` clean. `cargo test -p pumpkin --lib command::` passes, 90 tests.

Not re-verified on the live server, since it is running an older build.

### DIFF
--- a/pumpkin/src/command/context/command_context.rs
+++ b/pumpkin/src/command/context/command_context.rs
@@ -216,17 +216,26 @@ impl<'a> ContextChain<'a> {
     #[must_use]
     pub fn try_flatten(root: &CommandContext<'a>) -> Option<Self> {
         let mut modifiers = Vec::new();
-        let mut current = root;
+        let mut current = Arc::new(root.clone());
 
         loop {
-            if let Some(child) = current.get_child() {
-                modifiers.push(child.clone());
-                current = child;
-            } else {
-                return current
-                    .command
-                    .is_some()
-                    .then(|| Self::new(modifiers, Arc::new(current.clone())));
+            // A context holds the modifier of the node that redirected away from it,
+            // so it is the *parent* of each link that carries the modifier to run,
+            // never the child. Pushing the child here would skip the first modifier
+            // (dropping conditions such as `execute if block`) and would instead run
+            // the trailing executable context's own modifier.
+            let child = current.get_child().cloned();
+            match child {
+                Some(child) => {
+                    modifiers.push(current);
+                    current = child;
+                }
+                None => {
+                    return current
+                        .command
+                        .is_some()
+                        .then(|| Self::new(modifiers, current));
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Vaspyyy:
I prepared a focused tested follow-up in eshankiyer/Pumpkin#1.

It reports an empty modifier source set as command failure, so false conditions notify result consumers and `execute store success score ...` stores `0` instead of leaving the score unset. The patch changes only `command_context.rs` and includes a regression test.

Validation: formatting, clippy with warnings denied, all 151 Pumpkin library tests, and a live combined server smoke test.

aure31 (approved):
LGTM

PR #2616, opened 2026-07-27
fix: honour the simulation_distance config instead of a hardcoded radius

eshankiyer:
`World::update_active_chunks` built a hardcoded 17x17 box per player:

```rust
// TODO: gamerule for view distance/ticking distance
for dx in -8..=8 { for dy in -8..=8 { active_chunks.insert(center.add_raw(dx, dy)); } }
```

Meanwhile `simulation_distance` is defined in config with a default of 10 and read by nothing outside the Bedrock start game packet and the WASM plugin API. So mobs froze about two chunks closer than vanilla, and changing the setting did nothing.

This is not a new discovery. Merged PR #2341 already documents it verbatim in its notes: "active_chunks is currently a hardcoded 17x17 (radius 8) per player, slightly tighter than the default simulation distance (10), a pre-existing limitation shared with block-entity ticking and spawning."

The same `active_chunks` set gates entity AI ticking, block entity ticking, random ticks and mob spawning, so wiring it up fixes all four together. I grepped the four consumers and none of them carries its own hardcoded radius; they all just read the set. The one nearby literal, `MAGIC_NUMBER = 17 * 17` in `natural_spawner.rs`, is vanilla's own mob density normalisation divisor and is fixed in vanilla too, so it is left alone.

Radius is resolved per player inside the existing loop, so mixed Java and Bedrock sessions each contribute their own platform's configured value.

## On clamping rather than asserting

An earlier draft of this added startup asserts. I backed that out. `simulation_distance` was previously unread, so an existing server can have any value in its config and never have noticed, and an assert would turn that into a panic on upgrade over a value that used to be harmless. That also cuts directly against #2612, which removes two startup panics on recoverable conditions.

So the value is clamped into vanilla's documented range of 3 to 32 with a warning naming both the configured and the effective value. Clamping happens once in `validate`, which is the single choke point all four readers pass through, so the per-tick path gains no work.

On what vanilla does with an out of range value: the wiki documents the property as an integer 3 to 32 defaulting to 10, and states that invalid values are reset to defaults rather than rejected (https://minecraft.wiki/w/Server.properties, https://minecraft.wiki/w/Simulation_distance). Paper's chunk map patch shows the applied value passing through `Mth.clamp(watchDistance, 2, 32)`. I want to be honest that this second one is a patched fork rather than raw vanilla, which I could not obtain, so treat the clamp-versus-default question as supported but not proven. Either way we clamp rather than panic.

## Verification

`cargo fmt --check` and `RUSTFLAGS="-D warnings" cargo clippy -p pumpkin -p pumpkin-config --all-targets` clean.

No regression test: exercising this needs a full Server, World and Player fixture with a live ClientPlatform, and neither `chunker.rs` nor the config crate has existing test scaffolding to extend.

Not verified in game.

### DIFF
--- a/pumpkin-config/src/lib.rs
+++ b/pumpkin-config/src/lib.rs
@@ -62,12 +62,34 @@ impl LoadConfiguration for PumpkinConfig {
         Path::new("pumpkin.toml")
     }
 
-    fn validate(&self) {
+    fn validate(&mut self) {
         self.basic.validate();
         self.advanced.validate();
 
         let min_vd = NonZeroU8::new(2).unwrap();
         let max_vd = NonZeroU8::new(64).unwrap();
+
+        // Vanilla's simulation-distance server property accepts 3 to 32,
+        // default 10, for multiplayer:
+        //   https://minecraft.wiki/w/Server.properties (see the
+        //   `simulation-distance` row, "integer (3-32)")
+        //   https://minecraft.wiki/w/Simulation_distance#Range_of_simulation_distance
+        // Vanilla does not crash the server over an out-of-range value here.
+        // The server.properties page documents that keys with invalid
+        // values are reset to their defaults when the file is regenerated
+        // on startup, and the distance that actually gets applied to the
+        // chunk/tick system is clamped rather than rejected -- see the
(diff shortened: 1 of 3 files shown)


Rocraften:
Good find.

PR #2621, opened 2026-07-27
fix(block): correct berry bush and nether wart growth rates

### LINKED ISSUE #2582: Sweet berry bushes and nether wart grow slower than vanilla (extra random gate)
### What happened

Sweet berry bushes and nether wart advance a growth stage far less often than in vanilla, because a second random gate is applied on top of the one vanilla uses.

**Sweet berry bush.** `SweetBerryBushBlock::random_tick` (`BlockBehaviour`) already applies vanilla's gate:

```rust
if rand::rng().random_range(0..5) == 0 {
    <Self as CropBlockBase>::random_tick(self, args.world, args.position).await;
}
```

but the `CropBlockBase::random_tick` override it calls (`pumpkin/src/block/blocks/plant/crop/sweet_berry_bush.rs`) then applies another:

```rust
if rand::rng().random_range(0..=25) == 0 {
```

That is 1/5 * 1/26, so about 1/130 per random tick, where vanilla is 1/5:

```java
if (age < 3 && random.nextInt(5) == 0 && level.getRawBrightness(pos.above(), 0) >= 9) {
    level.setBlock(pos, state.setValue(AGE, age + 1), 2);
}
```

A bush therefore takes roughly 26x as many random ticks to ripen as it should.

**Nether wart.** `pumpkin/src/block/blocks/plant/nether_wart.rs` uses `random_range(0..=10)`, which is 1/11. Vanilla is `random.nextInt(10) == 0`, which is 1/10. Smaller, but it is the same inclusive-range off-by-one.

Worth noting the crop base case is correct: `random_range(0..=(25.0 / f).floor() as i64)` does match vanilla's `nextInt((int)(25.0F / f) + 1)`. So this is not a systematic convention problem, just these two sites.

### Also in the berry bush override

```rust
let state_above = world.get_block_state(&pos.up());
if state_above.is_full_cube() || state_above.is_solid() {
    return;
}
```

Vanilla's `SweetBerryBushBlock.randomTick` has no such check. It gates on `getRawBrightness(pos.above(), 0) >= 9`, which a solid block above will usually fail anyway, so the two overlap in the common case but are not equivalent: a bush under a transparent-but-solid block, or one lit from the side, behaves differently from vanilla.

### Version

master (`afbe839`). Both are pre-existing and unrelated to any open PR of mine; I ran into them while verifying the dispatch in #2567.

I am happy to open a PR for the probabilities if the approach looks right. My inclination is to drop the inner `0..=25` gate entirely rather than retune it, since the outer gate is already the vanilla one, and to change nether wart to `0..10`. The `is_full_cube` check I would rather leave for whoever added it to comment on, since removing it changes behaviour in a way the light check does not fully cover.

colinptr:
## Description

Fixes #2582.

Sweet berry bushes applied a second random growth gate after the
vanilla-equivalent 1-in-5 check, reducing their effective growth chance
to approximately 1 in 130. This removes the duplicate inner gate while
preserving the block-state update.

Nether wart used the inclusive range `0..=10`, producing a 1-in-11
growth chance. This changes it to `0..10`, matching vanilla's 1-in-10
chance.

## Testing

Passed locally on Windows 10 with Rust 1.97.1:

- `cargo fmt --check` — passed
- `cargo clippy --all-targets --all-features` — passed
- `cargo test` — passed

### DIFF
--- a/pumpkin/src/block/blocks/plant/crop/nether_wart.rs
+++ b/pumpkin/src/block/blocks/plant/crop/nether_wart.rs
@@ -82,7 +82,7 @@ impl CropBlockBase for NetherWartBlock {
     async fn random_tick(&self, world: &Arc<World>, pos: &BlockPos) {
         let (block, state) = world.get_block_and_state_id(pos);
         let age = self.get_age(state, block);
-        if age < self.max_age() && rand::rng().random_range(0..=10) == 0 {
+        if age < self.max_age() && rand::rng().random_range(0..10) == 0 {
             world
                 .set_block_state(
                     pos,
--- a/pumpkin/src/block/blocks/plant/crop/sweet_berry_bush.rs
+++ b/pumpkin/src/block/blocks/plant/crop/sweet_berry_bush.rs
@@ -188,15 +188,13 @@ impl CropBlockBase for SweetBerryBushBlock {
             if state_above.is_full_cube() || state_above.is_solid() {
                 return;
             }
-            if rand::rng().random_range(0..=25) == 0 {
-                world
-                    .set_block_state(
-                        pos,
-                        self.state_with_age(block, state, age + 1),
-                        BlockFlags::NOTIFY_NEIGHBORS,
-                    )
-                    .await;
-            }
+            world
+                .set_block_state(
+                    pos,
+                    self.state_with_age(block, state, age + 1),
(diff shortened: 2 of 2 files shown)


RoosterBooster007:
LGTM, @colinptr!

PR #2624, opened 2026-07-28
fix(item): prevent stacking items with different components

### LINKED ISSUE #2618: items with different components wrongfully stack with each other

### Operating System

Arch Linux

0.1.0-dev+26.2/33ab147

### What happened?

when picking up an item with custom data components while having another ItemStack of the same item but with no custom data components, they will stack

### To Reproduce

/give @p coal
/give @p coal[minecraft:item_model="minecraft:diamond"]

must be ran in this order, if ran the other way around the items will correctly not stack

### Expected behavior

The items don't stack together

colinptr:
## Description

Fixes #2618.

`ItemStack::are_items_and_components_equal` only checked component patches
present on the left-hand stack. When the existing stack had no component
patches, an incoming stack with additional components could incorrectly be
considered equal.

This requires both stacks to contain the same number of component patches
before comparing their component IDs and values, making the comparison
symmetric.

## Testing

- `cargo fmt --check` — passed
- `cargo test -p pumpkin-data` — 36 passed
- `cargo build -p pumpkin` — passed
- Manually tested plain then customized items — remained separate
- Manually tested customized then plain items — remained separate
- Manually tested two identically customized items — stacked correctly

### DIFF
--- a/pumpkin-data/src/item_stack/mod.rs
+++ b/pumpkin-data/src/item_stack/mod.rs
@@ -579,6 +579,11 @@ impl ItemStack {
         if self.item != other.item {
             return false;
         }
+
+        if self.patch.len() != other.patch.len() {
+            return false;
+        }
+
         for (id, data) in &self.patch {
             let mut not_found = true;
             'out: for (other_id, other_data) in &other.patch {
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Thanks, @colinptr!

Issue #2605, opened 2026-07-27
cant join on 1.21.4


catplemail-web:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

windows

### Server Software Version/Commit

8db9c01/release

### What happened?

error while joining on 1.21.4 (java) (server is 26.2)

disconnect error:
---- Minecraft Network Protocol Error Report ----
// Please insert The Internet CD #4
// Server bug reporting link: https://github.com/Pumpkin-MC/Pumpkin/issues

Time: 2026-07-27 16:58:34
Description: Packet handling error

io.netty.handler.codec.DecoderException: Failed to decode packet 'clientbound/minecraft:commands'
	at knot//net.minecraft.class_9136.method_56425(class_9136.java:38)
	at knot//net.minecraft.class_9136.decode(class_9136.java:14)
	at knot//net.minecraft.class_2543.decode(class_2543.java:31)
	at knot//io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:530)
	at knot//io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:469)
	at knot//io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:290)
	at knot//io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at knot//io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at knot//io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at knot//io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:107)
	at knot//io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at knot//io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at knot//io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at knot//io.netty.handler.flow.FlowControlHandler.dequeue(FlowControlHandler.java:202)
[...]

PR #2525, opened 2026-07-25
fix(skin): Missing outer skin layer (hat, jacket, sleeves, pants) upon joining the server

Q2297045667:
### Summary

**Fix: Second skin layer (hats, jackets, etc.) not rendering upon joining**
This issue affected 1.21.7+ and 26.1+ clients, occurring on both direct connections and via Velocity proxies.

**Root Cause:**

Different Minecraft versions handle skin data transmission differently. Our previous code only checked for `PLAYER_MODE_CUSTOMISATION`, causing 1.21.7+ clients to drop the packets entirely. For 26.1+ clients, the new protocol was used, but the entity metadata was still missing.
**The Fix:** We now send both the legacy and new Tracked IDs simultaneously. The underlying system automatically filters out the incompatible ones, ensuring all versions receive the correct data.

**Bug Fixes:**

1. **Alex models rendering as Steve:** We previously missed the `metadata` field when parsing the skin JSON, causing all slim models to be treated as wide-armed. This is now fixed.
2. **Default skin parts hidden:** The default config had `skin_parts` set to `0`. It has been reverted to the vanilla default of `0x7F` (all layers visible).

### Code Changes

- **Protocol (`player_action.rs` & `player_info_update.rs`)**: Added `UpdateHat` support. Packet writing is now version-aware; older versions automatically mask out the `UPDATE_HAT` and `UPDATE_LIST_PRIORITY` flags to prevent errors.
- **World Broadcasting (`world/mod.rs`)**: Updated join and metadata broadcasts to include the dual-ID sending logic.
- **Player Entity (`player.rs`)**: Added a `SkinMetadata` struct to correctly parse `"slim"` and set the arm size.
- **Default Config (`net/mod.rs`)**: Changed the default `skin_parts` value to `0x7F`.
- **Data Mapping (`tracked_data`)**: Added the missing `PLAYER_MODE_CUSTOMIZATION_ID: 17` mapping for versions 1.21 through 1.21.6.

### Compatibility

- **26.1+**: Uses `UPDATE_HAT` + new metadata.
- **1.21.7 - 1.21.11**: Uses legacy metadata ID (16/17).
- **1.21.6 and below**: Second skin layer metadata is not currently supported (ID is 255, automatically skipped).

### Testing

- [x] Code checks, Clippy, and formatting passed.
- [x] Direct connection tested with 26.2, 26.1.2, 1.21.11, 1.21.9, 1.21.7, 1.21.6, and 1.21.5 clients; skins render correctly.
[...]

PR #1290, opened 2026-01-13
Fix skin layers, cape & main hand rendering

blaxkwidow:
### Description
This PR addresses an issue where player skins (second layer/capes) were not rendering.

### Technical details
**Skin Parts:** Located at **Index 16**
**Main Hand:** Located at **Index 15**
**Data Type Mismatch:** The Main Hand field now utilizes the `HumanoidArm` enum ID 38. Previous implementations sending `Byte` (0) or `VarInt` (1) caused client-side crashes (`Invalid entity data item type`)

Players can now see their own skins, and capes immediately upon joining.

### testing
I tested these changes locally with a standard Minecraft 1.21.11 java client.

PR #2630, opened 2026-07-28
fix: restore main hand and skin customization metadata across versions

Necko1:
## Description

**What was changed?**
- Uncommented and updated the player entity metadata logic responsible for the main hand (left/right).
- Implemented cross-version support for hand metadata: sends `PLAYER_MAIN_HAND` for `v26.x` and `MAIN_ARM_ID` for `v1.21.x`.
- Added `PLAYER_MODE_CUSTOMIZATION_ID` metadata to ensure the second skin layer (hats, jackets, etc.) renders correctly on versions below `v26.x`.
- Replaced placeholder IDs (`255u8`) in `pumpkin-data/src/generated/tracked_data.rs` for `MAIN_ARM_ID` and `PLAYER_MODE_CUSTOMIZATION_ID` on versions `1.21` - `1.21.6` with the correct actual values.

The main hand metadata was previously commented out, causing players to always appear right-handed. Additionally, placeholder IDs in the tracked data meant that skin customizations were broken on older 1.21 versions. This PR restores these visual features and adapts them to the recent protocol changes. 

Players will now correctly see their configured main hand and full skin layers regardless of whether they join on newer (`v26.x`) or older (`v1.21.x`) clients.

I have successfully verified the functionality (left hand rendering and 2nd skin layer) on versions: `1.21.5`, `1.21.6`, `1.21.8`, `1.21.10`, `26.1.2`, and `26.2`. 
However, I could not test on versions strictly older than `1.21.5` due to an existing unrelated connection issue documented in #2605.

**Context:**
- Partially based on the approach from #2525.
- Checked against redundancy.

## Testing
- [x] Code passes `cargo clippy --all-targets` with no warnings.
- [x] Code passes `cargo test` successfully.
- [x] Tested in-game


Necko1:
> **Why were these changes necessary?**
> The main hand metadata was previously commented out, causing players to always appear right-handed.

fun fact: this exact main hand fix was actually proposed half a year ago in #1290, but it was unfortunately left out when that PR got closed

RoosterBooster007:
Looks good! Thanks, @Necko1!

PR #2631, opened 2026-07-28
perf: Make WasmPlugins be Arc<WasmPlugin> instead of Box<Arc<WasmPlugin>>

PilkeySEK:
## Description
Because the `Plugin` trait requires `&mut self` for both `on_load` and `on_unload`, it previously required that `WasmPlugin` was `Box<Arc<T>>` instead of just `Arc<T>`, even though `WasmPlugin` doesn't need a mutable reference (`Arc` is needed because there is a `Weak` being created). `Plugin` now only requires `&self`, which breaks compatibility with existing native plugins (like PatchBukkit), but is very easy to fix for these plugins (if they previously needed a mutable reference they can use interior mutability types).
(PatchBukkit doesn't require `&mut self` actually)

Overall the performance of WASM plugins may be very very slightly improved (assuming the compiler didn't optimize the double reference).

image

## Testing
I tested that WASM plugins work. I wasn't able to test whether the PatchBukkit plugin could be loaded (with a patch to make the plugin compatible again) because it failed to build, I might be able to test it later though.

### DIFF
--- a/pumpkin/src/plugin/api/mod.rs
+++ b/pumpkin/src/plugin/api/mod.rs
@@ -39,26 +39,16 @@ pub trait Plugin: Send + Sync + 'static {
     /// Asynchronous method called when the plugin is loaded.
     ///
     /// This method initializes the plugin within the server context.
-    ///
-    /// # Parameters
-    /// - `_server`: Reference to the server's context.
-    ///
-    /// # Returns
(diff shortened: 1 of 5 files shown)


Snowiiii:
LGTM, Thanks @PilkeySEK

Issue #2598, opened 2026-07-27
Soul sand and magma blocks do not generate a bubble column


Luminous-Journey:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows

### Server Software Version/Commit

0.1.00dev+26.2 Commit 5f12782/release

### What happened?

Soul sand and magma blocks do not generate bubble columns 

### To Reproduce

place either block under a water source block

### Expected behavior

bubble columns are generated and either push or pull the player

Luminous-Journey:
Tysm @HookWoods for resolving this issue and getting it merged!!!!

PR #2633, opened 2026-07-28
feat(block): implement bubble columns

HookWoods:
Implemented source-water bubble columns for Soul Sand and Magma Blocks. This resolve issue https://github.com/Pumpkin-MC/Pumpkin/issues/2598

   - propagates and removes columns with vanilla delays
   - transports entities and refreshes player air
   - added some test to make sure upward and downward system works correctly

The constant are from the decompiled java source so nothing new on that side.


Snowiiii:
LGTM, Thank you @HookWoods

Issue #2625, opened 2026-07-28
swimming freaks out


Mebecool1:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Fedora Linux 44 x86_64 Workstation

### Server Software Version/Commit

From the website

### What happened?

i tried to swim and it did this thing where it swims for like 100ms then stands up then swims again like that again and again

### To Reproduce

go into ocean try to swim

### Expected behavior

i expected normal consistant swimming

Simlesk:
Can confirm on a recent commit (0.1.0-dev+26.2-26.40, de8c7800) with a Java 26.2 client: the swim state toggles on/off repeatedly while swimming (as if the pose keeps getting reset server-side), which makes the player stutter constantly in water. Reproduces every time by just swimming in any body of water.

PR #2638, opened 2026-07-28
fix: swimming flickers between swimming and standing pose

TheRealMusawer:
the is_swimming check was gating on is_sprinting() which is wrong. in vanilla you just need to be moving in water to swim, you don't need to sprint. since the sprint state flickers back and forth while in water the pose would rapidly toggle between swimming and standing every few ticks.

added proper swimming state updates in tick_movement so the entity swimming flag gets set correctly based on movement input + water + not on ground, then changed player is_swimming to just read that flag. matches what vanilla does in LivingEntity#updateSwimming.

also fixed a small typo in set_swimming where the parameter had the wrong name.

### DIFF
--- a/pumpkin/src/entity/living.rs
+++ b/pumpkin/src/entity/living.rs
@@ -853,7 +853,6 @@ impl LivingEntity {
                 && !self.entity.on_ground.load(SeqCst)
                 && self.entity.water_height.load() > self.get_swim_height();
             self.entity.set_swimming(should_swim).await;
-            
         } else if self.entity.swimming.load(SeqCst) {
             self.entity.set_swimming(false).await;
         }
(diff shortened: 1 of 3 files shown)


TheRealMusawer:
fixes #2625

RoosterBooster007:
Hey @TheRealMusawer, thanks for digging into this one. It looks like #3247 already sorted it out, though (merged as 0a3c4994). The cause was fluid heights and the eye-submersion check, not the sprint gate, and that gate is actually what vanilla's `Entity#updateSwimming` does, so dropping it would put non-sprinting players into the swim pose. I'll close this out, but give latest master a try and shout if you still see it flicker.

Issue #2626, opened 2026-07-28
Chunk writer drops block_entities, structures and PostProcessing when saving a vanilla 26.2 world


leminkozey:
Loading a vanilla 26.2 world works fine, saving it destroys data. Every chunk a player visits loses its block entities on the next autosave. Chests, spawners, furnaces, signs, all gone.

## Setup

- nightly build, log says `Starting Pumpkin 0.1.0-dev+26.2 Minecraft (Protocol 776)`
- Windows 11 x64, `pumpkin-X64-Windows.exe` straight from the nightly release
- World: vanilla world exported from Aternos, `DataVersion 4903`, `ServerBrands: vanilla`, no mods, no plugins
- Converted the 26.1+ `dimensions/minecraft/<dim>/` layout to the `DIM-1`/`DIM1` layout that `into_level` expects
- `autosave_ticks = 6000`, `[world.chunk] type = "anvil"`, compression LZ4

## What happens

One player joined, flew around for about 20 minutes, left. Exactly one region file got rewritten, `r.-1.0.mca`, and 25 of its 973 chunks changed. I diffed those against the untouched original.

Chunk at index 302 (`xPos -18, yPos -4, zPos 9`):

| field | original | after Pumpkin saved it |
|---|---|---|
| `DataVersion` | 4903 | **4790** |
| `block_entities` | 2 entries (`minecraft:chest`, `minecraft:mob_spawner`) | **0 entries** |
| `structures` | Compound, 2 keys | **gone** |
| `PostProcessing` | List[24] | **gone** |
| `carving_mask` | LongArray[653] | **gone** |
| `InhabitedTime` | 0 | gone |
| `LastUpdate` | 1287805 | gone |
| `entities` | List[0] | gone |
| `Heightmaps` | 4 keys | 3 keys |
| `Status` | `minecraft:initialize_light` | `minecraft:light` |
| `sections` | List[24] | List[24], kept |

Uncompressed NBT dropped from 35173 bytes to 6880, so roughly 80 percent of the chunk is gone. Terrain survives, nothing else does. On a survival world that means every chest anyone walks past gets emptied and removed.

## The DataVersion is a second problem

[...]

Issue #2472, opened 2026-07-23
[Legal] Proprietary Mojang assets (NBT structures, and others) bundled in the repository — risk & proposed solution


Totgocpro:
### Description

The repository currently contains Minecraft files that are **direct, unmodified copies** of assets from Mojang's official `client.jar`.

Affected files/folders:

- `pumpkin-world/assets/` -> Many `.nbt` files are direct copy from Minecraft 26.1.2 official client [MOVED in `assets/datapacks/26_2/data/minecraft/structure`]
- `assets/loot_table/` -> Many `.json` files are direct copy from Minecraft 26.1.2 official client [MOVED in `assets/datapacks/26_2/data/minecraft/loot_table`]
- `assets/` -> Some files are just minified versions of the official client files
- And possibly others I may have missed

**Re-implementing the Minecraft protocol remains legal**

### Current risks :

- **DMCA takedown** of the repository by Mojang/Microsoft
- **Legal liability** for the project maintainers and contributors

### How should it work?

Copying official Mojang assets into the repository remains prohibited, but (as Minecraft patchers do, such as Paper, or Fabric), the code can be downloaded by the end user (automatically by the patcher or in this case Pumpkin), and used by the program.

Here is a clear example of what Pumpkin can do when the user first starts the server:

1. Inform the user that the server will download Mojang assets, and ask them to accept the Minecraft EULA (https://www.minecraft.net/en-us/eula).  
   *Good practice: Mojang does not technically block downloads, but using their assets requires EULA compliance.*
2. Fetch the latest version (or selected version) on https://piston-meta.mojang.com/mc/game/version_manifest_v2.json
3. Retrieve target version metadata: https://piston-meta.mojang.com/v1/packages/{sha1}/{version}.json
4. Legally download the official Minecraft client from https://piston-data.mojang.com/v1/objects/{sha1}/client.jar in a temporary file
5. Extract the assets from the client.jar (https://lib.rs/crates/zip can do the trick), and use them directly

### Proposed action items

- [ ] Remove all currently committed Mojang assets from the repository
- [ ] Purge them from Git history (using `git filter-repo` or BFG Repo-Cleaner)
- [ ] Implement the asset fetching flow at first startup
- [ ] Add a `pre-commit` hook to prevent accidental re-commits
- [ ] Update `CONTRIBUTING.md` with a legal notice

[...]

PR #2640, opened 2026-07-29
fix(world): read and write vanilla block and biome palettes again 🤖🤖🤖

### LINKED ISSUE #2639: Anvil reader and writer cannot handle vanilla block/biome palettes — vanilla worlds load as air and are destroyed on next save

### Operating System

Windows 11 (reproduced by unit test, platform-independent)

`0.1.0-dev+26.2` / `8db9c01` (current master)

### What happened?

Loading a world written by vanilla Minecraft produces **air everywhere and one wrong biome everywhere**. The next autosave writes that back over the original region file.

Since the 1.13 flattening, vanilla writes section block palettes as a list of `{Name, Properties}` compounds and biome palettes as a list of resource-location strings. We only understand raw numeric ids:

```rust
// pumpkin-world/src/chunk/format/mod.rs
pumpkin_nbt::tag::NbtTag::List(list) => {
    list.iter().map(|t| {
        let val = match t {
            NbtTag::Int(x) => *x as u16,
            NbtTag::Short(x) => *x as u16,
            NbtTag::Byte(x) => *x as u16,
            NbtTag::Long(x) => *x as u16,
            _ => 0,          // <-- a {Name, Properties} compound lands here
        };
        BlockStateId::new_or_air(val)
    })
}
```

A compound entry hits `_ => 0` and becomes state 0, air. `extract_u8_array` is the same shape for biomes and has no `NbtTag::String` arm at all, so every biome becomes id 0.

`README.md:52` lists `- [x] Chunk Loading (Vanilla, Linear, Pump)`, and the docs tell users to drop an existing world into the server directory, so this is advertised functionality failing silently and destructively.

### Measured, against a real world

I generated a world with Mojang's own `server.jar` for 26.2 (sha1 `823e2250...`, straight from the launcher manifest) and fed one chunk through `ChunkData::internal_from_bytes`.

The chunk, read with an independent NBT parser:

```
DataVersion : 4903        Status : minecraft:full
block_states.palette -> TAG_Compound  [{Name: "minecraft:bedrock"},
                                       {Name: "minecraft:deepslate", Properties: {axis: y}}, ...]
biomes.palette       -> TAG_String    ["minecraft:plains"]
```

Through our loader:

| | non-air blocks | distinct block types | biome |
|---|---|---|---|
| master | **0** / 98304 | 1 (air) | 0 |
| with the fix below | **31181** / 98304 | 27 | 40 (plains) |

### This is a regression, not a missing feature

`d048c338` ("chore: prepare 1.20.5", 2026-04-18) removed the palette name lookups from `pumpkin-world/src/chunk/palette.rs` in **both** directions:

```
-            .map(|entry| Biome::from_name(&entry.name).unwrap_or(&Biome::PLAINS).id)
-            .map(|entry| entry.get_state_id())
-    fn block_state_id_to_palette_entry(registry_id: u16) -> BlockStateCodec {
-                .map(|registry_id| PaletteBiomeEntry {
```

`git log -S from_name -- pumpkin-world/src/chunk/palette.rs` returns exactly two commits: the one that added it and `d048c338`. Nothing restored it.

That commit is **104 commits after** the 26.1 port (`0635b048`), so named palettes were being read and written correctly right up until it landed. The commit title refers to 1.20.5, which suggests this was collateral rather than intentional.

### To Reproduce

1. Generate or obtain any world written by vanilla Minecraft.
2. Start Pumpkin against it.
3. Fly around: the terrain is gone, everything is air.
4. Wait for one autosave. The region file is now permanently air.

### Expected behavior

A vanilla world loads with its terrain intact, and saving it produces a file vanilla can still read.

### Fix

PR linked below. Two parts, because `d048c338` broke two things:

- **Reader**: a `Compound` arm resolving `{Name, Properties}`, and a `String` arm resolving biome resource locations. The block half can reuse `BlockStateResolver::resolve_simple`, which already does namespace stripping and property-to-state-id lookup for structure templates. One gotcha: `Biome::from_name` matches bare names and does *not* strip `minecraft:` itself, unlike `Block::from_name`.
- **Writer**: emit `{Name, Properties}` and resource locations again. Without this, a vanilla world loads correctly and is then written back as numeric ids, so we can reread it but vanilla and third-party tools cannot — a one-way door into a format that still calls itself `.mca`.

The numeric reader arms stay, so worlds written by current builds keep loading.

Verified by round-tripping the real 26.2 chunk through load and save: **98304 of 98304 blocks identical**, palette tag types matching vanilla on both sides (`TAG_Compound` / `TAG_String`), output 30972 bytes against vanilla's 31809.

### Two related things worth flagging separately

- Nothing in CI could have caught this. The only test that reads a real region file, `load_java_chunk` in `pumpkin-world/src/chunk/format/anvil.rs`, is commented out and its asset was never committed. A committed fixture would be the obvious guard, but a real region file is Mojang-authored — see #2472 — so it probably needs a synthetic one.
- While confirming this I also found that the chunk writer drops every root NBT tag it does not model, including `InhabitedTime`, which it parses on load and then never writes. That is #2626 and has its own PR.

RicePollution:
Fixes #2639.

## What was broken

Since the 1.13 flattening, vanilla writes section block palettes as a list of `{Name, Properties}` compounds and biome palettes as a list of resource locations. We only understood raw numeric ids — `extract_u16_array` matched the numeric variants and sent everything else through `_ => 0`, and `extract_u8_array` had no `String` arm at all.

So a compound palette entry became state 0, air, and a biome resource location became id 0. **Loading a vanilla world produced air everywhere, and the next autosave wrote that back over the original.**

This is a regression. `d048c338` ("chore: prepare 1.20.5") removed the name lookups in both directions — `block_state_id_to_palette_entry` on the write side, `get_state_id` and `Biome::from_name` on the read side — and nothing restored them. `git log -S from_name -- pumpkin-world/src/chunk/palette.rs` shows exactly two commits: the one that added it, and that one. It landed 104 commits after the 26.1 port, so this worked correctly until then.

## What changed

Two commits, one per direction, because the regression broke both.

**Reader** — a `Compound` arm resolving `{Name, Properties}` through the existing `BlockStateResolver::resolve_simple`, which already handles namespace stripping and property-to-state-id lookup for structure templates; and a `String` arm for biome resource locations. `Biome::from_name` matches bare names and does not strip `minecraft:` itself, unlike `Block::from_name`, so the namespace is stripped explicitly — miss that and every biome silently becomes 0.

**Writer** — emit `{Name, Properties}` and resource locations again. Without this half, a vanilla world loads correctly and is then written back as numeric ids: we can reread it, vanilla and third-party tools cannot. That is a one-way door into a proprietary format that still calls itself `.mca`.

An unrecognised name costs one palette entry and logs a warning, rather than silently blanking the section. The numeric reader arms stay, so worlds written by current builds keep loading.

## Verification

I generated a world with Mojang's own `server.jar` for 26.2 (sha1 `823e2250...`, straight from the launcher manifest) and ran a real chunk through the loader.

Loading, `DataVersion 4903`, `Status: full`:

| | non-air blocks | distinct block types | biome |
|---|---|---|---|
| master | **0** / 98304 | 1 (air) | 0 |
| this PR | **31181** / 98304 | 27 | 40 (plains) |

Round-tripping that same chunk through load *and* save:

```
blocks compared : 98304
identical       : 98304  (100.0000%)
different       : 0
block palette elem type : vanilla=TAG_Compound  pumpkin=TAG_Compound
biome palette elem type : vanilla=TAG_String    pumpkin=TAG_String
```

Output is 30972 bytes against vanilla's 31809, so no size regression.

That fixture is Mojang-authored, so per #2472 it is **not** committed — the 11 tests here use synthetic chunks in the same shape, and they are pure functions with no server, disk, or tokio runtime.

## Impact, stated plainly

- Vanilla worlds load and save correctly. This is the headline.
- **The on-disk format of newly written chunks changes**, for every world including ones we generated ourselves. That is the intended fix — it is what makes output readable by vanilla — but it is a real change and worth a maintainer's eye. Existing numerically-written worlds still load, because the numeric reader arms are kept.
- `linear` and `pump` wrap the same chunk NBT, so they get the corrected palettes too.
- Slight CPU cost on save: a name string and property map per *palette entry*, not per block. A palette is the deduplicated list, typically a handful of entries per section.

## Known limitations

- No committed fixture covers a real region file, so nothing in CI would catch a future recurrence. The only test that would, `load_java_chunk` in `pumpkin-world/src/chunk/format/anvil.rs`, is commented out and its asset was never committed. Worth solving, but it needs a synthetic region file to stay clear of #2472, and that is its own change.
- An unknown block name still resolves to air. That is the best available answer without a block-state upgrade path, and it now logs instead of doing it silently.

## Conflicts

This touches the same file as my #2626 PR, and both add a test module to it, so they will conflict textually regardless of merge order. Happy to rebase whichever lands second — say the word and I will do it immediately.

### DIFF
--- a/pumpkin-world/src/chunk/format/anvil.rs
+++ b/pumpkin-world/src/chunk/format/anvil.rs
@@ -37,8 +37,31 @@ pub const CHUNK_COUNT: usize = REGION_SIZE * REGION_SIZE;
 /// The number of bytes in a sector (4 KiB)
 const SECTOR_BYTES: usize = 4096;
 
-// 26.1.2
-pub const WORLD_DATA_VERSION: i32 = 4790;
+/// The `DataVersion` stamped on every chunk we write.
+///
+/// This has to name the format `internal_to_bytes` actually produces, because it is
+/// what tells vanilla which datafixers to run when it opens a world we saved. Too low
+/// and vanilla migrates data that has already been migrated; too high and it skips
+/// fixes the data still needs.
+///
+/// Which makes it the same version `level.dat` claims through
+/// `MAXIMUM_SUPPORTED_WORLD_DATA_VERSION` — a world whose `level.dat` says 26.2 while
+/// its chunks say 1.21.11 is not a coherent world. This sat at 4790 while `level.dat`
+/// said 4903, which went unnoticed only because the numeric block palette meant nothing
+/// outside Pumpkin could read our chunks anyway.
+pub const WORLD_DATA_VERSION: i32 = 4903; // 26.2
+
+/// A guard, deliberately not an alias: what we emit and what we accept on load are
+/// different questions, and tying them together would mean the next widening of load
+/// support silently stamped a newer version onto chunks still written in the old shape.
+/// Failing the build instead forces the choice to be made on purpose.
+///
+/// If you are raising the load ceiling to a format `internal_to_bytes` does not yet
+/// produce, separate these two rather than bumping this one to match.
+const _: () = assert!(
+    WORLD_DATA_VERSION == crate::world_info::MAXIMUM_SUPPORTED_WORLD_DATA_VERSION,
+    "a chunk's DataVersion must name the format we write, which is also what level.dat claims"
+);
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(u8)]
--- a/pumpkin-world/src/chunk/format/mod.rs
+++ b/pumpkin-world/src/chunk/format/mod.rs
@@ -122,6 +122,33 @@ fn biome_id_from_palette_entry(name: &str) -> Option<u8> {
     Biome::from_name(bare).map(|biome| biome.id)
 }
 
+/// Renders a block state id as a vanilla palette entry,
+/// `{Name: "minecraft:stone", Properties: {..}}`.
+///
+/// `Properties` is omitted for a block that has none, matching vanilla.
+fn palette_entry_from_block_state_id(id: BlockStateId) -> pumpkin_nbt::tag::NbtTag {
+    let block = id.to_block_id().to_block();
+
+    let mut entry = NbtCompound::new();
+    entry.put_string("Name", format!("minecraft:{}", block.name));
+
+    if let Some(properties) = block.properties(id) {
+        let mut props = NbtCompound::new();
+        for (key, value) in properties.to_props() {
+            props.put_string(key, value.to_string());
+        }
+        entry.put_compound("Properties", props);
+    }
+
+    pumpkin_nbt::tag::NbtTag::Compound(entry)
+}
+
+/// Renders a biome id as a vanilla palette entry, a plain resource location.
+fn palette_entry_from_biome_id(id: u8) -> pumpkin_nbt::tag::NbtTag {
+    let registry_id = Biome::from_id(id).unwrap_or(&Biome::PLAINS).registry_id;
+    pumpkin_nbt::tag::NbtTag::String(format!("minecraft:{registry_id}").into())
+}
+
 fn extract_u16_array(tag: &pumpkin_nbt::tag::NbtTag) -> Option<Box<[BlockStateId]>> {
     match tag {
         pumpkin_nbt::tag::NbtTag::IntArray(arr) => Some(
@@ -524,7 +551,7 @@ impl ChunkData {
             let palette_tags: Vec<NbtTag> = block_states_nbt
                 .palette
                 .iter()
-                .map(|id| NbtTag::Int(BlockStateId::as_u16(*id) as i32))
+                .map(|id| palette_entry_from_block_state_id(*id))
                 .collect();
             bs_comp.put_list("palette", palette_tags);
             section_comp.put_compound("block_states", bs_comp);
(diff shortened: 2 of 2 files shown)


Sawlyer:
`WORLD_DATA_VERSION` is 4790 (1.21.11) in `pumpkin-world/src/chunk/format/anvil.rs:41`, and that is what `internal_to_bytes` stamps on every chunk, while `MAXIMUM_SUPPORTED_WORLD_DATA_VERSION` is 4903 (26.2).

Until now the numeric palette meant nothing outside Pumpkin could read the chunk, so the stamp did not matter. Once the palette is written in vanilla shape it does: a 26.2 world saved here becomes loadable by vanilla, which will then run datafixers from 4790 upward over data that is already 4903.

Bump it here, or does that belong with #2641?

RicePollution:
Agreed, and bumped here rather than in #2641 — this is the PR that makes it matter, so it shouldn't ship without it.

`WORLD_DATA_VERSION` is now 4903, matching what `level.dat` has been written with all along via `MAXIMUM_SUPPORTED_WORLD_DATA_VERSION`. Every chunk we saved was claiming to be four format revisions older than the world containing it; that went unnoticed precisely because the numeric palette meant nothing outside Pumpkin could read the chunks, so nothing ever acted on the stamp. Once they're vanilla-readable, the stamp is what selects the datafixers, and a 26.2 world saved here would get migrated from 4790 upward over data that's already 4903.

I also added a `const` assertion that the two constants are equal, so they fail the build rather than drifting apart again:

```rust
const _: () = assert!(
    WORLD_DATA_VERSION == crate::world_info::MAXIMUM_SUPPORTED_WORLD_DATA_VERSION,
    "chunk DataVersion must match the version written into level.dat"
);
```

One small note: the in-tree comment labelled 4790 as `26.1.2`, while you and the issue both call it 1.21.11. That stale label is gone with the line it annotated; the new constant is labelled `26.2`, matching how `world_info` already labels 4903.

PR #2641, opened 2026-07-29
fix(world): stop the chunk writer dropping NBT it does not model 🤖🤖🤖

RicePollution:
Part of #2626 — see "What this does not fix" below, which is most of the reported symptom.

## What changed

`ChunkData::internal_to_bytes` builds a fresh root compound and writes eleven keys. Every other root tag on the parsed chunk died with the local `let root_tag = nbt.root_tag;` binding, so saving a chunk stripped it. This PR addresses two separate losses in that function.

**`InhabitedTime` was parsed on load and never written back.** It is not foreign metadata — it is our own live field. `World::tick` increments it every tick for every chunk in a player's active set, and `MobSpawnEquipment` reads it for regional difficulty (`local_scale += chunk_inhabited_time / 3_600_000`). It was read at load, incremented in memory, and then silently dropped on save, so **the first save zeroed whatever had accumulated** and local difficulty could never rise above whatever was already on disk. It is now a twelfth written key, sourced from the live atomic rather than from the retained map, so a stale disk value cannot shadow it.

**Root tags we do not model are now preserved.** `structures`, `LastUpdate`, `blending_data` and anything a future version adds are cloned off the parsed root into a new `ChunkData::retained_nbt` and merged back at the end of the writer.

Merging last is load-bearing: `NbtCompound::put` is insert-if-absent, so a bug in the filter can only ever drop a foreign tag, never corrupt an owned one. There is a test pinning that ordering.

## Retention is gated at `features`, not `full`

An earlier revision of this PR gated retention on `Status: full`. That was wrong, and it made the whole change a no-op for the chunks actually reported — #2626's diff is of a chunk at `minecraft:initialize_light`. Thanks to @Sawlyer for catching it.

The cutoff is now `StagedChunkEnum::from(status) >= StagedChunkEnum::Features`.

Below `features`, the original reasoning stands: the chunk re-enters generation through `ProtoChunk::from_chunk_data`, which restores blocks and heightmaps but leaves `structure_starts` empty and `carving_mask` fresh, so it is finished *without* the structures the retained metadata describes. Writing that metadata back beside contradicting terrain would convert an honest data loss into a chunk that lies about its own contents. An unrecognised `Status` falls through to `ChunkStatus::Empty` and lands here too.

At or above `features` that cannot happen — the features/structures step never re-runs for such a chunk, three times over:

- `drop_satisfied_tasks` deletes every task at or below the loaded stage on IO completion
- `Cache::advance` early-returns when the centre chunk has already reached the requested stage
- the scheduler's dispatch loop drops queued nodes for already-reached stages

and `structure_starts` is only ever populated by `set_structure_starts`/`set_structure_references`, whose tasks are dropped for anything loaded at or above `StructureReferences`.

Widening the cutoff makes two keys reachable that could not appear under a `full`-only gate: `carving_mask` and `Lights`, both of which belong to chunks caught mid-generation. `Lights` — vanilla's per-section list of light sources still to be processed — was in neither the written nor the never-retained list, because it was previously unreachable. It is now excluded, since retaining it would write a work record onto a chunk we go on to stamp `Status: minecraft:full`.

**Keys that record work still to be done are never retained**, as opposed to keys that describe what is already present. `PostProcessing`, `below_zero_retrogen`, `UpgradeData`, `carving_mask` and `Lights` are instructions to re-run work on load. We have no code for any of them, so we can neither perform nor clear them — retaining one would re-assert the same directive on every save, against blocks that have since changed. Root `entities` is excluded separately because entities live in the `entities/` region folder, so a retained copy would be a second, never-updated set of the same mobs.

How those keys are classified is reasoning about vanilla, not something validated against a real region file — there is no `.mca` fixture in the workspace. The doc comment says so. **If you have a vanilla world and think I have misclassified a key, please say so.**

## What this does not fix

**Block entities, and block and fluid ticks, are still lost on the same chunks.** This is the headline symptom of #2626 — the emptied chests — and it is a different bug in a different place, so this PR does not close the issue.

`block_entities` is a *modeled* field, so it never routes through `retained_nbt` by design. It dies in `ProtoChunk::from_chunk_data`, which never copies `chunk_data.pending_block_entities`; `upgrade_to_level_chunk` then rebuilds the saved chunk from the empty vector `ProtoChunk::new` installed. `fluid_ticks` goes the same way, and `block_ticks` worse — `ProtoChunk` has no such field at all, so the rebuild hardcodes `ChunkTickScheduler::default()`.

Two things make that worse than the tag loss fixed here: it needs no player interaction, since Proto chunks skip the `is_dirty()` check that Level chunks get on the save path; and it is not limited to below-full chunks, since `needs_relighting` sends `full` chunks with `isLightOn: false` through the same `from_chunk_data`. Sent separately.

Note this is distinct from #2452, which persists *live* block entities out of `World::block_entities`. That one covers entities a player has touched; this one covers entities that were on disk and never made it back out. #2626 needs both.

## Known limitations

- Chunks below `features` still lose foreign root tags, deliberately, as above.
- Nested tags inside *owned* keys are still dropped — notably the `OCEAN_FLOOR` heightmap, since `ChunkHeightmaps` models three of vanilla's four. Root-level retention does not reach it. This is not a fully lossless round trip.
- An `initialize_light` chunk is written back as `Status: minecraft:light` with `isLightOn: true`, because `ChunkStatus::InitializeLight | Light` both map to `StagedChunkEnum::Lighting` and the Lighting task is then dropped as satisfied. So light initialization never actually runs for it. Pre-existing, orthogonal to this PR, and worth its own issue.

## Impact

- Vanilla-authored `structures` / `LastUpdate` and any unrecognised root tag now survive a save on any chunk at or above `features`, which is the population a half-generated imported world mostly consists of.
- `InhabitedTime` now persists, on every world including ones we generated ourselves. This is a behaviour change: regional difficulty will now actually accumulate across restarts rather than resetting.
- No new locks, no new allocations for generated chunks (`retained_nbt` is empty by construction and costs an empty `HashMap`). For a loaded chunk the retained subset is cloned once per save.

## Changes since first review

- Gate moved from `full` to `features`; `chunks_below_full_status_retain_nothing` replaced by a pair of tests covering both sides of the new boundary, including `initialize_light`.
- `Lights` added to the never-retained set, with coverage at both ends of the retention range.
- The `perf(world)` commit that released the section, heightmap and light guards early has been **dropped from this PR**. Its "no behaviour change" claim was false: `block_ticks`/`fluid_ticks` are read after the drops, and the section read guard was an incidental barrier against a concurrent `set_block_no_heightmap_update` plus its `schedule_tick`. It will be resent on its own with the tick queues snapshotted above the drops. (Block entities were never affected — they are snapshotted under their own mutex before any of the four guards are taken.)
- The `DataVersion` mismatch is fixed in #2640 rather than here, since that is the PR that makes our chunks readable by vanilla at all and therefore makes the stamp matter.

### DIFF
--- a/pumpkin-world/src/chunk/format/mod.rs
+++ b/pumpkin-world/src/chunk/format/mod.rs
@@ -152,6 +152,84 @@ where
     })
 }
 
+/// Root NBT keys that `ChunkData::internal_to_bytes` writes itself.
+///
+/// Some are recomputed from live state (`DataVersion`, `xPos`/`yPos`/`zPos`, `Status`,
+/// `sections`, `isLightOn`, `InhabitedTime`); the rest are echoed back out of a typed
+/// field that `internal_from_bytes` parsed them into (`Heightmaps`, `block_ticks`,
+/// `fluid_ticks`, `block_entities`). Either way the writer owns them, so they must
+/// never be sourced from [`ChunkData::retained_nbt`].
+///
+/// Keep in sync with the `root_compound.put_*` calls in `internal_to_bytes`;
+/// `writer_emits_exactly_the_written_root_keys` fails on drift.
+const WRITTEN_ROOT_KEYS: &[&str] = &[
+    "DataVersion",
+    "xPos",
+    "yPos",
+    "zPos",
+    "Status",
+    "Heightmaps",
+    "sections",
+    "block_ticks",
+    "fluid_ticks",
+    "block_entities",
+    "isLightOn",
+    "InhabitedTime",
+];
+
+/// Root keys we neither write nor carry over, because writing them back would be worse
+/// than losing them.
+///
+/// The rule: retain keys that *describe* what is already in the chunk, drop keys that
+/// record *work still to be done*. We have no code for any key below, so we can never
+/// clear one once it has been applied — we would re-assert the same directive on every
+/// save, against blocks that have since changed.
+///
+/// - `entities`: entities live in the separate `entities/` region folder, so a retained
+///   copy would be a second, never-updated set of the same mobs.
+/// - `PostProcessing`, `below_zero_retrogen`, `UpgradeData`: work records.
+/// - `carving_mask`: generation-stage scratch state. Unreachable in practice, since we
+///   only retain from chunks loaded at `Status: full`, but listed so the rule stays legible.
+///
+/// How these are classified is reasoning about vanilla rather than something validated
+/// against a real region file. What is verified is that we have no code for any of them.
+const NEVER_RETAINED_ROOT_KEYS: &[&str] = &[
+    "entities",
+    "PostProcessing",
+    "below_zero_retrogen",
+    "UpgradeData",
+    "carving_mask",
+];
+
+fn is_owned_root_key(key: &str) -> bool {
+    WRITTEN_ROOT_KEYS.contains(&key) || NEVER_RETAINED_ROOT_KEYS.contains(&key)
+}
+
+/// Clones the root tags we do not model out of a freshly parsed chunk, so they can be
+/// written back untouched instead of dropped.
+fn retain_foreign_root_tags(root: &NbtCompound) -> NbtCompound {
+    root.child_tags
+        .iter()
+        .filter(|&(key, _)| !is_owned_root_key(key))
+        .map(|(key, tag)| (key.clone(), tag.clone()))
+        .collect()
+}
+
+/// Writes retained tags back onto a root compound.
+///
+/// Must be called *after* every `root_compound.put_*` in `internal_to_bytes`:
+/// [`NbtCompound::put`] is insert-if-absent, so merging last makes it structurally
+/// impossible for a stale disk value to shadow live chunk state. A bug in the filter can
+/// then only ever drop a foreign tag, never corrupt an owned one.
+///
+/// Deliberately not `Extend`: `NbtCompound` has two `Extend` impls that disagree about
+/// overwriting, and the call site cannot show which one is selected.
+fn merge_foreign_root_tags(root: &mut NbtCompound, retained: &NbtCompound) {
+    for (key, tag) in &retained.child_tags {
+        root.put(key, tag.clone());
+    }
+}
+
 impl ChunkData {
     #[allow(clippy::too_many_lines)]
     pub fn internal_from_bytes(
@@ -367,6 +445,19 @@ impl ChunkData {
             _ => ChunkStatus::Empty,
         };
 
+        // Only a chunk loaded at `full` is guaranteed to be written back over the same
+        // terrain. Anything below `full` goes straight back into the generation pipeline
+        // through `ProtoChunk::from_chunk_data`, which restores blocks and heightmaps but
+        // leaves `structure_starts` empty and `carving_mask` fresh, so the chunk is
+        // finished *without* the structures the retained metadata describes. An
+        // unrecognised `Status` lands here too, via the fall-through above — precisely
+        // the case where we must not vouch for anything.
+        let retained_nbt = if status == ChunkStatus::Full {
+            retain_foreign_root_tags(&root_tag)
+        } else {
+            NbtCompound::new()
+        };
+
         Ok(Self {
             section,
             heightmap: std::sync::Mutex::new(heightmaps),
@@ -382,6 +473,7 @@ impl ChunkData {
             status,
             blending_data: None,
             inhabited_time: AtomicU64::new(root_tag.get_long("InhabitedTime").unwrap_or(0) as u64),
+            retained_nbt,
         })
     }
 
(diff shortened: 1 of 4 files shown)


Sawlyer:
The chunk in #2626's own diff is at `Status: minecraft:initialize_light`, which maps to `ChunkStatus::InitializeLight`, not `Full`. Capture is gated on `status == ChunkStatus::Full`, so `retained_nbt` stays empty for exactly that chunk and its `structures`, `PostProcessing`, `carving_mask` and `LastUpdate` are still dropped on save. `chunks_below_full_status_retain_nothing` pins that as intended.

Does the reported case need the gate widened, or is "a below-full chunk gets regenerated anyway" meant to cover it?

Separate: dropping the section, heightmap and light guards mid-function means `sections` is snapshotted at one point while `block_ticks` and `block_entities` are read after the guards are gone, so one save can mix two states. Deliberate scope here?

RicePollution:
Both points land. Taking them in turn, and the first one turned out to go deeper than the gate.

## The gate

Widened. You're right that `initialize_light` is exactly the status in #2626's diff and that the `Full` gate made this PR a no-op for the chunk that was actually reported — `chunks_below_full_status_retain_nothing` was pinning that as intended, which was the wrong call.

The gate is now `StagedChunkEnum::from(status) >= StagedChunkEnum::Features`. I went looking for whether the original justification ("a below-full chunk goes back through generation, which finishes it without the structures the retained metadata describes") actually holds, and it only holds strictly below `features`. A chunk resumed at or above that stage never re-runs structure placement, carvers or features, guaranteed three separate ways:

- `drop_satisfied_tasks` deletes every task at or below the loaded stage on IO completion (`schedule.rs`)
- `Cache::advance` early-returns when the centre chunk has already reached the requested stage (`generation_cache.rs`)
- the dispatch loop drops queued nodes for already-reached stages

And `structure_starts` is only ever populated by `set_structure_starts`/`set_structure_references`, whose tasks are dropped for anything loaded at or above `StructureReferences`, with `from_chunk_data` leaving it `default()`. So `Features` is the exact boundary where the contradiction stops being real. Below it the original reasoning stands and retention stays off.

One consequence worth calling out: widening makes `Lights` reachable. It's vanilla's per-section light-source work list, written for chunks below `light` status, and it was in neither `WRITTEN_ROOT_KEYS` nor `NEVER_RETAINED_ROOT_KEYS` because at a `full`-only gate it could never appear. Retaining it would write a work record onto a chunk we then stamp `Status: minecraft:full` — precisely the category the exclusion list exists for. Added to `NEVER_RETAINED_ROOT_KEYS` and to the spelled-out list in the test. Same caveat as the other entries there: that classification is reasoning about the vanilla format, not something I validated against a region file. What is verified is that we have no code for it anywhere.

## This PR still does not close #2626

Worth being blunt about, because widening the gate looks like it should and it doesn't. `block_entities` — the chest and spawner in the report, and the part that actually hurts — is a *modeled* field, so it never routes through `retained_nbt` at all; it's in `WRITTEN_ROOT_KEYS` by design.

Chasing where it actually dies: neither the reader nor the writer is at fault. `ProtoChunk::from_chunk_data` never copies `chunk_data.pending_block_entities`, so the ProtoChunk starts with the empty vector from `ProtoChunk::new`, and `upgrade_to_level_chunk` then rebuilds the saved `ChunkData` from it. `fluid_ticks` dies the same way, and `block_ticks` worse — `ProtoChunk` has no such field, so `upgrade_to_level_chunk` hardcodes `ChunkTickScheduler::default()`.

Two things make that nastier than the tag loss this PR fixes: it needs no player interaction, because Proto chunks skip the `is_dirty()` check on the save path that Level chunks get; and it isn't limited to below-full chunks, since `needs_relighting` sends `full` chunks with `isLightOn: false` through the same `from_chunk_data` and pins them at `Features`, so the entities are neither preserved nor regenerated.

That's a separate fix in `from_chunk_data`/`upgrade_to_level_chunk` rather than anything in this PR, and I'd rather send it on its own than bolt it on here. Flagging it so nobody merges this expecting the chests to come back. It'll conflict textually with this branch in both files — it inserts at the same spot in `from_chunk_data` and rewrites the lines next to the tick fields in `chunk_state.rs` — so whichever lands second needs a rebase.

## The guard drops

Half right, and the half that's wrong is my fault for putting a perf change in a data-integrity PR at all — I've dropped that commit from this branch.

Block entities were never at risk: `block_entities_nbt` is snapshotted under its own mutex *before* the four guards are acquired, so `block_lock` never covered it and releasing early couldn't widen anything there. Also, placing or breaking a chest doesn't write `pending_block_entities` at all — that map only holds not-yet-deserialized entities off disk.

The tick queues are a real problem though, and it's the same shape you described. `block_ticks`/`fluid_ticks` are read after the drops, and the section read guard was an incidental one-directional barrier against `set_block_no_heightmap_update` — so a block write plus its `schedule_tick` could land between the sections serialized above and the queues serialized below, saving a tick for a block the same file says isn't there. It's reachable: autosave clones the live `Arc` and leaves the chunk mapped, so serialization genuinely runs against a mutating chunk. The saved tick happens to be inert on reload, but "No behaviour change" in that commit message was simply false.

When I resend it separately it'll snapshot the two tick queues above the drops, and the message will lose both the "no behaviour change" claim and the "most expensive part of the function" one — the latter doesn't survive checking either, since block sections are `Indexed` rather than `Dense` on the common path and never take the palette scan I was implicitly attributing the cost to.

PR #2643, opened 2026-07-29
plugin-api: Rework filesystem permissions and move data folders.

QuantumSegfault:
Removes the ability for general `fs.read` and `fs.write`. There are very few legitimate reasons to read/write files outside a plugin's data directory, and such scenarios can be accounted for on a case-by-case basis later (e.g. one suggestion was maybe the need for access to world files).

The paths of the plugin data directories have been changed slightly. Instead of directly depositing data in `plugins/<plugin name>`, the data is placed in `plugins/data/<plugin name>` (to help avoid conflicts with other things that might go in there; e.g. the `cache`; or `config` in the future). Said path has also been obscured from the plugins. The plugin's data folder is mounted simply as `data` within the guest.

Finally, the two remaining two filesystem permissions have been tweaked. `fs.read.data` is no longer enabled by default, and `fs.write.data` no longer implies `fs.read.data`.

## Testing

The changes were tested thoroughly with some C plugins trying to `fopen` (and `fgetc` on the handle) files in `r` or `w` mode in (and out) of the data directory, with different combinations of `fs.read.data` and `fs.write.data` allowed.

### DIFF
--- a/pumpkin-plugin-api/src/permissions.rs
+++ b/pumpkin-plugin-api/src/permissions.rs
@@ -34,16 +34,14 @@ pub const NETWORK_OUTBOUND: &str = "network.outbound";
 /// This is separate from `network.outbound`. This allows the use of `wasi:http`; the other allows the more powerful `wasi:sockets`.
 pub const HTTP_OUTBOUND: &str = "http.outbound";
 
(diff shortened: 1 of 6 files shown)


Snowiiii:

PR #2646, opened 2026-07-29
fix: persist air supply across relog

MythicSorcerer:
Air was not saved to NBT, causing it to reset to full on every login. This allowed players to exploit drowning by relogging to restore lost air bubbles.

**Changes:**
- Save `AirSupply` and `DrowningTick` to NBT in `write_nbt()`
- Load them back in `read_nbt()`
- Send the restored air supply to the client on login


MythicSorcerer:
Addressed Copilot's clamping feedback — added clamping on both read and write sides for AirSupply and DrowningTick to handle edge cases like mid-tick saves or corrupted NBT.

Snowiiii:
LGTM, Thank you @MythicSorcerer

PR #2647, opened 2026-07-30
plugin-api: Improve `PluginInitError` messages

QuantumSegfault:
Modifies Wasm plugin initialization so `PluginInitError` provides more context in error messages. Most notably, the error recieved when a plugin/component's requested `import` don't match what Pumpkin provides (the most frequent kind of breakage) is now actually tracked as a `ApiVersionMismatch`. Previously only `export` mismatches were obviously so.

Before:
```
Failed to load plugin from "./plugins/pumpkin_fly_plugin.wasm": Wasm plugin initialization error: plugin failed to load with error: component imports instance `pumpkin:plugin/command@0.1.0`, but a matching implementation was not found in the linker
```

After:
```
Failed to load plugin from "./plugins/pumpkin_fly_plugin.wasm": Wasm plugin initialization error: Plugin is built against a different API version: component imports instance `pumpkin:plugin/command@0.1.0`, but a matching implementation was not found in the linker
```

As part of the rework, `.cwasm` cache filenames now use the hash of the original wasm (same as the permission system), rather than a combination file name, length, and timestamp. This was done mainly to reduce the amount of error cases.

-----------

Error messages were tested for
- `ApiVersionMismatch` for import mismatches
- `FileReadFailed` when file is present, but read permission is denied
- `ComponentCacheWriteFailed` when the `plugins/cache` directory can't be written to.
- `CallInitPluginFailed` for traps during `init_plugin`
- `CallGetMetadataFailed` for traps during `get_metadata`

The rest, e.g. `LinkerSetupFailed` aren't all easy to reproduce, and remain untested, but there's no reason why they shouldn't work.

### DIFF
--- a/pumpkin/src/plugin/cache.rs
+++ b/pumpkin/src/plugin/cache.rs
@@ -30,14 +30,19 @@ impl PermissionCache {
 }
 
 pub async fn calculate_hash(path: &Path) -> tokio::io::Result<String> {
+    let bytes = fs::read(path).await?;
+    Ok(calculate_hash_for_bytes(&bytes))
+}
+
+#[must_use]
+pub fn calculate_hash_for_bytes(bytes: &[u8]) -> String {
     use std::fmt::Write;
 
-    let bytes = fs::read(path).await?;
     let mut hasher = Sha256::new();
-    hasher.update(&bytes);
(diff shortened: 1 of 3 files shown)


Snowiiii:

PR #2650, opened 2026-07-30
fix(player): swap XP progress arguments when loading from NBT

Rito5060:
## Description

`pumpkin_util::math::experience::progress_in_level` is declared as
`progress_in_level(points: i32, level: i32)` and computes
`points / points_in_level(level)`.

`Player::read_nbt` called it with the arguments swapped:

    let (level, points) = experience::total_to_level_and_points(total_exp);
    let progress = experience::progress_in_level(level, points);

**What changed:** the arguments are now passed in the right order.

**Why:** the restored progress was computed as `level / points_in_level(points)`,
which has no relation to the player's actual progress within their level.
`read_nbt` runs whenever a player is loaded, so the experience bar was wrong
after every rejoin — at 1000 total XP it showed a completely full bar instead
of ~3%.

## Testing

- `cargo clippy -p pumpkin --all-targets` — no new warnings.


Snowiiii:
@Rito5060 please fix conflicts

PR #2651, opened 2026-07-30
feat(plugins): Configuration API

PilkeySEK:
## Description
Adds a configuration API for plugins. There is `get-config` on `context`, which returns the `config` resource. Through the resource, plugins can call `get`, `get-or-default`, `set`, `remove`, `overwrite` and `has-key` (and their `path-*`-variants).
Currently, configs are saved in the TOML format in a new `config/` directory (where the plugin name is the name of the config file, with `.toml` extension).

Related Discord thread: https://discord.com/channels/1268592337445978193/1531403803813875913

Note that this PR depends on a change to `pumpkin-plugin-wit`, which is PR https://github.com/Pumpkin-MC/pumpkin-plugin-wit/pull/24.

## Testing
I tested that the config API works and there are also some small tests for single functions (like path parsing).

## Summary by CodeRabbit

* **New Features**
  * Added persistent TOML-based configuration for plugins.
  * Plugins can read, set, remove, overwrite, and check configuration values.
  * Supports booleans, numbers, strings, lists, objects, null values, nested keys, and array indices.
  * Added default-value operations for missing settings.
  * Configuration is loaded automatically and saved when changed, including during shutdown.
  * Exposed configuration access through the plugin context and WebAssembly plugin API.
  * Added support for path-based configuration access using keys and array indices.


PilkeySEK:
I forgot to mention that currently, plugin configs are only saved when the server stops... maybe they should be auto saved from time to time though?
Also I should probably check if /save-all is implemented in pumpkin already

PilkeySEK:
~~Another thing is that I assume the toml crate won't sort keys in any particular way so we might want to sort them alphabetically ourselves before saving~~

PilkeySEK:
The `PartialEq` implementations now make this possible:

```rust
impl Plugin for MyPlugin {
    // ...
    
    fn on_load(&mut self, context: pumpkin_plugin_api::Context) -> pumpkin_plugin_api::Result<()> {
        let config = context.get_config();
        let enabled =
            ConfigVal::from(config.get_or_default("enabled", &ConfigVal::Bool(true).into()));

        // without needing to compare `ConfigVal`, instead comparing `bool` directly:
        if enabled != true {
            return Ok(());
        };

        // more logic

        Ok(())
    }
}
```

PilkeySEK:
And now you can also convert from some T to ConfigTree (usually):
`let tree: ConfigTree = true.into();`

This is useful for many things such as the above `get_or_default`:
```rust
let enabled = ConfigVal::from(config.get_or_default("enabled", &true.into()));
```

PilkeySEK:
i hate github

PilkeySEK:
image
merging works locally, idk what github is on.
to merge this PR circumventing github it might be necessary to squash & merge locally then push

vyPal:
I'm pretty sure the merge conflicts are due to Alex recently doing a bit of refactoring, where he also moved all the crates (and the wit folder) to a new `crates/` folder

PilkeySEK:
The merge conflicts were before that change... I will work on resolving the conflicts in a few days

PilkeySEK:
Since the unused dependencies check is failing because of upstream commit https://github.com/Pumpkin-MC/Pumpkin/commit/c7c8d10d4709fcd1066c4a7d58404d2e0196f12c I won't be fixing it and instead waiting for Snowii to fix it and then merge again

crates/pumpkin/src/plugin/mod.rs
```diff
@@ -36,6 +40,7 @@ pub type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
 pub const PLUGIN_API_VERSION: u32 = 2;
 
 const PLUGIN_DIR: &str = "./plugins";
+const PLUGIN_CONFIG_DIR: &str = "./config";
```
vyPal: Wouldn't it be better to have plugin configs stored under `./plugins/config/` rather than just `./config/` since plugin data has moved to `./plugins/data`

QuantumSegfault:
The WIT changes should no longer be in a separate PR. Once you rebase, the to `crates/pumpkin-plugin-wit` should be done directly in this PR.

See https://github.com/Pumpkin-MC/pumpkin-plugin-wit/issues/44

PR #2659, opened 2026-07-31
fix(combat): apply vanilla fist damage and attack speed

MythicSorcerer:
Fixes fist (empty hand) dealing 2 damage instead of 1. The AIR item has no attribute modifiers, so the attack function used base_damage (2.0) + 0.0 = 2.0. Vanilla applies -1.0 base_attack_damage and -2.4 base_attack_speed for the fist, resulting in 1.0 damage and 1.6 attack speed.

This caused entities like cows (10 HP) to die in 5 hits instead of 10.

Also fixes fist attack speed being 4.0 (creative-mode speed) instead of vanilla 1.6.

Additionally fixes the cooldown scaling formula: `base_damage + add_damage * multiplier` incorrectly kept base_damage un-scaled, so weapon damage was reduced by cooldown but fist damage wasn't. Changed to `(base_damage + add_damage) * multiplier` to match vanilla.

### DIFF
--- a/pumpkin/src/entity/player.rs
+++ b/pumpkin/src/entity/player.rs
@@ -940,7 +940,11 @@ impl Player {
 
         {
             let stack = item_stack.lock().await;
-            if let Some(modifiers) = stack.get_data_component::<AttributeModifiersImpl>() {
+            if stack.is_empty() {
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thank @MythicSorcerer :D

PR #2700, opened 2026-08-01
fix: Support nested TextComponent styles in console

PilkeySEK:
## Description
Previously, the following code would print as white in the console instead of blue:
```rust
sender.send_message({
    let text = TextComponent::text("");
    text.add_child({
        let text = {
            let text = TextComponent::text("");
            text.add_child(TextComponent::text("some text"));
            text
        };
        text.color_rgb(RgbColor {
            r: 85,
            g: 85,
            b: 255,
        });
        text
    });
    text
});
```
This PR fixes that by passing the parent style to the child style, which applies styles on top of that style.

## Testing
I confirmed that the text now prints as blue in the console.


Snowiiii:
@PilkeySEK

RoosterBooster007:
Thanks, @PilkeySEK!

PR #2701, opened 2026-08-01
Fix mob target visibility checks

eshankiyer:
## Summary
- prevent mobs from acquiring targets through solid blocks
- retain unseen targets only for the existing 60-tick memory window
- preserve visibility-ignoring revenge behavior and reset unseen memory when sight returns

## Validation
- `cargo fmt -p pumpkin`
- focused target-memory tests (2 passed)
- `cargo test -p pumpkin --lib` (188 passed)
- focused tests on current upstream (2 passed)
- `git diff --check`


Snowiiii:
LGTM

PR #2703, opened 2026-08-01
Enable spider wall climbing

eshankiyer:
## Summary
- update spider and cave-spider climbing state from horizontal collisions
- synchronize the climbing state for current protocol versions
- preserve the existing climbing movement behavior

## Validation
- `cargo fmt -p pumpkin`
- focused spider climbing regression
- `cargo test -p pumpkin --lib` (190 passed)
- codegen tests
- focused test on current upstream
- `git diff --check`


Snowiiii:
@eshankiyer, please fix conflicts, also oh boy these meta data hacks become annoying

PR #2708, opened 2026-08-01
Apply cave spider poison on melee hits

eshankiyer:
## Summary

- run mob-specific effects only after a successful melee hit
- apply Poison I for 7 seconds on Normal and 15 seconds on Hard
- leave Peaceful and Easy attacks without poison

## Validation

- cave-spider difficulty-duration regression test
- `cargo test -p pumpkin --lib` (168 passed)
- `cargo build -p pumpkin`
- `git diff --check`


eshankiyer:

PR #2711, opened 2026-08-01
fix(fishing): deliver caught fish and experience

eshankiyer:
Deliver caught fish and experience when reeling in a bite.


eshankiyer:

PR #2712, opened 2026-08-01
fix(item): consume wind charges

eshankiyer:
Consume wind charges and apply their impact on hit.


eshankiyer:

PR #2713, opened 2026-08-01
fix(net): flush block sequence acknowledgements on every game tick

### LINKED ISSUE #2698: Lectern Book Takes Time To Appear
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

a3b66be/release

### Operating System

Linux Mint 22.3

When attempting to place a book in a lectern, the book takes a few seconds (varying each time) to render on the client.

I was looking for a first contribution to make, and found this bug. I've tested it and it seems like the book registers on the server immediately after clicking on the lectern, but only renders a few seconds after.  

### Steps to Reproduce

1. Boot up a Pumpkin server on the latest release branch
2. Obtain a lectern (You can do this through creative mode: Do /gamemode creative {playername})
3. Grab a "Book and Quill"
4. Place the lectern down
5. Attempt to place the Book and Quill inside the lectern

aprettygoodprogramer:
Server now sends CAcknowledgeBlockChange updates every tick as vanilla does.

Lecterns (and other blocks with similar behavior) now immediately show a visual change on the block, rather than after each keep_alive_interval.

## Description

Before this code change, interactive block states like lecterns took upwards of 15 seconds to visually update because `CAcknowledgeBlockChange` packets were bound directly to the `keep_alive_interval`. While underlying data changes registered immediately on the server, the client paused local visual prediction while waiting for sequence validation. 

I updated the player network packet loop to evaluate and dispatch sequence receipts every game tick, resolving the visual desync and aligning behavior with the vanilla Minecraft implementation.

Closes #2698

## Testing

```bash
cargo test
RUSTFLAGS='-D warnings' cargo clippy -p pumpkin --all-targets
git diff --check
```


Snowiiii:
We already do in World

aprettygoodprogramer:
I grepped through the codebase but couldn’t find what you’re referring to. Could you point me to where that is? I’d really appreciate it. This commit also seemed to fix a few issues I’d been running into.

PR #2716, opened 2026-08-02
fix(projectile): attribute trident damage to the trident

eshankiyer:
Attribute trident hits to the projectile instead of the victim.


eshankiyer:

PR #2721, opened 2026-08-02
fix: claim both parents' love state atomically before breeding

### LINKED ISSUE #2532: Breeding results in 2 babies instead of 1

eshankiyer:
Fixes #2532.

Entities tick as independent concurrent tokio tasks, so both breeding parents' BreedGoal::tick can reach the timer>=60 && dist_sq<9.0 threshold in the same world tick and each independently call breed(), spawning two children for one breeding event. Vanilla's single-threaded tick gives finalizeSpawnChildFromBreeding implicit mutual exclusion that Pumpkin's concurrent model does not provide, which matches what the issue reports: cows always produce 2 babies, pigs often do.

Fix: add MobEntity::try_claim_love, a real compare-and-swap (AtomicI32::fetch_update from a positive value to 0) exposed through EntityBase so it is callable on a mate reference. breed() now claims both parents atomically before spawning the child; a losing concurrent call sees its own or the mate's claim fail and returns without spawning.

Test plan:

### DIFF
--- a/pumpkin/src/entity/ai/goal/breed.rs
+++ b/pumpkin/src/entity/ai/goal/breed.rs
@@ -62,6 +62,13 @@ impl BreedGoal {
 
     async fn breed(mob: &dyn Mob, mate: &dyn EntityBase) {
         let mob_entity = mob.get_mob_entity();
+        if !mob_entity.try_claim_love() {
+            return;
+        }
+        if !mate.try_claim_love() {
+            return;
+        }
+
         let entity = mob.get_entity();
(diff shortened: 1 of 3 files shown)


eshankiyer:

PR #2723, opened 2026-08-02
fix: let primed TNT take explosion knockback

eshankiyer:
TNTEntity overrode is_immune_to_explosion to return true, so the explosion entity loop skipped primed TNT entirely, giving them neither damage nor knockback. In vanilla, PrimedTnt does not override Entity.ignoreExplosion (base returns false), so an explosion pushes primed TNT like any other entity. ServerExplosion.hurtEntities even selects the knockback origin with 'entity instanceof PrimedTnt ? position() : getEyePosition()', which Pumpkin already mirrors in Explosion::damage_entities but could never reach because the immunity check short-circuited first.

Fix: remove the override so primed TNT inherits the default false, restoring explosion-driven TNT momentum (cannons, flying machines, chain detonations).

Test plan:

### DIFF
--- a/pumpkin/src/entity/tnt.rs
+++ b/pumpkin/src/entity/tnt.rs
@@ -122,8 +122,4 @@ impl EntityBase for TNTEntity {
     fn cast_any(&self) -> &dyn std::any::Any {
         self
     }
-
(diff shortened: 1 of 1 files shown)


eshankiyer:

PR #2724, opened 2026-08-02
fix: sync item motion at vanilla's velocity-delta threshold

eshankiyer:
sync_motion_if_dirty compared this-tick velocity against the pre-gravity value with a squared-length threshold of 0.1, ten times looser than vanilla ItemEntity.tick's 0.01 (needsSync = delta.lengthSqr() > 0.01). Thrown and flowing items suppressed position/velocity packets vanilla would send, producing visible client-side stutter.

Test plan:


eshankiyer:

PR #2727, opened 2026-08-02
fix: deal starvation damage on Peaceful when health is high

eshankiyer:
HungerManager::tick hardcoded should_starve to false on Peaceful. Vanilla FoodData's starvation check is a single boolean expression, health > 10.0F || difficulty == HARD || (health > 1.0F && difficulty == NORMAL), which for Peaceful reduces to health > 10.0F, not an unconditional false. Exhaustion cannot normally drop hunger to zero on Peaceful, but a food level forced to zero by NBT or command should still starve down to 10 health as vanilla does.

Test plan:


eshankiyer:

PR #2730, opened 2026-08-02
feat: let experience orbs follow nearby players

eshankiyer:
ExperienceOrbEntity::tick had no equivalent of vanilla's followNearbyPlayer: orbs only moved under gravity and could only be collected by literal bounding-box contact.

Port the vanilla algorithm each tick: find the closest non-spectator player within 8 blocks (ExperienceOrb.java's search radius), aim at their eye height halved above their feet as vanilla's getY() + eyeHeight / 2.0 does, and add an ease-in velocity of normalize(delta) * (1 - dist / 8)^2 * 0.1 toward them.

This uses a fresh closest-player lookup each tick rather than vanilla's cached followingPlayer field (re-searched only when the cached player goes out of range or becomes a spectator); the two are behaviorally equivalent since the lookup already selects the nearest candidate.

Test plan:

### DIFF
--- a/pumpkin/src/entity/experience_orb.rs
+++ b/pumpkin/src/entity/experience_orb.rs
@@ -82,6 +82,21 @@ impl EntityBase for ExperienceOrbEntity {
 
             let mut velo = original_velo;
 
+            let world = entity.world.load();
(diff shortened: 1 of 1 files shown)


eshankiyer:

PR #2731, opened 2026-08-02
fix: wrap time-of-day before the sleep night-window check

eshankiyer:
can_sleep compared the raw cumulative level_time.time_of_day against the night-window constants (12542..23459) without reducing it modulo 24000 first. time_of_day increments unconditionally every tick with no wraparound, so outside of thunderstorms sleeping only ever worked during day 0 and became permanently impossible from day 1 onward on every server.

Every other consumer of the same night-window constants in this codebase (mob/mod.rs, world/mod.rs) already reduces time_of_day mod 24000 first; can_sleep was the one place that did not.

This is high severity: on a real server running past day 1, players can never sleep at all outside of a thunderstorm.

Test plan:


eshankiyer:

PR #2732, opened 2026-08-02
fix: correct ridden-minecart on-rail friction

eshankiyer:
The ridden on-rail slowdown factor was 0.99 instead of vanilla's 0.997 (OldMinecartBehavior.getSlowdownFactor, used since MINECART_IMPROVEMENTS is off by default in 26.2). The unridden branch (0.96) already matched vanilla exactly, isolating this to the ridden case: a ridden minecart lost velocity roughly 3x faster per tick than vanilla, capping travel distance well short.

Test plan:


eshankiyer:

PR #2735, opened 2026-08-02
refactor(tag) remove registry-keys / values

PigTurtleDev:
Removes the `&'static [&'static str]` from tags, but keeps them in doc comments for reference.
Pumpkin doesn't actually use them and they're duplicated in the actual struct definitions for whatever is tagged, so they can always be obtained from there.

This allows `Tag`s to just be `&'static [u16]` slices, which removes some double derefs introduced through the use of `&'static Tag`s.

- Codegen now checks for unchanged tags between versions to avoid duplicates
- Codegen now uses the [`phf_codegen`](docs.rs/phf_codegen) crate to inline the `phf_map!` macro invocations for tags

This reduces the compile time for pumpkin-data by 2.5s on debug builds, a ~6.5% speedup.
0.45s are macro inlining, the rest is removing the `str`s.

### Testing
- `cargo check && cargo clippy` pass
- `cargo fmt --check` passes


PigTurtleDev:
Closing for now because git isn't playing nice with merging tag.rs. If there is interest in this change let me know and I'll reimplement the changes on top of the current tag.rs

Issue #2104, opened 2026-05-07
Player floating and stuck in swim state(WARN: Terible english)


supanika721-beep:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Android 13, Vivo y02t

### Server Software Version/Commit

0.1.0-dev+26.1/idk and /pumpkin didn't show ny commit

### What happened?

1. The player is floating on top of water(see the screenshot)
2. Player is in swim state like there a bubble when you swim but the player is on the ground
3. unknown game packets, logs:
Unknown Game packet: 307
2026-05-07 10:12:20  WARN tokio-rt-worker ThreadId(08) pumpkin::net::bedrock: Bedrock: Received Unknown Game packet: 307

Received Unknown Game packet: 31
2026-05-07 10:12:14  WARN tokio-rt-worker ThreadId(09) pumpkin::net::bedrock: Bedrock: Received Unknown Game packet: 31
2026-05-07 10:12:14  WARN tokio-rt-worker ThreadId(04) pumpkin::net::bedrock: Bedrock: Received Unknown Game packet: 31

Received Unknown Game packet: 152

026-05-07 10:11:55  WARN tokio-rt-worker ThreadId(09) pumpkin::net::bedrock: Bedrock: Received Unknown Game packet: 30

<img width="1600" height="720" alt="Image" src="https://github.com/user-attachments/assets/37629eb5-e052-4256-89ff-acae480e9e0d" />

### To Reproduce

1. player shouldn't float on water, expected: swim
2. the unknown packets might triggered by an action, the 30 and 31 packets may triggered by changing the inventory slot

### Expected behavior

1. player swims when in water or fall to the water
2. no bubble indicator when on the ground

QuantumSegfault:
This is a known issue (and one that's proving illusive). Bedrock support is still in early stages.

A temporary work around is to leave the world and rejoin! After said second join, gravity and air bubbles work as expected.

supanika721-beep:
owh okay, and thank to all the devs and contributors for making my dream come true

GreenedDev:
> owh okay, and thank to all the devs and contributors for making my dream come true

why did you reopen?

mtthidoteu:
I can confirm this is still an issue, after disconnecting/reconnecting, the bedrock player can't see other players. But java players can't see blocks placed by bedrock players (but sees them floating if they jump on one)

[...]

PR #2738, opened 2026-08-03
fix(bedrock): preserve HAS_GRAVITY when a player uses an item

eshankiyer:
Bedrock applies gravity through a per-entity actor flag (HAS_GRAVITY, index 49), not server-side like Java. It is set on spawn, but set_living_flag built its Bedrock SetActorData from an empty EntityMetadata carrying only USING_ITEM. Because the FLAGS field is a full bitfield that replaces the clients whole flag set, using an item wiped HAS_GRAVITY and the client stopped applying gravity, so the player floated (the "anti-grav" bug). The partial update now seeds FLAGS with the entitys accumulated bedrock_flags before toggling USING_ITEM, with a unit test that a seeded flag survives a single-flag update.

Related: #2104 (Bedrock player floating). This addresses the gravity-flag clobber specifically; the swim-state part of that report may be separate.


eshankiyer:

PR #2744, opened 2026-08-03
fix(protocol): actually write the resource-pack prompt message

eshankiyer:
CConfigAddResourcePack::write_packet_data had a TODO where, if prompt_message was Some, it wrote the has-prompt boolean flag as true but then wrote zero bytes for the actual NBT-encoded text component.

This desyncs the packet stream for any server config with a non-empty prompt_message, since login.rs only passes Some(...) for that field when the configured string is non-empty, so the bad branch is reachable in normal use, not dead code.

Fix encodes the component the same way every other packet in this codebase already does (TextComponent::encode()), matching the pattern in e.g. actionbar.rs.

Verified with cargo check / cargo fmt --check / cargo clippy --all-targets -D warnings, all clean.


eshankiyer:

PR #2750, opened 2026-08-03
fix(protocol): add block state and custom model data codecs

Vaspyyy:
## Summary

- implement Java protocol serialization and deserialization for `BlockStateImpl`
- implement Java protocol serialization and deserialization for the current `CustomModelDataImpl` fields
- register both implementations in the shared data-component codec dispatch
- reject negative or excessively large collection lengths while decoding
- add round-trip tests through the public component dispatch

## Why

Both data-component implementations already exist in `pumpkin-data`, but the protocol codec currently falls through to the generic "not yet implemented" error. This prevents item stacks containing block-state or custom-model-data components from being encoded or decoded through the shared component path.

## Validation

- `cargo test`

### DIFF
--- a/crates/pumpkin-protocol/src/codec/data_component.rs
+++ b/crates/pumpkin-protocol/src/codec/data_component.rs
@@ -5,19 +5,19 @@ use crate::ser::{NetworkReadExt, NetworkWriteExt, ReadingError, WritingError};
 use pumpkin_data::Enchantment;
 use pumpkin_data::data_component::DataComponent;
 use pumpkin_data::data_component_impl::{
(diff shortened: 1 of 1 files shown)


Snowiiii:
@Vaspyyy please resolve conflicts

RoosterBooster007:
Hey @Vaspyyy, thanks for the block state + custom model data codec work. However, it looks like both of those landed on master already in commit d1cfc0d18, and that file's been restructured a fair bit since. So, I'll close this out. If you still want the defensive length caps and the round-trip tests in, they'd be a great small PR against the current codec layout.

PR #2751, opened 2026-08-03
fix(recipe): show ghost recipe for missing ingredients

Vaspyyy:
## Summary

- add the clientbound `place_ghost_recipe` packet using the 26.2 `RecipeDisplay` wire format
- share crafting display serialization between recipe-book entries and ghost recipes
- send the full ghost pattern when the server cannot assemble one complete recipe
- avoid partially filling a crafting grid when only some ingredients are available

## Root cause

`handle_place_recipe` attempted to place one of every ingredient for a normal recipe-book click, then moved each ingredient independently. When the inventory did not contain a complete craft, available ingredients were moved into the real grid while missing slots stayed empty. Pumpkin also had no clientbound ghost-recipe packet implementation, so it could not ask the client to render the missing pattern.

## Impact

Recipe-book placement with insufficient ingredients now follows vanilla behavior: real grid contents are returned to the inventory and the complete recipe is displayed as ghost slots. Recipes that can be assembled continue to place real ingredients normally. The handler is shared by built-in and dynamically loaded crafting recipes.

## Validation

- `cargo test`
- `cargo check -p pumpkin`
- Minecraft 26.2 in-game test with a datapack recipe: the complete ghost pattern rendered and the available real ingredients stayed in the inventory


RoosterBooster007:

PR #2739, opened 2026-08-03
fix(redstone): prevent panic when querying vertical redstone power

aprettygoodprogramer:
Fixes #2734

When querying power from a block above the wire, `args.direction` is `BlockDirection::Down`. Calling `.opposite()` results in `BlockDirection::Up`. Because `Up` is vertical, `to_horizontal_facing()` returns `None`. Calling `.unwrap()` on this `None` value caused a panic.

## Description

This PR fixes a crash that would occur if a redstone component was placed above redstone dust. 

The PR specifically changes the `get_weak_redstone_power` function in `redstone_wire.rs`.  I removed the '.unwrap()' in place of a Some() to fix the error. 

## Testing

```bash
cargo fmt --check
cargo test
RUSTFLAGS='-D warnings' cargo clippy -p pumpkin --all-targets
git diff --check
```

Please follow our [Coding Guidelines](https://github.com/Pumpkin-MC/Pumpkin/blob/master/CONTRIBUTING.md#coding-guidelines)

PR #2758, opened 2026-08-04
fix(redstone): prevent panic when querying strong power from above

### LINKED ISSUE #2734: Putting redstone dust below many redstone componets crashes server
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

f474756/debug

### Operating System

Linux Mint 22.3

When you place a redstone component and then place redstone dust below it, the server crashes. 

In vanilla, placing redstone dust below any redstone component will not crash the server. 

### Steps to Reproduce

1. Start up the pumpkin server and obtain redstone dust and a piston
2. Place the piston
3. Break one block below the piston
4. Place redstone dust below the piston

catdum200:
Relates to #2734. Deliberately **not** using a closing keyword: this fixes one of the two
panicking call sites, and #2739 fixes the other. #2734 is only fully resolved once both land.

## Description

`RedstoneWireBlock::get_strong_redstone_power` calls:

```rust
args.direction.opposite().to_horizontal_facing().unwrap()
```

`to_horizontal_facing()` returns `None` for the vertical directions. When the querying block
sits directly above the wire, `args.direction` is `Down`, its opposite is `Up`, and the
`unwrap()` panics, taking down the Tokio worker and the server with it.

The path is reachable without involving weak power at all:

```
get_max_strong_power(pos, dust_power = true)          redstone/mod.rs:102
  -> iterates BlockDirection::all(), which includes Down
  -> get_strong_power(..., side = Down, dust_power = true)   redstone/mod.rs:157
     -> the `!dust_power && block == REDSTONE_WIRE` guard at :165 does NOT fire
     -> get_strong_redstone_power(..., Down)                 redstone/mod.rs:170
        -> Down.opposite() == Up -> to_horizontal_facing() == None -> unwrap() panics
```

So any solid block with redstone dust directly beneath it panics the server as soon as
something asks that block for its power — a note block, repeater, door, or piston next to it
is enough.

### Relationship to #2739

#2739 fixes the byte-identical `unwrap()` in the sibling function
`get_weak_redstone_power` (line 203) and closes #2734. That patch is correct, but it does not
touch `get_strong_redstone_power` (line 219), and the trace above shows that site is reached
independently through `get_max_strong_power`. If #2739 merges alone, #2734 gets closed while
this crash is still live.

These two PRs are complementary, not competing. This one deliberately leaves
`get_weak_redstone_power` untouched so it cannot conflict with #2739 — the diffs are in
different hunks and can merge in either order.

### The change

The condition moves into a free function, `strong_power_toward`, which matches the `Option`
instead of unwrapping it. Extracting it also makes the vertical cases reachable from unit
tests — the trait method itself needs a live `&World`, so it cannot be tested directly.

Behaviour is otherwise unchanged:

| Query direction | Before | After |
| --- | --- | --- |
| `Up` | `wire.power` | `wire.power` |
| Horizontal | depends on connected side | depends on connected side |
| `Down` | **panic** | `0` |

## Testing

Reproduced against a real client (Java, offline mode) on an unpatched build: place redstone
dust, put a solid block directly on top of it, then place a note block beside that solid
block. The server dies immediately:

```
Message: called `Option::unwrap()` on a `None` value
Panic Location: pumpkin\src\block\blocks\redstone\redstone_wire.rs:219:92

--- Panicking Thread ---
Name: tokio-rt-worker
```

Note the line number: **219**, `get_strong_redstone_power` — not the 203 site that #2739
addresses. The identical setup on a patched build does not crash and redstone behaves
normally.

Three unit tests cover the vertical and horizontal cases. They were verified to fail against
the original code before being committed — reinstating the `unwrap()` makes
`queried_from_above_yields_no_strong_power` panic at the unwrap, while the other two continue
to pass, confirming the fix does not change the non-panicking behaviour.

```bash
cargo fmt --check
RUSTFLAGS='-Dwarnings' cargo clippy --all-targets --all-features
RUSTFLAGS='-Dwarnings' cargo clippy --release --all-targets --all-features
cargo test --workspace   # 484 passed, 0 failed
```

### DIFF
--- a/pumpkin/src/block/blocks/redstone/redstone_wire.rs
+++ b/pumpkin/src/block/blocks/redstone/redstone_wire.rs
@@ -215,13 +215,7 @@ impl BlockBehaviour for RedstoneWireBlock {
     ) -> BlockFuture<'a, u8> {
         Box::pin(async move {
             let wire = RedstoneWireProperties::from_state_id(args.state.id, args.block);
-            if args.direction == BlockDirection::Up
-                || wire.is_side_connected(args.direction.opposite().to_horizontal_facing().unwrap())
-            {
-                wire.power
-            } else {
-                0
-            }
+            strong_power_toward(wire, args.direction)
         })
     }
 
@@ -235,6 +229,24 @@ impl BlockBehaviour for RedstoneWireBlock {
         Box::pin(async move {
             update_wire_neighbors(args.world, args.position).await;
         })
+    }
+}
+
+/// Strong power the wire emits to the block that queried it, where `direction`
+/// points from the querying block toward the wire.
+///
+/// `to_horizontal_facing` returns `None` for the vertical directions, so the
+/// result has to be matched instead of unwrapped: a block sitting directly above
+/// the wire queries it with `BlockDirection::Down`, whose opposite is
+/// `BlockDirection::Up`.
+fn strong_power_toward(wire: RedstoneWireProperties, direction: BlockDirection) -> u8 {
+    if direction == BlockDirection::Up {
+        return wire.power;
+    }
+
+    match direction.opposite().to_horizontal_facing() {
+        Some(facing) if wire.is_side_connected(facing) => wire.power,
+        _ => 0,
     }
 }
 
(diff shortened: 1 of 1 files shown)


catdum200:
Closing this one - the panic is already gone from master. 49360e53 ("chore: clippy enforce `unwrap_used`, `expect_used` and `panic`") replaced `.to_horizontal_facing().unwrap()` with `is_some_and(...)` in both `get_weak_redstone_power` and `get_strong_redstone_power`, so there is nothing left here to fix.

PR #2759, opened 2026-08-04
feat: emit lectern redstone signals and add the lectern screen

Mcxiaocaibug:
Part of #1402 (Lectern).

## What

- Turning a page emits the vanilla redstone pulse: `powered` for 2 game ticks, weak power 15 on all sides, strong power 15 to the block below, plus the page-turn world event. Neighbors below are updated when the pulse starts/ends and when a powered lectern is broken (`on_state_replaced`).
- Comparator output uses the vanilla formula `floor(page / (pageCount - 1) * 14) + 1` when a book is present (single-page books read 1, no book reads 0).
- Adds the lectern screen (`WindowType::Lectern`): previous/next/jump-to-page buttons sync the current page through container property 0, and the take-book button hands the book to the player (or drops it) and clears `has_book`. Every page turn fires the redstone pulse.
- Placing a book uses the `minecraft:lectern_books` item tag (no hardcoded item list) and plays the vanilla book-put sound; a freshly placed book always opens on its first page.

## Implementation notes

- The screen handler lives in `pumpkin-inventory`, which cannot depend on `World`, so it talks to the block through a small `LecternController` trait. The block-side implementation clamps and stores the page, marks the block entity dirty, and fires the pulse.
- Page count is derived from the book's `WrittenBookContentImpl`/`WritableBookContentImpl` data components; the stored page is clamped on load.

Verified with `cargo clippy` under the workspace lint set (`all`/`nursery`/`pedantic`/`cargo` denied); the redstone and comparator behavior mirrors vanilla's `LecternBlock`/`LecternScreenHandler`.

### DIFF
--- a/pumpkin-inventory/src/lectern_screen_handler.rs
+++ b/pumpkin-inventory/src/lectern_screen_handler.rs
@@ -1,0 +1,143 @@
+use std::any::Any;
+use std::sync::Arc;
+
+use crate::screen_handler::{
+    InventoryPlayer, ItemStackFuture, ScreenHandler, ScreenHandlerBehaviour, ScreenHandlerFuture,
+    ScreenProperty, offer_or_drop_stack,
+};
+use crate::slot::NormalSlot;
+
+use pumpkin_data::item_stack::ItemStack;
+use pumpkin_data::screen::WindowType;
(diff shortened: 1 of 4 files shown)


Snowiiii:
LGTM

PR #2762, opened 2026-08-04
feat: add ranged bow combat for skeletons

q2qr:
## Summary

Adds proper ranged combat for skeletons equipped with bows.

Previously, skeletons would run up to players and attack in melee even when holding a bow. They now move into range, aim at their target, and fire arrows with cooldowns and accuracy based on the current difficulty.

Wither skeletons still use their existing melee behavior.

### Skeleton ranged AI

* Added a dedicated `BowAttackGoal` for skeletons.
* Replaced the melee attack goal for bow-wielding skeleton variants.
* Kept the existing melee behavior for wither skeletons.
* Skeletons now look toward their target while attacking.
* Skeletons move toward targets that are outside their 15-block firing range.
* They stop moving once the target is within range.
* Dead players and players in creative or spectator mode are ignored.

### Bow timing

* Added a 20-tick draw time before the first shot.
* Added difficulty-based cooldowns between shots:

  * Easy: 40 ticks
  * Normal: 40 ticks
  * Hard: 20 ticks
* Fixed the ranged attack goal restarting every tick and skipping its cooldown.

### Arrow behavior

* Skeletons now fire real arrow entities with the skeleton set as the owner.
* Arrows are aimed toward the target with trajectory compensation.
* The skeleton shooting sound plays when an arrow is fired.
* Arrows fired by skeletons cannot be picked up.
* Arrow accuracy now changes based on difficulty:

  * Easy: 10.0 divergence
  * Normal: 6.0 divergence
  * Hard: 2.0 divergence

### Collision and damage

* Players are now included when checking for arrow collisions.
* Fixed skeleton arrows passing directly through players.
* Arrow damage now uses the full damage context:

  * The arrow is recorded as the direct damage source.
  * The skeleton is recorded as the attacker.
  * The exact impact position is passed to the damage system.
* Existing critical-hit, flame, punch, potion, spectral-arrow, and piercing behavior is unchanged.

## Testing

* Added focused tests for difficulty-based bow timing.
* Added focused tests for difficulty-based arrow accuracy.
* Verified the skeleton bow timing tests pass.
* Ran `cargo fmt`.
* Ran Clippy with warnings treated as errors.
* Verified the patch with `git diff --check`.

## Breaking changes

None.


Snowiiii:
@q2qr please fix conflicts

RoosterBooster007:
Hey, thanks for this, @q2qr. The skeleton bow combat is a nice one to knock out, and the cooldown/accuracy tuning is solid. Looks like it already landed in #2818 a couple hours after your last push, so this is superseded. I'm closing this out, but if the worldgen and arrow bits are still worth doing, I'm happy to help you spin those into their own PRs.

PR #2768, opened 2026-08-04
refactor: enchant command code refactor

TNXWasTaken:
## Summary

Refactors the enchant command and removes the TODO in the `itemstack.enchant` method since this has been fixed

## Description

I was looking at updating the client when the enchant command was called since the client wasn't receiving that their item was now enchanted. This was fixed in commit 938a627 (fix(item): prevent stacking items with different components (#2624)), which adds checking for itemstacks that have different data components. While I was reading the enchant command code, I also refactored it to remove the `only_one` var. Once I saw that the enchant method was patched, I removed the TODO message in the item_stack/mod.rs since any item that is enchanted will now sync with the client

## Functonality

This should show no functional changes since this is only designed to be a refactor

## Testing

When I was looking at adding tests, I tried just creating fake values to feed into execute, but going down the path of creating new fake values just ended up with effectively needing to run a whole server, and I don't have the capabilities to add that to pumpkin and did not see another option for creating a dummy server

### DIFF
--- a/pumpkin-data/src/item_stack/mod.rs
+++ b/pumpkin-data/src/item_stack/mod.rs
@@ -547,7 +547,6 @@ impl ItemStack {
     }
 
     pub fn enchant(&mut self, enchantment: &'static Enchantment, level: i32) {
-        // TODO itemstack may not send update packet to client
         if level <= 0 {
             return;
         }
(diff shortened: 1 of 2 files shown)


Snowiiii:
LGTM, Thank you @TNXWasTaken :D

PR #2748, opened 2026-08-03
feat: Added central registry

stijn-boer:
## Description

Adds the initial foundation for a central registry system

This is a first step toward #731. It introduces the core registry needed to represent different Minecraft registries through a common root registry while retaining their concrete entry types

## What changed?

* Added `DataKey<T>` for identifying typed registry values

  * Supports nested key paths
  * Uses a type marker so keys for different value types cannot be mixed accidentally
* Added core registry
* Added support for unsized registry entry types, including trait objects such as `dyn Block`
* Added a root registry that can contain registries with different entry types
* Added checked downcasting when retrieving a typed registry from the root registry

  * Incorrect key types return an error instead of panicking
* Added per-Minecraft-version numeric ID mappings
* Added error types for

  * duplicate registrations
  * missing registries or entries
  * type mismatches

## Why is this necessary?

Pumpkin currently stores different categories of game data through separate, unrelated mechanisms. This makes it difficult to:

* expose registries through a consistent API
* resolve references between registry entries
* load vanilla and custom data packs

The root registry provides a common location for all registries while typed keys and checked downcasting preserve type safety

## Impact

This PR primarily introduces infrastructure and does not yet replace Pumpkin's existing registries

It establishes the API required for later work on:

* registering Pumpkin's built-in registries
* loading the vanilla data pack
* loading user-provided data packs
* migrating existing generated assets to the unified registry system

## Known limitations

* Existing Pumpkin data sources have not yet been migrated.
* This is only the registry foundation and does not fully resolve #731.

## Related issue

Progresses #731.

Issue #731, opened 2025-04-12
Create unified registry system to load data packs, including the vanilla data pack


EpixScripts:
In Minecraft, a **registry** is a "table describing static, gameplay-related objects of some kind," such as the types of entities, blocks, or biomes. ([source](https://minecraft.wiki/w/Java_Edition_protocol/Packets))

Most registries are managed through a unified registry system. Both registries and entries within registries are identified by [resource locations](https://minecraft.wiki/w/Resource_location). The root registry—the registry of all registries—is identified by `minecraft:root`.

Each registry belongs to one of two groups:
- **Static** - registries that are **not** modifiable by data packs. These are fixed for a specific game version.
  - Initialized in `net.minecraft.core.registries.BuiltInRegistries`
  - <details>
    <summary>List of static registries as of 1.21.5</summary>

    - `game_event`
    - `sound_event`
    - `fluid`
    - `mob_effect`
    - `block`
    - `entity_type`
    - `item`
    - `potion`
    - `particle_type`
    - `block_entity_type`
    - `custom_stat`
    - `chunk_status`
    - `rule_test`
    - `rule_block_entity_modifier`
    - `pos_rule_test`
    - `menu`
    - `recipe_type`
    - `recipe_serializer`
    - `attribute`
    - `position_source_type`
    - `command_argument_type`
    - `stat_type`
    - `villager_type`
    - `villager_profession`
    - `point_of_interest_type`
    - `memory_module_type`
    - `sensor_type`
    - `schedule`
    - `activity`
    - `loot_pool_entry_type`
    - `loot_function_type`
    - `loot_condition_type`
    - `loot_number_provider_type`
    - `loot_nbt_provider_type`
    - `loot_score_provider_type`
    - `float_provider_type`
    - `int_provider_type`
    - `height_provider_type`
    - `block_predicate_type`
    - `worldgen/carver`
    - `worldgen/feature`
    - `worldgen/structure_placement`
    - `worldgen/structure_piece`
    - `worldgen/structure_type`
    - `worldgen/placement_modifier_type`
    - `worldgen/block_state_provider_type`
    - `worldgen/foliage_placer_type`
    - `worldgen/trunk_placer_type`
    - `worldgen/root_placer_type`
    - `worldgen/tree_decorator_type`
    - `worldgen/feature_size_type`
    - `worldgen/biome_source`
    - `worldgen/chunk_generator`
    - `worldgen/material_condition`
    - `worldgen/material_rule`
[...]

PR #2772, opened 2026-08-04
feat: Data migration to central registry

stijn-boer:
## Description

**THIS IS STILL WORK IN PROGRESS**

Migrates Pumpkin’s existing game data to the central registry system introduced in #2748.

This continues the work toward #731 by replacing the separate lookup mechanisms used by existing generated data with a common root registry.

## What changed?

* Added initialization for Pumpkin’s built-in registries
* Registered the migrated registries in the central root registry
* Migrated existing generated registry data to `Registry<T>`
* Migrated numeric protocol ID lookups to the per-version registry mappings
* Updated call sites to resolve entries through the central registry

## Why is this necessary?

Pumpkin currently exposes much of its generated game data through separate static tables and registry-specific lookup functions.

Although #2748 introduces a shared registry abstraction, the existing data continues to bypass it. Maintaining both systems would duplicate functionality and make data-pack support more difficult.

Migrating the existing data provides a single API for:

* resolving registry entries by identifier
* resolving protocol IDs for a specific Minecraft version
* referring to entries through typed keys
* accessing registries through the root registry
* allow for the addition of data from vanilla and user-provided data packs

## Implementation details

The migrated built-in registries are initialized during startup and inserted into the root registry.

Generated numeric IDs are stored as version mappings on their corresponding registries. Code that previously accessed generated ID tables directly now requests the mapping for the relevant `MinecraftVersion`.

Entry lookups remain type checked. Requesting a registry using an incompatible `DataKey<T>` returns a type mismatch error rather than panicking or producing an incorrectly typed value.

## Impact

This changes the internal source of existing registry data, but is not intended to alter gameplay or protocol behavior.

The main effects are:

* registry data is accessed through one consistent API
* protocol ID mappings are owned by their corresponding registries
* registry references can use typed keys instead of unrelated identifiers or indexes
* future data-pack loading can populate the same registry structures used by built-in data

## Known limitations

* Vanilla data packs are not loaded yet
* User-provided data packs are not loaded yet
* Registry reloading and data-pack precedence are outside the scope of this PR

## Dependencies

This PR is based on and requires #2748.

## Related issue

Progresses #731.


aure31:
just letting you know that this pull request would never be merged because it’s way too big, so you’ll need to split it into several much smaller pull requests. Also, it might be better to wait for #2748 to be merged before opening other pull requests that depend on it.

RoosterBooster007:
There are a lot of conflicts here... and yeah this is huge. Also, I'm closing due to: https://github.com/Pumpkin-MC/Pumpkin/pull/2748#issuecomment-5497659696 and inactivity. Please sort that comment out with @Snowiiii before continuing.

PR #2773, opened 2026-08-04
fix(worldgen): BlockDirection random sampling parity with vanilla

sugoijan:
## Description

`BlockDirection::random` and `random_horizontal` had two bugs.

**Wrong bound.** Both used `next_bounded_i32(len - 1)`, so the last entry was never drawn - `East` could not come up. Vanilla uses `values[random.nextInt(values.length)]` over the whole array.

**Wrong order.** `random_horizontal` read `horizontal()`, which is `[North, South, West, East]`. Vanilla `Direction.Plane.HORIZONTAL` is `[North, East, South, West]`. So even valid draws gave the wrong direction.

This adds `horizontal_worldgen()` with the vanilla order. `horizontal()` stays as is, because most callers do not care about order.

## Why it matters

Pumpkin tries to generate the same world as vanilla for the same seed. These features did not. They drew the right random numbers but applied them to the wrong direction, so coral, dripstone, mangrove roots and five tree shapes grew the wrong way.

The two gameplay bugs show up without a seed. A gourd stem could never grow its fruit to the east, one case in four that never happened. Fire picked the portal's Z axis twice as often as X, because two of the three directions it could reach were north and south.

## Impact

These worldgen features now match vanilla for the same seed. Fire portal lighting and gourd stem growth now pick evenly from four directions instead of three; both seed from `rand::rng()`, so that is a distribution fix, not seed parity.

Structures are unaffected. They use `pumpkin_util::BlockDirection::get_random_horizontal_direction`, which was already correct.

Coral tree and claw still miss vanilla's shuffle, which spends extra draws. They are not bit-exact yet, and the `TODO`s say so.

## Testing

Draw sequences pinned for seed `12345` against `java.util.Random`, the same through `Xoroshiro` (what production uses), plus reachability tests for all six and all four directions. The reachability test fails on the old bound.

### DIFF
--- a/crates/pumpkin-data/src/block_direction.rs
+++ b/crates/pumpkin-data/src/block_direction.rs
@@ -81,11 +81,20 @@ impl BlockDirection {
     }
 
     pub fn random(random: &mut RandomGenerator) -> Self {
-        Self::all()[random.next_bounded_i32(Self::all().len() as i32 - 1) as usize]
+        // Vanilla `Direction.getRandom` = `values[nextInt(values.length)]` over all six
+        // directions, in declaration order (DOWN, UP, NORTH, SOUTH, WEST, EAST).
+        Self::all()[random.next_bounded_i32(Self::all().len() as i32) as usize]
     }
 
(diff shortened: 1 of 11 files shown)


Snowiiii:
Hey @sugoijan, Sorry it took so long, LGTM

PR #2774, opened 2026-08-05
fix(protocol): accept BungeeCord-forwarded handshake addresses

### LINKED ISSUE #1293: Bungeecord fails to handshake when ip_forward is true

### Operating System

Docker with Artix Linux host

Pumpkin 0.1.0-dev+1.21.11

### What happened?

Connection fails and the following gets written to Pumpkin's logs:

```
2026-01-14 05:28:55 [ERROR] Failed to read incoming packet with id 0: too large: string
2026-01-14 05:28:55 [ERROR] Failed to read incoming packet with id 0: incomplete: failed to fill whole buffer
```

### To Reproduce

In Bungeecord's config.yml, set `ip_forward: true`

### Expected behavior

Handshake succeeds and the player joins the server

catdum200:
`SHandShake::read` bounds `server_address` at 255 characters, the limit a vanilla
client observes. `bungeecord_login` expects that same field to carry the host, the
client's IP, its UUID and its signed profile properties separated by NUL bytes —
exactly as its own doc comment says:

```rust
/// It utilizes the `server_address` received in the `SHandShake` packet,
/// which may contain optional data about the client:
///
/// 1. IP address (if `ip_forward` is enabled on the `BungeeCord` server)
/// 2. UUID (if `ip_forward` is enabled on the `BungeeCord` server)
/// 3. Game profile properties (if `ip_forward` and `online_mode` are enabled ...)
```

With `ip_forward` enabled the signed properties alone run well past a thousand
characters, so the read fails before `handle_handshake` ever runs: the connection
state never advances, `server_address` stays empty, and the proxy login path is
unreachable. That is the `too large: string` in the log in #1293 — it is
`ReadingError::TooLarge`'s `Display`.

The same struct already *writes* this field with `write_string`, which bounds it at
`i16::MAX`. The reader was stricter than the writer, and rejected the payload the
proxy layer is written to parse. This reads it with that same bound, as Spigot does.

## Evidence

Two tests, each failing before the change and passing after.

**The reader** — `reads_bungeecord_forwarded_server_address`, in
`pumpkin-protocol`, feeds a handshake carrying a realistic forwarded address:

```
TooLarge("string has too many bytes (1226 > 765)")
```

**The whole path** — `logs_in_from_a_handshake_decoded_off_the_wire`, in `pumpkin`,
encodes a handshake the way BungeeCord puts it on the wire, decodes it with the real
packet reader, and hands the resulting address to `bungeecord_login`, asserting the
forwarded IP, the forwarded UUID and the signed skin all survive. Reverting only the
bound makes it fail inside `SHandShake::read`, so `bungeecord_login` is never
reached at all.

Two more act as controls, and pass either way:

- `reads_plain_server_address` — ordinary hostnames still parse.
- `rejects_oversized_server_address` — 32768 characters is still refused, so the
  bound moved rather than disappeared.

## Notes

- No new DoS surface: `MAX_PACKET_SIZE` (2 MiB) is enforced in the packet decoder
  before the payload reaches `SHandShake::read`, and the wider bound admits at most
  ~98 KB.
- #2559 (BungeeGuard) appends a fifth NUL-delimited token to this same field, and
  its tests build that string in-process rather than reading it off the wire — so it
  needs this fix to work over a real connection.

Closes #1293

### DIFF
--- a/pumpkin-protocol/src/java/server/handshake/mod.rs
+++ b/pumpkin-protocol/src/java/server/handshake/mod.rs
@@ -26,7 +26,12 @@ impl<'a> ServerPacket<'a> for SHandShake {
     fn read(read: &mut &'a [u8], _version: &JavaMinecraftVersion) -> Result<Self, ReadingError> {
         Ok(Self {
             protocol_version: read.get_var_int()?,
-            server_address: read.get_str_bounded(255)?,
+            // A vanilla client never sends more than 255 characters here, but
+            // `BungeeCord`-style IP forwarding appends the client's IP, its UUID
+            // and its signed profile properties to this field, which easily
+            // exceeds that. Read it with the same bound `write_string` already
+            // writes it with, as Spigot does.
+            server_address: read.get_str_bounded(i16::MAX as usize)?,
             server_port: read.get_u16_be()?,
             next_state: read
                 .get_var_int()?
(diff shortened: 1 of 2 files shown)


Snowiiii:
LGTM

Snowiiii:
Reference: https://github.com/PaperMC/Paper/blob/0a993450f129c4942c2a9ed45ba047412b4667cf/paper-server/patches/sources/net/minecraft/network/protocol/handshake/ClientIntentionPacket.java.patch

PR #2782, opened 2026-08-05
feat(pumpkin): add /debug start and stop commands

GabrielDieterly:
Implements the vanilla `/debug start` and `/debug stop` commands, which are part of the unchecked `/debug` entry in the command tracking issue #15.

## What

- **`/debug start`** — starts one server-wide tick profiling session and fails with `commands.debug.alreadyRunning` if profiling is already active.
- **`/debug stop`** — stops the active session and reports its elapsed seconds, processed ticks, and average TPS. It fails with `commands.debug.notRunning` when no session is active.
- The command result for `/debug stop` is the floored average TPS, matching Vanilla.

`/debug function` is intentionally out of scope because Pumpkin does not yet implement the `/function` command or the data-pack function runtime it requires.

## How

- Added a server-wide `DebugProfiler` that records the start `Instant` and server tick count behind a mutex, preserving the original session when duplicate starts are attempted.
- Tick deltas use wrapping subtraction so profiling remains correct when the server tick counter wraps.
- Registered `minecraft:command.debug` at OP level 3 using the current-style command builder and existing Java/Bedrock translation keys.
- Added one focused lifecycle test covering inactive stop, start, duplicate start, elapsed-time/TPS calculation, and duplicate stop.

## Testing

- `cargo check --workspace --all-targets`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test -p pumpkin --lib` — 181 passed, 0 failed
- Verified on a running Java Edition 26.2 server:
  - `/debug start` -> "Started tick profiling"
  - `/debug start` again -> "The tick profiler is already started"
  - `/debug stop` -> "Stopped tick profiling after 11.10 second(s) and 222 tick(s) (20.00 tick(s) per second)"
  - `/debug stop` again -> "The tick profiler hasn't started"


aure31 (approved):
the implementation is good for me

Snowiiii:
LGTM, Thank you @GabrielDieterly :D

GabrielDieterly:
> LGTM, Thank you @GabrielDieterly :D 

Happy to contribute 🫡

PR #2783, opened 2026-08-05
perf: cache surface noise threshold samplers

nobuildersnotools:
## Description

Cache surface-rule noise threshold samplers in the material-rule context, avoiding repeated sampler construction during generation. This was previously marked as todo in the code touched by this PR

## Testing

Benches were ran for surface_generation, overworld surface (which is a separate bench), and full_chunk_generation. Results:
1. surface_generation: 2.134ms -> 1.492ms
2. overworld surface: 10.006ms -> 9.175ms
3. full_chunk_generation: 23.9ms -> 23.3ms

## Summary by CodeRabbit

* **Performance**
  * Improved noise threshold processing by reusing cached samplers, reducing repeated setup during world generation.


JamieLittle16:
Is `DoublePerlinNoiseParameters::id` guaranteed to uniquely identify the entire parameter set, including `lo`, `hi`, octave and amplitude data?

`get_noise_sampler_for_id` derives the random source from `lo`/`hi` and constructs the sampler from the complete parameter value, while this cache is keyed only by `id`. If that uniqueness is a generated-data invariant, could we document it here or add a small regression assertion? Otherwise the cache key may need to include the complete parameter identity.

Other than that looks good to me

RoosterBooster007:

RoosterBooster007 (approved):

RoosterBooster007:
Thanks, @nobuildersnotools!

PR #2784, opened 2026-08-05
perf: read nbt byte arrays in bulk

nobuildersnotools:
## Description

Reads NBT byte arrays through the bulk reader method instead of decoding each byte individually, which was done previously

## Testing

On the chunk_nbt_deserialization benchmark, time spent decreases from 106.4 µs to 88.3 µs.


RoosterBooster007:
Please fix conflicts.

RoosterBooster007:
Thanks, @nobuildersnotools!

PR #2785, opened 2026-08-05
perf: write nbt byte arrays in bulk

nobuildersnotools:
## Description

Replaces per-byte nbt serialization with a single bulk write. This preserves the encoded output and applies to both Java and Bedrock nbt

## Testing

Benchmarks for a 64 KiB byte array:

Java: 19.32 µs ->3.01 µs
Bedrock: 18.77 µs -> 2.97 µs

### DIFF
--- a/crates/pumpkin-nbt/src/tag.rs
+++ b/crates/pumpkin-nbt/src/tag.rs
@@ -141,9 +141,11 @@ impl NbtTag {
                 }
 
                 w.write_i32(len as i32)?;
-                for int in byte_array {
-                    w.write_i8(int)?;
-                }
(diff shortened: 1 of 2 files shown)


JamieLittle16:
Non-blocking suggestion: the `i8` to `u8` reinterpretation looks sound to me. Since this introduces an `unsafe` block, would it be worth adding a small byte-for-byte regression test containing values such as `-128`, `-1`, `0` and `127`, ideally for both Java and Bedrock NBT?

This is mainly to document and protect the safety assumptions rather than because I suspect the implementation is incorrect.

RoosterBooster007:
Please fix conflicts.

RoosterBooster007:
Thanks, @nobuildersnotools!

PR #2788, opened 2026-08-06
feat(plugin-api): Initial inter-plugin communication

QuantumSegfault:
Adds support for basic one-to-one synchronous inter-plugin communication. Plugins may send a "message" (`list<u8>`; untyped byte packet) to another plugin, which the target plugin acts on immediately and returns a response.

This PR also adds the plugin name to `PluginHostState` to help simplify things (avoid threading `context` through). In the future, the use of names might be replaced with a more appropriate, restricted UID, in which case this addition would be only the plugin UID, not the display name.

## Testing

Testing was done with a simple pair of Rust plugins.

Plugin A declared a dependency (in the metadata) on plugin B to ensure favorable load order. During `on_load`, plugin A sends a message to plugin B, which reverses the payload and returns it as the response. The message send was successful and the returned payload was as expected.

The special-casing of not allowing plugins to send messages to themselves was also checked.

In short, it works as expected.

### DIFF
--- a/crates/pumpkin-plugin-api/src/lib.rs
+++ b/crates/pumpkin-plugin-api/src/lib.rs
@@ -99,8 +99,8 @@ pub use wit::pumpkin::plugin::{
     data_components, entity,
     entity_types::EntityType,
     event::{self as events_wit, EventType},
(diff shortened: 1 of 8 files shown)


vyPal (approved):
LGTM, I'll update the wit commit reference to latest master and merge

PR #2798, opened 2026-08-06
fix: flatten hover event nbt fields

xRookieFight:


Snowiiii:
LGTM, Thanks @xRookieFight :D

PR #2799, opened 2026-08-06
feat: add config to enable/disable commands and set their permission level

darkstarworks:
## What

Adds a `[commands.overrides]` section to the config so a server owner can turn individual commands off, or change the permission level needed to use them, without touching code. This is the last unchecked item on the command tracking issue (#15).

Each entry is keyed by the command name:

```toml
# Only server owners can change gamemodes
[commands.overrides.gamemode]
permission_level = 4

# Turn the /tp command off completely
[commands.overrides.tp]
enabled = false
```

- `enabled = false` hides the command everywhere: it cannot be run, does not show up in `/help`, and is left out of tab-completion on both Java and Bedrock. Players who try it get the normal "unknown command" reply, exactly as if it was never registered. Disabling any alias (like `tp`) takes the whole command down, including its main name and other aliases.
- `permission_level` sets who is allowed to use the command: 0 = everyone, 2 = normal ops, 3 = admins, 4 = owner. It works by rewriting the default of the command's permission node, so it applies to both the newer node-based commands and the legacy ones.

Both settings are checked on startup. An override for a command that does not exist logs a warning so typos are caught, and the changes that were applied are confirmed in the console in plain language.

## How

Pumpkin has two command dispatchers (a newer node tree and a legacy fallback) sitting behind one wrapper. The changes hook in at that wrapper so both layers behave the same:

- Disabled command names are kept in a set on the wrapper dispatcher and filtered out of execution, `/help` listings, and the client command graph. Legacy commands are also unregistered outright so they drop from every legacy path. Node-based aliases (which are modelled as redirecting root nodes) are resolved and flagged too.
- The disable gate lives in both `handle_command` (the normal entry point) and `execute_input` (the lower entry point used by `/execute run <command>`), so a disabled command cannot be reached that way either.
- Permission overrides rewrite the permission node's default in the registry after the commands are registered. Command requirements look their permission up at execution time, so no per-command wiring is needed and both dispatcher styles are covered.

## Known limitations

- Overrides apply to the built-in commands registered at startup. Plugin commands load later and are not covered by this.
- Disabling a node-based command by one of its aliases (only `/help` has any today: `h` and `?`) flags that alias but does not cascade back to the main name. Disabling by the main name works in every case, which is what the config comments recommend.
- Pumpkin writes `config.toml` from defaults without comments, so the option is not pre-filled in a fresh file. It takes effect once the `[commands.overrides.*]` table is added, the same as the other config options.

## Verification

`cargo test` and `cargo clippy --all-targets -- -D warnings` are clean on Windows with stable Rust. Added nine tests covering legacy and node-based disable, the alias cascades, unknown-command handling, case-insensitive names, permission rewrites for both dispatcher styles, level 0 meaning everyone, and the `/execute run` path.

### DIFF
--- a/crates/pumpkin-config/src/commands.rs
+++ b/crates/pumpkin-config/src/commands.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use pumpkin_util::PermissionLvl;
 use serde::{Deserialize, Serialize};
 
@@ -19,6 +21,26 @@ pub struct CommandsConfig {
     pub broadcast_console_to_ops: bool,
     /// The `op` permission level of everyone that is not in the `ops` file.
     pub default_op_level: PermissionLvl,
+    /// Per-command settings, so you can turn individual commands off or change
+    /// who is allowed to use them.
+    ///
+    /// Each entry is named after the command it applies to (without the leading
+    /// slash), for example `gamemode` or `tp`. You only need to list the
+    /// commands you actually want to change; everything you leave out keeps its
+    /// normal behaviour.
+    ///
+    /// Example:
+    ///
+    /// ```toml
+    /// # Only server owners may change gamemodes
+    /// [commands.overrides.gamemode]
+    /// permission_level = 4
+    ///
+    /// # Turn the /tp command off completely
+    /// [commands.overrides.tp]
+    /// enabled = false
+    /// ```
+    pub overrides: HashMap<String, CommandOverride>,
 }
 
 impl Default for CommandsConfig {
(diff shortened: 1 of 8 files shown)


Q2297045667:
LGTM

Snowiiii:
I actually like that, @darkstarworks could you fix conflict

darkstarworks:
Conflict is fixed, rebased onto current master.

The conflict was small: `fetchprofile::register` landed in the same spot as the `apply_command_overrides` call, so both are kept, with the overrides applied after every command is registered. I also updated the new `pumpkin_command_aliases` test to pass the commands config, since `default_dispatcher` takes it as a third argument now.

~~Heads up that `master` does not currently build~~ - this has been resolved

image

Almost done

darkstarworks:
waiting for your go-ahead

PR #2802, opened 2026-08-07
fix(world): drive neighbor updates iteratively to stop stack overflows

### LINKED ISSUE #2016: /filling a large cube of saplings crashes the server

### Operating System

nixos 25.11

0.1.0-dev+26.1 5e616a185e86a9c9920b6a93068bcf6b5f152657

### What happened?

when filling an area larger than roughly 576 blocks the server crashes with a core dump
Image

 

### To Reproduce

run `/fill ~ ~ ~ ~15 ~15 ~15 minecraft:acacia_sapling`

### Expected behavior

i would expect the saplings to place and then break

mcbabel:
## Description

`/fill ~ ~ ~ ~15 ~15 ~15 minecraft:acacia_sapling` aborts the server with a stack overflow
and a core dump.

The block update cascade was a mutually recursive set of async functions with no depth bound:

```
World::replace_with_state_for_neighbor_update
  -> World::break_block
    -> World::set_block_state
      -> BlockRegistry::update_neighbors
        -> World::replace_with_state_for_neighbor_update   (Box::pin)
```

`Box::pin` only made that cycle representable as a type; every level still consumed a poll
frame for four large async futures. A core dump of the reported command shows that exact
four-frame cycle repeated ~230 times: a sapling's random tick shape-updates its neighbours,
every unsupported sapling replaces itself with air, and breaking it shape-updates its own
neighbours, depth-first through the whole 4096-block blob until the 2 MiB worker stack is
gone. Inert blocks such as stone are unaffected because they never replace themselves.
Pressure plates reproduce it through `on_neighbor_update` handlers that break their own block.

`World`'s three cascade entry points (`update_neighbors`, `update_neighbor` and
`replace_with_state_for_neighbor_update`) now submit work items to a per-task update stack
instead of calling into each other. The first submission in a task drains the stack in a
loop, so stack depth stays constant regardless of cascade size. A `MAX_CHAINED_UPDATES`
budget matching vanilla's `max-chained-neighbor-updates` bounds the queue.

Ordering follows vanilla's `CollectingNeighborUpdater`: the stack is LIFO and items staged by
the running item are pushed on top in submission order, reproducing the depth-first order of
the old recursion. What does change is when a handler's own code runs relative to the updates
it submits — a handler now runs to completion before those updates execute. That is vanilla's
semantics, but not the old Pumpkin behaviour.

`RedstoneWireTurbo` is where that is observable: its walk now reads neighbour powers that
queued updates recompute after the walk rather than during it. The updates still propagate,
but intermediate states and tick timing differ, and I have not verified that path in game.

Two bounds I left out of scope: vanilla's `recursionLeft = 512` threaded through
`Level.setBlock` is still missing, so a very large blob unravels further than vanilla would;
and `MAX_CHAINED_UPDATES` is hard-coded rather than exposed as a server property.

## Testing

Reproduced the crash on master, then confirmed the fix on a running server: repeated
`/fill ~ ~ ~ ~15 ~15 ~15` with `minecraft:acacia_sapling` and
`minecraft:spruce_pressure_plate` over 4096 blocks. The server stays up and the blocks place
and break as they do in singleplayer.

Three new unit tests cover submission ordering, updates staged while a drain is running, and
the budget cut-off.

and `cargo test --doc` pass.

Not covered: no benchmark for the changed path, and no in-game redstone verification
(repeaters, observers, pistons, comparators).

Fixes #2016

## Summary by CodeRabbit

- **New Features**
  - Added correct explosion-survival loot behavior for wall, soul, redstone, and copper wall torches.
  - Improved cascading block and shape updates for more consistent world behavior.

- **Bug Fixes**
  - Block physics events now report the correct originating position.
  - Improved handling of block breaking so drop behavior is applied correctly.
  - Added safeguards to prevent excessive chained block updates from impacting performance.

### DIFF
--- a/crates/pumpkin/src/block/registry.rs
+++ b/crates/pumpkin/src/block/registry.rs
@@ -1176,12 +1176,9 @@ impl BlockRegistry {
         for direction in BlockDirection::abstract_block_update_order() {
             let pos = position.offset(direction.to_offset());
 
-            Box::pin(world.replace_with_state_for_neighbor_update(
-                &pos,
-                direction.opposite(),
-                flags,
-            ))
-            .await;
+            world
+                .replace_with_state_for_neighbor_update(&pos, direction.opposite(), flags)
+                .await;
         }
     }
 
--- a/crates/pumpkin/src/world/mod.rs
+++ b/crates/pumpkin/src/world/mod.rs
@@ -23,6 +23,7 @@ pub mod chunker;
 pub mod explosion;
 pub mod loot;
 pub mod map;
+mod neighbor_updater;
 pub mod portal;
 pub mod time;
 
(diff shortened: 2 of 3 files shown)


ydw1904:
Heads up: I opened #3130 which converged on essentially the same iterative-stack design as this PR (per-task `UpdateStack`, LIFO drain, `MAX_CHAINED_UPDATES`, `execute_neighbor_update`/`execute_shape_update` split). Now that both exist, they overlap almost entirely, so it doesn't make sense to keep both open — this one can absorb the difference.

The one behavioral gap: #3130 also fixes #3120. When a support block is broken, the cascade currently propagates `SKIP_DROPS` into the chained breaks, so attached blocks (torches, levers, buttons, rails) never drop their loot in creative. The fix is a one-liner where the replacement is applied:

```rust
Some(self.set_block_state(position, new_state_id, flags - BlockFlags::SKIP_DROPS))
```

If you fold that in here, this PR covers both #2016 and #3120 and I'll close #3130 as superseded. Happy to push it as a suggestion or a small commit to your branch, whichever you prefer.

RoosterBooster007:
Hey @mcbabel, thanks for your work here. One thing it needs before it can land: master split `update_neighbors` into `update_neighbors_at` (287404dde) after you branched, so a plain rebase would leave `set_block_state` and the piston call sites on the old recursive loop and the fix wouldn't actually bite (`update_neighbors_at` needs to go through the stack too). It's a big change to update semantics with the redstone-wire timing shift you flagged yourself, so I'd want @Snowiiii to sign off on the approach, and while you're in there, the `SKIP_DROPS` one-liner from ydw1904's now-closed #3130 is worth folding in so #3120 gets fixed as well.

Issue #1458, opened 2026-02-07
Crashes and End portal/spectator-mode


Adioc14:
### I've searched existing issues and couldn't find a duplicate.

- [x] I confirm this is not a duplicate.

### Operating System

Windows

### Server Software Version/Commit

b1f62f3

### What happened?

- End portal
endportals dont spawn aligned correctly, and completing the portal with eye of ender crashes the server with no logs to explain

- spectator mode parity mismatch
spectator mode doesn't have its transparent head look, which causing unable to see through blocks if you went to third person

### To Reproduce

- Endportal
find an endportal you will see it misaligned, build another endportal with correct alignment and fill the eyes and will crash the server
- Spectator mode
enter spectator mode you will see your head not transparent, and also go through blocks you will not see anything in thirdperson

### Expected behavior

should NOT crash , and portal should atleast be aligned correctly

spectator mode head should be transparent and you can see through blocks in third person

kokofixcomputers:
With the end portal crash, it seems that if i enable rust backtrace, you will actually see some logs:

```
thread 'Gen-0' (2713000) panicked at pumpkin-world/src/generation/noise/router/chunk_density_function.rs:257:13:
index out of bounds: the len is 99 but the index is 129
stack backtrace:
   0:        0x1009ecec4 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h10dcb5e2ebe8f0ac
   1:        0x1002ad580 - core::fmt::write::hd5926bdf73ee24f4
   2:        0x1009eb470 - std::io::Write::write_fmt::h7c97e47276bac25e
   3:        0x1009eb76c - std::panicking::default_hook::h5bd341aa6d010dc8
   4:        0x100ac2024 - pumpkin::main::{{closure}}::{{closure}}::hd6a0679b5748bce4
   5:        0x1009eb0b0 - std::panicking::panic_with_hook::hbf4b5b50eb72ae21
   6:        0x100a0acd8 - std::panicking::panic_handler::{{closure}}::h24ca9bac65f2240f
   7:        0x100a0ac94 - std::sys::backtrace::__rust_end_short_backtrace::hf48afbd2b4eb8a2d
   8:        0x100a0ac88 - __rustc[9e6a08e89e4b9111]::rust_begin_unwind
   9:        0x100ac4170 - core::panicking::panic_fmt::h702fd765199c1b52
[...]

PR #2803, opened 2026-08-07
fix(world): size proto chunks by the dimension, not by the noise settings

MCbabel:
## Description

Generating a chunk in the End or the Nether panics on every generation thread with an index
out of bounds, so the chunks a player needs after stepping into an End portal never finish
generating.

`ProtoChunk::new` took its height from the noise settings (`settings.shape.height`), while
`Chunk::build_level_sections` builds `dimension.height / 16` sections and reads all of them
back out of the proto chunk. The End and the Nether are the two dimensions whose dimension
type is taller than their noise settings (256 vs 128), so their proto chunks had 8 sections
of storage for the 16 the level chunk asks for. `get_block_state_raw` then indexed a
32768-entry `flat_block_map` with the stride of a 128-block-tall chunk for y up to 255 and
panicked at index 32768.

The same mismatch left chunk light data covering only half the column, read biome sections
out of range, and silently dropped every block a feature placed above y = 128. The End city
ship in `fixed_seed_generates_vanilla_end_ship_chunk` lost 45 of its 59 blocks, which is why
that expectation had been lowered to 14.

A chunk spans its dimension, exactly like vanilla's `LevelHeightAccessor`; the noise settings
only bound the vertical window the terrain generator writes into, which is vanilla's
`WorldGenerationContext`. `ProtoChunk` now carries both — storage sized by the dimension,
plus a generation window built with `GenerationShapeConfig::trim_height`. `YOffset` anchors
resolve against that window, so surface rules (the Nether's bedrock roof and floor
gradients), carver height providers and feature placements keep resolving against y < 128 in
the End and the Nether instead of following the taller storage.

This covers the crash only. The issue also reports End portal misalignment and spectator-mode
transparency; both are untouched.

The panic quoted in the issue (`chunk_density_function.rs`, len 99 index 129) was the same
conflation in `populate_noise` and was already fixed in 3c43f971 — its cell counts come from
the noise sampler now. This fixes the remaining occurrence.

## Testing

Verified on a running server: completed an End portal and travelled to the End. Generation
completes and the dragon fight initialises. Previously every generation thread died.

New regression test `dimensions_taller_than_their_noise_setting`, and the End ship
expectation is restored from 14 blocks to the full 59.

and `cargo test --doc` pass.

Refs #1458

### DIFF
--- a/crates/pumpkin-world/src/chunk_system/generation.rs
+++ b/crates/pumpkin-world/src/chunk_system/generation.rs
@@ -152,6 +152,51 @@ mod tests {
     }
 
     #[test]
+    fn dimensions_taller_than_their_noise_settings_generate_all_sections() {
+        for (dimension, terrain_state) in [
+            (
+                Dimension::THE_NETHER,
+                pumpkin_data::Block::NETHERRACK.default_state.id,
+            ),
+            (
+                Dimension::THE_END,
+                pumpkin_data::Block::END_STONE.default_state.id,
+            ),
+        ] {
+            let seed = Seed(42);
+            let block_registry = Arc::new(BlockRegistry);
+            let world_gen =
+                get_world_gen(seed, dimension.clone(), false, Vec::new(), String::new());
+            let biome_mixer_seed = hash_seed(world_gen.seed());
+
+            let chunk = generate_single_chunk(
+                &dimension,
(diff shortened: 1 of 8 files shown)


Snowiiii:
LGTM, Thanks @MCbabel :D

PR #2709, opened 2026-08-01
Add centralized unix timestamp functions

O2theC:
refactor: add centralized unix timestamp functions

## Description
Adds five helper functions for unix timestamps, one returning a `Duration` and the other 4 for seconds, milliseconds, microseconds, and nanoseconds.
Refactored all relevant call sites to use these functions.
this centralizes the unwrap/expect required for getting the current unix timestamp and ensures a descriptive expect msg.

PR #2658, opened 2026-07-31
fix: scoreboard.dat saving/loading, working render types, criteria, operations, improved API, etc.

RoosterBooster007:
## Description
Overhauls the scoreboard system to fix long-standing issues and bring it much closer to vanilla parity, plus a greatly expanded scoreboard plugin API.

## Fixes
- [x] Scoreboard persistence (objectives, teams, scores, and display
  slots saved/loaded from `scoreboard.dat`)
- [x] Render types (integer/hearts)
- [x] Criteria (e.g. `dummy`, `health`, `trigger`, custom), not just hardcoded "dummy"
- [x] Display slots (added the missing team-colored sidebar slots and a new `DisplaySlot`
  command argument type)

## Commands
- [x] `/scoreboard` - objectives add/list/remove/modify, players get/set/add/remove/reset, teams create/join/leave/remove, and score operations.
- [x] `/trigger` - works with trigger-criteria objectives, so [future datapacks](https://github.com/Pumpkin-MC/Pumpkin/pull/2642) can use objective-based triggers.

## Plugin API (`scoreboard.wit`)
- [Check out my PR in the `pumpkin-plugin-wit` submodule](https://github.com/Pumpkin-MC/pumpkin-plugin-wit/pull/25) (waiting on this)

## Testing
Example: I got the hearts list display working:

PR #2804, opened 2026-08-07
fix(world): load level.dat instead of discarding it

### LINKED ISSUE #1291: Serde error: missing field `Series`

MCbabel:
## Description

`AnvilLevelInfo::read_world_info` validated only `DataVersion` and `version`, then returned
`LevelData::default(Seed(0))` overlaid with the `data/minecraft/*.dat` side files. Everything
actually stored in level.dat - `LevelName`, `Difficulty`, spawn position, world border,
`DataPacks`, `Version` - was discarded on every start, and the symmetric writer emitted only
`DataVersion`, `version` and `LastPlayed`, truncating the rest on the next save.

A world whose seed lives in level.dat's legacy `Data.WorldGenSettings` compound, which is
what converters such as Chunker emit, therefore loaded with seed 0 and silently generated
foreign terrain into the existing region files.

The reported `missing field Series` symptom is unreachable as written: pumpkin-nbt has no
serde support, so `LevelData`'s derived `Deserialize` is dead code. The underlying complaint
- level.dat contents being replaced by defaults - is real and worse, so this fixes it at the
root. The serde `default` attributes are added anyway as insurance for the public types.

What changed:

- level.dat is parsed field by field through pumpkin-nbt, each field falling back to its
  vanilla default when absent, so an optional tag such as `Version.Series` can no longer fail
  a load.
- Saving is a read-modify-write merge of the existing `Data` compound, so tags Pumpkin does
  not model (`Player`, `DragonFight`, `hardcore`, `GameType`, `ServerBrands`) survive.
- When neither `world_gen_settings.dat` nor level.dat carries a seed, startup aborts with
  `MissingWorldSeed` instead of quietly generating with seed 0. To keep that recoverable the
  seed is also written into level.dat's `Data.WorldGenSettings`, since level.dat is the only
  world file Pumpkin backs up (`level.dat_old`, refreshed on every successful start) while
  `world_gen_settings.dat` is written non-atomically and never backed up.
- Version stamping follows vanilla: `DataVersion`, `version` and the `Version` compound are
  re-stamped with the current constants on every save rather than echoing back what was read.
  Chunks are stamped unconditionally with `WORLD_DATA_VERSION`, so preserving an imported
  world's older version would desynchronise level.dat from the data Pumpkin actually wrote.
- A failure to read the existing level.dat at save time is propagated instead of downgraded
  to a warning, so a transient error can no longer replace a complete level.dat with a lossy
  one.

Known gaps: legacy `Data.GameRules` is still not read, so a converted world's game rules load
as defaults — that needs a real format converter, since the legacy format uses unprefixed
string values. `NbtCompound::get_*` returns `None` for both "absent" and "wrong tag type", so
a converter writing `Difficulty` as an Int silently yields the vanilla default. level.dat is
still written non-atomically.

Overlap with open PRs: #2658 touches the same four files
(`world_info/{anvil,data_files,mod}.rs` and `server/mod.rs`) for scoreboard.dat handling. The
two are functionally independent — that PR does not change level.dat parsing — but they will
conflict textually, and I am happy to rebase on whichever lands first. #2437 and #2709 also
touch `world_info/anvil.rs` in smaller ways.

## Testing

Verified on a running server: the first start creates the world, later starts load it. The
`Failed to get level_info, using default instead` warning is gone and `/seed` stays constant
across restarts. Previously every start silently fell back to defaults.

Six unit tests: a converted level.dat without `Version.Series`, the missing-seed rejection,
seed recovery after `world/data` is deleted, version re-stamping of both level.dat and the
side files, changed-value plus unmanaged-tag survival across a rewrite, and a full round trip.

and `cargo test --doc` pass.

Fixes #1291

### DIFF
--- a/crates/pumpkin-world/src/world_info/anvil.rs
+++ b/crates/pumpkin-world/src/world_info/anvil.rs
@@ -1,23 +1,29 @@
 use std::{
     fs::File,
-    io::{Cursor, Read},
+    io::ErrorKind,
     path::Path,
     time::{SystemTime, UNIX_EPOCH},
 };
 use tracing::error;
 
-use flate2::read::GzDecoder;
+use pumpkin_nbt::{
+    compound::NbtCompound,
+    nbt_compress::{read_gzip_compound_tag, write_gzip_compound_tag},
+    tag::NbtTag,
+};
+use pumpkin_util::{Difficulty, world_seed::Seed};
 use serde::{Deserialize, Serialize};
 
 use crate::world_info::{
-    MAXIMUM_SUPPORTED_LEVEL_VERSION, MAXIMUM_SUPPORTED_WORLD_DATA_VERSION,
-    MINIMUM_SUPPORTED_LEVEL_VERSION, MINIMUM_SUPPORTED_WORLD_DATA_VERSION,
+    DataPacks, MAXIMUM_SUPPORTED_LEVEL_VERSION, MAXIMUM_SUPPORTED_WORLD_DATA_VERSION,
+    MINIMUM_SUPPORTED_LEVEL_VERSION, MINIMUM_SUPPORTED_WORLD_DATA_VERSION, WorldVersion,
     data_files::{
         minecraft_data_dir, read_game_rules, read_wandering_trader, read_weather,
         read_world_clocks, read_world_gen_settings, write_custom_boss_events_stub,
         write_game_rules, write_scheduled_events_stub, write_wandering_trader, write_weather,
         write_world_clocks, write_world_gen_settings,
     },
+    default_data_packs,
 };
 
 use super::{LevelData, WorldInfoError, WorldInfoReader, WorldInfoWriter};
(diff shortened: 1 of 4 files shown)


Snowiiii:
LGTM, Danke @MCbabel :D

PR #2808, opened 2026-08-07
chore(protocol): Clean up Bedrock by using more `derive`

QuantumSegfault:
## Description

- Adds derive support for `PacketReadSlice`, and makes some tweaks to the other derive macros to properly support packets with up to a single lifetime parameter (which are used in packets that need to implement `PacketReadSlice`)
- Fixes up `PacketRead` (and its derive) support for `Vec<T>` (only allowed when prefixed by standard VarUInt length).
- Adds `Packet*` impls for `Cow<'_, str>`
- Replaces manual `Packet*` impls with `derive` wherever possible.
- Replaces a couple of manual uses of `[u8; 16]` for UUID with actual `uuid::Uuid`.
- Moves incorrectly labelled `SActorEvent` over as `CActorEvent` where it belongs (as at least currently, it is only sent to clients; never to the server).
  - Running `pumpkin-codegen` to update the WIT has been skipped for now.

## Testing

These changes have been tested in a basic capacity (login, mine, place blocks, craft, find & hit mob) and have worked as expected.

### DIFF
--- a/crates/pumpkin-macros/src/lib.rs
+++ b/crates/pumpkin-macros/src/lib.rs
@@ -509,8 +509,18 @@ pub fn derive_serialize(input: TokenStream) -> TokenStream {
             .into();
     };
 
+    let type_generic = match input.generics.params.len() {
+        0 => quote! {},
+        1 => quote! { <'_> },
+        _ => {
+            return syn::Error::new(name.span(), "Only up to one lifetime parameter is supported.")
+            .to_compile_error()
+            .into();
+        }
+    };
+
     let expanded = quote! {
-        impl PacketWrite for #name {
+        impl PacketWrite for #name #type_generic {
             fn write<W: std::io::Write>(&self, writer: &mut W) -> Result<(), std::io::Error> {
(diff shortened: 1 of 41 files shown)


ZlordHUN:
I tested this last night and found no issues. LGTM

Snowiiii:
How much this affects runtime speed?, because i recently also moved from Serde to manual write and reads which was quiet a lot faster

QuantumSegfault:
> How much this affects runtime speed?, because i recently also moved from Serde to manual write and reads which was quiet a lot faster

I don't expect this to have any notable impact on runtime speed. This is just using `derive` to create the same `impl` automatically, instead of by hand. No extra machinery that wasn't already being invoked manually

The ONE potential slow down is that `Vec<u8>` is no longer read whole, but rather byte-by-byte (as you can't specialize impls...). If this proves to be a problem, manual impls can still be written out for any containing large `Vec<u8>`

QuantumSegfault:
Ping @Snowiiii

PR #2809, opened 2026-08-07
feat(plugin): add persistent key-value storage API for WASM plugins

darkstarworks:
## What

Adds a persistent key-value storage API for WebAssembly plugins. Each plugin gets a private store, saved to its data folder, that survives server restarts:

```rust
pumpkin_plugin_api::storage::{set, get, remove, contains, keys, clear}
```

Values are strings, so plugins serialize structured data (JSON and so on) themselves. `keys(prefix)` supports enumerating a group, for example per-player state stored under `player:<uuid>`.

The store is a single `storage.json` in the plugin's data folder (`plugins/data/<name>/`), loaded on first use and written back after every change through a temporary file plus rename, so a crash mid-write cannot leave a corrupt file behind.

Access follows the existing data-folder permissions: reads require `fs.read.data` and writes require `fs.write.data`, so the convenience store cannot be used to get around the file-system permission gate.

## Why

Plugins run as WebAssembly and have no database or file handling of their own, so anything that must survive a restart (cooldowns, statistics, progress) has nowhere to live. This gives every plugin a small persistent store without each one reinventing file access.

## How

- New `storage` WIT interface (companion PR Pumpkin-MC/pumpkin-plugin-wit#28), imported by the plugin world; this PR bumps the submodule to include it.
- Host implementation in `wasm_host/wit/v0_1/storage.rs`: a small `PluginKvStore` (load, get, set, remove, keys, clear, with atomic writes), the `storage::Host` impl, and the permission checks.
- `PluginHostState` gains the plugin name (needed to find the data folder) and a lazily-loaded store, so plugins that never touch storage pay nothing for it.
- Plugin-facing re-export in `pumpkin-plugin-api`.

## Tests

Six unit tests: set/get/remove round-trip, persistence across a reload, prefix-filtered `keys` and `clear`, plugin-name sanitization against path traversal, storage-path containment, and permission enforcement.

## Notes

Depends on Pumpkin-MC/pumpkin-plugin-wit#28. The submodule pointer here references that change, which is not yet on the canonical submodule repo, so CI cannot build until #28 is merged. Left as a draft for now; I will re-point the submodule to the merged commit once #28 lands.

`cargo test` and `cargo clippy --all-targets -- -D warnings` are clean on Windows with stable Rust.

### DIFF
--- a/crates/pumpkin-plugin-api/src/lib.rs
+++ b/crates/pumpkin-plugin-api/src/lib.rs
@@ -63,7 +63,7 @@ pub use wit::pumpkin::plugin::{
     entity_types::EntityType,
     event::{self as events_wit, EventType},
     gui, i18n, item_stack, java_dialogs, java_packets, particles, permission, player, scoreboard,
-    server, text, uuid, world,
+    server, storage, text, uuid, world,
 };
 
 // Convenience re-exports of commonly-used plugin types so plugin authors can
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
@@ -69,6 +69,12 @@ pub struct PluginHostState {
     pub plugin: Option<Weak<WasmPlugin>>,
     pub server: Option<Arc<Server>>,
     pub permissions: Vec<String>,
+    /// The name of the plugin this state belongs to, used to locate its private
+    /// data folder. Set once the plugin's metadata has been read.
+    pub plugin_name: Option<String>,
+    /// The plugin's persistent key-value store, loaded from disk the first time
+    /// the plugin uses it.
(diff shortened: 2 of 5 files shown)


vyPal:
> Plugins run as WebAssembly and have no database or file handling of their own, so anything that must survive a restart (cooldowns, statistics, progress) has nowhere to live.

Pumpkin plugins (using WASM) are built to use WASI and the WIT component interface. Thanks to WASI, plugins do in fact have storage access. The `fs.read.data` and `fs.write.data` permissions give the plugin access to read and/or write data to its dedicated directory. To fetch the path to this directory, you must call `context.get_data_folder()` from the `on_load()` or `on_unload()` handlers. To actually interact with the folder, you can use the native file API provided by your language (so `fs::read()` in rust, or `open()` in python for example).

darkstarworks:
You're right, thank you for the correct. I'd missed that the pre-opened data folder together with `fs.read.data` / `fs.write.data` already gives plugins everything this was trying to provide through the native file API. 
Thank you for taking the time to explain it. Rust isn't my native language

PR #2820, opened 2026-08-07
docs(plugin-api): explain how plugins persist data

darkstarworks:
## What

Adds a "Persisting data" section to the plugin API crate docs. It explains that a plugin stores data that survives restarts by using its language's normal file API (for example `std::fs` in Rust) against the folder returned by the context's `get_data_folder` method, and that this requires the `fs.read.data` / `fs.write.data` permissions.

## Why

This came out of #2809, where I proposed a key-value storage interface without realizing that plugins already have file access to their data folder through WASI. The capability was there the whole time; it just was not documented anywhere, which is what led to the confusion. This writes it down so the next person does not miss it.

No code changes, docs only.

### DIFF
--- a/crates/pumpkin-plugin-api/src/lib.rs
+++ b/crates/pumpkin-plugin-api/src/lib.rs
@@ -28,6 +28,32 @@
 //!
 //! register_plugin!(MyPlugin);
 //! ```
+//!
+//! # Persisting data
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thank you @darkstarworks

PR #2821, opened 2026-08-07
fix: Bedrock NetherNet offline mode

ZlordHUN:
## Description
Fixes Bedrock NetherNet connections when Pumpkin is running in offline mode.

## What
- Allows offline-mode clients to use unverified identity-provider metadata.
- Keeps token expiration, client public-key, and signed DTLS fingerprint validation enabled.
- Preserves strict identity-provider and OIDC validation in online mode.
- Adds a regression test covering both offline acceptance and online rejection.

## How
Identity-provider validation now runs only when an OIDC verifier is configured. Offline mode skips centralized provider verification while retaining the cryptographic binding between the client identity and its WebRTC connection.

## Testing
- `cargo test -p pumpkin net::bedrock::nethernet::tests --lib`
- Verified in-game with online_mode = false


Snowiiii:
LGTM, Thank you @ZlordHUN :D

PR #2797, opened 2026-08-06
feat(bedrock): update to 1.26.40 and add NetherNet transport

ZlordHUN:
## Description
Updates Pumpkin’s Bedrock support to Minecraft 1.26.40 and implements NetherNet transport. I'm going to have to split this into multiple PRs aren´t I?

## What
- Updates Bedrock to version 1.26.40, protocol 2168.
- Implements NetherNet HTTP signaling and WebRTC DataChannel transport.
- Retains the existing RakNet transport through a shared transport abstraction.
- Updates affected 1.26.40 packet codecs.
- Regenerates the Bedrock block, item, runtime-item, and creative inventory mappings.
- Adds configuration for the NetherNet signaling address and persistent identity key.

## How
- Implements NetherNet discovery and SDP negotiation endpoints.
- Validates client identity assertions, token expiration, DTLS fingerprints, and ES384 signatures.
- Generates and persists a P-384 server identity key.
- Negotiates reliable and unreliable WebRTC DataChannels.
- Implements NetherNet packet fragmentation and reassembly.
- Routes NetherNet traffic through Pumpkin’s existing Bedrock login and gameplay pipeline.
- Keeps RakNet acknowledgement and retransmission handling isolated to RakNet clients.
- Ports the changed 1.26.40 packet formats and generated registry data.

## Testing
- `cargo fmt --all -- --check`
- `cargo test --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`
- Tested using a Minecraft Bedrock 1.26.40 client.
- Verified server discovery and NetherNet negotiation.
- Verified login, resource loading, world entry, and stable-ish gameplay, this PR includes no bugfixes Im afraid.
- Verified the world renders using the correct 1.26.40 block and item mappings.

PR #2825, opened 2026-08-07
fix(bedrock): negotiate NetherNet over LAN discovery instead of HTTP

xRookieFight:
Bedrock clients could never join over NetherNet. The transport added in #2797 negotiated WebRTC over an HTTP endpoint (`POST /v1/join/{network_id}` on an axum server bound to 19132), but that is not how the game speaks NetherNet - the client never sends such a request, so nothing ever connected.

Vanilla negotiates over the LAN discovery protocol instead: an encrypted UDP protocol on port 7551. This replaces the HTTP endpoint with a real implementation of it.

### What NetherNet actually does

Discovery datagrams are `HMAC-SHA256 checksum || AES-256-ECB(length | packet id | sender id | padding | payload)`, keyed with the SHA-256 of `0xdeadbeef`. Three packets matter:

- `DiscoveryRequestPacket` - broadcast by clients looking for servers.
- `DiscoveryResponsePacket` - our answer, carrying a hex-encoded `ServerData` blob with the world name, player counts and transport info.
- `DiscoveryMessagePacket` - carries the signaling messages (`CONNECTREQUEST`, `CONNECTRESPONSE`, `CANDIDATEADD`, `CONNECTERROR`) that negotiate the WebRTC session.

So the flow is: client broadcasts a request, we respond with our world card, the client sends an SDP offer as `CONNECTREQUEST`, we answer with `CONNECTRESPONSE` and trickle our ICE candidates as `CANDIDATEADD`, and the game traffic then runs over the `ReliableDataChannel`.

### Changes

- New `net/bedrock/nethernet` module: `crypto` (discovery cipher and checksum), `discovery` (packet codec and `ServerData`), `signal` (signaling messages), `session` (data channels and fragment reassembly) and `identity` (the existing ES384 assertion code, moved).
- Removed the axum dependency from `pumpkin`; the signaling server is gone.
- ICE candidates are now trickled in both directions. Candidates that arrive before the remote description is applied are buffered instead of dropped.
- Negotiations that stall are dropped after 15 seconds.
- `bedrock.nethernet.address` now defaults to `0.0.0.0:7551`. The RakNet-style server-list pong got its own `bedrock.address` (`0.0.0.0:19132`) since it no longer shares a socket with signaling.

### Getting it to actually work

A few things only showed up against a real client, checked against a 1.26.40 dedicated server on the wire:

- `ServerData` is at version 6, not 5 as documented in the reference implementations. Version 6 adds a world id (16 hex digits) between the auth flags and the transport layer. While we sent version 5 the client silently ignored the response and the world never appeared in the LAN list. The encoder test reproduces a real dedicated server response byte for byte.
- The `a=identity:` assertion is optional. Clients joining over LAN omit it entirely, and we were rejecting the offer outright. It is now only verified when present, and the login chain is bound to the transport identity only if there is one.
- The Bedrock connection timeout went from 10 to 30 seconds. Clients send nothing at all while they load the world, so a slow initial chunk load looked like a dead connection and got the player kicked right after joining.

### Testing

Joined from a 1.26.40 Windows client over LAN: the world card shows up, the connection negotiates, and the player spawns in and stays connected. Unit tests cover the discovery codec, the cipher, `ServerData` against real server bytes, signal parsing and fragment reassembly.


crates/pumpkin/src/net/bedrock/nethernet/mod.rs
```diff
@@ -0,0 +1,506 @@
...
+        ServerData {
+            server_name: &config.motd,
+            level_name: &self.server.basic_config.default_level_name,
+            game_type,
+            player_count,
+            max_player_count: config.max_players as i32,
+            editor_world: false,
+            hardcore: self.server.basic_config.hardcore,
+            accepts_online_auth: config.online_mode,
+            accepts_self_signed_auth: !config.online_mode,
```
QuantumSegfault, line 441: This should probably also take `networking.bedrock.authentication.enabled` into account.

Like if online_mode is on, it accepts online auth (as it does right now).

If authentication is disabled, we also allow self signed (rather than only when online_mode is disabled).

--------

Maybe the client adapts. But the `authentication` config should be taken into account and this should be tested in multiple combinations (of logged in vs logged out; online on and off, authentication on and off)

crates/pumpkin/src/net/bedrock/nethernet/discovery/server_data.rs
```diff
@@ -0,0 +1,96 @@
...
+            level_name: "world",
+            game_type: 0,
+            player_count: 1,
+            max_player_count: 20,
+            editor_world: false,
+            hardcore: false,
+            accepts_online_auth: true,
+            accepts_self_signed_auth: false,
+            transport_layer: TRANSPORT_LAYER_NETHERNET,
+            connection_type: CONNECTION_TYPE_LAN_SIGNALING,
```
QuantumSegfault, line 80: Does this interact correctly with the server list (not just the LAN games menu)?

xRookieFight: It does iirc

RoosterBooster007:
@xRookieFight, are you still interested in getting this through?

QuantumSegfault:
We're no longer pursuing this direction; we're sticking with HTTP negotiation.

If there are other useful changes you made you still want in, feel free to open separate PRs for those.

PR #2829, opened 2026-08-08
fix: Bedrock block breaking feedback

ZlordHUN:
## Description
Excited about this one.
Current master branch has issues in which if a bedrock player starts to break blocks/breaks blocks - the progression isnt displayed, the blocks dont crack. After breaking blocks there are no particles either. On top of that the floating/swimming penalty is always applied to bedrock players so they break blocks at a significantly reduced speed. This PR addresses all of these.

## What
- Displays the complete cracking animation while Bedrock players mine blocks.
- Uses the correct mining speed for grounded players, fixing blocks such as sand taking much longer than expected.
- Prevents blocks from briefly reappearing when the client predicts their destruction.
- Completes Bedrock block destruction server-side when mining progress reaches 100%.
- Sends the correct block-specific destruction particles and sound.
- Keeps block-breaking progress synchronized for nearby Java and Bedrock players.

## How
- Sends Bedrock’s start, update, and stop breaking events using the calculated progress-per-tick rate.
- Tracks each player’s active block, accumulated progress, and current mining speed.
- Updates Bedrock’s grounded state from authenticated movement input so Vanilla’s airborne mining penalty is only applied when appropriate.
- Accepts reasonable client-side destruction predictions to avoid a restore-then-destroy flicker.
- Maps the destroyed Java block state to its Bedrock runtime ID for the destruction particle event.

## Testing
- Verified in-game with a Bedrock client:
  - Cracking progresses smoothly through the full animation.
  - Sand breaks at the expected speed.
  - Blocks no longer flicker before disappearing.
  - Block-specific destruction particles appear correctly.
- `cargo check -p pumpkin`

### DIFF
--- a/crates/pumpkin-protocol/src/bedrock/client/level_event.rs
+++ b/crates/pumpkin-protocol/src/bedrock/client/level_event.rs
@@ -14,6 +14,7 @@ pub struct CLevelEvent {
 #[repr(i32)]
 pub enum LevelEvent {
     // There are hundreds of these, adding only what we need for now
+    ParticlesDestroyBlock = 2001,
     BlockStartBreak = 3600,
     BlockStopBreak = 3601,
     BlockUpdateBreak = 3602,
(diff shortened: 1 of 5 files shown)


Snowiiii:
LGTM, Thank you @ZlordHUN :D

PR #2832, opened 2026-08-08
fix: Bedrock dropped item synchronization

ZlordHUN:
## Description
Fixes dropped-item synchronization for Bedrock players, including inventory pickup updates, item movement, and cross-edition visibility.

## What
- Picked-up items now appear in Bedrock inventories instead of disappearing.
- Bedrock inventory slots are mapped correctly between Java’s player screen layout and Bedrock’s 36-slot inventory.
- Dropped items receive smooth position updates on Bedrock and settle on the correct supporting block. (This is in line with bedrock singleplayer behavior)
- Java retains Vanilla’s 20-tick item tracking interval.
- Players joining later receive edition-appropriate entity spawn packets.
- Java players receive the item-stack metadata required to render existing dropped items.

## How
- Connects Bedrock players to Pumpkin’s existing screen-handler synchronization path.
- Sends full Bedrock inventory contents in hotbar-first order and translates incremental player-screen slot updates.
- Sends moving item positions to Bedrock every server tick while preserving Vanilla’s Java tracking cadence.
- Adds edition-aware late entity spawning and includes item metadata after Java item spawns.

## Testing
- `cargo check -p pumpkin`

Verified in-game that:
- Bedrock players can drop and recover items without losing them.
- Picked-up items immediately return to the Bedrock inventory.
- Dropped items move smoothly and rest on the correct block.
- Java players joining after an item was dropped can see and collect it.

### DIFF
--- a/crates/pumpkin/src/entity/item.rs
+++ b/crates/pumpkin/src/entity/item.rs
@@ -11,7 +11,7 @@ use pumpkin_protocol::bedrock::network_item::ItemStackWrapper;
 use pumpkin_protocol::codec::item_stack_seralizer::ItemStackSerializer;
 use pumpkin_protocol::codec::var_long::VarLong;
 use pumpkin_protocol::codec::var_ulong::VarULong;
-use pumpkin_protocol::java::client::play::Metadata;
+use pumpkin_protocol::java::client::play::{CSetEntityMetadata, Metadata};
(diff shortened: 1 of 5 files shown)


Snowiiii:
LGTM, Thank you @ZlordHUN :D

PR #2839, opened 2026-08-08
fix(block): notify neighbors when a block is removed by a neighbor update

### LINKED ISSUE #2837: Doors underwater
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

ee6e047

### Operating System

Windows 11

Doors don't update correctly:
- If you break the bottom half the door top breaks but does not update the water around the top half (but sand on the door does fall so it might be a water issue) same other way round.
- doors don't instantly break when the block below is destroyed.
- breaking the bottom half of a door under water causes the water to flow down one layer (server side the client does not see the water but can drown) but not further than that.

### Steps to Reproduce

1. break the bottom half of a door under water.
2. place a door underwater and break the block below the door.

xRookieFight:
Fixes #2837

When a block is destroyed as a result of a neighbor update (the second half of a door, a torch losing its support, ...), `replace_with_state_for_neighbor_update` passed the incoming flags straight to `break_block`. Those flags come from `set_block_state`, which already stripped `NOTIFY_NEIGHBORS` before running the neighbor pass, so the removal itself never notified anything around it.

Vanilla does the opposite here: `Block.replace` calls `World.breakBlock`, which always sets the new state with `NOTIFY_ALL`. Only the `setBlockState` branch keeps the reduced flags.

The visible effect underwater: breaking the lower half of a door removes the upper half, but nothing around the upper half is updated, so the water next to it never schedules a fluid tick. The column above only advances a single layer and then stops.

Also fixed the door support check. `can_place_at` required the block above to be replaceable, which is a placement-time condition - for an already placed door the block above is its own upper half, so the check could never pass. It is now split: placement keeps both conditions, while `get_state_for_neighbor_update` only asks whether the block below can still support the door.

### DIFF
--- a/crates/pumpkin/src/block/blocks/doors.rs
+++ b/crates/pumpkin/src/block/blocks/doors.rs
@@ -181,7 +181,11 @@ impl BlockBehaviour for DoorBlock {
     }
 
     fn can_place_at(&self, args: CanPlaceAtArgs<'_>) -> bool {
-        can_place_at(args.block_accessor, args.position)
+        has_support(args.block_accessor, args.position)
+            && args
+                .block_accessor
+                .get_block_state(&args.position.up())
+                .replaceable()
     }
 
     fn placed<'a>(&'a self, args: PlacedArgs<'a>) -> BlockFuture<'a, ()> {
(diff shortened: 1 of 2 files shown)


Snowiiii:

PR #2840, opened 2026-08-08
fix(worldgen): skip jigsaw blocks whose final state is a structure void

### LINKED ISSUE #2836: villages spawn with structure voids
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

ee6e047

### Operating System

Windows 11

Structure voids should be removed when placing a structure

### Steps to Reproduce

Find a village seed: 

Image

(1787870112421306378) pos: (265, 69, 551)

xRookieFight:
Fixes #2836

`place_template` filtered structure voids by looking at the raw palette entry, but a jigsaw block is only turned into its `final_state` a few lines later. Village templates use `"final_state": "minecraft:structure_void"` on most of their jigsaw blocks - 1124 of them across the embedded village templates - which means those positions were resolved to a real structure void state and written into the world.

The void check now runs on the entry that is actually placed, so it covers both a structure void coming from the palette and one coming from a jigsaw final state. The same substitution in the jigsaw block entity (used by `/place jigsaw` and the jigsaw block generate button) got the same check.

Added a test that runs every embedded template through `place_template` and asserts no structure void ends up at the placer. It fails on master and passes with this change.

### DIFF
--- a/crates/pumpkin-world/src/generation/structure/template/mod.rs
+++ b/crates/pumpkin-world/src/generation/structure/template/mod.rs
@@ -86,10 +86,8 @@ pub fn place_template(
     for block in &template.blocks {
         let palette_entry = &template.palette[block.state as usize];
 
-        // Structure blocks are data markers and structure void preserves the existing block.
-        if palette_entry.name == "minecraft:structure_void"
-            || palette_entry.name == "minecraft:structure_block"
-        {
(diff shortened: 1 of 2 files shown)


Snowiiii:
LGTM, Thanks @xRookieFight

PR #2844, opened 2026-08-08
fix: bedrock and java players disconnecting when joining the same server

ZlordHUN:
## Description
Fixes Java and Bedrock players disconnecting when players from the other edition join, while ensuring Java skins render correctly for Bedrock players.

## What
- Prevents Bedrock clients from rejecting player-list packets and disconnecting when Java players join.
- Prevents Java clients from receiving invalid chat-session information for Bedrock players.
- Synchronizes existing and newly joined players across both editions.
- Displays Java skins to Bedrock players using the correct wide or slim geometry.
- Preserves legacy `64×32` Java skins so limbs are rendered correctly.
- Gives fallback skins stable per-player identifiers to avoid skin collisions.
- Logs Bedrock packet-violation warnings with the rejected packet and reason.

## How
- Uses Bedrock’s valid unknown-platform value instead of an invalid build-platform enum value.
- Omits Java chat-session initialization when a Bedrock player has no valid signed chat session.
- Includes Bedrock-compatible player geometry and selects the correct model from the Java skin metadata.
- Applies the required opaque-pixel masks while preserving legacy skin dimensions.
- Splits larger skin and player-list payloads into NetherNet-compatible fragments.

## Testing
- Verified in-game with Java and Bedrock players joining in both orders.
- Verified both clients remain connected when the other edition joins or reconnects.
- Verified Java skins, including legacy skins, render correctly for Bedrock players.


crates/pumpkin-protocol/src/bedrock/client/mod.rs
```diff
@@ -47,6 +47,9 @@ pub mod update_abilities;
 pub mod update_attributes;
 pub mod update_block;
 
+/// `BuildPlatform::Unknown` in the Bedrock protocol.
+pub const UNKNOWN_BUILD_PLATFORM: i32 = -1;
```
xRookieFight: As far as I know, this should be moved inside of build platform enum

xRookieFight (changes requested):

crates/pumpkin-protocol/src/bedrock/client/mod.rs
ZlordHUN: Fair and done.

xRookieFight (approved):
LGTM

Snowiiii:
LGTM, Thanks @ZlordHUN :D

ZlordHUN:
@jansegre sorry about the regression, my bad. Would you be willing to fix it or should I?

jansegre:
I'll open a fix soon.

PR #2845, opened 2026-08-08
Fix runtime lighting updates

eshankiyer:
Fixes runtime light propagation and client updates to match vanilla behavior.


tools/pumpkin-codegen/src/block.rs
```diff
@@ -630,13 +631,46 @@ impl BlockState {
...
         });
         tokens
     }
 }
 
+/// Mirrors vanilla's explicit `useShapeForLightOcclusion` overrides. The generated
+/// state data already carries the vanilla light opacity, which distinguishes the
+/// state-dependent slab and piston-base overrides without allocating property
+/// objects during lighting.
+fn uses_shape_for_light_occlusion(block_name: &str, state: &BlockState) -> bool {
```
Snowiiii: No, you may have to update our extractor to extract some new data

eshankiyer:
Updated the extractor to read vanilla’s useShapeForLightOcclusion value directly for each block state instead of inferring it from block names. Generated data now matches the decompile, including slab and piston states.

PR #2847, opened 2026-08-09
feat(chunk): improve chunk generation speed ~1.27x

BjornTheProgrammer:
## Description

A few optomization have been made to improve chunk generation speed. These were validated using a benchmark. 

I had also found some dead code along the way, so I just removed it.

The aquifer changes have been made due to the fact that beyond being a performance improvement, it actually helps the chunk gen maintain parity with vanilla.

## Testing

I spun up a vinilla fabric server and compared the generated chunks to one another to see if there were any issues, additionally I added some tests here to verify noise generation stays the same with no regressions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3424,6 +3424,7 @@ dependencies = [
  "base64 0.23.1",
  "bytes",
  "colored",
+ "criterion",
  "crypto-bigint 0.7.5",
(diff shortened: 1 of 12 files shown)


Snowiiii:
LGTM, Thanks @BjornTheProgrammer :D

PR #2848, opened 2026-08-09
Notify neighbors on block flag update

notrootshark:
Fixes an issue where water does not flow back into a block position after a block is broken underwater.


crates/pumpkin/src/world/mod.rs
```diff
@@ -4637,6 +4654,7 @@ impl World {
             } else {
                 flags.insert(BlockFlags::SKIP_DROPS);
             }
+            flags.insert(BlockFlags::NOTIFY_NEIGHBORS);
```
xRookieFight, line 4640: This seems no-op for the reported bug. Every player break path already passes NOTIFY_NEIGHBORS. Real cause is likely the neighbor-update to fluid-tick scheduling path. Also, force-inserting removes caller opt-out. Piston moves pass only SKIP_DROPS on purpose and now fire neighbor updates mid-move.

xRookieFight (changes requested):
Also, on fill command this means one neighbor update per block on large fills. This issue should be fixed from another way.

RoosterBooster007:
Thanks for the fix. The bug is real, but forcing `NOTIFY_NEIGHBORS` in `break_block` [adds a neighbor update per block on large fills](https://github.com/Pumpkin-MC/Pumpkin/pull/2848#pullrequestreview-4891131560), and the gap is now covered by #2839, so I'm closing this as superseded. If the water bug still reproduces on master, a targeted fluid-side fix would be welcome!

PR #2849, opened 2026-08-09
fix(mobs): match vanilla AI control timing

eshankiyer:
Matches mob AI control timing and jump handling with vanilla 26.2.


eshankiyer:
Closing this because the fork divergence makes the PR include unrelated commits. The mob AI changes remain on PumpkinPie.

PR #2818, opened 2026-08-07
fix: make skeletons shoot arrows with their bow

xRookieFight:
Fixes #2800

PR #2850, opened 2026-08-09
fix: adapt bow attack goal to new equipment API

xRookieFight:
Master is currently failing to compile:

```
error[E0308]: mismatched types
  --> crates/pumpkin/src/entity/ai/goal/bow_attack.rs:49:9
   = note: expected struct `Arc<Mutex<ItemStack>>`
              found struct `ItemStack`
```

The bow attack goal from #2818 was written against the old `EntityEquipment::get`, which returned `Arc<Mutex<ItemStack>>`. It now returns an owned `ItemStack`, and the two PRs landed without ever being built together.

Fix is straightforward: `main_hand_item` returns `ItemStack`, the two call sites drop their `.lock().await`, and the now unused `tokio::sync::Mutex` import goes away. No behavior change.


Snowiiii:

Issue #2827, opened 2026-08-08
Crash related to Storage blocks


CivilizationUndecided:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.40 (Commit: 63d97e2/release) - 266 Contributors

### Operating System

CachyOS Latest

### Bug Description & Expected Behavior

I have a bunch of hopefully useful log files. 

The issue seems to happen anytime I start up a server and walk around enough. I'm assuming its something related to worldgen and storage blocks ex; chests, barrels, dispensers, etc.

[5bc3btsq-crash-2026-08-08_06.13.49.txt](https://github.com/user-attachments/files/30851979/5bc3btsq-crash-2026-08-08_06.13.49.txt)
[r395xgk8-crash-2026-08-08_05.18.06.txt](https://github.com/user-attachments/files/30851983/r395xgk8-crash-2026-08-08_05.18.06.txt)
[shbvpckz-crash-2026-08-08_05.08.00.txt](https://github.com/user-attachments/files/30851981/shbvpckz-crash-2026-08-08_05.08.00.txt)
[gt7dh44q-crash-2026-08-08_05.05.33.txt](https://github.com/user-attachments/files/30851982/gt7dh44q-crash-2026-08-08_05.05.33.txt)
[hvww3wgn-crash-2026-08-08_05.02.42.txt](https://github.com/user-attachments/files/30851980/hvww3wgn-crash-2026-08-08_05.02.42.txt)
[2gnhqw2f-crash-2026-08-08_05.01.41.txt](https://github.com/user-attachments/files/30851985/2gnhqw2f-crash-2026-08-08_05.01.41.txt)
[zhkdw4p8-crash-2026-08-08_04.57.17.txt](https://github.com/user-attachments/files/30851984/zhkdw4p8-crash-2026-08-08_04.57.17.txt)

### Steps to Reproduce

1. Start server
2. Load into server
3. Walk around a bit, or crash on join.

PR #2851, opened 2026-08-09
fix: don't block the async runtime when reading inventories

xRookieFight:
Follow-up to #2827.

The crashes in that issue came from `blocking_write()` calls inside `from_nbt`, which panic with *"Cannot block the current thread from within a runtime"* whenever a chunk with a chest/barrel/dispenser got ticked. Those particular calls are already gone on master (49360e5 replaced them with `get_mut()`), but the same pattern is still alive in two places, so the crash can still be hit:

**`SimpleInventory::size()`** used `stacks.blocking_read()`. `size()` is a sync trait method and it is called from `ScreenHandler` while building slots, so it always runs on a tokio worker - opening an anvil, a stonecutter, an enchanting table or a villager trade panics the server the same way. The vector length never changes after construction, so the size is now just stored as a field, matching how every other inventory implements `size()` (a constant).

**`BrewingStandBlockEntity::from_nbt()`** still went through `futures::executor::block_on` to grab the item lock. It doesn't panic, but it parks a runtime worker thread while loading a chunk for no reason. The entity is still owned locally at that point, so `get_mut()` works there, same as barrel/dispenser/chest.

No behavior change beyond not blocking - loading and saving still read the exact same NBT.

Verified with `cargo check` and `cargo clippy` on `pumpkin` and `pumpkin-world`.

One thing I left alone: `chunk_data_nbt()` in several block entities still uses `futures::executor::block_on(self.items.read())`. It only takes `&self`, so `get_mut()` isn't an option, and making it async would ripple into `add_block_entity`/`update_block_entity` and their callers. Worth a separate PR.

### DIFF
--- a/crates/pumpkin-world/src/inventory/simple_inventory.rs
+++ b/crates/pumpkin-world/src/inventory/simple_inventory.rs
@@ -5,6 +5,7 @@ use tokio::sync::RwLock;
 
 pub struct SimpleInventory {
     pub stacks: RwLock<Vec<ItemStack>>,
+    size: usize,
 }
 
 impl SimpleInventory {
@@ -12,6 +13,7 @@ impl SimpleInventory {
     pub fn new(size: usize) -> Self {
         Self {
             stacks: RwLock::new(vec![ItemStack::EMPTY.clone(); size]),
+            size,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Snowiiii:
LGTM

PR #2856, opened 2026-08-09
fix: save block entities before their chunk is dropped

### LINKED ISSUE #2771: furnace items lost after rejoin
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0

### Operating System

win 11

items in furnace arent saved what rejoining the server 

### Steps to Reproduce

https://github.com/user-attachments/assets/f01515fa-482d-4d39-a5a6-7d2153e530b7

xRookieFight:
Fixes #2771.

`World::get_block_entity` takes the block entity's NBT *out* of the chunk (`pending_block_entities.remove`) when it wakes the entity up, so from that moment the live `World::block_entities` map is the only copy of its state. Nothing ever writes that copy back: when the chunk stops being watched, `remove_entities_in_chunks` just drops the map, and the chunk is then saved without the entity's data. Anything the entity did while it was loaded is gone - for a furnace that means the items, the burn timers and the recipes used for XP.

This mirrors what already happens for regular entities: `remove_entities_in_chunks` calls `save_entity` before dropping them, and `World::shutdown` saves them all. Block entities were missing the equivalent step, so this adds `save_block_entities`, which serializes each live block entity of a chunk with `write_internal` (the same NBT `from_nbt` reads back) and puts it into the chunk's block entity data. It runs in both places:

- `remove_entities_in_chunks`, before the chunk's entry is dropped - this covers a player walking away and the periodic `clean_memory` pass
- `World::shutdown`, for chunks that are still loaded when the server stops

`write_internal` is async and takes the lock properly, so unlike `chunk_data_nbt` it can't silently skip an inventory it failed to `try_read`.

Also adds a test that a furnace holding an item survives the `write_internal` -> `block_entity_from_nbt` round trip, which is the payload the flush relies on.

Verified with `cargo test -p pumpkin`, `cargo clippy -p pumpkin --all-targets` and `cargo fmt`. I could not drive this end to end from the console: chunks only load around players, `forceload` marks a chunk active without requesting a load, and there's no `data get block` yet, so `setblock` at an unwatched position just answers "That position is not loaded". Testing with a real client would be welcome.

### DIFF
--- a/crates/pumpkin/src/block/entities/mod.rs
+++ b/crates/pumpkin/src/block/entities/mod.rs
@@ -433,3 +433,40 @@ pub fn create_block_entity(
         _ => None,
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::{BlockEntity, block_entity_from_nbt, furnace::FurnaceBlockEntity};
+    use pumpkin_data::{item::Item, item_stack::ItemStack};
+    use pumpkin_nbt::compound::NbtCompound;
+    use pumpkin_util::math::position::BlockPos;
+    use pumpkin_world::inventory::Inventory;
+    use std::sync::Arc;
+
+    /// A loaded block entity is serialized back into its chunk with
+    /// `write_internal`, so whatever it holds has to survive that round trip or
+    /// it is gone the next time the chunk is read.
+    #[tokio::test]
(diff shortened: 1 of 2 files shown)


Snowiiii:

PR #2857, opened 2026-08-09
fix: don't deadlock when equipping an item from the hand

### LINKED ISSUE #2855: Elytra. latest version (Linux if that makes a difference)
EDIT: THIS HAPPENS WITH ANY EQUITABLE PIECE OF ARMOR OR SUCH 

### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

last commit as of download (dedce1c)

### Operating System

Bazzite (Fedora based) linux

okay so i can hold elytra
i can put on elytra through inventory
but if i right click to equip elytra it freezes the server with no errors in logs.
last commit as of download (dedce1c)

okay so i can hold elytra
i can put on elytra through inventory
but if i right click to equip elytra it freezes the server with no errors in logs.

also side addition to this bug you can toggle flight while in creative while flying with the elytra. doing so will make you lose no height or gain no height unless using normal creative flight controls. however wasd will now work during this. either could make it behave like vanilla where this just doesnt happen or make it into a addition and have wasd gain momentum.
last commit as of download (dedce1c)

okay so i can hold elytra
i can put on elytra through inventory
but if i right click to equip elytra it freezes the server with no errors in logs.

also side addition to this bug you can toggle flight while in creative while flying with the elytra. doing so will make you lose (almost)no height or gain (almost) no height unless using normal creative flight controls. however wasd will now work during this. either could make it behave like vanilla where this just doesnt happen or make it into a addition and have wasd gain momentum. 

video:
https://cdn.discordapp.com/attachments/1536018437925773342/1536018440702267473/2026-08-09_10-26-25.mp4?ex=6a79e026&is=6a788ea6&hm=3475379f68e8b307691ef71c2352ef1e3b24da2bd50def67de2619cd9997d2d0&

### Steps to Reproduce

grab elytra
put in hotbar
right click to equip
bam server freeze with no errors

xRookieFight:
Fixes #2855.

Right-clicking to equip an elytra (or any equippable) froze the whole server. `prepare_hand_item_for_use` held the `entity_equipment` lock while it called `set_stack_in_hand`, and for the off hand that method locks `entity_equipment` again - tokio's mutex isn't reentrant, so the task parked forever holding the lock and every later access to that player's equipment parked behind it.

The reason a *main* hand click ended up on the off hand path is the hand conversion. `Hand::try_from` reads the dominant-hand value from the client settings, where `0` means left; the `hand` field of a play packet uses the opposite convention, where `0` means the main hand. `handle_use_item` used `try_from` on the packet field, so every main hand use was treated as the off hand. It papered over that when reading the item (`hand == Hand::Left` picked the main hand item) but everything downstream - the equipment write-back, the active hand flag - still got the wrong hand.

So this does two things:

- Adds `Hand::from_packet_id`, documents how it differs from the settings conversion, and uses it in `handle_use_item`, which lets the item read go through `get_stack_in_hand` instead of the inverted special case.
- Drops the long-lived equipment guard: the current item is read, the lock is released, and the new item is written back in a second short lock, so nothing else is called while it's held.

`handle_use_item_on` and `handle_swing_arm` convert the same packet field the same wrong way and have their own compensations; they don't deadlock, so I left them out of this fix, but they should get the same treatment.

Verified with `cargo check`, `cargo clippy` and `cargo fmt` on `pumpkin` and `pumpkin-util`. I don't have a Java client here, so the deadlock itself is reasoned from the lock order rather than reproduced - testing an equip from both hands would be welcome.

The creative-flight-while-gliding behaviour also mentioned in the issue is a separate problem and isn't touched here.

### DIFF
--- a/crates/pumpkin-util/src/lib.rs
+++ b/crates/pumpkin-util/src/lib.rs
@@ -271,6 +271,21 @@ impl Hand {
     pub const fn all() -> [Self; 2] {
         [Self::Right, Self::Left]
     }
+
+    /// Converts the `hand` field of a play packet, where `0` is the main hand.
+    ///
+    /// This is the opposite of [`TryFrom<i32>`], which reads the dominant hand
+    /// out of the client settings, where `0` is the left hand.
+    ///
+    /// # Errors
+    /// Returns `InvalidHand` if the value is not 0 or 1.
+    pub const fn from_packet_id(value: i32) -> Result<Self, InvalidHand> {
+        match value {
+            0 => Ok(Self::Right),
+            1 => Ok(Self::Left),
+            _ => Err(InvalidHand),
+        }
+    }
 }
 
 /// Error type for invalid hand conversion.
--- a/crates/pumpkin/src/net/java/play.rs
+++ b/crates/pumpkin/src/net/java/play.rs
@@ -2521,17 +2521,13 @@ impl JavaClient {
         player.update_last_action_time();
 
         let inventory = player.inventory();
-        let Ok(hand) = Hand::try_from(use_item.hand.0) else {
+        let Ok(hand) = Hand::from_packet_id(use_item.hand.0) else {
             self.kick(TextComponent::text("InvalidHand")).await;
             return;
         };
         self.update_sequence(player, use_item.sequence.0);
 
-        let mut item_in_hand = if hand == Hand::Left {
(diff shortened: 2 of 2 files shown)


CrayzyEyezz:
WWWWWWWW

Snowiiii:

PR #2858, opened 2026-08-09
feat(build): decrease incremental build times by ~1.5x

BjornTheProgrammer:
## Description

Should decrease build times by doing some optomizations for build and dev packages, and removing some debugging information that can only be accessed using gdb.

## Testing

I ran it on my machine with all 3 changed and measured the before and after. Incremental compilation feel from 26.8s to 17s. 

## Summary by CodeRabbit

* **Chores**
  * Updated development build settings to reduce debug information and optimize dependencies.
  * Improved development build performance and output efficiency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,6 +105,19 @@ edition = "2024"
 rust-version = "1.95"
 license = "GPL-3.0"
 
+[profile.dev]
+# `line-tables-only` keeps file:line in panics, backtraces and crash-reports,
+# and only drops local-variable inspection in a step debugger.
+debug = "line-tables-only"
+
+[profile.dev.package."*"]
+opt-level = 2
+debug = false
+
+[profile.dev.build-override]
+opt-level = 2
+debug = false
+
 [profile.release]
 lto = true
 strip = "debuginfo"


Cargo.toml
```diff
@@ -105,6 +105,19 @@ edition = "2024"
 rust-version = "1.95"
 license = "GPL-3.0"
 
+[profile.dev]
+# `line-tables-only` keeps file:line in panics, backtraces and crash-reports,
+# and only drops local-variable inspection in a step debugger.
+debug = "line-tables-only"
+
+[profile.dev.package."*"]
+opt-level = 2
```
zhangxuan2011, line 114: AFAIK, using optimizations in debug mode might increase the compile time. Just asking, why add the opt level...

Cargo.toml
```diff
@@ -105,6 +105,19 @@ edition = "2024"
...
+debug = false
+
+[profile.dev.build-override]
+opt-level = 2
```
zhangxuan2011, line 118: Same as above

Cargo.toml
BjornTheProgrammer: I tested just this change and it decreased the compilation time. I believe that it optimizes the dependencies that are using macros.

Cargo.toml
BjornTheProgrammer: Same reason as above. I tested this separately as well and it decreased the compilation time.

Cargo.toml
zhangxuan2011: Okay, let me do a research about this...

I have asked doubao, seems only `pumpkin` is using opt-level=0. some dependencies, like `serde`, are using `proc-macro`, which will be run in compile time.

So, for those dependencies, using `opt-level=2` means the time of running these macros are decreased, consequently the compile time is lower than before.

Am I right? If so, i can use this skill to decrease my project's compile time as well :)

zhangxuan2011 (approved):

zhangxuan2011:
LGTM

Cargo.toml
BjornTheProgrammer: I believe that you are right 😃

Q2297045667:
GOOD!!!

Snowiiii:
@coderabbitai full review

PR #2859, opened 2026-08-09
fix(bedrock): block breaking and item crafting

ZlordHUN:
## Description
Fixes Bedrock block-breaking progress not resetting and several inventory synchronization issues affecting crafting outside of and inside the workbench.

## What
- Clears block-breaking progress when a Bedrock player stops, aborts, or changes the targeted block.
- Correctly maps Bedrock inventory and hotbar slots while a workbench is open.
- Prevents a recursive screen-handler lock that could freeze inventory interaction.
- Prevents crafting ingredients from being consumed twice.
- Uses the server’s authoritative stack IDs in item-stack responses.
- Reports workbench crafting slots using Bedrock’s expected `32–40` range.
- Keeps the cursor, crafting grid, hotbar, and inventory synchronized after taking crafted results.

## Testing
- `cargo test -p pumpkin net::bedrock::play::tests`
- Verified in-game with a Bedrock 1.26.40 client:
  - Interrupted block breaking resets immediately.
  - Items can be moved into the 3×3 workbench grid.
  - Crafted results remain in the inventory.
  - Inventory and workbench slots remain interactive after crafting without reopening the screen.

### DIFF
--- a/crates/pumpkin-inventory/src/screen_handler.rs
+++ b/crates/pumpkin-inventory/src/screen_handler.rs
@@ -157,7 +157,12 @@ pub trait InventoryPlayer: Send + Sync {
     ) -> PlayerFuture<'a, ()>;
 
     /// Sends a single slot update packet.
-    fn enqueue_slot_packet<'a>(&'a self, packet: &'a CSetContainerSlot) -> PlayerFuture<'a, ()>;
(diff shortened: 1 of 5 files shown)


Snowiiii:
Much better now, Thank you @ZlordHUN :D

PR #2868, opened 2026-08-10
refactor(deps) reduce `pumpkin-macros` dependencies

PigTurtleDev:
## Description

You may have noticed me change this PR; 1ee7be3f8c7cb2314047d4e6444f62e95fa90143 made it basically impossible to keep as is.

proc-macro dependencies are always compiled separately from normal crate dependencies. having `pumpkin-macros` depend on `pumpkin-data` therefore duplicated compiling the dependency tree below pumpkin-data.

This PR gates `pumpkin-data`'s dependencies as needed (based on enabled features), meaning `pumpkin-macros` has to far less of them than before.

This PR also declares "pumpkin" as the only `default-member` of the workspace so `cargo build` without args will no longer have to compile `pumpkin-plugin-api` & dependencies.

Due to parallel compilation, this "only" reduces (debug) compile times by 5s on my machine. Those 5s are ones of full cpu utilization though, so I presume this change has far greater effects on machines with less available parallelism or weaker hardware.

## Changes

- made `pumpkin-codegen` generate additional files for `pumpkin-macros`, because compiling the full `tag`/`block` files seems excessive
  - It's a separately compiled instance of `pumpkin-data` anyway, so using different files isn't an issue
- changed the `pumpkin-data` dependency in `pumpkin-codegen` 
  - the features only require `pumpkin-data` & `phf`, so no other crates need compiling
  - Removes ~50 dependencies on build
- added `default-members = ["crates/pumpkin"]` to `Cargo.toml`.
  - this prevents building the `pumpkin-api` crate without an explicit `-p pumpkin-api` or `cargo build --workspace`
  - Removes another ~30 dependencies on build

Dependencies for `cargo build`: 700 -> 614

## Testing

- [x] `cargo fmt --check --all`
- [x] `cargo check --workspace`
- [x] `cargo clippy --workspace`


Snowiiii:
> You may have noticed me change this PR; https://github.com/Pumpkin-MC/Pumpkin/commit/1ee7be3f8c7cb2314047d4e6444f62e95fa90143 made it basically impossible to keep as is

so is this still worth it?

PigTurtleDev:
Yeah, the effect is still the same. The change was that instead of having `pumpkin-macros` not depend on `pumpkin-data` at all it now depends on a version of `pumpkin-data` with all dependencies disabled (except `phf`). 

The dependency numbers in the PR might be off by one, I'm not sure I corrected them when I edited it.

Snowiiii:
Actually found a more "elegant" way, still thanks @PigTurtleDev :D

PR #2869, opened 2026-08-10
fix(block): apply vanilla position offsets to block shapes

MegalithOfficial:
## Description
Pumpkin extracts collision and outline shapes at `BlockPos::ZERO` and was reusing those shapes at every world position. Vanilla dynamically offsets certain blocks based on their X/Z coordinates, so the server collision could disagree with the shape rendered by the client. This was especially noticeable with bamboo, where walking between stalks caused the server to push the player back.

Though affects more than bamboo. Vanilla 26.2 applies position-based offsets to 39 blocks, including pointed dripstone and sulfur spikes for collision, along with flowers, grass and other vegetation for outlines and raycasts.

I added the vanilla offset metadata to the block data pipeline and generate the runtime lookup from it. The offset calculation uses the same coordinate hash, XZ/XYZ behavior and limits as the official server, including the smaller horizontal limit used by pointed dripstone and sulfur spikes.

Collision checks, entity interactions, placement obstruction, dismount calculations, outlines, raycasts and plugin block-state queries now use position-aware shapes. The server geometry is shared across every supported client version, while block-state remapping remains version-aware from 1.20.5 through 26.2.

## Testing
I compared the implementation directly against the official 26.2 server source and verified the complete offset-block registry, coordinate hash results and vanilla offset limits.

I tested the original bamboo issue in game and confirmed that players can now move between stalks without being pushed back.


Snowiiii:
from where that json file is from?

MegalithOfficial:
> from where that json file is from?

I made it myself from the official 26.2 server jar. I checked the blocks using offsetType in Blocks class and added them to this JSON. Mojang doesnt provide this as a JSON file.

Snowiiii:
> > from where that json file is from?
> 
> I made it myself from the official 26.2 server jar. I checked the blocks using offsetType in Blocks class and added them to this JSON. Mojang doesnt provide this as a JSON file.

This will break when a new version will be released, we use our Extractor for such things, i think we could also use Block extractor

MegalithOfficial:
Alright, The offsetType data is now fetched through the extractor.

Snowiiii:
LGTM, Thanks @MegalithOfficial

PR #2873, opened 2026-08-10
feat(block): implement bonemeal interactions

MegalithOfficial:
## Description

Bonemeal was handled directly inside the item implementation with a few hardcoded cases. Crops, saplings and grass had partial behavior, while most other supported blocks did nothing.

I moved bonemeal handling into the block behavior system. Blocks can now decide whether they are valid bonemeal targets, whether the attempt succeeds and what should happen after a successful attempt. Bonemeal is only consumed when the targeted block accepts it, and the vanilla growth particles and sound are sent through the plant growth world event.

This implements bonemeal behavior for grass blocks, crops, stems, sweet berry bushes, torchflowers, bamboo, short grass, ferns, dry vegetation and rooted dirt.

Grass blocks now use the same 128 spread attempts and position offsets as the official server. Flowers are selected from the biome's configured features using the vanilla `can_spawn_from_bone_meal` tag instead of a hardcoded flower list. The Extractor change that provides this tag is in https://github.com/Pumpkin-MC/Extractor/pull/53.

Tree and sapling generation are intentionally not included because Pumpkin's current tree generation still needs separate work.

## Testing
I compared the supported block behavior, growth amounts, success chances, grass spread offsets and placement conditions directly against the official 26.2 server implementation.

I tested grass near trees and other obstructed areas, repeated bamboo and crop growth, short grass and fern conversion, dry vegetation, and rooted dirt after repeatedly placing and breaking hanging roots.


Snowiiii:

PR #2874, opened 2026-08-10
fix(protocol): send block change acknowledgements every tick

MegalithOfficial:
## Description

Pumpkin was only sending block change acknowledgements alongside the keep-alive packet, which is sent every 15 seconds.

These acknowledgements tell the client that the server has finished processing an interaction. While waiting for one, block state changes could appear several seconds late or multiple changes could appear at once. This affected any interaction that changes a block state, including crops growing from bonemeal and rails
  connecting to each other.

I moved block change acknowledgements to the regular player tick, matching the official server. Since Pumpkin processes packets asynchronously, the sequence is kept pending until its packet handler finishes. The next player tick then acknowledges the highest completed sequence.

## Testing

I tested bonemealing farmland crops and confirmed that they grow immediately instead of updating several seconds later. I also tested placing rails next to each other and confirmed that they connect immediately.

### DIFF
--- a/crates/pumpkin/src/entity/player.rs
+++ b/crates/pumpkin/src/entity/player.rs
@@ -2097,6 +2097,10 @@ impl Player {
 
     #[expect(clippy::too_many_lines)]
     pub async fn tick(self: &Arc<Self>, server: &Server) {
+        if let ClientPlatform::Java(client) = self.client.as_ref() {
+            client.acknowledge_block_changes().await;
+        }
+
         if let Some(camera_id) = self.camera_target_id.load() {
             if camera_id == self.entity_id() {
                 self.camera_target_id.store(None);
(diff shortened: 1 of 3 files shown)


Snowiiii:
Again this is wrong, they should be already send in the World::tick function

MegalithOfficial:
> Again this is wrong, they should be already send in the World::tick function

Welp, Vanilla sends block-change acknowledgements from `ServerGamePacketListenerImpl.tick()`, which is the per-player connection tick, not the world tick. Pumpkin's World tick already calls Player tick for every player. Keeping it in Player tick also ensures acknowledgements continue during /tick freeze, when World tick is skipped but player/network ticking continues. What specifically requires this to be moved into World tick?

Issue #2875, opened 2026-08-10
Placing a block with 2+ items in hand dupes into offhand


FireAcidCow:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

2ffe470

### Operating System

Fedora 44

### Bug Description & Expected Behavior

When you are holding two blocks, placing one puts one item in your offhand and one in your main hand. Same thing happens if you have 10 blocks in your hand and place one, 9 are now in your main hand and 9 are in your offhand. Not a duplicate of #2831 because this happens without pressing the offhand keybind. I suspect this may have to do with 734f79e as the main/offhand code was updated there.

### Steps to Reproduce

1. Get 2+ blocks, such as dirt
2. Place one
3. Duplicated into your offhand

Phoenixxo:
fix seemed to be merged from #2876 , this can probably be closed...

Issue #2862, opened 2026-08-10
Infinite Oak Logs and when I place the logs automatically duplicate to my other hand


AarafGaming:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.40 (Commit: 604d597/release) - 268 Contributors

### Operating System

DietPi Server, Windows 11 Client (Docker)

### Bug Description & Expected Behavior

I place a log it decreases for a second and then goes back to 4 but the log stays placed. Also after I place the logs duplicate to my other hand.
I added a video to show what it looks like
https://github.com/user-attachments/assets/75f7b64e-5df5-4dab-8d97-520893a6994d

### Steps to Reproduce

1. Make a PumpkinMC server
2. Log on
3. Get Oak Logs
4. Place them

ZlordHUN:
Doesnt seem limited to oak logs. All blocks are affected in my case. When I pick up a stack of blocks it immediately goes to my off hand instead of my main hand, duplicating the stack.

AarafGaming:
> Doesnt seem limited to oak logs. All blocks are affected in my case. When I pick up a stack of blocks it immediately goes to my off hand instead of my main hand, duplicating the stack.

I havent tested other blocks yet

Issue #2831, opened 2026-08-08
Block Dupe


Us3ful-dev:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

ee6e047

### Operating System

Linux Mint 22.3 - Xfce 64-bit

### Bug Description & Expected Behavior

Blocks can get duped, which should not be possible.

### Steps to Reproduce

1. get a block in your main hand
2. place the block on the ground
3. press offhand hotkey at the same time (as if you are attribute swapping)

Phoenixxo:
Fixed in #2876 , can probably be closed

Us3ful-dev:
> Fixed in #2876 , can probably be closed

Thx, hadn't realized.

PR #2876, opened 2026-08-11
fix(inventory): decode block-placement and swing hands via from_packet_id

2982136527:
## Summary

Fixes the "infinite items / duplication into the off-hand" bug reported in #2831, #2862 and #2875.

When placing a block (or swinging), the packet's `hand` field was decoded with `Hand::try_from`, which maps `0 -> Hand::Left`. That conversion is meant for the client's **dominant-hand setting** (`0 = left hand`). But the packet field uses `0 = main hand`, so a main-hand placement produced `Hand::Left` — which means **off-hand** in the inventory API.

Result: the decremented stack was written back to the **off-hand** instead of the main hand. The main-hand slot was never actually decremented on the server, so:
- blocks only "consumed" once visually, then stayed at the same count (infinite blocks),
- a copy of the stack appeared in the off-hand (duplication),
- swing animation fired on the wrong arm.

## Changes

- Use `Hand::from_packet_id` (added in #2857) for the block-placement (`use_item_on`) and swing (`swing_arm`) packet handlers.
- Read the in-hand item via `inventory.get_stack_in_hand(hand)` and flip the main-hand checks from `Hand::Left` to `Hand::Right`, matching the existing `use_item` handler.

## Testing

Verified with a live client on a fresh world: placing sand, dirt and logs decrements the stack on every placement and no longer duplicates into the off-hand. `cargo check -p pumpkin` passes.

### DIFF
--- a/crates/pumpkin/src/net/java/play.rs
+++ b/crates/pumpkin/src/net/java/play.rs
@@ -1352,7 +1352,7 @@ impl JavaClient {
         swing_arm: SSwingArm,
     ) {
         player.update_last_action_time();
-        let Ok(hand) = Hand::try_from(swing_arm.hand.0) else {
+        let Ok(hand) = Hand::from_packet_id(swing_arm.hand.0) else {
             self.kick(TextComponent::text("Invalid hand")).await;
             return;
         };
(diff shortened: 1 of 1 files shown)


zhangxuan2011 (approved):

zhangxuan2011:
LGTM

AarafGaming:
nice

Snowiiii:
LGTM, Thank you @2982136527 :D

Issue #1096, opened 2025-07-31
Random block tick tracking


teknostom:
Here is an issue to track current progress on random tick.

Random ticks seem to still cause deadlocks, so to test you have to uncomment 670-682 in `pumpkin/src/world/mod.rs` to develop with them.

- [x] BambooBlock
- [x] OxidizableSlabBlock
- [x] OxidizableStairsBlock
- [x] CactusBlock
- [x] BambooShootBlock
- [x] NetherPortalBlock
- [x] OxidizableGrateBlock
- [x] BuddingAmethystBlock
- [x] NyliumBlock
- [x] BeetrootsBlock
- [x] CocoaBlock
- [x] RedstoneOreBlock
- [x] PointedDripstoneBlock
- [x] SweetBerryBushBlock
- [x] LeavesBlock
- [x] SugarCaneBlock
- [x] FlowerPotBlock
- [x] PropaguleBlock
- [x] SpreadableBlock
- [x] MushroomPlantBlock
- [x] TurtleEggBlock
- [x] OxidizableTrapdoorBlock
- [x] CropBlock
- [x] EyeblossomBlock
- [x] DriedGhastBlock
- [x] NetherWartBlock
- [x] SnowBlock
- [x] OxidizableBulbBlock
- [x] IceBlock
- [x] SaplingBlock
- [x] OxidizableDoorBlock
- [x] ChorusFlowerBlock
- [x] VineBlock
- [x] FarmlandBlock
- [x] StemBlock
- [x] OxidizableBlock

PR #2878, opened 2026-08-11
fix(block): use random ticks for redstone ore

Phoenixxo:
## Description

Use the random block tick callback to extinguish lit redstone ore and
deepslate redstone ore instead of scheduling a shortened deterministic tick.

This aligns the behavior with the random-tick tracking work.

Part of #1096.

## Testing

- `cargo check --locked -p pumpkin`
- Manually verified that redstone ore lights on interaction and later
  extinguishes through random ticking


Snowiiii:
LGTM, Thank you @Phoenixxo :D

PR #2886, opened 2026-08-11
feat(block): implement cocoa block behavior

Phoenixxo:
## Description

- Registers dedicated block behavior for cocoa pods.
- Restricts placement to horizontal faces of blocks that support cocoa.
- Removes cocoa when its supporting block is removed.
- Advances cocoa growth through random ticks.
- Makes bone meal advance cocoa by one stage, up to its maximum age of 2 (as per the Java code).

## Testing
- All passed: 
  - `cargo check -p pumpkin`
- Manually verified placement on supported and unsupported blocks.
- Manually verified unrelated neighbor updates preserve cocoa and removing its support removes it.
- Manually verified natural growth from age 0 through age 2.
- Manually verified bone meal advances one stage at a time and does nothing at maximum age.


Phoenixxo (commented):
These changes match vanilla 26.2 cocoa growth more closely by using an exact bounded 1-in-5 random roll and listener-only state updates for both natural and bonemeal growth. This removes modulo bias, ensures clients receive the updated growth state, and avoids unnecessary neighbor updates.

Wanted to bump this PR and make sure it's seen @Snowiiii

PR #2888, opened 2026-08-11
feat(bedrock): add support for vibrant visuals

ZlordHUN:
## Description
Adds the Bedrock biome metadata required for Vibrant Visuals rendering.

## What
- Implements the Bedrock `BiomeDefinitionList` packet.
- Sends biome definitions during Bedrock world initialization.
- Includes definitions for all 88 Bedrock 1.26.40 vanilla biomes, including:
  - Temperature and downfall
  - Water and foliage colors
  - Rain support
  - Biome depth and scale
  - Biome tags
  - Optional chunk-generation metadata
- Bundles the biome registry locally, requiring no external service or runtime download.

## How
- Uses the gzip-compressed `biome_definitions.nbt` registry for Bedrock 1.26.40.
- Adds a `pumpkin-codegen` stage that validates the NBT and converts it into Bedrock’s current `BiomeDefinitionList` wire format.
- Generates a static, pre-encoded packet payload in `pumpkin-data`.
- Sends the generated payload directly, avoiding NBT decompression, parsing, allocation, and encoding when clients connect.
- Preserves the registry’s biome ordering, shared strings, tags, and optional generation fields.
- Causes malformed or incompatible registry data to fail during code generation rather than when a Bedrock player joins.
- Sources the registry from `Kaooot/bedrock-network-data`, with attribution included in the repository.
- Replaces the previous JSON asset and removes the now-unused direct `serde_json` dependency from `pumpkin-protocol`.

## Testing
- `cargo test -p pumpkin-protocol biome_definition`
- Connected to Pumpkin with Vibrant Visuals enabled. The visuals were indeed vibrant.


assets/bedrock/biome_definitions.json
```diff
@@ -0,0 +1,2543 @@
+{
```
xRookieFight: Would be better if you can use the .nbt file, more efficient in my opinion. I can give the .nbt file to you, i have it in my computer

xRookieFight (changes requested):

assets/bedrock/biome_definitions.json
ZlordHUN: Oh okay, please do :D

Snowiiii:
I think it would be worth to have that pre-parsed in pumpkin codegen similar to the registry nbt. this is safer and will be much faster

ZlordHUN:
@Snowiiii Good enough?

Snowiiii:
Yeah, also looks much better now in game, Thank you @ZlordHUN :D

PR #2892, opened 2026-08-12
fix(plugin): apply WASM sent packet payloads

Bluezly:
## Description

Apply raw payload changes returned by blocking WASM packet sent handlers.

Outgoing Java and Bedrock packet paths now queue the payload returned by `PacketSentEvent` instead of always sending the originally serialized bytes. The existing public cancellation helper remains unchanged.

No WIT changes. `packet-id` remains metadata in this change; raw payload replacement uses the existing serialized clientbound payload.

### DIFF
--- a/crates/pumpkin/src/entity/player.rs
+++ b/crates/pumpkin/src/entity/player.rs
@@ -2385,18 +2385,26 @@ impl Player {
         false
     }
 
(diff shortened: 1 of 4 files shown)


Snowiiii:
LGTM, Thank you @Bluezly :D

@vyPal i remember you wanted this

PR #2903, opened 2026-08-12
fix(bedrock): restore NetherNet LAN discovery

ZlordHUN:
## Description
Restores NetherNet-native LAN discovery and signaling for Bedrock clients.

## What
- Advertises Pumpkin in Bedrock’s LAN Worlds list through NetherNet discovery.
- Includes the server name, world name, game mode, player counts, online-mode state, and transport information.
- Handles LAN connection requests and ICE candidate signaling.
- Prevents one Pumpkin instance from appearing twice in the LAN Worlds list.
- Fixes LAN connections failing with `Door` while direct-IP connections work.

## How
- Listens for encrypted NetherNet discovery traffic on UDP port `7551`.
- Responds using Bedrock 1.26.40’s discovery packet and server-data formats.
- Routes LAN SDP and ICE signaling through Pumpkin’s existing NetherNet negotiation path.
- Uses Pumpkin’s stable server GUID as both the advertised server identity and NetherNet network ID.
- Ensures RakNet status and NetherNet discovery identify the same server, allowing clients to deduplicate the entries and target the correct signaling recipient.

## Testing
- `cargo test -p pumpkin nethernet::discovery`
- Verified the encoded discovery request, response, server data, and signaling message round trips.
- Confirmed direct-IP connections remain unaffected.

### DIFF
--- a/crates/pumpkin/src/lib.rs
+++ b/crates/pumpkin/src/lib.rs
@@ -11,7 +11,7 @@ use crate::data::VanillaData;
 use crate::logging::{GzipRollingLogger, PumpkinCommandCompleter, ReadlineLogWrapper};
 use crate::net::bedrock::{
     BedrockClient,
-    nethernet::{NetherNetListener, load_or_create_identity_key},
+    nethernet::{NetherNetListener, discovery::NetherNetDiscovery, load_or_create_identity_key},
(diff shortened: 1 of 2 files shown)


xRookieFight (approved):
LGTM

PR #2907, opened 2026-08-12
fix(world): release chunk holders, dependency edges and tickets during travel

xRookieFight:
## Description

Chunk holders, DAG edges and loading tickets were never released when a player travelled fast enough to outrun chunk generation, and resident memory never came back down afterwards. Measured with a throwaway protocol-level load client that logs in and flies at a set chunk rate.

**1. Chunks are never released once they have served as a generation dependency.**

`ChunkHolder::dependency_stage` is a high-water mark - `ensure_dependency_chain` raises it, nothing lowers it. Since `effective_target = target_stage.max(dependency_stage)`, such a holder is never queued for unload and its tasks are never cancelled. `garbage_collect_dependencies` can't recover: the stranded holders are mutually reachable, so the pass has no seed.

`recompute_dependency_stages` rebuilds the requirement from the live `target_stage` values instead, as a worklist closure over `get_direct_radius` / `get_direct_dependencies`, run once per second alongside the existing GC. Max direct radius is 1, so cost scales with the loaded set. `purge_dropped_queue_entries` drops heap entries whose node was cancelled, and the periodic unload pass now runs even when `unload_chunks` is empty, since the GC is what fills that queue.

**2. A holder's `occupied_by` edge chain is leaked when the holder unloads.**

`ensure_dependency_chain` pushes an edge onto `holder.occupied_by` to record which task is waiting on this chunk's data; `process_unload_queue` dropped the holder without freeing the chain. Unlike the edges reachable from `Node::edge` these carry no `in_degree`, so freeing them cannot unblock anything. At 40 chunks/s this leaked ~37k edges/s.

**3. `ChunkManager` never released its simulation ticket.**

`update_center_and_view_distance` adds a view ticket and a simulation ticket, but `clean_up` and `change_world` only removed the view one, so every disconnect and every dimension change stranded a ticket and the ~440 chunks under it. It now records the pair it added and drops exactly that, which also fixes the view ticket being recomputed from a possibly-changed view distance.

**4. Freed chunk buffers were not returned to the OS.**

With 1-3 fixed, a counting `GlobalAlloc` shows the live heap flat at 83 / 95 / 104 MB after each of three flights, and every chunk-system structure drained to zero - but RSS still went 721 -> 1139 -> 1256 MB. Nothing was retained; the system allocator was holding the freed buffers. Switching the global allocator to mimalloc returns them.

## Testing

`cargo test -p pumpkin-world` (146 pass), `cargo clippy --all-targets --all-features`, `cargo fmt --check`.

Chunk-system internals over a 40 chunks/s flight at view distance 16:

| | `chunk_map` | `DAG::edges` | task queue |
|---|---|---|---|
| before | 1089 -> 6887 | 47k -> 5.27M | 1082 -> 13793 |
| after | ~3900 peak, 0 when idle | ~160k peak, 0 when idle | ~3800 peak, 0 when idle |

`unload_chunks` was empty for the whole "before" run, which is what pointed at the dependency accounting. After a disconnect the level now drains to zero tickets and zero loaded chunks; previously one ticket and ~440 chunks stayed resident.

Resident memory after each of three 150-second flights, and time for a freshly joined client to receive 1000 chunks:

| | RSS after flights 1 / 2 / 3 | time to 1000 chunks |
|---|---|---|
| master | - | 18.0 / 19.9 / 19.4 s |
| fixes 1-3, system allocator | 721 / 1139 / 1256 MB | 18.5 / 20.9 / 21.4 s |
| fixes 1-3 + mimalloc | 828 / 664 / 676 MB, 683 / 470 / 750 MB | 15.8 / 17.8 s |

One intended behaviour change: because chunks now actually unload, backtracking re-reads them from disk. The one-second unload batching still covers short backtracking.

The allocator swap is a separate commit so it can be dropped independently. Without it the leak fixes still hold - every chunk-system structure drains - but resident memory keeps climbing on glibc.

## Known issue, not addressed here

When all generation slots are busy, `work()` pushes the task back and breaks out of the dispatch loop, but the only blocking wait is gated on `self.queue.is_empty()`, which is false at that point, so the scheduler thread spins on a full core for as long as generation stays behind (59560 ms of CPU over a 60 s run). This pre-dates the PR. Blocking on `recv_chunk.recv_timeout` there fixes the spin but cost ~30% chunk throughput in the benchmark above, so it needs a better answer than a plain blocking wait.


Snowiiii:
Should be fixed now, thank you @xRookieFight for your effort

PR #2908, opened 2026-08-12
fix(entity): clamp health to 1 when max health is set below 1

### LINKED ISSUE #2883: set-max-health allows values below 1
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

cda8f19

### Operating System

Fedora 44

When you tell a plugin to set a player's max health to a value below 1, it works. This causes the player to die the instant they join/respawn. In vanilla, it accepts a value below 1, but the player's health is capped at 1 instead.

### Steps to Reproduce

1. Set player's max health to 0 or below via a plugin. One has been attached. Use the command /no_hp
```rust
use pumpkin_plugin_api::{self, Context, Plugin, PluginMetadata, Server, command::{Command, CommandError}, commands::CommandHandler, permission::{Permission, PermissionDefault}};

struct CommandTesting;
impl Plugin for CommandTesting {
    fn new() -> Self { CommandTesting }

    fn metadata(&self) -> PluginMetadata {
        PluginMetadata {
            name: "Command Testing".into(),
            version: env!("CARGO_PKG_VERSION").into(),
            authors: vec!["FireAcidCow".into()],
            description: "set player max health to 0".into(),
            dependencies: vec![],
            permissions: vec![]
        }
    }

    fn on_load(&mut self, context: Context) -> pumpkin_plugin_api::Result<()> {
        context.register_permission(&Permission {
            node: "Command Testing:use".into(),
            description: "Allows running /hello".into(),
            default: PermissionDefault::Allow,
            children: vec![],
        })?;

        let mut command = Command::new(&["no_hp".into(), "a".into()], "proof of concept for hp issue");
        command = command.execute(Command1); 
        context.register_command(command, "Command Testing:use");

        Ok(())
    }
}

struct Command1;
impl CommandHandler for Command1 {
    fn handle(&self,sender: pumpkin_plugin_api::command::CommandSender,_server: Server,_args: pumpkin_plugin_api::command::ConsumedArgs,) -> pumpkin_plugin_api::Result<i32, CommandError>
    {
        sender.as_player().unwrap().set_max_health(0.0);
        Ok(1)
    }
}


pumpkin_plugin_api::register_plugin!(CommandTesting);
```

eacherey:
## Description

Clamps current health to a minimum of 1 when `set_max_health` is called with a value below 1, matching vanilla Minecraft behavior.

- The `max_health` attribute itself updates to the requested value (including 0 or negative), which vanilla allows.
- Only the living/current health is capped at 1 instead of being clamped to `[0, max_health]`.
- This prevents players from instantly dying on join/respawn when a plugin sets their max health below 1.

Closes #2883.

## Testing

- `cargo check -p pumpkin`
- `cargo clippy --all-targets`
- `cargo test`
- Manually verified setting max health to 0 via plugin no longer kills the player on respawn.
- Manually verified setting max health to 20 restores normal health behavior.
- Manually verified setting max health to 0.5 keeps the player alive at 1 HP.


zhangxuan2011 (approved):
LGTM, thank you for your hard work!

eacherey:
accidentally

zhangxuan2011:
> accidentally

What's going on bro😁

RoosterBooster007:
This fix works, but it's patching the symptom at the health layer rather than where vanilla enforces it. Vanilla clamps per-attribute values in the attribute registry.

PR #2918, opened 2026-08-13
fix(bedrock): advancement translations

### LINKED ISSUE #2901: Advancements show up in unparsed form
### Before Submitting

### Minecraft Edition

Bedrock Edition

### Server Version / Commit

7eb49e2

### Operating System

Nintendo Switch 2

When you get a java advancement, it sends `{PLAYERNAME} just earned the achievement %advancements.adventure.sleep_in_bed.title` (note that the playername does show up correctly) as opposed to the java `{PLAYERNAME} has made the advancement [Sweet Dreams]` (hoverable). This should show up as a parsed string for bedrock clients as it can't be displayed correctly.

### Steps to Reproduce

1. `/time set night`
2. `/give @s bed`
3. Place the bed
4. Sleep in the bed

ZlordHUN:
## Description
Add localized Bedrock advancement announcements and toast notifications using an embedded resource pack.

## Changes
- Adds an embedded Bedrock resource pack containing:
  - Java advancement title translations.
  - Bedrock advancement toast frame labels.
- Serves the pack through Bedrock’s resource-pack protocol without external downloads.
- Adds resource-pack metadata, chunk transfer, and chunk-request handling.
- Uses Bedrock’s UUID wire format and correct resource-pack hash serialization.
- Keeps advancement chat announcements client-side using Bedrock rawtext translation components.
- Resolves toast title and content on the server using the Bedrock player’s locale, since `CToastRequest` carries plain strings.
- Waits for the Bedrock player-initialized acknowledgement before sending advancement feedback.
- Stops waiting when the client closes before initialization.
- Sends the appropriate toast label for task, goal, and challenge advancements.
- Explicitly persists advancement data for players on disconnect using the existing advancement storage.
- Keeps Java advancement evaluation and Java chat announcements intact.

## Testing
- `cargo check -p pumpkin`
- `cargo test -p pumpkin advancement --lib`
- `cargo test -p pumpkin-protocol --lib`
- Verified a fresh resource-pack download in game.
- Verified localized advancement chat messages and toast notifications.
- Verified advancement progress remains completed after disconnecting and reconnecting.

## Summary by CodeRabbit

* **New Features**
  * Added Bedrock advancement translations for supported languages.
  * Bedrock players now receive localized advancement completion toasts.
  * Added chunked transfer support for the advancement translation pack.
  * Advancement progress is saved when Bedrock players disconnect.
  * Added support for the `nb_NO` Norwegian Bokmål locale.

* **Bug Fixes**
  * Corrected UUID encoding and decoding to match Bedrock’s expected byte order.

* **Documentation**
  * Documented the advancement translation pack and its update process.


Snowiiii:
This kinda does not make sense cause now there is no translation at all and its just sending plain English

ZlordHUN:
That is definitely my bad. Probably going to take a break after this, getting sloppy.

ZlordHUN:
@Snowiiii @QuantumSegfault ready for re-review

crates/pumpkin-util/src/translation.rs
```diff
@@ -547,6 +559,9 @@ impl FromStr for Locale {
 
     #[expect(clippy::too_many_lines)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.eq_ignore_ascii_case("nb_no") {
+            return Ok(Self::NoNo);
```
QuantumSegfault: Why not put this in the match??? It's already compared lowercase. If it's not an already supported locale, add it in there.

QuantumSegfault (commented):
I'm not particularly acquainted with the translation system, but looks good for the most part.

Just one question

QuantumSegfault (approved):

Snowiiii:
1. Are you sure there are no client-side translations for advancements, would kinda not make sense since there is a translation system, 
or if there are no advancements send in chat in bedrock we might also should not send them to bedrock players
2. Also don't add random files to assets, we need to know from where they are from and how we can update them when a new mc version comes out

ZlordHUN:
I ripped the asset for the languages supported by bedrock but I did find some repos with the same file - more or less. They had a file for each language. If necessary those files can be used but I'll look into client side translations first. If client side translations are out would you prefer I add the advancement translation files from an external repo? @Snowiiii

ZlordHUN:
So, update:
Bedrock supports translations, but it doesn’t include the translation keys for Java advancements. 
The Bedrock client translates the surrounding message, while Pumpkin translates the advancement name itself. I also documented exactly where the translation files come from and how to update them for future Minecraft versions.

I believe this is the best I can do right now. Couldn't find any better alternatives but am open to suggestions.

Snowiiii:
There has to be a better way..., From my understanding it still translates server side and sends it based on player local, Ideally we would be able to send translations keys for advancements like before

**But**, iirc Bedrock advancements are Never broadcasted to the in-game chat anyway

ZlordHUN:
> There has to be a better way..., From my understanding it still translates server side and sends it based on player local, Ideally we would be able to send translations keys for advancements like before
> 
> **But**, iirc Bedrock advancements are Never broadcasted to the in-game chat anyway

I'll check for alternatives and get back to you. To be completely honest the purpose of this PR was just to fix the issue that was reported.

ZlordHUN:
@Snowiiii I've done the research, the only alternative would be to distribute a resource pack with the necessary missing java advancement keys but I don't think that would be better and I dont even know how to begin approaching that.

Snowiiii:
> @Snowiiii I've done the research, the only alternative would be to distribute a resource pack with the necessary missing java advancement keys but I don't think that would be better and I dont even know how to begin approaching that.

Let's just don't send advancements in chat to bedrock players, like in vanilla bedrock

QuantumSegfault:
> > @Snowiiii I've done the research, the only alternative would be to distribute a resource pack with the necessary missing java advancement keys but I don't think that would be better and I dont even know how to begin approaching that.
> 
> Lets just don't send advancements in chat to bedrock players, like in vanilla bedrock

As a Bedrock player, I would hate to miss out.

I was looking forward to Java advancements as substitute for achievements (which require XBL login)

ZlordHUN:
Yeah same here plus the achievements can only be unlocked once per profile while advancements can be unlocked per world. Would be a bummer to miss out.

Snowiiii:
> > > @Snowiiii I've done the research, the only alternative would be to distribute a resource pack with the necessary missing java advancement keys but I don't think that would be better and I dont even know how to begin approaching that.
> > 
> > 
> > Lets just don't send advancements in chat to bedrock players, like in vanilla bedrock
> 
> As a Bedrock player, I would hate to miss out.
> 
> I was looking forward to Java advancements as substitute for achievements (which require XBL login)

I saw there an achievement packet send to Bedrock players, we should use that instead?

ZlordHUN:
@Snowiiii I reworked this PR. It now uses a resource pack instead of the translation asset. Writing up a PR description right now.

ZlordHUN:
## Description
Fixes advancement announcements for Bedrock players while keeping translation client-side.

## What
- Displays localized Java advancement titles in Bedrock chat.
- Shows Bedrock toast notifications when advancements are completed.
- Preserves advancement progress across disconnects and restarts.
- Prevents raw translation keys from appearing during the first connection.
- Supports every locale included in the bundled translation pack.
- Does not change Java advancement behavior.

## How
- Bundles the missing Java advancement title keys in a small Bedrock resource pack.
- Transfers the pack through Bedrock’s resource-pack protocol without external downloads.
- Uses Bedrock `rawtext` translation components so each client selects its own language.
- Waits for Bedrock’s explicit player-initialized acknowledgement before sending announcements, ensuring newly downloaded translations are active.
- Saves Bedrock advancement progress through Pumpkin’s existing advancement storage.
- Uses Bedrock’s correct UUID wire encoding for resource-pack identifiers.
- Removes the previous server-side advancement translation table.

## Testing
- `cargo check -p pumpkin`
- `cargo test -p pumpkin advancement --lib`
- `cargo test -p pumpkin-protocol --lib`
- Verified a fresh resource-pack download in game.
- Verified localized advancement chat messages and toast notifications.
- Verified advancement progress remains completed after disconnecting and reconnecting.

crates/pumpkin-protocol/src/codec/uuid.rs
```diff
@@ -6,6 +6,31 @@ use crate::serial::PacketWrite;
 
 impl PacketWrite for Uuid {
     fn write<W: Write>(&self, writer: &mut W) -> Result<(), Error> {
-        writer.write_all(self.as_bytes())
+        let uuid = self.as_u128();
+        writer.write_all(&((uuid >> 64) as u64).to_le_bytes())?;
+        writer.write_all(&(uuid as u64).to_le_bytes())
```
QuantumSegfault:
```suggestion
        ((uuid >> 64) as u64).write(writer)?;
        (uuid as u64).write(writer)
```

crates/pumpkin-protocol/src/bedrock/client/resource_pack_data_info.rs
QuantumSegfault, line 1: Any reason to not using `String` and `Vec` here?

crates/pumpkin-protocol/src/bedrock/client/toast_request.rs
QuantumSegfault, line 1: Ditto (why not just owning `String`)

crates/pumpkin-protocol/src/bedrock/client/resource_pack_chunk_data.rs
QuantumSegfault, line 1: Here...borrowing the `chunk_data` probably DOES make sense. And since we have to have a manual decoder anyway, I suppose `&'a str` doesn't hurt.

assets/bedrock/advancement_translations.mcpack
QuantumSegfault, line 1: What's with the `"capabilities": ["pbr"]` in the manifest? Is that necessary?

crates/pumpkin-protocol/src/bedrock/client/resource_pack_data_info.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub resource_name: String,
+    pub chunk_size: u32,
+    pub number_of_chunks: u32,
+    pub file_size: u64,
+    pub file_hash: Vec<u8>,
+    pub is_premium_pack: bool,
+    pub pack_type: u8,
+}
+
+impl PacketWrite for CResourcePackDataInfo {
```
QuantumSegfault, line 18: This can be replaced with `derive`.

crates/pumpkin/src/entity/player/advancement.rs
```diff
@@ -424,6 +426,82 @@ impl PlayerAdvancement {
...
+                let title_key = match display.frame_type {
+                    FrameType::Task => "advancements.toast.task",
+                    FrameType::Challenge => "advancements.toast.challenge",
+                    FrameType::Goal => "advancements.toast.goal",
+                };
+                player
+                    .send_packet(&CToastRequest {
+                        title: TextComponent::translate(title_key, []).get_text(),
+                        content: display.get_title().get_text(),
+                    })
```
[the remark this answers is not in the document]
QuantumSegfault: Coderabbit has a point here...I looked a little deeper and it's comment seems valid. The toast will end up being sent always-English. Did you test this part more? With other languages.

QuantumSegfault (commented):
It seems wrong to be hardcoding the advancement resource pack like this. Feels like we should have a mechanism for discovering available resource packs at runtime from disk and a builtin registry and grabbing name, uuid, etc. from the pack itself.

But that can always be improved later.

assets/bedrock/advancement_translations.mcpack
ZlordHUN: I believe I added it cause without it, this implementation breaks Vibrant Visuals. Players can still have it enabled and join but it just wont actually be applied.

crates/pumpkin-util/src/translation.rs
```diff
@@ -589,7 +589,14 @@ impl FromStr for Locale {
 
     #[expect(clippy::too_many_lines)]
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s.to_lowercase().as_str() {
+        let lowercase = s.to_lowercase();
+        // Bedrock spells Norwegian Bokmål `nb_NO`, which shares Java's `no_no` translations.
+        let code = if lowercase == "nb_no" {
```
QuantumSegfault, line 594: I feel I've mentioned this before...

Why not just have `nb_no` inside the match. No need for special casing here.

crates/pumpkin/src/entity/player/advancement.rs
```diff
@@ -424,6 +443,74 @@ impl PlayerAdvancement {
...
+                .level_info
+                .load()
+                .game_rules
+                .show_advancement_messages;
+        if !display.show_toast && !announce_to_chat {
+            return;
+        }
+
+        tokio::spawn(async move {
+            if let ClientPlatform::Bedrock(client) = player.client.as_ref() {
```
QuantumSegfault, line 462: Is it really worth the effort to spawn a task and WAIT for initialization? If a player isn't fully loaded and ready, why not just drop the message (in the unlikely case they earn an advancement while still loading into the same)?

QuantumSegfault (commented):
This is going to sound dumb...but...why the resource pack, again???

What is the actual value of doing this client-side, if toast messages are still translated server-side anyway? Does it help anything?

Issue #2810, opened 2026-08-07
Plugin API: world-editing follow-ups for WorldEdit-style plugins


darkstarworks:
### Context

The bulk world-editing operations being added to the plugin `world` API (`fill-blocks`, `get-blocks`, `set-region`, `set-blocks`) already cover plain-block WorldEdit-style work: cuboid and arbitrary-shape selections (the plugin computes the position set and calls `set-blocks`), undo/redo (snapshot with `get-blocks`, restore with `set-region`), masks and patterns (read, decide in-plugin, write), and copy/paste of plain blocks.

This issue collects the follow-up work needed for fuller WorldEdit-style support so the maintainers can prioritize. None of these block the initial bulk ops; they build on top of them.

### Follow-ups

- [ ] **Block-entity NBT in clipboards.** `get-blocks`/`set-region` carry only block-state ids today, so copy/paste loses chest contents, sign text, spawner config, banners, and the like. A sparse pair such as `get-block-entities(start, end) -> list<(block-pos, nbt)>` plus `set-block-entities(list<(block-pos, nbt)>)` would fix this. Depends on generic block-entity NBT read/write being available (a read-side counterpart to the in-progress `set-block-entity-nbt`).
- [ ] **Entities in a region.** WorldEdit copies armor stands, item frames, paintings, and so on. Needs entity serialization; larger and separate.
- [ ] **Bulk set-biome.** Only `get-biome` exists today; a `set-biome` / `fill-biome` would enable the equivalent of `//setbiome`. Small and self-contained.
- [ ] **Block-state rotate/flip helpers.** Transforming a clipboard means remapping directional block states. A plugin can do this itself given block-property data (overlaps with the proposed blocks-query interface).
- [ ] **Chunk-section-level bulk writes for performance.** The current ops go through `set-block-state` per block. A FastAsyncWorldEdit-style path that writes chunk sections directly would be needed for multi-million-block operations. Deeper, engine-level work.

### Notes

WorldGuard-style region protection is a separate concern (event interception and cancellation) and is not covered here.

PR #2919, opened 2026-08-13
feat(plugin): add bulk world editing ops to the world API

darkstarworks:
## Description

Adds bulk block editing to the plugin world API so plugins can read and write regions of blocks in a single call instead of one block at a time. New functions on the world resource:

- fill-blocks: set every block in a cuboid to one state.
- get-blocks: read the state ids of a cuboid, ordered by x, then z, then y.
- set-region: write a flat array of state ids back into a cuboid, in the same order as get-blocks, so a snapshot read with get-blocks round-trips through set-region.
- set-blocks: set many blocks from an explicit list of position and state pairs.

Each operation is capped at 2097152 blocks, yields to the async runtime every few thousand blocks so a large call cannot starve the server tick, and validates every block state before changing anything so an invalid entry cannot leave a region half updated.

This depends on the WIT interface change in Pumpkin-MC/pumpkin-plugin-wit#31. The submodule pointer here targets that branch, so this is a draft until the WIT PR merges and the pointer can move to the merged commit.

Further work (block-entity handling, entities, biomes, rotate and flip, chunk-section performance) is tracked in #2810.

## Testing

cargo fmt --all --check, cargo clippy --all-targets --all-features, typos, and cargo test --workspace all pass. New unit tests cover the region iteration order, inclusive volume counting, saturating behaviour on extreme spans, and the block limit.

### DIFF
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
@@ -66,7 +66,7 @@ use crate::plugin::loader::wasm::wasm_host::wit::v0_1::pumpkin::plugin::game_rul
 use crate::plugin::loader::wasm::wasm_host::wit::v0_1::pumpkin::plugin::world::{
     BlockDirection as WitBlockDirection, BlockEntity, BlockEntityType, BlockFlags as WitBlockFlags,
     BlockPos as WitBlockPos, BlockState as WitBlockState, BlockStateInfo as WitBlockStateInfo,
-    BoundingBox as WitBoundingBox, Chunk as WitChunk,
+    BlockUpdate as WitBlockUpdate, BoundingBox as WitBoundingBox, Chunk as WitChunk,
     NoteblockInstrument as WitNoteblockInstrument, PistonBehavior as WitPistonBehavior,
     WorldBorder as WitWorldBorder,
 };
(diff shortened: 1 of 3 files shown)


crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
```diff
@@ -243,6 +243,99 @@ impl pumpkin::plugin::world::Host for PluginHostState {
...
+/// This is a safety limit, not a game-mechanics one: it bounds how big a
+/// `get-blocks` result can get (2 bytes per block copied into the plugin's
+/// memory) and guards against a single call asking for an unreasonable amount
+/// of work. It is far larger than vanilla's `/fill` limit; larger jobs should
+/// be split across several calls.
+const MAX_BULK_BLOCKS: u64 = 2_097_152;
+
+/// How many blocks a bulk operation processes before yielding back to the async
+/// runtime, so a large operation cannot starve the server tick.
+const BULK_YIELD_INTERVAL: u64 = 4_096;
```
vyPal: Any reason for these two specific values in particular? It might also not be a bad idea to have these values determined by config values, since (at least for the `BULK_YIELD_INTERVAL`), more performant servers can handle more blocks in a shorter time and yielding too frequently could actually hurt performance.

Edit: the `BULK_YIELD_INTERVAL` could probably be bumped up to at least around 16-64k without being an issue

vyPal: Might also be worthwhile to get @Snowiiii's opinion on this

vyPal (changes requested):
The API changes look pretty good overall, but it would be nice if you'd also extend the plugin-api crate to provide helpers/wrappers around the `get_region()`/`set_region()` methods as well. Interacting with the raw array would be painful, so most users will use some helper methods to allow for easier getting/setting of a specific block using it's coordinates relative to the origin of the region. You could maybe check out #2351 as a reference/inspiration

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
darkstarworks: Good points, thanks. `MAX_BULK_BLOCKS` is a memory bound rather than a mechanics one: `get-region` returns a `list<u16>` (2 bytes per block), so 2^21 caps the copy into plugin memory at 4 MiB. You're right that `BULK_YIELD_INTERVAL` at 4096 was too eager, I've bumped it to 32768 and clarified both comments. Making them config-driven seems reasonable as a follow-up, happy to defer to @Snowiiii on whether that's wanted.

darkstarworks:
Added the ergonomic helpers in a new `region` module on the plugin-api crate: a `Region` type that wraps the flat array so plugins `get`/`set` by absolute or relative coordinates and iterate `(BlockPos, state)`, plus a `WorldExt` trait with `read_region`/`write_region`. It uses the same x/z/y ordering as the host, with a unit test to keep the two in sync. Thanks for the suggestion!

darkstarworks:
@vyPal this should be ready to merge now: both points are addressed (get-blocks renamed to get-region, the constants documented and BULK_YIELD_INTERVAL bumped to 32k, and the ergonomic `Region` helpers added to the plugin-api crate), rebased onto current master and green. Mind giving it another look?

darkstarworks:
## Closed with reason:

PR #2922, opened 2026-08-13
feat(plugin): add status effect methods to the entity API

darkstarworks:
## Description

Adds status effect methods to the plugin entity API so plugins can apply and query potion effects on any living entity, not only players. New methods on the entity resource:

- add-effect
- remove-effect
- clear-effects
- has-effect
- get-effect
- get-active-effects

These mirror the methods the player resource already has. Internally they surface the existing living entity effect system, so effects apply, sync to clients, and persist across reload exactly as they already do for players. Calling them on an entity that cannot hold effects, such as a dropped item or a projectile, is a no-op, matching how the existing mob-only entity methods handle non-matching entities.

The only supporting change is a small get-active-effects accessor moved onto the living entity, which the player method now delegates to, removing a duplicate.

This depends on the WIT interface change in Pumpkin-MC/pumpkin-plugin-wit#32. The submodule pointer here targets that branch, so this is a draft until the WIT PR merges and the pointer can move to the merged commit.

## Testing

cargo fmt --all --check, cargo clippy --all-targets --all-features, typos, and cargo test --workspace all pass. The methods are thin pass-throughs to already-tested living entity methods, the same pattern as the existing player effect bindings, so no new unit tests are added.

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -732,6 +732,12 @@ impl LivingEntity {
         effects.get(&effect).cloned()
     }
 
+    /// Returns every status effect currently applied to this entity.
+    pub async fn get_active_effects(&self) -> Vec<Effect> {
+        let effects = self.active_effects.lock().await;
+        effects.values().cloned().collect()
+    }
+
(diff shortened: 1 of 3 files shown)


darkstarworks:
## Closed with reason:

PR #2924, opened 2026-08-13
refactor(plugin): move NBT types to a dedicated nbt interface

darkstarworks:
## Description

Moves the shared NBT types (nbt-tree, nbt-tag, nbt-entry) out of the item-stack WIT interface into a dedicated nbt interface, and bumps the plugin-wit submodule to that change. See Pumpkin-MC/pumpkin-plugin-wit#33.

These types are generic, not item specific. Giving them their own interface fixes the dependency direction and lets other interfaces reuse structured NBT rather than depending on item-stack for it. The immediate motivation is upcoming per-entity custom data.

Host changes are minimal and mechanical:

- item_stack.rs imports the NBT types from the nbt module instead of the item_stack module.
- An empty Host impl is added for the new nbt interface, matching how the other types-only interfaces (enchantments, data-components, biomes) are wired.

No behaviour changes. The type representation is unchanged, and the component model matches types by shape, so this does not affect the ABI.

This depends on the WIT change above, so it is a draft until that merges and the submodule pointer can move to the merged commit.

## Testing

cargo fmt --all --check, cargo clippy --all-targets --all-features, typos, and cargo test --workspace all pass.


QuantumSegfault (approved):
Needs to be rebased, but otherwise good.

darkstarworks:
Closing as superseded: upstream has since moved the NBT types (nbt-tree/nbt-tag/nbt-entry) out of item-stack into `common` directly, which achieves the same neutral-home goal this PR aimed for. Entity custom data will build on `common.nbt-tree` instead. Thanks for the review @QuantumSegfault!

darkstarworks:
## Closed with reason:

PR #2926, opened 2026-08-13
Add API to get world spawn location

Neol3108:
For my hub plugin I want to TP players to spawn on join, but there is currently no way to get the world spawn location.

Depends on https://github.com/Pumpkin-MC/pumpkin-plugin-wit/pull/35

### DIFF
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
@@ -68,7 +68,7 @@ use crate::plugin::loader::wasm::wasm_host::wit::v0_1::pumpkin::plugin::world::{
     BlockPos as WitBlockPos, BlockState as WitBlockState, BlockStateInfo as WitBlockStateInfo,
     BoundingBox as WitBoundingBox, Chunk as WitChunk,
     NoteblockInstrument as WitNoteblockInstrument, PistonBehavior as WitPistonBehavior,
-    WorldBorder as WitWorldBorder,
+    WorldBorder as WitWorldBorder, WorldSpawnLocation as WitWorldSpawnLocation,
 };
 use crate::plugin::loader::wasm::wasm_host::{
     state::{
@@ -510,6 +510,22 @@ impl pumpkin::plugin::world::HostWorld for PluginHostState {
             .dimension
             .minecraft_name
             .to_string())
+    }
+
+    async fn get_spawn_location(
+        &mut self,
+        world: Resource<World>,
+    ) -> wasmtime::Result<WitWorldSpawnLocation> {
+        let (pos, yaw, pitch) = self.get_world_res(&world)?.provider.get_spawn_location();
+        Ok(WitWorldSpawnLocation {
+            pos: WitBlockPos {
+                x: pos.0.x,
+                y: pos.0.y,
+                z: pos.0.z,
+            },
+            yaw,
+            pitch,
+        })
     }
 
     async fn get_top_block_y(
--- a/crates/pumpkin/src/world/mod.rs
+++ b/crates/pumpkin/src/world/mod.rs
@@ -447,6 +447,17 @@ impl World {
             .unwrap_or("world")
     }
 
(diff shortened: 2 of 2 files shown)


QuantumSegfault (commented):
I think this should be unified more with (made more similar to) the `player` respawn point, one way or another.

Right now the world provides yaw and pitch, while the player doesn't. Actually, it seems like the player one is messed up anyway, interacting with a completely disconnected `respawn_location`, rather than the true `respawn_point`

Though `respawn_point` doesn't have pitch...only yaw...

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
```diff
@@ -487,6 +487,22 @@ impl pumpkin::plugin::world::HostWorld for PluginHostState {
             .to_string())
     }
 
+    async fn get_spawn_location(
+        &mut self,
+        world: Resource<World>,
+    ) -> wasmtime::Result<WitWorldSpawnLocation> {
+        let (pos, yaw, pitch) = self.get_world_res(&world)?.provider.get_spawn_location();
+        Ok(WitWorldSpawnLocation {
```
QuantumSegfault, line 520: FYI, WIT does have `tuple`, if you'd prefer to make it more similar to the new method you added Rust side

vyPal: It's probably better to use a named structure here for the spawn data, so that it's a little easier to orient in in the API

vyPal (approved):
LGTM, if you have no other additions to make, I'll merge the wit changes tomorrow, and after you update the submodule reference, we can merge this as well

zzniki:
Could you add another function to be able to set the spawn location? That would also be very useful, thanks.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
```diff
@@ -487,6 +487,22 @@ impl pumpkin::plugin::world::HostWorld for PluginHostState {
...
+            pos: WitBlockPos {
+                x: pos.0.x,
+                y: pos.0.y,
+                z: pos.0.z,
+            },
+            yaw,
+            pitch,
+        })
+    }
+
```
zzniki, line 530:
```rust
async fn set_spawn_location(
        &mut self,
        world: Resource<World>,
        wit_pos: WitWorldSpawnLocation
    ) -> wasmtime::Result<WitWorldSpawnLocation> {
        self.get_world_res(&world)?.provider.set_spawn_location(
            wit_pos.pos,
            wit_pos.yaw,
            wit_pos.pitch
        );
        
        Ok(())
    }
```

crates/pumpkin/src/world/mod.rs
```diff
@@ -422,6 +422,17 @@ impl World {
...
+    #[must_use]
+    pub fn get_spawn_location(&self) -> (BlockPos, f32, f32) {
+        let level_info = self.level_info.load();
+        (
+            BlockPos::new(level_info.spawn_x, level_info.spawn_y, level_info.spawn_z),
+            level_info.spawn_yaw,
+            level_info.spawn_pitch,
+        )
+    }
+
```
zzniki, line 460:
```rust
#[must_use]
    pub fn set_spawn_location(&self, pos: BlockPos, yaw: f32, pitch: f32) {
        let level_info = self.level_info.load();
        self.level_info.store(Arc::new(LevelData {
            spawn_x: pos.0.x,
            spawn_y: pos.0.y,
            spawn_z: pos.0.z,
            spawn_yaw: yaw,
            spawn_pitch: pitch,
            ..(**level_info).clone()
        }));
    }
```

Neol3108:
> Could you add another function to be able to set the spawn location? That would also be very useful, thanks.

@zzniki I agree that method should exist, but I was planning on doing that in a separate PR. Getting this one merged is already taking some time, I don't want to add more stuff that needs an entire new review

Snowiiii:
Needed to fix this cause then master would have fail, this thank you @Neol3108 :D

PR #2928, opened 2026-08-13
fix(bedrock): stop bedrock players from flying after eating

ZlordHUN:
## Description
Fixes Bedrock players gaining underwater bubbles and losing gravity after eating.

## What
- Prevents Bedrock players from floating after consuming food.
- Keeps gravity, breathing, collision, and other entity states intact while using an item.
- Correctly starts and stops the Bedrock eating animation.

## How
- Updates the `USING_ITEM` bit within the player’s existing Bedrock entity flag mask.
- Sends the complete updated flag mask instead of replacing it with partial metadata.
- Keeps the change limited to Bedrock metadata synchronization.

## Testing
- Verified in survival that eating food no longer displays air bubbles or allows the player to float.


Snowiiii:
LGTM, Thank you @ZlordHUN :D

PR #2929, opened 2026-08-13
feat(command): implement /return command 🤖🤖🤖

yusgsysh:
Implements the vanilla `/return` command.

## Changes
- `return <value>` - ends execution with a success and sets the return value to the given integer
- `return fail` - ends execution with a failure and a return value of 0
- `return run <command>` - executes the command and returns its success/result value

Follows the node-based command style (see `seed.rs`), permission level 2 (`minecraft:command.return`), matching vanilla's `LEVEL_GAMEMASTERS` requirement. Reference: `mcsource/net/minecraft/server/commands/ReturnCommand.java`.

## Verification
- `cargo check -p pumpkin --lib` ✅
- `cargo clippy -p pumpkin --lib` ✅
- `cargo test -p pumpkin --lib` ✅ (234 passed)

### DIFF
--- a/crates/pumpkin/src/command/commands/mod.rs
+++ b/crates/pumpkin/src/command/commands/mod.rs
@@ -52,6 +52,7 @@ mod plugins;
 mod pumpkin;
 mod random;
 mod recipe;
+mod r#return;
 mod ride;
 mod rotate;
 mod saveall;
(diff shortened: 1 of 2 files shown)


zhangxuan2011:
Why close it? you can `cargo fmt` to deal with the CI failing

PR #2932, opened 2026-08-14
 fix(death):  Fixed some behavior in player death

### LINKED ISSUE #1583: Dying in the Nether without setting a spawnpoint, respawns on top of nether roof.

### Operating System

Pop!_OS 24.04

8433f05

### What happened?

The Nether does not properly teleport the player to the overworld when killed without a spawnpoint set.

### To Reproduce

Tp into the nether away from the nether spawn. Then do `/kill` on yourself without clicking a bed or a nether anchor. Instead of going back to nether spawn, you instead appear on the top of the nether roof at 0, 0.

### Expected behavior

You should spawn back at the world spawn.

zhangxuan2011:
## Description
Fixes #1583.

Now when you died in other dimentions, if no anchor or bed helps you reset your spawn point, it will forcely teleport you back to the `minecraft:overworld` instead of using the dimension where you died.

Simultaneously, this PR fixes that a player killed another player, the death message doesn't show correctly.

## Testing
 - [x] `cargo clippy --all-targets` 
 - [x] `cargo test`
 - [x] `cargo fmt --check`
 - [x] In-game test

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -1376,8 +1376,8 @@ impl LivingEntity {
                         format!("death.attack.{}.player", damage_type.message_id)
                     };
                     TextComponent::translate_cross(
-                        key.clone(),    // Java
-                        key.clone(),    // Bedrock
+                        key.clone(), // Java
+                        key.clone(), // Bedrock
                         [
                             dyn_self.get_display_name().await,
                             cause.get_display_name().await,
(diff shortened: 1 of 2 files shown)


RoosterBooster007:
Looks like the overworld respawn fix landed in [d8c7c23c0](https://github.com/Pumpkin-MC/Pumpkin/commit/d8c7c23c0), so I'll close this out. The death message bug might still be there, though. You could open a separate PR for that.

PR #2933, opened 2026-08-14
fix(bedrock): one block crawling

### LINKED ISSUE #2898: Crouching/jumping into a one-block space lets you crawl
### Before Submitting

### Minecraft Edition

Bedrock Edition

### Server Version / Commit

7eb49e2

### Operating System

Nintendo Switch 2

When you jump towards a one block space, you will sometimes enter crawl mode automatically

https://github.com/user-attachments/assets/56c5d812-cf17-48ce-91e4-05d31034c4c9

### Steps to Reproduce

1. Find/make a one block gap
2. Run up against/jump into the gap
3. Be forced into crawl mode

ZlordHUN:
## Description
Fixes Bedrock players automatically entering crawl mode when jumping or crouching toward a one-block gap.

## What
- Prevents Pumpkin’s Java pose fallback from forcing Bedrock players into crawl mode.
- Handles Bedrock’s explicit start and stop crawling inputs.
- Preserves normal intentional crawling.
- Does not change Java Edition pose behavior.

## How
- Tracks `START_CRAWLING` and `STOP_CRAWLING` from Bedrock’s `PlayerAuthInput` packet.
- Uses the Bedrock client’s reported crawling state when selecting its pose.
- Keeps collision-derived standing, crouching, and swimming pose fallback exclusive to Java players.

## Testing
- Verified in game that jumping or crouching toward a one-block gap no longer forces crawling.
- Verified that normal intentional crawling still works.
- Resolves: https://github.com/Pumpkin-MC/Pumpkin/issues/2898

### DIFF
--- a/crates/pumpkin/src/net/bedrock/play/player_auth_input.rs
+++ b/crates/pumpkin/src/net/bedrock/play/player_auth_input.rs
@@ -1,5 +1,6 @@
 #[allow(clippy::wildcard_imports)]
 use super::*;
+use pumpkin_data::entity::EntityPose;
 
 impl BedrockClient {
     #[expect(clippy::too_many_lines)]
@@ -160,6 +161,12 @@ impl BedrockClient {
             entity.set_sneaking(true).await;
         } else if input_data.get(InputData::StopSneaking as usize) {
             entity.set_sneaking(false).await;
+        }
+
+        if input_data.get(InputData::StartCrawling as usize) {
+            entity.set_pose(EntityPose::Swimming);
+        } else if input_data.get(InputData::StopCrawling as usize) {
+            player.update_player_pose().await;
         }
 
         if input_data.get(InputData::StartFlying as usize) {


Snowiiii:
can't we just set the already existing Pose to crawling?

ZlordHUN:
> can't we just set the already existing Pose to crawling?

Yes that also seems to work. Tested and pushed it.

yunuservices:
Looks good and minimal, but two things before I approve:

1. Does this actually resolve #2898? The description mentions making the collision fallback Java-exclusive, but the diff only adds StartCrawling/StopCrawling handling and doesn't touch update_player_pose(). The auto-crawl-on-jump-into-a-gap case looks like it comes from that fallback.

2. Could you add a short comment explaining why EntityPose::Swimming is used for crawling (and whether the CRAWLING actor-data flag should be used instead)?

PR #2937, opened 2026-08-14
fix(worldgen): square biome parameter distances

Br0bserver:
## Description

Pumpkin's multi-noise biome search calls this a squared-distance search, but each parameter range was returning the raw distance outside the range. Those per-parameter values are then added together, so the search was using a different distance metric from vanilla and could choose the wrong biome near parameter boundaries.

This squares the distance contributed by each parameter before the values are summed. Values inside the range still contribute zero. The checked-in generated biome data is updated together with the code generator.

I found this while comparing fixed-seed world-generation output against vanilla. A wrong biome selection can also affect structure biome checks, even when the structure placement itself is otherwise correct.

The regression test covers a value inside the range and values three units above and below it. The outside values must contribute `9`, not `3`.

## Testing

- Added `parameter_distance_is_squared_outside_range`

### DIFF
--- a/crates/pumpkin-data/src/generated/biome.rs
+++ b/crates/pumpkin-data/src/generated/biome.rs
@@ -9100,13 +9100,14 @@ pub struct ParameterRange {
 }
 impl ParameterRange {
     pub fn calc_distance(&self, noise: i64) -> i64 {
-        if noise > self.max {
+        let distance = if noise > self.max {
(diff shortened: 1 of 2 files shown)


RoosterBooster007:
It looks like this was covered in beb6947df, so I'll close this out (along with #2938 and #2939). Feel free to reopen otherwise.

PR #2938, opened 2026-08-14
fix(worldgen): preserve vanilla jigsaw candidate ordering

Br0bserver:
## Description

This fixes three places where Pumpkin fed Jigsaw candidates to the random source in a different order from vanilla.

- When a structure names a start jigsaw, vanilla shuffles the matching anchors before choosing one. Pumpkin always took the first matching block from the template.
- Target jigsaws are shuffled and then ordered by `selection_priority`. Pumpkin already did this for source jigsaws, but not for the targets being attached.
- Vanilla's template loader orders block-entity entries by Y, X, and Z. Jigsaw blocks all carry NBT, so that coordinate order is the starting point before they are shuffled. Pumpkin was using the raw template block order instead.

These details are easy to miss because the structure still generates, but a different candidate order changes both the selected connection and the random numbers consumed afterward. From that point on, the rest of the structure can diverge even with the same world seed.

I found the differences while comparing fixed-seed ancient-city and bastion layouts against vanilla. Keeping the three ordering rules together makes the random sequence and the chosen pieces line up for the same reason, rather than special-casing either structure.

## Testing

- Added a test that checks embedded templates expose jigsaws in vanilla's Y, X, Z order
- Added a test that higher `selection_priority` entries are tried first while equal priorities keep their shuffled order

### DIFF
--- a/crates/pumpkin-world/src/generation/structure/structures/jigsaw_placement.rs
+++ b/crates/pumpkin-world/src/generation/structure/structures/jigsaw_placement.rs
@@ -136,7 +136,11 @@ impl JigsawPlacement {
         let mut anchored_position = position;
         if let Some(target_jigsaw_id) = start_jigsaw {
             let mut found_anchor = None;
-            let jigsaws = get_jigsaw_blocks(&template);
+            let mut jigsaws = get_jigsaw_blocks(&template);
+            for index in (1..jigsaws.len()).rev() {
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
https://github.com/Pumpkin-MC/Pumpkin/pull/2937#issuecomment-5555625662

PR #2939, opened 2026-08-14
fix(worldgen): preserve vanilla jigsaw bounds

Br0bserver:
## Description

This fixes two Jigsaw boundary checks that were using a larger area than vanilla.

First, `BlockBox` uses inclusive maximum coordinates. The horizontal search bounds already include the block at `center + max_distance`, so adding another block on the positive X and Z sides made the allowed area one block too wide. A piece touching that extra edge could be accepted by Pumpkin even though vanilla rejects it.

Second, the expansion hack can temporarily grow a piece's collision box to reserve room for children. That expanded space is useful while checking whether a new piece fits, but it is not part of the source template itself. Pumpkin was also using the expanded box to decide whether a jigsaw connection belonged to the source piece's interior. This could put the connection in the wrong collision space and change which pieces were allowed.

The interior check now uses the source piece's real bounding box. The separate list of expanded per-piece boxes is no longer needed after that change, so it has been removed.

I found both differences while tracing fixed-seed Jigsaw layouts against vanilla. They show up at structure edges and around pieces that use the expansion hack, which makes them easy to miss in normal visual testing.

## Testing

- Added a test for the inclusive horizontal distance limit
- Added a test showing that space added by the expansion hack is not source-piece interior

### DIFF
--- a/crates/pumpkin-world/src/generation/structure/structures/jigsaw_placement.rs
+++ b/crates/pumpkin-world/src/generation/structure/structures/jigsaw_placement.rs
@@ -183,15 +183,11 @@ impl JigsawPlacement {
 
         let center_y = bottom_y + local_anchor_position.y;
 
-        let global_bounding_box = BlockBox::new(
-            center_x - max_distance_from_center.horizontal,
-            (center_y - max_distance_from_center.vertical)
-                .max(context.min_y + dimension_padding.bottom),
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
https://github.com/Pumpkin-MC/Pumpkin/pull/2937#issuecomment-5555625662

PR #2940, opened 2026-08-14
fix(net): acknowledge block changes after handling each packet

Max-109:
## Description

The problem is we only sent the block change confirmation from the keepalive timer, so once every 15s. Until it lands the client keeps showing its own prediction for that block and throws away our updates for that position.

Vanilla sends it at the end of the packet that carried the sequence. Moved it there.

TNT is where you notice it, but it's any interaction where we disagree with the client's guess.

## Testing

Placed TNT, hit it with flint and steel.

### Before
before

### After
after


Snowiiii:
LGTM, Thank you @Max-109 :D

PR #2941, opened 2026-08-14
fix(entity): let explosions push TNT

Max-109:
## Description

The problem is TNT was immune to explosions. `damage_entities` throws out immune entities right at the top, before it gets anywhere near the knockback, so a chained TNT never got pushed. Vanilla TNT isn't immune, so the override is gone.

`tick` was also grabbing the velocity before it moved and writing that old value back at the end, so any push that came in while it was moving got wiped. It reads the velocity again after `move_entity` now, like vanilla does.

### Before
before

### After

after

### DIFF
--- a/crates/pumpkin/src/entity/tnt.rs
+++ b/crates/pumpkin/src/entity/tnt.rs
@@ -41,13 +41,17 @@ impl EntityBase for TNTEntity {
     ) -> EntityBaseFuture<'a, ()> {
         Box::pin(async move {
             let entity = &self.entity;
-            let original_velo = entity.velocity.load();
 
-            let mut velo = original_velo;
+            let mut velo = entity.velocity.load();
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thank you @Max-109 :D

O2theC:
so the bug of tnt not affecting each other is back, seems this time it's related to ray casts and things.
main thing is exposure points, the base logic finds that none of the 27 exposure points are "exposed" (likely a clear path between points) and thus have an exposure score of 0, meaning explosion logic for the tnt entity gets cut off early and no knockback is applied.
i have little to no knowledge regarding ray casts and collision logic and my best guess for where the logic error is, is in all the core ray cast and collision logic, the stuff that is kind of core to all ray casts and block collision (i think).

PR #2942, opened 2026-08-14
Add option to disable plugin signature verification

Neol3108:
This PR adds a configuration option to disable plugin signature verification.

I want to optimise my startup times as much as possible, and currently the request takes a few seconds, and is done per plugin. I'm planning on only running proprietary plugins so I won't actually need/use plugin signature verification unless there will one day be an option to verify it against a self generated key.

On my M5 16GB running in Docker, with 2 proprietary WASM plugins, this makes about ~5 seconds difference.

If wanted I could open another PR after this to make sure the key gets cached instead of retrieved for each plugin.

### DIFF
--- a/crates/pumpkin-config/src/plugins.rs
+++ b/crates/pumpkin-config/src/plugins.rs
@@ -45,6 +45,8 @@ pub struct PluginsConfig {
     /// MY_API_KEY = "secret_key"
     /// ```
     pub overrides: HashMap<String, PluginOverride>,
+    /// Whether Pumpkin should verify WASM plugin signatures before loading.
+    pub verify_signatures: bool,
 }
 
 impl Default for PluginsConfig {
@@ -60,6 +62,7 @@ impl Default for PluginsConfig {
             loopback_only: false,
             max_memory_mb: None,
             overrides: HashMap::new(),
+            verify_signatures: true,
         }
     }
 }
@@ -117,6 +120,7 @@ mod tests {
         assert!(!config.loopback_only);
         assert_eq!(config.max_memory_mb, None);
         assert!(config.overrides.is_empty());
+        assert!(config.verify_signatures);
     }
 
     #[test]
@@ -131,6 +135,7 @@ mod tests {
             inherit_env = true
             loopback_only = true
             max_memory_mb = 256
+            verify_signatures = false
 
             [overrides.my_plugin]
             enabled = false
@@ -154,6 +159,7 @@ mod tests {
         assert!(config.inherit_env);
         assert!(config.loopback_only);
         assert_eq!(config.max_memory_mb, Some(256));
+        assert!(!config.verify_signatures);
 
         let override_cfg = config.overrides.get("my_plugin").unwrap();
         assert!(!override_cfg.enabled);
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
@@ -89,29 +89,36 @@ impl PluginRuntime {
     pub async fn init_plugin<P: AsRef<Path>>(
         &self,
         path: P,
+        verify_signatures: bool,
     ) -> Result<(Arc<WasmPlugin>, PluginMetadata), PluginInitError> {
         let wasm_bytes = std::fs::read(&path).map_err(PluginInitError::FileReadFailed)?;
-
-        let verification =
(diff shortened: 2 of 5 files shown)


QuantumSegfault:
I think it would make more sense to simply cache the key, rather than provide a way to opt-out of these checks entirely.

Make it fast AND safe(r) by default.

Q2297045667:
> I think it would make more sense to simply cache the key, rather than provide a way to opt-out of these checks entirely.
> 
> Make it fast AND safe(r) by default.

I don't see any issues with this. Some users might want to disable these options when they don't need key verification, so I think it's perfectly fine to provide a configuration to turn them off.

When this config is disabled, we could output a warning log during server startup to indicate the current authentication status.

@Neol3108 Could you add a log output when turning off this verification

Neol3108:
I agree, adding a log would make sense. This should be an advanced use case. A sort of "Make sure you know what you're doing" log. Will add that later

Neol3108:
image
Added the log

crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
vyPal, line 1: I don't think a full test case for what is essentially just an if statement is necessary. For simplicity it would be best to just wrap the original function with the if statement directly and the tests.
```rs
if verify_signatures {
  signature::verify_wasm_plugin(&wasm_bytes, &path.as_ref().to_string_lossy());
}
```

crates/pumpkin-config/src/plugins.rs
```diff
@@ -1,9 +1,46 @@
...
+            r#"
+verify_signatures = false
+blocked_permissions = ["pumpkin:fs.read"]
+"#,
+        )
+        .expect("config should deserialize");
+
+        assert!(!config.verify_signatures);
+        assert_eq!(config.blocked_permissions, vec!["pumpkin:fs.read"]);
+    }
```
vyPal, line 103: The tests here are not necessary

vyPal (changes requested):
I'd remove the two test sections as they're not needed, but otherwise LGTM.

If you wanted to make a PR later for caching plugin signatures, that would probably be nice

crates/pumpkin-config/src/plugins.rs
Neol3108: Removed the tests

Neol3108:
@vyPal I'll start on the caching when this PR gets merged. What kind of caching do we want here? I suppose in-process right? Not actually cache the key to the disk, that would make subsequent launches faster, but if the key has to be rotated there is no way to enforce retrieval of the new key on peoples machines. Unless you still make a request to verify the key, which would make caching obsolete.

Neol3108:
I think Alex just added his own implementation for this. Will double check tomorrow and otherwise close

Neol3108:
He didn't add this exact behaviour, so I'll have to update my PR to work with this changes. Will do that a little bit later.

Snowiiii:
Startup time should now be back to normal, the problem was mostly in the bedrock key fetching not the marketplace key

Neol3108:
> Startup time should now be back to normal, the problem was mostly in the bedrock key fetching not the marketplace key

@Snowiiii I'll try tonight, but as of yesterday night I still had 11s startup time with the two "this plugin is unsigned..." messages being 5 seconds apart from other logs each. Surely that isn't for the Bedrock key?

Tho, I also had runs where it's nearly instant. For me it's kind of flaky

Neol3108:
@Snowiiii just tried Docker image version `sha-399bf14` and got:
```
2026-08-19 17:44:13  INFO            main ThreadId(01) pumpkin::net::bedrock::nethernet: Bedrock NetherNet ICE is listening on 0.0.0.0:19132
2026-08-19 17:44:13  WARN            main ThreadId(01) pumpkin::plugin::loader::wasm::wasm_host::signature: Plugin './plugins/<plugin1>.wasm' is unsigned. Unsigned plugins may come from untrusted sources.
2026-08-19 17:44:18  WARN            main ThreadId(01) pumpkin::plugin::loader::wasm::wasm_host::signature: Plugin './plugins/<plugin2>.wasm' is unsigned. Unsigned plugins may come from untrusted sources.
2026-08-19 17:44:23  next log...
```

Snowiiii:
@Neol3108 would like to merge this, please fix conflicts

Snowiiii:
LGTM, Thank you @Neol3108 :D

PR #2943, opened 2026-08-14
fix(bedrock): restore eating sounds

ZlordHUN:
## Description
Fixes missing eating sounds for Bedrock players.

## What
- Handles Bedrock’s `ActorEvent` packet during item consumption.
- Relays native `Feed` events to nearby Bedrock players.
- Plays the burp sound after successfully consuming food.
- Ignores invalid feed events unless the player is actively eating.

## How
- Adds decoding for the serverbound Bedrock `ActorEvent` packet.
- Reconstructs the event using the authenticated player’s entity ID instead of trusting the client-provided ID.
- Preserves the client-provided item data needed for the correct eating feedback.
- Uses Bedrock’s native event cadence instead of maintaining a separate server-side sound timer.

## Testing
- `cargo test -p pumpkin-protocol reads_feed_event`
- `cargo check -p pumpkin`
- Verified in game that Bedrock players hear eating sounds while consuming food and the burp sound when consumption finishes.


crates/pumpkin-protocol/src/bedrock/server/actor_event.rs
QuantumSegfault, line 1: So this is a bidirectional packet? -_-

It was only ever sent TO clients, so I moved it to `CActorEvent`. Guess that should be moved back (even though it's really neither...).

I'll do that and you can rebase so it's just adding the test.

crates/pumpkin/src/world/mod.rs
```diff
@@ -5919,7 +5849,11 @@ impl World {
         Self::broadcast_java_grouped(packet, recipients_by_version);
     }
 
-    fn broadcast_to_chunk_bedrock<P: BClientPacket>(&self, chunk_pos: Vector2<i32>, packet: &P) {
+    pub(crate) fn broadcast_to_chunk_bedrock<P: BClientPacket>(
```
QuantumSegfault: This should probably just be normal `pub`, considering `broadcast_to_chunk` is

crates/pumpkin-protocol/src/bedrock/server/actor_event.rs
QuantumSegfault: #2949

QuantumSegfault: Merged. You can rebase now

crates/pumpkin-protocol/src/bedrock/server/actor_event.rs
```diff
@@ -1,12 +1,14 @@
...
-use crate::codec::var_int::VarInt;
+use crate::{
+    codec::{var_int::VarInt, var_ulong::VarULong},
+    serial::{PacketRead, PacketWrite},
+};
 
-#[derive(Debug, PacketWrite)]
+#[derive(Debug, PacketRead, PacketWrite)]
 #[packet(27)]
 pub struct SActorEvent {
```
QuantumSegfault, line 13: Enum values are out of date (and now writing is out of sync with reading).

Thus `Feed` won't be broadcast properly.

I was going to be updating this myself, but I suggest you do it here for completeness

QuantumSegfault (changes requested):

crates/pumpkin-protocol/src/bedrock/server/actor_event.rs
QuantumSegfault: Ah...I see. 

I'm looking at a future version of the docs/protocol (the official ones are at 26.50 for some reason... -_-),

These look correct for 26.40 (particularly after the renames).

QuantumSegfault (approved):

PR #2946, opened 2026-08-15
Fix missing Handshake packet ID for protocol 26.1/26.2

darkzmaj:
\`assets/packet/26_1_packets.json\` and \`26_2_packets.json\` both have an empty \`serverbound.handshake\` object, while every earlier tracked version (e.g. \`1_21_11_packets.json\`) has \`{"intention": 0}\`. This makes the generated \`HANDSHAKE_INTENTION.v26_1\`/\`.v26_2\` constants resolve to \`-1\` (\`crates/pumpkin-data/src/generated/packet.rs\`), even though the Handshake packet's wire ID has been \`0\` since its introduction and doesn't change between versions - confirmed against every other tracked version in the same file, and against \`crates/pumpkin/src/net/java/pending.rs::handle_handshake_packet\`, which hardcodes \`match packet.id { 0 => ... }\` rather than using this table at all.

Server-side impact is limited (Pumpkin never calls \`to_id()\` for its own Handshake handling, so this never affected real players), but it breaks any external tool that builds on \`pumpkin-protocol\`/\`pumpkin-data\` the way the crate's own docs suggest (using \`to_id()\` generically for outgoing packets) - found while building a protocol-accurate benchmark client, which failed every 26.x handshake with a literal wire-level \`-1\` packet ID until this was fixed.

**Fix:** add the missing \`"intention": 0\` entry to both JSON files and regenerate \`packet.rs\` via \`cargo run -p pumpkin-codegen -- packet\`.

## Test plan
- [x] Regenerated \`packet.rs\` from the fixed JSON and confirmed \`HANDSHAKE_INTENTION.v26_1\`/\`.v26_2\` are now \`0\` instead of \`-1\`
- [x] \`cargo check -p pumpkin-data --features packet\` passes
- [x] Diff is minimal - only the intended 4-line change (reverted an unrelated \`LOWEST_SUPPORTED_MC_VERSION\` drift the regeneration also picked up, which is pre-existing and unrelated to this fix)


QuantumSegfault:
Seems like https://github.com/Pumpkin-MC/Pumpkin/commit/c7c8d10d4709fcd1066c4a7d58404d2e0196f12c subsumed the main purpose of this PR.

The remaining changes are unrelated.

Feel free to open new/separate PRs if there are still issues or you still want to make those two other tiny tweaks.

PR #2948, opened 2026-08-15
Fix persistent mobs counting toward spawn caps

eshankiyer:
Keep persistent mobs, passengers, and leashed mobs out of natural spawn cap accounting like vanilla. Also preserve PersistenceRequired through NBT and clean up mount and leash state when entities are removed.


Q2297045667:
Could you add a configuration file for this stuff? Something like Bukkit where you can control the number of creatures generated and their tick rate

PR #2950, opened 2026-08-15
Match ocelot spawning

eshankiyer:
Match ocelot spawning with vanilla in normal and chunk-generation paths. Checked with cargo check --locked -p pumpkin.


Necko1:
finally kitties

Issue #2870, opened 2026-08-10
Cant join on 1.21.11


AdmerPRO:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

1.0.0/366c9cb

### Operating System

Windows 10

### Bug Description & Expected Behavior

There’s an issue with version **1.21.11** (at least for me) when I try to join using it I get a "Network Protocol Error" and can't join I tried version **26.2 and everything works** fine. It might be a problem with ViaBackwards but Im not sure. The error occurs on both **Fabric and Vanilla**.

### Steps to Reproduce

1. Launch Minecraft 1.21.11
2. Join Server

K8bus:
maybe im not right, but im pretty sure you dont need viabackwards here

zhangxuan2011:
> maybe im not right, but im pretty sure you dont need viabackwards here

I tested, 1.21.10 is able to join

Q2297045667:
[disconnect-2026-08-23_17.41.55-client.txt](https://github.com/user-attachments/files/31346156/disconnect-2026-08-23_17.41.55-client.txt)
Adding the client log for reference.

Q2297045667:
Fixed it

PR #2954, opened 2026-08-15
fix(player): change MAIN_ARM_ID metadata to ARM type for 1.21.11

Beengoo:
Fix for https://github.com/Pumpkin-MC/Pumpkin/issues/2870 where you can't join from 1.21.11 because MAIN_ARM_ID send wrong data type

### DIFF
--- a/crates/pumpkin/src/entity/player.rs
+++ b/crates/pumpkin/src/entity/player.rs
@@ -3896,9 +3896,10 @@ impl Player {
                     MetaDataType::BYTE,
                     config.skin_parts,
                 ),
+                // 1.21.11
                 Metadata::new(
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Hey @Beengoo. Thanks for chasing this down! We can actually close this out, as the `MAIN_ARM_ID` fix already landed on master.

The `tracked-data` system recently got completely rewritten (by Snowiiii and necko). `MAIN_ARM_ID` now automatically maps to the right wire types via generated data, and the old code path you patched here is totally gone.

So, I'll be closing this as superseded. Let us know if you spot any other 1.21.11 join issues on the latest master, and we can get a fresh PR going!

PR #2958, opened 2026-08-15
feat(pumpkin): add /perf command

Mcxiaocaibug:
Part of #15.

Adds the vanilla `/perf` command (permission level 4).

- `/perf start` begins a metrics recording session. As in Vanilla, the run ends on its own after 10 seconds unless it is stopped early.
- `/perf stop` ends the running session immediately.

Stopping reports the vanilla `commands.perf.stopped` feedback with the elapsed seconds, the number of ticks that passed and the resulting ticks per second. Starting while a run is active gives `commands.perf.alreadyRunning`, and stopping without one gives `commands.perf.notRunning`.

## Implementation notes

- The session lives on `Server` as a small `PerfProfiler` (`server/perf_profiler.rs`) so `/perf start` and `/perf stop` share one server-wide run, matching Vanilla's single-profiler behavior.
- Each run gets a generation token. The scheduled auto-stop only ends the run it was started for, so this sequence behaves correctly: start, stop manually, start again — the first run's timer cannot cut the second run short. There is a unit test covering that.
- Non-player command sources get an already-rendered message instead of the translation, since the console cannot resolve the placeholders server-side.

Vanilla also writes a profiling report to disk (`commands.perf.reportSaved` / `commands.perf.reportFailed`); those keys are left unused for now, as Pumpkin has no report format to write yet.

### DIFF
--- a/crates/pumpkin/src/command/commands/mod.rs
+++ b/crates/pumpkin/src/command/commands/mod.rs
@@ -45,6 +45,7 @@ mod op;
 mod pardon;
 mod pardonip;
 mod particle;
+mod perf;
 mod place;
 mod playsound;
 mod plugin;
@@ -188,6 +189,7 @@ pub async fn default_dispatcher(
     help::register(&mut dispatcher, registry);
     kill::register(&mut dispatcher, registry);
     op::register(&mut dispatcher, registry);
+    perf::register(&mut dispatcher, registry);
     place::register(&mut dispatcher, registry);
     random::register(&mut dispatcher, registry);
     list::register(&mut dispatcher, registry);
--- a/crates/pumpkin/src/command/commands/mod.rs
+++ b/crates/pumpkin/src/command/commands/mod.rs
@@ -48,6 +48,7 @@ mod op;
 mod pardon;
 mod pardonip;
 mod particle;
+mod perf;
 mod place;
 mod playsound;
 mod plugin;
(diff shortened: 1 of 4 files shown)


Mcxiaocaibug:
This is ready for review whenever someone has time.

CI is fully green now — the earlier red mark was the `cargo-machete` failure coming from master, which has since been fixed there. The branch is rebased on current master (c13c515) and all jobs pass, including the build and test jobs on all five platforms that were previously being skipped.

Happy to adjust anything: the auto-stop duration, the permission level, or how the stopped message is rendered for non-player senders. Thanks!

aure31 (approved):
it's match vanilla for the message and the implementation is good for me

Mcxiaocaibug:
Thanks for the review @aure31! Conflicts are resolved — @Snowiiii this one is ready to merge whenever you have a moment.

The conflict was with the `/debug` command that landed in the meantime: it added `debug_profiler` to `Server` in the same place this adds `perf_profiler`. Both are kept, since the two commands own separate sessions. The `register` signature also moved from `&mut PermissionRegistry` to `&PermissionRegistry`, which is adapted here.

Rebased on current master (8be5ed1e), and `cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test -p pumpkin` (266 passing) are all clean locally.

RoosterBooster007 (approved):

RoosterBooster007:
Looks good. Waiting for @Snowiiii's review.

PR #2960, opened 2026-08-15
feat(pumpkin): add /reload command

Mcxiaocaibug:
Part of #15.

Adds the vanilla `/reload` command (permission level 2, same as Vanilla).

Vanilla reloads datapacks and then pushes the rebuilt command tree to every player. Pumpkin has no datapack system yet, so the reloadable server data here is its plugins: `/reload` unloads and reloads them, then resends the command tree to all online players so commands that plugins registered or dropped during the reload appear immediately. Once datapack support lands (#2642), reloading those fits into the same command.

Feedback follows Vanilla: `commands.reload.success` ("Reloading!") is sent up front, and `commands.reload.failure` ("Reload failed; keeping old data") is returned if unloading or loading fails — the command tree is resent in that case too, so clients never keep stale entries for plugins that did unload.

### DIFF
--- a/crates/pumpkin/src/command/commands/mod.rs
+++ b/crates/pumpkin/src/command/commands/mod.rs
@@ -55,6 +55,7 @@ mod plugins;
 mod pumpkin;
 mod random;
 mod recipe;
+mod reload;
 mod ride;
 mod rotate;
 mod saveall;
@@ -187,6 +188,7 @@ pub fn default_dispatcher(
     forceload::register(&mut dispatcher, registry);
     ride::register(&mut dispatcher, registry);
     recipe::register(&mut dispatcher, registry);
+    reload::register(&mut dispatcher, registry);
     help::register(&mut dispatcher, registry);
     kill::register(&mut dispatcher, registry);
     op::register(&mut dispatcher, registry);
--- a/crates/pumpkin/src/command/commands/mod.rs
+++ b/crates/pumpkin/src/command/commands/mod.rs
@@ -52,6 +52,7 @@ mod plugins;
 mod pumpkin;
 mod random;
 mod recipe;
+mod reload;
 mod ride;
 mod rotate;
 mod saveall;
@@ -185,6 +186,7 @@ pub async fn default_dispatcher(
     forceload::register(&mut dispatcher, registry);
     ride::register(&mut dispatcher, registry);
     recipe::register(&mut dispatcher, registry);
+    reload::register(&mut dispatcher, registry);
     help::register(&mut dispatcher, registry);
     kill::register(&mut dispatcher, registry);
     op::register(&mut dispatcher, registry);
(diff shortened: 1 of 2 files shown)


Mcxiaocaibug:
The red check here is not coming from this PR. The only failing job is `Detect unused dependencies`, which fails on master itself (a2b92e6) because the new `pumpkin-fuzzer` crate declares three dependencies it does not use. Since that job runs before the build and lint jobs, those are skipped and the PR cannot go green.

#2961 removes those dependencies and is fully green there, including the build and test jobs on all five platforms. Once it is merged, a re-run of the checks on this PR (and on #2958) should be all that is needed — no changes are required here.

For what it is worth, this branch is rebased on current master and I ran the CI steps locally against it: `cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test -p pumpkin` (236 passed) are all clean.

Thanks for taking a look!

Mcxiaocaibug:
This is ready for review whenever someone has time.

CI is green now — the earlier red mark was the `cargo-machete` failure coming from master, which has since been fixed there. The branch is rebased on current master (c13c515) and the formatting, lint, and build/test jobs all pass.

One design point worth a second opinion: Vanilla's `/reload` reloads datapacks and then pushes the rebuilt command tree to every player. Pumpkin has no datapack system yet, so this reloads the plugins instead and resends the command tree, which keeps the observable behaviour close to Vanilla. If you would rather this command wait for datapack support (#2642) and land together with it, that is completely reasonable — just say so and I will close it or rework it. Thanks!

aure31:
The PR is really simple, so I don't see any issue with adding it before datapacks. But I think the only issue for me is that the vanilla `/reload` command only reloads datapacks. So I think we should split it into different commands: one that matches vanilla behavior and only reloads datapacks, and one for plugins. Maybe another one could reload both. For example, we could have `/reload` and `/reload datapack` reload datapacks only, `/reload plugin` reload plugins, and `/reload all` reload both. So that `/reload` alone still matche vanilla behavior. What do you think ?

aure31:
I just realise that the /plugin commands let you reload plugins so this pull request does not have any value you can close it

Snowiiii:
@Mcxiaocaibug we now have simple datapack support actually

Mcxiaocaibug:
Thanks both — @Snowiiii that was the missing piece, so I have reworked this rather than closing it.

`/reload` now does what Vanilla's does: it reloads **datapacks**, not plugins. It goes through the same `Server::reload_datapacks` path that `/datapack enable` and `/datapack disable` already use, so it re-reads the enabled packs, re-runs the `#minecraft:load` function tag and pushes the rebuilt recipes to every player.

That resolves both of @aure31's points at once:

- The original objection was that Vanilla's `/reload` only reloads datapacks — it now only reloads datapacks, so no `/reload plugin` or `/reload all` split is needed to keep `/reload` vanilla-accurate.
- The redundancy with `/plugin` is gone too, since plugin reloading is no longer part of this command. `/plugin` stays the way to reload plugins.

The command dropped from 95 lines to 53 as a result, and `master` still has no `/reload`, so this fills that entry in #15.

Rebased on current master (8be5ed1e); `cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test -p pumpkin` (265 passing) are clean.

One thing worth a second opinion: `/datapack enable|disable` call `reload_datapacks` without resending the command tree, so I kept that behaviour here for consistency. Vanilla does resend it on `/reload`; if you would rather this command do that too, I am happy to add it.

PR #2959, opened 2026-08-15
fix(bedrock): Java player respawn state

ZlordHUN:
## Description
Fixes an issue where Bedrock clients continued seeing a Java player lying down after they respawned, preventing interaction and attacks.

## What
- Removes a dead Java player’s actor from nearby Bedrock clients after the death animation.
- Recreates the actor after the Java player respawns.
- Synchronizes the player’s current position, rotation, skin, metadata, and game mode.
- Prevents Bedrock clients from retaining the dead, non-interactable actor state.
- Leaves the server-side player entity intact so reconnecting while dead continues to work.

## How
- Extracts the relevant player lifecycle handling previously tested as part of PR #2853.
- Sends Bedrock clients an actor removal when the Java player dies.
- After respawning and teleporting the player, sends an ordered actor removal, player-list addition, and player spawn.
- Uses Pumpkin’s existing Java-to-Bedrock player and skin data without adding resource-pack functionality.

## Testing
- `cargo check -p pumpkin`
- `cargo test -p pumpkin --lib`
- `cargo clippy -p pumpkin --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- Verified in game that a Java player displays normally to Bedrock clients after respawning and can be interacted with and attacked.
- Resolves: https://github.com/Pumpkin-MC/Pumpkin/issues/2935

PR #2961, opened 2026-08-15
chore(fuzzer): drop unused dependencies

Mcxiaocaibug:
`pumpkin-fuzzer` declares `bytes`, `pumpkin-protocol` and `pumpkin-util`, but the crate uses none of them — its only imports are `clap`, `colored`, `rand`, `std` and `tokio`. The `bytes` matches in the source are the `bytes_sent` counter and `to_be_bytes()` calls, not the crate.

This fails the `Detect unused dependencies` job, which currently makes the Cargo workflow red on master (a2b92e6) and on every pull request whose CI has run since. Because that job runs first, the build and lint jobs are skipped, so PRs cannot go green at all.

Verified with `cargo machete` (clean afterwards) and `cargo check -p pumpkin-fuzzer`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3511,11 +3511,8 @@ dependencies = [
 name = "pumpkin-fuzzer"
 version = "0.1.0-dev+26.2-26.40"
 dependencies = [
- "bytes",
  "clap",
  "colored",
- "pumpkin-protocol",
- "pumpkin-util",
  "rand 0.9.5",
  "tokio",
 ]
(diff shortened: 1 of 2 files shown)


Mcxiaocaibug:
Heads up that this is currently blocking the whole repository, not just one PR: the `Detect unused dependencies` job runs before the build and lint jobs, so when it fails those are skipped and no PR can reach a green state.

The Cargo workflow is red on master itself (a2b92e6), and every PR whose CI has run since that commit landed fails the same way — for example #2958, #2959 and #2960 — while PRs that last ran before it (#2951, #2953, #2954) are still green.

The three dependencies really are unused (`bytes` only matches the `bytes_sent` counter and `to_be_bytes()` calls), so removing them makes `cargo machete` clean again and `cargo check -p pumpkin-fuzzer` still passes.

PR #2965, opened 2026-08-15
fix(bedrock): transaction decoding fix

ZlordHUN:
## Description
Fixes Bedrock block placement and interaction transactions being decoded incorrectly. I think the regression was introduced with this PR https://github.com/Pumpkin-MC/Pumpkin/pull/2808

## What
- Restores block placement and interaction for Bedrock players.
- Prevents valid transactions from failing with `failed to fill whole buffer`.
- Fixes held items being misread as invalid item IDs such as `1025`.
- Ensures placed blocks are handled server-side and persist correctly.

## How
- Decodes the Bedrock block-face field as its actual one-byte wire value instead of a four-byte integer.
- Keeps subsequent transaction fields correctly aligned.
- Adds regression tests using captured empty-hand and crafting-table placement packets.

## Testing
- `cargo check -p pumpkin`
- Verified in game that a Bedrock player can place and interact with a newly placed crafting table.

### DIFF
--- a/crates/pumpkin-protocol/src/bedrock/server/inventory_transaction.rs
+++ b/crates/pumpkin-protocol/src/bedrock/server/inventory_transaction.rs
@@ -130,7 +130,7 @@ pub struct UseItemTransactionData {
     pub action_type: VarInt,
     pub trigger_type: u8,
     pub block_position: BlockPos,
-    pub block_face: i32,
+    pub block_face: u8,
     pub hot_bar_slot: VarInt,
     pub item_in_hand: NetworkItemDescriptor,
     pub player_position: Vector3<f32>,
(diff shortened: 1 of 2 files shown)


crates/pumpkin-protocol/src/bedrock/server/inventory_transaction.rs
QuantumSegfault: The correct solution would to be make this `u8`, rather than create a custom deserializer to coerce it.

QuantumSegfault (changes requested):

crates/pumpkin-protocol/src/bedrock/server/inventory_transaction.rs
ZlordHUN: Done.

crates/pumpkin/src/net/bedrock/play/inventory_action.rs
```diff
@@ -235,7 +235,7 @@ impl BedrockClient {
                                 pumpkin_protocol::java::server::play::SUseItemOn {
                                     hand: VarInt(0),
                                     position: data.block_position,
-                                    face: VarInt(data.block_face),
+                                    face: VarInt(i32::from(data.block_face)),
```
QuantumSegfault, line 238: Could use simply `data.block_face.into()` here. VarInt supports `From<u8>`. But I won't block on it.

QuantumSegfault (approved):

QuantumSegfault:
Will merge once CI finishes

PR #2966, opened 2026-08-15
fix(rcon): make max_connections actually limit connections

Lstarsky0:
## What was changed

`max_connections` is now tracked in an `AtomicU32` shared with the connection tasks, decremented when a task ends instead of one line after the spawn.

## Why

The counter went up and back down inside the same loop iteration, while the connection was handled in a detached `tokio::spawn`. It therefore read zero on every pass and the limit never applied. `RCONConfig` defaults to `max_connections: 10` and documents it as "the maximum number of concurrent RCON connections allowed", so every server has a cap that does nothing.

With `max_connections = 2` and five clients authenticating:

| | before | after |
| --- | --- | --- |
| authenticated | 5 | 2 |
| still usable afterwards | 5 | 2 |

Opening two more once those are closed succeeds, so the count comes back down rather than latching.

## Impact

Servers that set `max_connections` get it enforced; a refused connection is hung up on, since RCON has no "busy" reply. `0` still means unlimited. `closed RCON connection` now logs on close rather than on spawn.

`cargo test` 599 passed, `cargo clippy --all-targets` and `cargo fmt --check` clean.

## Known issues or limitations

The check and the increment are not one atomic step, which is fine while a single accept loop owns them, but would need revisiting if RCON ever accepts from more than one task.

---

Used AI assistance to navigate the codebase. The change and the measurements above were done by hand.

### DIFF
--- a/crates/pumpkin/src/net/rcon/mod.rs
+++ b/crates/pumpkin/src/net/rcon/mod.rs
@@ -1,4 +1,7 @@
-use std::{net::SocketAddr, sync::atomic::Ordering};
+use std::{
+    net::SocketAddr,
+    sync::atomic::{AtomicU32, Ordering},
+};
 
 use packet::{ClientboundPacket, Packet, PacketError, ServerboundPacket};
 use pumpkin_config::RCONConfig;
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Hey @Lstarsky0, thanks for this. Turns out it already landed on master in 66a5f5340, which does the same `Arc<AtomicU32>` + `fetch_sub`-in-the-task thing and even logs the refusal at `warn!`. I'll close this one as superseded.

PR #2971, opened 2026-08-15
Fix wrong argument name in worldborder error messages

Smorki:
## Description

 when the `time` arg fails to parse in `/worldborder set` and `/worldborder add`, the error says "distance is out of bounds" instead of "time". The error message uses `distance_consumer().default_name()` but the block is parsing `time`.

Fixed both spots.

## Testing

Not tested on a running server


Snowiiii:
Actually ideally we would have this at argument level and use plain translations

Smorki:
Hi, thanks for the review. Youre right that argument-level translations would be a better solution but thats a bigger refactor Do you want me to close this PR and open an issue about it, or keep this small fix for now?

Snowiiii:
Hey, I actually just refactored it my own, but the time consumer might be missing, still thank you for your PR

PR #2972, opened 2026-08-16
fix(inventory): stop the item count arithmetic wrapping in screen_handler

Lstarsky0:
## Description

Item counts in `screen_handler.rs` are `u8` and the arithmetic on them wraps, so a stack holding more than a slot can take corrupts the counts around it instead of being skipped.

Merging: `insert_item` adds the two counts with `+` (`screen_handler.rs:761`). A slot of 200 taking 100 wraps to 44, which then passes the `combined_count <= max_slot_count` check, so the incoming stack is zeroed and the slot is rewritten to 44. 300 items become 44 and nothing is logged.

Clicking: the "how much more fits" expression `max - current` appears four times in this file. One already uses `saturating_sub` (`:1234`); the other three subtract directly. With a cursor already past its own limit, `64 - 200` wraps to 120 and a double-click pulls another stack in on top of a full cursor — measured, 200 becomes 210. The `max(0, ..)` around one of them does nothing, since the operands are `u8` and the subtraction has already wrapped by the time `max` sees it.

Debug builds panic at each of these instead of wrapping.

Counts above 127 are reachable through the plugin API, which clamps nothing: `item-stack.new(registry_key, count)` and `set-count(count)` take a raw `u8` (`wit/v0_1/item_stack.rs:143`, `:173`), and `gui.set-item` writes the stack straight into a slot (`wit/v0_1/gui.rs:83`). From there the swap at `:1222` moves such a stack onto the cursor.

`checked_add` leaves the merge case to the partial-fill arm right below it, which already handles a combination too large to merge outright, and `saturating_sub` leaves no headroom on a slot that is already over. Nothing changes for counts that fit.

This fixes the corruption, not the source of the oversized counts — clamping in the plugin API would be a separate change.

## Testing

`cargo test -p pumpkin-inventory`, with four new tests: topping up a matching stack, filling to the slot limit and carrying the rest over, the merge overflow, and the cursor-headroom underflow.

Both new failure cases fail on `master`. The merge one gives `attempt to add with overflow` in debug and `left: 44, right: 200` in release; the cursor one gives `attempt to subtract with overflow` in debug and `left: 210, right: 200` in release. The two baseline tests pass on `master`, so they pin existing behaviour rather than the fix.

`cargo fmt --all -- --check` and `cargo clippy --all-targets` are clean, and `cargo test --workspace` is 603 passed / 0 failed.

### DIFF
--- a/crates/pumpkin-inventory/src/screen_handler.rs
+++ b/crates/pumpkin-inventory/src/screen_handler.rs
@@ -53,9 +53,9 @@ use pumpkin_world::{
     block::entities::PropertyDelegate,
     inventory::{ComparableInventory, Inventory},
 };
+use std::pin::Pin;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::{any::Any, collections::HashMap, sync::Arc};
-use std::{cmp::max, pin::Pin};
 use tokio::sync::Mutex;
 use tracing::warn;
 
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thank

PR #2973, opened 2026-08-16
fix(world): reject region entries that underflow when read

Lstarsky0:
## Description

Two values read straight out of a region file are subtracted from without a check, and both wrap.

Sector offset: sectors 0 and 1 hold the location and timestamp tables, so a chunk can only start at sector 2 or later. The presence check at `anvil.rs:568` skips offset 0 but not offset 1, and `(sector_offset - 2) * SECTOR_BYTES` at `:578` underflows on it. Release is worse than a wrong read rather than better: `bytes_offset` becomes 2^64 − 4096, adding `bytes_count` wraps it back to 0, so the bounds check at `:581` is satisfied by the same overflow that broke it, and the slice at `:593` dies inside `bytes` with `range start must not be greater than end: 18446744073709547520 <= 0`. Debug panics at the subtraction itself.

Chunk length: the length prefix covers the compression byte, so `:251` subtracts one from it. A declared length of 0 underflows. This one is debug-only — in release `usize::MAX > bytes.len()` holds, so it lands in the "Chunk length is greater than available bytes" error path at `:256` and is reported correctly.

Offset 1 now joins the presence check, and the length uses `checked_sub` and returns that same parsing error. A bad entry is skipped or reported like any other and the rest of the region still loads, which is what the surrounding code already does for offset 0 and for an over-long length.

Both values come from the file, so a truncated or corrupt region reaches them. I have not tried to produce one from a real save — the tests build the region bytes directly.

## Testing

`cargo test -p pumpkin-world`, with three new tests: a location entry pointing at sector 1, the same entry at sector 2 as a control, and a chunk declaring length 0.

Reverting either fix on its own fails only its own test and leaves the control passing, so neither test is passing for an unrelated reason.

`cargo fmt --all -- --check` and `cargo clippy --all-targets` are clean, and `cargo test --workspace` passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed chunk data to prevent parsing failures caused by invalid length declarations.
  - Region entries pointing into reserved file headers are now treated as absent instead of being parsed as chunk data.
  - Added coverage for invalid, reserved, and valid chunk-location scenarios.

### DIFF
--- a/crates/pumpkin-world/src/chunk/format/anvil.rs
+++ b/crates/pumpkin-world/src/chunk/format/anvil.rs
@@ -247,8 +247,16 @@ impl AnvilChunkData {
 
     fn from_bytes(bytes: Bytes) -> Result<Self, ChunkReadingError> {
         let mut bytes = bytes;
-        // Minus one for the compression byte
-        let length = bytes.get_u32() as usize - 1;
+        // Minus one for the compression byte, which the length covers, so
+        // anything below one does not describe a chunk at all.
+        let declared_length = bytes.get_u32() as usize;
+        let Some(length) = declared_length.checked_sub(1) else {
+            return Err(ChunkReadingError::ParsingError(
+                ChunkParsingError::ErrorDeserializingChunk(
+                    "Chunk length does not cover its compression byte".to_string(),
+                ),
+            ));
+        };
 
         if length > bytes.len() {
             return Err(ChunkReadingError::ParsingError(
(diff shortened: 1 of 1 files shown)


RoosterBooster007 (approved):

RoosterBooster007:
Hey @Lstarsky0, thanks for this. One thing I noticed on the way through: the length check runs before the compression byte is consumed, so a declared length of payload−3 still panics at the slice; happy to see that as a follow-up if you want it.

PR #2974, opened 2026-08-16
feat(hurt_sound): add hurt sound for each mobs

zhangxuan2011:
## Description
In the previous version, if you hurt other mobs (except zombie, skeleton, ...), it will play generic hurt sounds.

To address this, this PR fixes this problem. Currently, in this PR-fixed version, at least in my own environment, if i hurt some other mobs, such as cows, sheep, pigs, villagers, it will automatically play their specific hurt sounds instead of playing generic hurt sounds.

## Changes
 - Updated entities config to add each sounds
 - Used `pumpkin-codegen` to re-generate the data

## Notes
1. This PR is not a complete fix. If you want to align to vanilla's hurt sound behavior **completely**, you must do these:
   - Each hurt of mobs shall use different pitch. For example, if I hit a cow, it will play almostly same specific hurt sounds (at least in my environment, in PR-fix version); Same as sheep, villagers and so on;
   - Added `death_sound` field. Once a mob died, it is still using the hurt sound -- That's incorrect. A mob died should use its death sound instead of its hurt sound.

    The to-do list here might not very full. If need, you can extend it in follow comments :)

2. Some other files are generated simultaneously, so the changes might very huge, but the re-generated data can work in my environment. If you mind that... hmm please forgive that qwq! I am no longer online T_T

## Testing
 - [x] `cargo clippy --all-targets`
 - [x] `cargo build`
 - [x] `cargo test`
 - [x] In-game test (in expected behavior)


kefu84510-cmd (approved):

kefu84510-cmd (approved):

kefu84510-cmd:
Very good

Snowiiii:
don't change the json assets manauly

PR #2912, opened 2026-08-12
Use TextComponent for ServerListPingEvent motd

Neol3108:
In addition to my last PR, this allows you to style your MOTDs.

Depends on https://github.com/Pumpkin-MC/pumpkin-plugin-wit/pull/30

PR #2989, opened 2026-08-18
Fix text component errors in events

Neol3108:
I will happily admit this is an AI PR 🤖🤖🤖 I couldn't figure out why setting my ServerListPingEvent MOTD to a TextComponent made my server crash on ping. I recently added the possibility to use TextComponent as MOTD in [this PR](https://github.com/Pumpkin-MC/Pumpkin/pull/2912), I tested it and it worked. But now it doesn't. AI suggested this, I personally feel like this isn't a good fix, but that's up to maintainers to decide I guess. But feel free to add a better fix.

I believe this commit/file caused it: https://github.com/Pumpkin-MC/Pumpkin/commit/8ce513ff8978609d3ecf66d5e83e98346ff89b37#diff-fba470b4a7a98ebb687a5a9e6858cf222740f0d673459e47e4813c35a01e74e8


Neol3108:
I think this also has to to with this:
image

Neol3108:
I guess this will be fixed with the re-entry fix(?)

PR #2990, opened 2026-08-18
fix(plugin): handle missing field conversions in packet codegen

Bluezly:
## Description

Some Java packet mappings were skipped because `packet_mapping.rs` didn't handle
`LpVector3d` or `Box<[u8]>`, even though both are already represented in the
packet WIT.

This adds the missing conversions:

- `Box<[u8]>` to and from WIT `list<u8>`
- `LpVector3d` to and from the existing WIT string representation using `[x, y, z]`

Regenerating `generated_packets.rs` now produces mappings for `CSpawnEntity`,
`CEntityVelocity`, `CSetEntityMetadata`, and other packets using boxed byte slices.

No packet WIT changes are required.

## Testing

- [x] `cargo run -p pumpkin-codegen`
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features`
- [x] `cargo test`


Snowiiii:
LGTM, Thank you @Bluezly :D

PR #2995, opened 2026-08-19
fix(piston): stop block swapping from corrupting pistons

### LINKED ISSUE #2191: Strange piston behaivor with falling blocks

### Operating System

Linux

Pumpkin 0.1.0-dev+26.1 (Commit: fa5e07f/release)

### What happened?

In certain situations, the main piston block breaks and a falling block appears in its place

Image

Image

### To Reproduce

1.
```
/give @s piston
/give @s redstone_torch
/give @s sand
```
2. Place a piston up and sand above it.
3. Quickly place and break a redstone torch to activate the piston (or place two observers).

### Expected behavior

In the vanilla game, the block is usually thrown up and doesn't have time to fall, but I also had it fall out as an item.

Image

Mcxiaocaibug:
Works towards the "Pistons (Sticky/Regular) & Block Swapping" item in #1402, and fixes #2191.

Pistons swap blocks around more than anything else in the game, so they are where these three divergences from Vanilla show up.

### Scheduled ticks ran against the wrong block

`tick_chunks` looked up the block currently at the position and ticked that, ignoring the block the tick was actually scheduled for — even though `ScheduledTick` already carries it. When a piston moved a block out from under a pending tick, the tick fired on whatever took the position over: a repeater pushed into a spot where sand had a pending fall check would run that check as a repeater tick.

Vanilla drops these in `ServerLevel::tickBlock` (`if (blockState.is(block))`) and `ServerLevel::tickFluid` (`if (fluidState.getType() == fluid)`). Both are now checked.

### Falling blocks destroyed pistons (#2191)

`FallingEntity` overwrote the block at its landing position unconditionally. A piston mid-animation occupies its own position with `moving_piston`, so a falling block landing on one replaced the piston: the piston block was gone and the falling block was left sitting where it had been — exactly what the screenshots in #2191 show.

Vanilla guards this with `if (!blockState.is(Blocks.MOVING_PISTON))` and leaves the entity alone, so it settles once the animation finishes. Same here.

### Pistons pushed blocks out of the world

`is_movable` carried a `TODO: more checks` and had no position, so the build-limit checks from Vanilla's `PistonBaseBlock::isPushable` were missing: positions outside the world are unpushable, pushing down at the bottom and up at the top is refused. Without them a block pushed past the limit was silently destroyed. The function now takes the world and position, and a small wrapper on `PistonHandler` keeps the call sites short.

### Verification

`cargo fmt --check`, `cargo clippy --all-targets --all-features` and `cargo test -p pumpkin` (257 passed) are all clean.

The first change is the one worth a careful look: dropping stale ticks is what Vanilla does, but if anything in Pumpkin was relying on a tick surviving a block change, it would now stop firing. I could not find such a case.

### DIFF
--- a/crates/pumpkin/src/block/blocks/piston/mod.rs
+++ b/crates/pumpkin/src/block/blocks/piston/mod.rs
@@ -1,6 +1,7 @@
 use piston::PistonBlock;
 use pumpkin_data::Block;
 use pumpkin_data::BlockDirection;
+use pumpkin_data::BlockState;
 use pumpkin_data::block_state::PistonBehavior;
 use pumpkin_util::math::position::BlockPos;
 
@@ -51,7 +52,8 @@ impl<'a> PistonHandler<'a> {
         self.broken_blocks.clear();
         let (block, block_state) = self.world.get_block_and_state(&self.pos_to);
 
-        if !PistonBlock::is_movable(
+        if !self.is_movable(
+            &self.pos_to,
             block,
             block_state,
             self.motion_direction,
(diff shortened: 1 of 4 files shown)


RoosterBooster007:
Please fix conflicts.

PR #3001, opened 2026-08-19
fix(pumkin): portal deadLock and finding portal logic

aure31:
## Description
There was a deadlock involving `world.portal_poi`, which was already locked and then attempted to be locked again inside `world.set_block_state()`.

Also, in the lookup logic, `world.get_block()` was being called while the chunk was not loaded, causing it to always return Air and failling to validate it.

link to  #2560 
## Testing
- [x] ingame testing
- [x] cargo suite 


Snowiiii:
LGTM, Thank you @aure31 :D

PR #3007, opened 2026-08-20
fix(net): validate the encryption response verify token

BryanFRD:
## Problem

`handle_login_start` generates a 4-byte verify token and sends it in
`CEncryptionRequest`, but the token is never stored and
`handle_encryption_response` never looks at `SEncryptionResponse::verify_token`.
The field is decoded off the wire and dropped.

Vanilla decrypts the echoed token and compares it against the one it sent, and
refuses the login on mismatch. Skipping that check drops the only anti-replay
binding between the encryption request and the response.

It also means `SEncryptionResponse` is accepted on a connection that was never
sent a `CEncryptionRequest` at all, which is reachable even when
`networking.java.encryption` is disabled.

## Change

`PendingConnection` now remembers the token it sent. `handle_encryption_response`
verifies it before anything else: RSA-decrypt the echoed token, compare with the
expected one, kick on mismatch.

Rejected cases that previously went through:

- a token that does not match the one sent
- a token that fails to decrypt
- an encryption response with no encryption request pending

The token is consumed with `take()`, so a second `SEncryptionResponse` on the
same connection is rejected rather than re-verified.

## The 1.19.3 to 1.20.1 exception

In that version window a client that owns a profile key signs `token + salt` with
it instead of echoing the token, and the decoder surfaces that as an empty token
(`SEncryptionResponse::read`). Those clients are still accepted, but **only**
inside that version range, otherwise a modern client could skip the check
entirely by sending a zero-length token. `may_omit_verify_token` is what enforces
that, and the test is what keeps it enforced.

Verifying the signature itself would require collecting and validating the
player's profile public key during login, which the server does not do today.
That is left as is, so the check remains weaker for those versions than for
current ones.

## On constant-time comparison

Left as a plain comparison on purpose. The token is a single-use random nonce,
regenerated per connection, and a failure closes the session, so there is no
byte-by-byte oracle to walk. This is unlike a long-lived shared secret.

## Tests

Added `only_profile_key_versions_may_omit_the_verify_token`, which pins the
version window and fails if the empty-token exception is ever widened to all
versions. Full `cargo test -p pumpkin --lib` is green (258 passed).


Snowiiii:
LGTM, Thank you @BryanFRD :D

PR #3009, opened 2026-08-20
refactor(world): remove the unused get_fluid_collisions

BryanFRD:
## Change

`World::get_fluid_collisions` has no callers. The only occurrence of the name in
the repository is its own definition, and it is not part of the plugin surface,
which goes through the WIT interface.

It is also a verbatim copy of `check_fluid_collision` sitting directly below it,
the only difference being that one accumulates into a `Vec` while the other
short-circuits. Deleting the unused one removes the duplication on its own, so
there is nothing left to factor out.

## Tests

No new test, this only removes code. `cargo clippy --all-targets` is clean, which
confirms no import was left dangling, and the existing suite passes (257).

### DIFF
--- a/crates/pumpkin/src/world/mod.rs
+++ b/crates/pumpkin/src/world/mod.rs
@@ -1739,34 +1739,6 @@ impl World {
         }
     }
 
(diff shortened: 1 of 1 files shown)


Snowiiii:
LGTM, Thank you @BryanFRD :D

PR #3011, opened 2026-08-20
fix(protocol): fix stream desync and lost formatting for named items in bundles

darkzmaj:
## Description

Putting a bundle containing a crafted (non-default) item - e.g. a firework rocket with real explosion data - into your inventory disconnects the client with `Error while handling incoming packet component_id_N (TODO)`.

**Root cause 1 - stream desync:** `minecraft:bundle_contents` encodes its nested items using the plain Slot format (`Prefixed Array of Slot`, per the protocol spec - no per-component byte-length prefix). `deserialize_item_stack_template` read and discarded a VarInt as if such a prefix existed, which actually consumed the first bytes of the component's real payload. Every read after the first non-default component on a bundled item was then offset, and the corrupted bytes eventually got misread as a bogus component ID - surfacing as the unrelated "component_id_N (TODO)" disconnect. The same bug pattern existed in the currently-unused `ItemStackSerializer::read`, fixed for consistency.

**Root cause 2 - lost/broken text formatting:** `CustomNameImpl`/`ItemNameImpl` serialize their text component as raw NBT (matching the wire format and the `item_stack_seralizer.rs` decode path used for Set Creative Mode Slot), but their generic `deserialize` read a plain length-prefixed wire string instead. This asymmetry meant any named item round-tripping through the generic codec (e.g. nested inside a bundle) would still fail even after fixing the length-prefix bug above. Added a small `NetworkReadExt -> Read + Seek` adapter so these parse the self-delimiting NBT payload directly off the wire, matching their `serialize` counterparts.

Checked existing issues/PRs first - nothing open covers this. The closest prior fix (#2223 / #2301) addressed a related but different anvil-rename path and didn't touch the bundle-contents codec.

## Testing

- Added a regression test (`bundle_with_named_firework_round_trips`) that serializes a bundle containing a renamed, non-default firework rocket and asserts it decodes back correctly. Confirmed it fails against the pre-fix code and passes against the fix.
- `cargo test -p pumpkin-protocol --lib` - 92 passed, 0 failed
- `cargo check -p pumpkin-protocol -p pumpkin -p pumpkin-inventory -p pumpkin-data` and `cargo check --workspace`
- `cargo clippy -p pumpkin-protocol` - clean
- `cargo fmt --check -p pumpkin-protocol` - clean
- Built and ran the server locally to confirm it starts cleanly with the change


QuantumSegfault:
Please remove all the unrelated `clippy` supressions. If they are still relavent, they should be done as a separate PR.

darkzmaj:
Fair call — reverted the clippy suppressions from this PR (and from #2946, where I'd also bundled them) and split them into their own PR: #3014.

darkzmaj:
Note for anyone (or anything) watching CI here: "Run lints (debug/release)" is expected to stay red on this PR until #3014 merges — that's the whole point of the split, this PR no longer carries the suppressions that were masking it. Not something to re-fix here.

RoosterBooster007:
Hey @darkzmaj, thanks for your work here! The bundle `_byte_len` desync is definitely a real bug and still present on master, so we definitely want this fix in.

The main challenge is that the branch has snowballed to 8 commits, and several parts have already landed elsewhere on master (the lodestone codec, entity-metadata versioning, and the legacy component ID mapping in `data_component_type_id_remap.rs`). Untangling and rebasing all of that in-place is going to be pretty messy.

Would you mind closing this and splitting it into two clean, focused PRs?
1. The template/bundle desync + `ItemName` NBT fix
2. The stale block-entity guards

Happy to review both once they're up!

PR #3012, opened 2026-08-20
Fix/chorus fruit teleport

### LINKED ISSUE #3010: Chorus Fruit doesn't teleport the player
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.40 (Commit: 0844e92)

### Operating System

Fedora 44

When a player eats a Chorus Fruit, they are not teleported.

According to [https://minecraft.wiki/w/Chorus_Fruit](https://minecraft.wiki/w/Chorus_Fruit):
An entity (such as a player or a fox) who eats a Chorus Fruit may also be teleported: up to 16 attempts are made to choose a random destination within ±8 on all three axes in the same way as Enderman teleportation,

### Steps to Reproduce

1. Get a Chorus Fruit (for exemple by giving it to you)
2. Try to eat it
3. You won't be teleported

matheustavarestrindade:
## Description

Fixes https://github.com/Pumpkin-MC/Pumpkin/issues/3010

Chorus Fruit consumption updated the server-side entity position without using the player teleport flow. As a result, Java clients remained in place.

This change implements the corresponding Java Edition behavior:

- Routes players through the client-aware teleport path.
- Tries up to 16 destinations within the configured diameter.
- Scans downward from each sampled height to find solid ground.
- Rejects unloaded, obstructed, and fluid-filled destinations.
- Resets fall distance and emits the teleport game event and sound after a successful teleport.
- Dismounts entities before teleporting.
- Sends only one teleport confirmation ID when teleporting from a vehicle, preventing confirmation-related disconnects.

The behavior remains in the data-driven `TeleportRandomly` consume-effect handler instead of being hardcoded for Chorus Fruit.

## Testing

- `cargo fmt --check`
- `cargo test -p pumpkin consumable_effect_tests`
- `cargo test -p pumpkin`
- `cargo clippy -p pumpkin --all-targets --all-features -- -D warnings`
- `cargo build -p pumpkin`
- Verified normal Chorus Fruit teleportation with a local Java client.
- Verified destinations do not place the player inside solid blocks or fluids.
- Verified enclosed destinations can fail after 16 attempts, matching Java Edition behavior.
- Verified eating Chorus Fruit in a minecart dismounts and teleports the player without a teleport-confirmation disconnect.

Quick note, i used AI to generate this description as my main language is not english, Thanks for the amazing project guys, im happy to contribute to it :)


yunuservices:
routing through Player::teleport is looks good overall and i want to ask you smh

in the `TeleportRandomly` handler the game event and sound positions are swapped relative to vanilla. vanilla emits `GameEvent.TELEPORT` at the entity's **new** position and plays `CHORUS_FRUIT_TELEPORT` at the **original** position(captured before the attempts loop). currently `emit_game_event("teleport", center)` uses the old position and `play_sound(..., &destination)` uses the new one, so they should be swapped.

and also...
- Emit the teleport particles on success(vanilla `randomTeleport` broadcasts entity event `46`)
- The `reposition=false` dismount path only sends `CSetPassengers` directly to java clients. is the bedrock dismount handled elsewhere?

matheustavarestrindade:
I checked the vanilla code. The game event position is captured before the teleport happens (old position), and the sound uses the entity's live position after the teleport (new position). Pumpkin does the same thing, so no change needed there.

Abou t the particles, Vanilla sends entity event 46 on a successful teleport, and we were missing that. Fixed in the newest commit , now the player gets the teleport particle effect.

Pumpkin's whole riding system is Java only right now. Mounting and dismounting only send Java packets (CSetPassengers), and no Bedrock link packet (CSetActorLink) is sent anywhere. So this isn't a new gap from my change as Bedrock clients can't ride anything on the server at all today. 

Adding a Bedrock unlink packet here would be dead code until riding itself works on Bedrock. I'd say that's a separate issue. what do you think?

Snowiiii:
LGTM, Thank you @matheustavarestrindade, Please follow the conventional commits spec for titles next time

PR #3014, opened 2026-08-21
chore(lints): resolve the pre-existing clippy::unused_async_trait_impl wall

darkzmaj:
## Summary
Split out per QuantumSegfault's feedback on #3011 — this was previously bundled into that PR and #2946, both unrelated.

`wasmtime`'s `bindgen!` requires every Host trait method to be `async fn`, regardless of whether a given implementation needs to `.await` anything. That triggers `clippy::unused_async_trait_impl` 578 times across the WASM plugin bridge (`crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1`), plus a handful of similarly-shaped trait impls elsewhere (redstone pressure plates, several plant blocks, the pathfinder, three `Entity` setter methods). This has been blocking "Run lints (debug/release)" CI on every branch regardless of what it actually changes.

- Allowed the lint at the module level for `wasm_host`'s submodules (one attribute per module covers every impl inside it) and per-function for the scattered non-WASM occurrences, rather than restructuring any of the affected code.
- Also fixes three unrelated pre-existing lint failures uncovered once the above stopped masking them:
  - `pumpkin-macros::count_placeholders` needed to be a `const fn` (`clippy::missing_const_for_fn`) — this alone was blocking every other lint check from running at all, since `pumpkin-macros` failing to compile takes the whole workspace down with it.
  - `serde_json5` unused dependency in `pumpkin-codegen` (`cargo-machete`).
  - `Result<_, ()>` in `Plugin::send_message` (`clippy::result_unit_err`).
  - `.ok().is_some_and(..)` that `clippy::manual_is_variant_and` wants as `.is_ok_and(..)`.

No behavior changes — this is purely lint suppression plus two semantically-equivalent one-liners.

## Test plan
- [x] `cargo clippy --all-targets --all-features` (debug) — clean
- [x] `cargo clippy --release --all-targets --all-features` — clean
- [x] `cargo test -p pumpkin --lib` — 257 passed

### DIFF
--- a/crates/pumpkin/src/block/blocks/plant/big_dripleaf.rs
+++ b/crates/pumpkin/src/block/blocks/plant/big_dripleaf.rs
@@ -229,6 +229,7 @@ impl PlantBlockBase for BigDripleafBlock {
         let support_block = block_accessor.get_block(pos);
         can_plant_dripleaf_on_top(support_block)
     }
+    #[allow(clippy::unused_async_trait_impl)]
     async fn get_state_for_neighbor_update(
         &self,
         block_accessor: &dyn BlockAccessor,
--- a/crates/pumpkin/src/block/blocks/plant/big_dripleaf_stem.rs
+++ b/crates/pumpkin/src/block/blocks/plant/big_dripleaf_stem.rs
@@ -49,6 +49,7 @@ impl PlantBlockBase for BigDripleafStemBlock {
         can_plant_dripleaf_on_top(support_block)
     }
 
+    #[allow(clippy::unused_async_trait_impl)]
     async fn get_state_for_neighbor_update(
         &self,
         block_accessor: &dyn BlockAccessor,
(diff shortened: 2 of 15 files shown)


QuantumSegfault:
Why suppress the warning instead of following `clippy`'s suggestion? (I'm still not SUPER familiar with Rust and particularly async, so maybe there is one).

darkzmaj:
For the wasm_host ones it's not really a choice for those Host trait methods are generated by wasmtime's bindgen! macro from the WIT interface, so the signature (including async) is fixed by the trait definition. Clippy's suggested rewrite (dropping async, wrapping the body in std::future::ready(...) and returning impl Future) would change the function signature, which breaks the trait impl entirely — it just won't compile against the generated trait.

The handful outside wasm_host (the three Entity setters, a couple of block/pathfinder methods) are inherent methods, not trait impls, so in principle they could drop async. I left them alone because they already had a pre-existing #[expect(clippy::unused_async)] before I touched anything — looks like a deliberate choice to keep entity mutators async-by-convention even when the current implementation doesn't need to await, so nothing has to change signature later if it starts doing network I/O. Didn't want to unilaterally reverse that pattern in a PR that's supposed to be just the lint cleanup.

QuantumSegfault (approved):

Snowiiii:
LGTM, Thank you @darkzmaj :D

PR #3015, opened 2026-08-21
fix(item): remove incorrect can_mine implementation for shears

LegendsOfXania:
This PR removes the custom can_mine() implementation from ShearsItem to restore parity with Vanilla.

Previously, can_mine was restricted based on player game mode (player.gamemode.load() != GameMode::Creative), which does not align with Vanilla behavior where shears work the same in Survial or in Creative mode.

# Testing
- [x] Verified block breaking with shears in Survival mode.
- [x] Verified block breaking with shears in Creative mode.
- [x] Tested that cargo test passes cleanly.


RoosterBooster007:
Looks good, @LegendsOfXania. Please fix conflicts.

LegendsOfXania:
Done!

RoosterBooster007:
Thanks, @LegendsOfXania!

PR #3017, opened 2026-08-22
feat(block): add farmland trampling

Simlesk:
## What

Implements vanilla farmland trampling: falling onto farmland now has a chance to turn it back to dirt, matching `FarmlandBlock.fallOn` in vanilla 26.2.

- Added `on_landed_upon` to `FarmlandBlock` (`crates/pumpkin/src/block/blocks/farmland.rs`)
- Made `Entity::get_pos_with_y_offset` `pub(crate)` so block behaviours can resolve the block position an entity landed on (`OnLandedUponArgs` doesn't carry a position, unlike vanilla's `fallOn(level, state, pos, entity, fallDistance)`)

## Why

Farmland could never be trampled, so crop farms were immune to landing damage on the soil — a gameplay gap vs vanilla.

## Vanilla accuracy

Logic verified line-by-line against the decompiled official 26.2 server (`FarmlandBlock.fallOn`):

- trample chance: `random.nextFloat() < fallDistance - 0.5`
- only living entities; non-players additionally require the `mobGriefing` game rule
- entity size gate: `width * width * height > 0.512` (small mobs never trample)
- trample happens before regular fall damage is applied (same order as vanilla)
- farmland becomes `dirt` (default state)

## How I verified it

- `cargo clippy --all-targets`: no warnings
- `cargo test`: all tests pass
- In-game (26.2 client, survival): repeatedly jumping on farmland converts it to dirt at roughly the expected rate; higher falls convert near-always; fall damage unchanged

## Known limitations

- Crops on top currently do NOT pop when the farmland under them converts to dirt. This is a pre-existing bug, not introduced here: the neighbor-update survival check for crops resolves to the generic `PlantBlockBase::can_plant_on_top` (which accepts any `supports_vegetation` block, including dirt) instead of `CropBlockBase::can_plant_on_top` (which requires farmland). I'd like to fix that in a separate follow-up PR to keep this diff small.
- Vanilla's `turnToDirt` calls `pushEntitiesUp`; skipped here, consistent with the existing `// TODO: push up entities` markers already in `farmland.rs`
- Vanilla emits `GameEvent.BLOCK_CHANGE` (sculk vibration) on trample; no block behaviour currently emits game events, so left out for consistency
- A cleaner long-term shape might be adding `position` to `OnLandedUponArgs` (vanilla's `fallOn` receives the pos directly) — left out to keep this diff minimal

### DIFF
--- a/crates/pumpkin/src/block/blocks/farmland.rs
+++ b/crates/pumpkin/src/block/blocks/farmland.rs
@@ -3,6 +3,7 @@ use std::sync::Arc;
 use crate::block::BlockBehaviour;
 use crate::block::CanPlaceAtArgs;
 use crate::block::GetStateForNeighborUpdateArgs;
+use crate::block::OnLandedUponArgs;
 use crate::block::OnPlaceArgs;
 use crate::block::OnScheduledTickArgs;
 use crate::block::RandomTickArgs;
@@ -20,6 +21,7 @@ use pumpkin_util::math::vector3::Vector3;
 use pumpkin_world::tick::TickPriority;
 use pumpkin_world::world::BlockAccessor;
 use pumpkin_world::world::BlockFlags;
+use rand::RngExt;
 
 type FarmlandProperties = FarmlandLikeProperties;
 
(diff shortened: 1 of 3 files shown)


RoosterBooster007:
Please fix conflicts.

RoosterBooster007:
Please fix conflicts.

PR #3022, opened 2026-08-22
fix(command): send vanilla-accurate feedback for /spawnpoint

### LINKED ISSUE #3018: Message without coordinates: Set spawn point to %s, %s, %s [%s, %s] in %s for %s
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.40 beb6947

### Operating System

Ubuntu

When player sets player's spawn location with /spawnpoint, it sends a message without parsed coordinates:
`Set spawn point to %s, %s, %s [%s, %s] in %s for %s`

Image

Client: 26.2
Server: 26.2 (0.1.0-dev+26.2-26.40 beb6947)

### Steps to Reproduce

1. Join server
2. Type /spawnpoint

Simlesk:
Fixes #3018

## What

Makes `/spawnpoint` feedback and argument handling match vanilla 26.2:

- Sends the 7 arguments vanilla sends, in vanilla order: x, y, z, yaw, pitch, dimension, player name (was: 4 args in the order name, x, y, z — the client fails formatting and displays the raw `%s` pattern, the symptom in #3018)
- Distinguishes `commands.spawnpoint.success.single` (1 target, shows name) from `commands.spawnpoint.success.multiple` (N targets, shows count, one message total) like vanilla
- Passes the pitch through to the respawn point (was hardcoded to 0) and applies vanilla's `wrapDegrees(yaw)` / `clamp(pitch, -90, 90)`
- Uses zero rotation for the no-angle variants and the sender's position for the no-pos variants, like vanilla (`WorldCoordinates.ZERO_ROTATION`, `source.getPosition()`); falls back to the first target's position for senders without a position (console)
- Fixes all `/spawnpoint <targets> ...` variants being broken with "internal error": the argument is registered as `targets` but was resolved via `find_arg_default_name`, which looks up the consumer's default name `target`. Now resolved with `find_arg(args, ARG_TARGETS)` like other commands (e.g. `experience.rs`)

## How I verified it

- Decompiled official 26.2 server: `SetSpawnCommand.setSpawn` sends `commands.spawnpoint.success.single` with `{x, y, z, yaw, pitch, dimensionId, displayName}` (or `.multiple` with count)
- In-game (26.2 client): `/spawnpoint` shows "Set spawn point to X, Y, Z [0, 0] in minecraft:overworld for <name>"; `/spawnpoint <player> 100 64 100 45 10` works (previously "internal error") and shows `[45, 10]`
- `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` all pass

## Known limitations

- Vanilla uses the player's display name (team-colored); this uses the profile name, consistent with the previous code
- Bedrock's `commands.spawnpoint.success.single` has a different argument order (name first); `translate_cross` shares one argument list across editions, so the bedrock rendering is best-effort — pre-existing limitation of the shared-args mechanism

### DIFF
--- a/crates/pumpkin/src/command/commands/spawnpoint.rs
+++ b/crates/pumpkin/src/command/commands/spawnpoint.rs
@@ -2,17 +2,17 @@ use std::sync::Arc;
 
 use pumpkin_data::translation;
 use pumpkin_util::math::position::BlockPos;
+use pumpkin_util::math::wrap_degrees;
 use pumpkin_util::text::TextComponent;
 
 use crate::command::args::players::PlayersArgumentConsumer;
 use crate::command::args::position_block::BlockPosArgumentConsumer;
 use crate::command::args::rotation::RotationArgumentConsumer;
-use crate::command::args::{Arg, ConsumedArgs, FindArgDefaultName};
+use crate::command::args::{Arg, ConsumedArgs, FindArg};
 use crate::command::dispatcher::CommandError;
 use crate::command::tree::CommandTree;
 use crate::command::tree::builder::argument;
 use crate::command::{CommandExecutor, CommandResult, CommandSender};
-use crate::entity::EntityBase;
 use crate::entity::player::Player;
 
 const NAMES: [&str; 1] = ["spawnpoint"];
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Thanks for the fix, but it looks like it already landed in d1cfc0d18. I'll close this out as superseded.

PR #3024, opened 2026-08-22
fix(block): break crops when the block below is no longer farmland

Simlesk:
## What

Crops (wheat, carrots, potatoes, beetroots, torchflower crop) now break — with their drops — when the block below them stops being farmland, matching vanilla.

Each of the five crop blocks gets the same 5-line override: `PlantBlockBase::can_plant_on_top` now delegates to the farmland check from `CropBlockBase`, following the exact pattern already used by `nether_wart` (soul sand check) and the melon/pumpkin stems.

## Why

The neighbor-update survival path goes through `PlantBlockBase::get_state_for_neighbor_update`, whose `self.can_plant_on_top(...)` call statically resolves to `PlantBlockBase`'s own method (generic check: any block tagged `supports_vegetation`, which includes dirt) — not to `CropBlockBase::can_plant_on_top` (requires farmland), which is a separate trait method that only shadows it where called explicitly.

As a result, when farmland converts to dirt under a crop (trampling — see #3017 — or drying out), the crop happily survives on dirt instead of popping.

No changes needed on the breaking side: `World::replace_with_state_for_neighbor_update` already routes an air result through `break_block`, which rolls the crop's loot table.

## How I verified it

- In-game (26.2 client, combined with #3017 locally): jumping on planted farmland converts it to dirt and the wheat pops with its drops; breaking farmland under a crop by hand does the same
- `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` all pass

## Notes

`nether_wart` and the gourd stems already override `can_plant_on_top` correctly and are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved crop placement and survival validation for beetroot, carrots, potatoes, torchflowers, and wheat.
  * Crops can now grow on blocks tagged as supporting crops, rather than only on standard farmland.
  * Updated plant checks to consistently apply crop-specific soil requirements.


Lstarsky0 (commented):
Diagnosis checks out against the shipped tag data. `assets/tags/26_2_tags.json` has `minecraft:supports_vegetation` as dirt, coarse_dirt, rooted_dirt, mud, muddy_mangrove_roots, moss_block, pale_moss_block, grass_block, podzol, mycelium, farmland — so the default really does keep a crop alive on plain dirt.

Coverage is complete. There are eight `impl CropBlockBase`: the five here, plus `NetherWartBlock` and `StemBlock` which already carry their own `PlantBlockBase::can_plant_on_top`, plus `SweetBerryBushBlock`, which correctly keeps the `supports_vegetation` default since berries are not a farmland crop. `AttachedStemBlock` overrides it directly at `gourds/attached_stem.rs:73` and is unaffected.

Two things around the edges.

`CropBlockBase::can_plant_on_top` is the only identity comparison in this tree — `block == &Block::FARMLAND`. `PlantBlockBase` uses `supports_vegetation`, `NetherWartBlock` uses `supports_nether_wart`, `StemBlock` and `AttachedStemBlock` use `supports_*_stem`, all through `has_tag`. The tag file already ships `minecraft:supports_stem_crops` as exactly `["farmland"]`, so the same answer is available in the same form as everything around it, and it moves with the data instead of being pinned to one block id.

The reason five files need the same five lines is that `CropBlockBase::can_plant_on_top` has the same name as the supertrait method without being an override, so `can_place_at` resolves the `PlantBlockBase` default and never sees it. That is what produced this bug and it will produce the next one when a sixth crop is added. Giving the `CropBlockBase` method a distinct name would make the delegations read as necessary rather than as copy-paste.

Separate from this PR, but in the file the pattern is copied from: `StemBlock::can_plant_on_top` (`gourds/stem.rs:154`) branches on `block == &Block::PUMPKIN_STEM`, and `can_place_at` passes `block_pos.down()`, so `block` is the support and that branch never fires — every stem falls through to the melon check. It happens to be harmless because `supports_pumpkin_stem` and `supports_melon_stem` are both `["farmland"]`, but it would stop being harmless the moment they differ.

Simlesk:
Thanks for the thorough review — both suggestions adopted, with one correction on the tag name.

- **Tag-based check**: done, but using `minecraft:supports_crops` rather than `supports_stem_crops` — decompiled 26.2 `CropBlock.mayPlaceOn` is `state.is(BlockTags.SUPPORTS_CROPS)`, and `supports_stem_crops` is what `StemBlock`/`AttachedStemBlock` use. Both ship as `[farmland]` today, so behaviour is identical, but this matches vanilla per-block.
- **Distinct name**: `CropBlockBase::can_plant_on_top` is now `can_plant_crop_on_top`, with a comment explaining why it must not share the supertrait method's name. Renamed the existing overrides in `nether_wart` and `sweet_berry_bush` accordingly, and switched the five crops' `BlockBehaviour::can_place_at` to go through `PlantBlockBase::can_place_at` so there is a single dispatch path (same shape as `nether_wart`).

Agreed on the `StemBlock::can_plant_on_top` dead branch — leaving it out of this PR as you suggested; happy to follow up separately.

RoosterBooster007 (approved):

RoosterBooster007:
Thanks, @Simlesk!

PR #3032, opened 2026-08-23
fix(bedrock): unsupported version disconnect

ZlordHUN:
## Description
Rejects unsupported Bedrock client versions with the correct native update message instead of allowing the connection to time out.

## What
- Rejects Bedrock clients older or newer than Pumpkin’s supported protocol.
- Displays Bedrock’s native version mismatch message.
- Correctly tells outdated clients that the server is running a newer version.
- Closes unsupported NetherNet sessions immediately after sending the response.
- Handles older clients that attempt a legacy RakNet connection instead of NetherNet.

## How
- Uses Bedrock’s `PlayStatus` response when the client reaches the normal NetherNet login sequence.
- Detects RakNet `OpenConnectionRequest1` packets from older clients.
- Returns the minimal RakNet incompatible-protocol response without restoring the legacy RakNet game transport.
- Derives the response version from the requesting client so Bedrock displays the correct update direction.

## Testing
- `cargo test -p pumpkin-protocol encodes_raknet_incompatible_protocol_response`
- `cargo test -p pumpkin net::bedrock::status --lib`
- Tested with Bedrock 26.0 and up until 26.40 (which is supported) and confirmed it receives the correct message stating that the host uses a newer Minecraft version.

### DIFF
--- a/crates/pumpkin-protocol/src/bedrock/status.rs
+++ b/crates/pumpkin-protocol/src/bedrock/status.rs
@@ -61,6 +61,26 @@ impl PacketWrite for CUnconnectedPong {
     }
 }
 
+#[derive(PacketWrite)]
+#[packet(0x19)]
+pub struct CIncompatibleProtocolVersion {
+    protocol_version: u8,
+    magic: [u8; 16],
+    #[serial(big_endian)]
+    server_guid: u64,
+}
+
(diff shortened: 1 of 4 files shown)


Snowiiii:
LGTM, Thank you @ZlordHUN

PR #3033, opened 2026-08-23
fix(bedrock): game logic deadlock

ZlordHUN:
## Description
Fixes Bedrock sessions freezing after loading only part of the initial world.

## What
- Restores continuous chunk loading and gameplay updates for Bedrock players.
- Ensures Bedrock packets use each client’s negotiated compression settings.
- Prevents chunk-cache handling from holding a lock during network operations.

## How
- Encodes Bedrock broadcasts through the connected client’s packet encoder instead of an uncompressed default encoder.
- Uses the awaited encoder path for asynchronous packet delivery.
- Releases the blob-cache lock before sending cache-miss responses, preventing lock contention and network backpressure from blocking chunk processing.

## Testing
- `cargo test --workspace --all-targets`
- `cargo test --workspace --doc`
- `cargo check -p pumpkin`
- Verified in game that Bedrock chunks continue loading, mobs and block interactions function as intended.

### DIFF
--- a/crates/pumpkin/src/entity/mod.rs
+++ b/crates/pumpkin/src/entity/mod.rs
@@ -3054,11 +3054,9 @@ impl Entity {
                 },
                 tick: VarULong(0),
             };
-            if let Ok(packet_data) =
-                pumpkin_protocol::bedrock::packet_encoder::serialize_packet(&packet)
-            {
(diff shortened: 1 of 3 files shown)


Snowiiii:
LGTM, Thank you @ZlordHUN :D

PR #2614, opened 2026-07-27
Client version control

sirfusebox:
https://github.com/Pumpkin-MC/Pumpkin/issues/2613

Adds a configurable version-control system for Java Edition clients under networking.java.versions in pumpkin.toml.

PR #3035, opened 2026-08-24
fix(config): reject view distance of 1, which crashed the server on first join

catdum200:
## Description

`PumpkinConfig::validate` contains a guard that cannot fail:

```rust
let min_vd = NonZero::<u8>::MIN;
// ...
assert!(
    self.advanced.networking.java.view_distance >= min_vd,
    "Java View distance must be at least 2"
);
```

`view_distance` is a `NonZero<u8>` and `NonZero::<u8>::MIN` is `1`, so the comparison
is true by construction. The message promises a minimum of 2, but 1 is accepted. The
Bedrock assert a few lines below has the same bound.

**Why this matters.** `chunker::get_view_distance` evaluates

```rust
let fallback = NonZero::new(2).unwrap_or(NonZero::<u8>::MIN); // 2
// ...
player.config.load().view_distance.clamp(fallback, max_view_distance)
```

where `max_view_distance` is the configured value. `Ord::clamp` opens with
`assert!(min <= max)`. That is a plain `assert!`, so it is live in release builds, and
the `clippy::panic` / `unwrap_used` denies do not see it because it sits inside `std`.
With `view_distance = 1` it evaluates `assert!(2 <= 1)`.

`get_view_distance` is called from the join and chunk-tracking paths (eight call sites in
`world/mod.rs`, plus `chunker::update_position`), and the panic hook in `main.rs` calls
`stop_server()` on the first panic from a worker thread. So a server configured with
`view_distance = 1` starts normally and then **shuts down completely when the first player
connects**, before that player finishes joining.
`net/bedrock/play/request_chunk_radius.rs` clamps to the same floor of 2 and has the same
exposure.

**What changed.** `min_vd` is now 2 — the bound the assert message already documents, and
the same floor that both `get_view_distance` and the Bedrock chunk-radius handler clamp to
— written with the let-else form already used for `max_vd` on the next line. A misconfigured
server now fails fast at startup with `Java View distance must be at least 2` instead of
dying on first join with an opaque `min > max. min = 2, max = 1` that points into
`core/src/num/nonzero.rs` and names no Pumpkin code.

**Impact.** One line of behaviour change. A `view_distance` of 1 was never usable — it
guaranteed a crash on first join — so no working configuration is rejected. Values 2..=64
are unaffected. The configured maximum is read straight from `advanced_config` and has no
runtime setter, so validation is the only place it needs enforcing.

**Known limitations.** The neighbouring `<= max_vd` assert says "must be less than 64"
while checking `<= 64`. That is a one-off in the message only, it is harmless, and I left
it alone to keep this diff on the crash. `simulation_distance` has no validation at all,
but it never reaches a `clamp`, so it is a separate concern and is not touched here.

**Difference from existing PRs.** No open PR touches this validation or
`get_view_distance`. Of the five open PRs that touch any of the files involved (#2985,
#2772, #2614, #2521, #2488), none modifies the bound or the clamp; #2521 only removes an
unrelated `get_view_distance` call site in `world/mod.rs` as part of per-player entity
tracking. #2908 also clamps a value, but for entity health, with no shared code.

## Testing

Four new tests in `crates/pumpkin-config/src/lib.rs`, two of which fail on `master`.

Before the change, `cargo test -p pumpkin-config --lib`:

```
test tests::bedrock_view_distance_below_minimum_is_rejected - should panic ... FAILED
test tests::java_view_distance_below_minimum_is_rejected - should panic ... FAILED
test tests::default_config_is_valid ... ok
test tests::view_distance_of_two_is_accepted ... ok

---- tests::java_view_distance_below_minimum_is_rejected stdout ----
note: test did not panic as expected at crates\pumpkin-config\src\lib.rs:441:8

test result: FAILED. 13 passed; 2 failed; 0 ignored
```

After the change: `test result: ok. 15 passed; 0 failed; 0 ignored`.

The other two are controls that pass both before and after, so they show the bound moved to
exactly 2 rather than the assert simply becoming trigger-happy: `default_config_is_valid`
keeps the shipped defaults valid, and `view_distance_of_two_is_accepted` pins the boundary.

Full CI gate on `4c7100061` plus this commit, Rust 1.98.0:

| Command | Result |
| --- | --- |
| `cargo fmt --check` | exit 0 |
| `RUSTFLAGS=-Dwarnings cargo clippy --all-targets --all-features` | exit 0 |
| `RUSTFLAGS=-Dwarnings cargo clippy --release --all-targets --all-features` | exit 0 |
| `cargo test --workspace` | exit 0 — 713 passed / 0 failed / 16 ignored (709 on master, plus the 4 above) |

### Live server, release build, real Java client

Same `pumpkin.toml` throughout; only the binary and `view_distance` differ.

**Unpatched, `view_distance = 1`.** Server starts normally, then a player connects and the
whole server goes down. It never logs `joined the game` — the panic lands mid-join.

```
Message: min > max. min = 2, max = 1
Panic Location: ...\core\src\num\nonzero.rs:277:49
--- Panicking Thread ---
ID: ThreadId(2)
Name: tokio-rt-worker
```

followed by `Gracefully shutting down...` and `The server has stopped.`

**Patched, same `view_distance = 1`.** Refuses to start, exit code 1:

```
Message: Java View distance must be at least 2
Panic Location: crates\pumpkin-config\src\lib.rs:103:9
--- Panicking Thread ---
ID: ThreadId(1)
Name: main
```

**Patched, `view_distance = 2`** (the new minimum). Starts in 26 ms, the client joins and
plays normally, and no crash report is produced:

```
INFO pumpkin::world: <player> joined the game
INFO pumpkin::world: <player> left the game
```

The `assert!(min <= max)` inside `Ord::clamp` was also confirmed to fire in an optimised
build independently of Pumpkin (`rustc -O`, no debug assertions), which is why this is a
release crash rather than a debug-only one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Configuration validation now consistently enforces a minimum view distance of 2 for Java and Bedrock configurations.


Snowiiii:
Nonsense tests

Q2297045667:
If it's just a few lines of code change, there's no need to test it. Unless there are major changes to the functionality or logic, testing isn't necessary

RoosterBooster007:
@catdum200, are you still interested in contributing here? Would you like me to remove the tests for you?

catdum200:
Yes, still interested - thanks for the nudge, and no need to do it for you: I've pushed a commit removing the tests.

The PR is now just the two-line bound fix:

```diff
-        let min_vd = NonZero::<u8>::MIN;
+        let Some(min_vd) = NonZero::new(2) else {
+            return;
+        };
```

`NonZero::<u8>::MIN` is 1, so for a `NonZero<u8>` field the assert was true by construction and never fired, despite its message saying the minimum is 2. A `view_distance = 1` config was therefore accepted, and `chunker::get_view_distance` then ran `clamp(2, 1)`, which panics in release and takes the server down on first join.

RoosterBooster007:
Thanks, @catdum200!

PR #3041, opened 2026-08-24
fix(worldgen): generate both halves of double-height plants

Love0118:
## Description

Updates `SimpleBlockFeature` to place both the lower and upper states of double-height plants such as sunflowers, lilacs, and tall grass, matching vanilla world generation.

Previously, `SimpleBlockFeature` wrote only the single configured block state. This produced an orphaned lower half and could still place that lower half when the position above was obstructed.

### Changes

- Identified double-height plants through `TallSeagrassLikeProperties` and resolved them into explicit `half=lower` and `half=upper` state pairs.
- Added upper-position validation before either state is written, placing neither half when the position above is obstructed.
- Ensured that unrelated blocks with a `half` property, such as stairs, are not treated as double-height plants.

## Testing

- Added unit tests in `simple_block::tests`:
  - Resolves a sunflower into lower and upper state pairs.
  - Rejects unrelated blocks that expose a `half` property.
- `cargo fmt --all -- --check` (passed)
- `cargo test -p pumpkin-world simple_block::tests --lib -- --nocapture` (2 passed)
- `cargo check -p pumpkin-world` (passed)
- `cargo test --workspace --lib --bins --tests --examples` (17 suites, 723 passed on the isolated PR branch)
- `cargo clippy -p pumpkin-world --lib --all-features -- -D warnings` (clean)
- `cargo build --release -p pumpkin --bin pumpkin` (passed)
- In-game verification completed: generated double-height plants contained both halves, obstructed upper positions did not leave orphaned lower halves, and ordinary single-block vegetation generation did not regress.

### DIFF
--- a/crates/pumpkin-world/src/generation/feature/features/simple_block.rs
+++ b/crates/pumpkin-world/src/generation/feature/features/simple_block.rs
@@ -1,4 +1,7 @@
-use pumpkin_data::Block;
+use pumpkin_data::{
+    Block, BlockState,
+    block_properties::{BlockProperties, DoubleBlockHalf, TallSeagrassLikeProperties},
+};
 use pumpkin_util::{math::position::BlockPos, random::RandomGenerator};
 
 use crate::generation::proto_chunk::GenerationCache;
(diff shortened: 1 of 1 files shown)


RoosterBooster007:

PR #3045, opened 2026-08-25
feat(plugin-api): Better method chaining (Rust)

QuantumSegfault:
## Description

Updates `wit-bindgen` for the `pumpkin-plugin-api` (Rust bindings) to `0.61` to take advantage of https://github.com/bytecodealliance/wit-bindgen/pull/1602

The configuration has been updated so that only some resources enable method chaining, and that said methods/resources use owning method chaining, rather than borrowing (they return `Self` rather than `&Self`).

## Testing

Testing was done to ensure bindings still compile, and that a Rust plugin using commands and text components can compile successfully (after modification).

Some of the methods in the crate wrap around generated methods from the WIT, so they show evidence that the transition to owning borrowing is working.

`wit-bindgen` itself has tests for whether the chaining functions correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5298,16 +5298,6 @@ dependencies = [
 
 [[package]]
 name = "wasm-encoder"
-version = "0.254.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "09480d646178e5fdd12bb06e812d0af9a3a191dbc9cd697fdc86687beade7393"
-dependencies = [
- "leb128fmt",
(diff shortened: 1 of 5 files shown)


QuantumSegfault:
CI fails cause of a new issue introduced into `0.61`. I have a patch here: https://github.com/bytecodealliance/wit-bindgen/pull/1697

In the meantime, I've but in a stub for `cabi_realloc` to satisfy the linker (the symbol is otherwise never called).

Snowiiii:

PR #3050, opened 2026-08-25
fix(worldgen): existing chunk state loading

ZlordHUN:
### Description
Prevents existing chunks from appearing missing or being overwritten with empty chunk data during loading and regeneration. This PR was made cause I've been using the same world with Pumpkin and testing for a while now. After recent changes to the main branch a bunch of my chunks got corrupted which resulted in my world having holes that resulted in entities and me falling out of the map. This PR addresses that issue and could potentially come in handy in the future for long term worlds with the same issue.

### What
- Distinguishes genuinely missing chunks from chunks that failed to load.
- Prevents empty, ungenerated `ProtoChunk`s from being saved.
- Preserves pending block entities and scheduled block/fluid ticks when existing chunks pass through `ProtoChunk`.
- Restores state-dependent block entity data such as hopper facing.
- Prevents stale or mismatched hopper block entities from panicking during chunk transitions.

### How
- Reports chunk read failures through the scheduler instead of treating them as new chunks.
- Only persists `ProtoChunk`s that have progressed beyond the empty generation stage.
- Copies block entities and scheduled ticks during full chunk-to-proto conversion.
- Validates loaded block entities against their current block state before ticking them.
- Uses loaded-only, validated block states when ticking or updating hoppers.

### Testing
- `cargo test -p pumpkin-world -p pumpkin`
- `cargo clippy -p pumpkin-world -p pumpkin --all-targets -- -D warnings`
- Verified that existing chunks load without the previously observed missing rectangular regions.

### DIFF
--- a/crates/pumpkin-world/src/chunk_system/chunk_state.rs
+++ b/crates/pumpkin-world/src/chunk_system/chunk_state.rs
@@ -327,7 +327,7 @@ impl Chunk {
             x: proto_chunk.x,
             z: proto_chunk.z,
             dirty: AtomicBool::new(true),
-            block_ticks: ChunkTickScheduler::default(),
+            block_ticks: ChunkTickScheduler::from_iter(proto_chunk.block_ticks),
             fluid_ticks: ChunkTickScheduler::from_iter(proto_chunk.fluid_ticks),
             pending_block_entities: Mutex::new(pending_block_entities),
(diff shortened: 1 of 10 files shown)


ZlordHUN:
Started a new Pumpkin world. Will revisit if the issue occurs in that one as well.

PR #3056, opened 2026-08-25
Fire cancellable entity interaction event on Bedrock

Neol3108:
This event worked on Java player interaction, but not for Bedrock

## Summary by CodeRabbit

- **New Features**
  - Added cancellable events for player interactions with entities, allowing interactions to be modified or prevented before execution.
  - Entity interactions now support updated action handling, including standard interaction, targeted interaction, and attacks.

- **Bug Fixes**
  - Invalid interaction types or entity targets are rejected safely with warnings instead of being processed.
  - Interaction behavior now follows the final approved action, improving consistency between plugins and gameplay.


RoosterBooster007:
Hey @Neol3108, thanks for hooking the Bedrock path into `PlayerInteractEntityEvent`. One small thing before merge: that `unreachable!` for `InteractAt` is actually reachable, since a WASM plugin can hand back `EntityInteractionAction::InteractAt` and it gets mapped straight onto the event. Java just treats `Interact | InteractAt` together, so could you fold it into the same arm? Should be good to go after that.

RoosterBooster007:
Thanks, @Neol3108!

PR #3057, opened 2026-08-25
feat!: Correct Piston behaviour

zinedinp:
Goal of this fork: get TNT dupers working on Pumpkin. That means pistons, redstone, and minecarts have to behave like vanilla on a concurrent async tick, without races. Along the way a few related crashes and timing bugs showed up and were fixed.

Vanilla does all of that single-threaded, in creation order. Pumpkin ticks block entities and entities on a JoinSet. Without a sequential island around the moving piston, two placeholders can place into the same neighbour, and a cart can tick against last tick’s rail.

**Pistons and tick order**

- PistonBatch: order-sensitive block entities (PistonMovingBlockEntity) leave the concurrent BE JoinSet and tick one at a time, oldest-first (blockEntityTickers creation order; NBT-restored BEs tie-break by y, z, x).
- Entities whose AABB overlaps the construction (cell plus one-cell sweep, PistonMath.getMovementArea) also leave the entity JoinSet and tick sequentially by entity id. A cart on piston-driven slime therefore reads the rail the placeholders rewrite this tick.
- World tick order matches vanilla: pending ticks, then runBlockEvents (pistons), then entities, then block entities. /tick freeze no longer flushes block events.
- handlingTick is true during scheduled ticks and block events, so 0-tick sticky retract (getRetractType / lastTicked) works.
- Piston push/pull closer to vanilla: height-limit unpushable, sticky resolve() re-evaluates list size while discovering slime branches, retract does not destroy DESTROY blocks, neighbor/shape updates after FORCE_STATE, captured pre-move block as the update source (not air). Quasi-connectivity hasSignal is on the piston cell, not the block below.

**Piston motion and collision**

- PistonMovingBlockEntity ports moveCollidedEntities, getMovementArea, slime (not honey) launch, skip markers/displays, no ServerPlayer fling, source never launches.
- Retracting source uses the stationary head/base shape. finalTick does not strip waterlogged.
- Entity::move_entity_piston (MoverType.PISTON) is the path carts take when carried; rail collision-ignore cells are recomputed from current pos, not last tick’s.
- MOVING_PISTON collision comes from the live BE’s animated boxes. BlockAccessor getters resolve the animated block; raw world getters stay as stored.

**Redstone and rails**

- Detector rail was mostly a stub; it now has getSearchBB (0.2 inset), checkPressed, updatePowerToConnected (including slope connections), 20-tick unpower, strong power only up, analog 0 unless a chest/hopper cart is on the rail.
- Detector POWERED is not re-derived before carts load: tick_chunks waits on entity_ready_chunks.
- Scheduled ticks use signed subTickOrder (vanilla LevelChunkTicks.unpack): NBT-restored ticks sort ahead of live ones, same-tick same-priority order is stable after reload.

**Minecarts**

- Track motion follows OldMinecartBehavior: halt_track brakes before move, power_track boosts after move + friction.
- getMaxSpeed clamps the step, not deltaMovement (slime launch is 1 block/tick).
- Off-track: ground halves before move; air drag after; never both. Occupied vehicle scale 0.75.
- Rail snap is a teleport setPos. Collision context ignores the cell under the rail (and the slope climb cell).
- Explosions treat all MINECRAFT_RAILS as rails, not a hard-coded four-block list (TNT minecart rail protection).

**TNT**

- Prime is atomic: swap TNT to air first under the section lock. Concurrent flint-and-steel vs redstone neighbor update cannot ignite the same block twice. Cancelled plugin events restore the block.
- Primed TNT applies air drag 0.98 every tick, then ground (0.7, -0.5, 0.7) on top (vanilla, not either/or).
- Fuse metadata is marked dirty every tick, so the tracker resyncs almost every tick (vanilla setFuse), not on the 10-tick interval.

**Bugs caught along the way**

- BE JoinSet ticked a hopper after chunk unload; HopperLikeProperties::from_state_id(air) panicked and the panic hook killed the process. Guard + skip unloaded BEs.
- Chunk NBT entities went into entities but not entities_by_chunk, so detector rails / get_entities_at_box missed carts. Index on load.
- get_entities_at_box walked every entity (hot on rails/pistons). Now a chunk-column index plus one-column slack.
- Hottest CPU path was block_in_place inside the async tick; made async.
- Fences/walls stick out of their cell; queries that only visit overlapped cells missed them (hasLargeCollisionShape).
- is_dirty is also consumed by the WASM API; comparators now use a separate is_comparator_dirty poke each tick.
- Frozen worlds still flushed block events on the leftover path; freeze now skips that.

I know this is a huge commit. feel free to change and ask me. Take your time.

Servus


Cargo.toml
KairuDeibisu, line 1: This change is a little out of scope. It looks to me kind of like test code you forgot to remove. Can you explain why this should be changed to achieve what you're trying to do in this pull request?

KairuDeibisu (changes requested):

Cargo.toml
zinedinp: I wanted to achive a working Vanilla like TNT-Duper.
Therfore many things needed to come together:

- Minecart behaviour
- Minecart Collision
- Detector Rails
- Block Update/Quasi-Connectivity sequence
- Piston checking behavior (x/z/y)
- Slimeblock pushing behaviour
- Minecarts on Piston Kontraptions
- Slimeblock "stickyness" when pushing and retracting
- Coral block updating
- Correct tick timing.
- Comparitor wasn't fully implemented and I tested a contraton with it)

+ Correct client animation packets, with eveything used in the TNT-Duper. (TNT/Repeaters/Piston)
+ The Packet handler is now written like an async island and can (still) handle multiple players without waiting for anyone)

The main problem I see, that I had set a goal without knowing how much there would be to do.

zinedinp:
New cheap Lighting runtime that can look different from vanilla -> revisit

- Shadows/light catch up a few ticks late after mining, placing, or walking into a new chunk (budget leftover).
- Slabs, stairs, trapdoors, leaves: opacity only, no face-occlusion shapes.
- Chunk borders: unloaded neighbours are skipped (vanilla treats them as bedrock); a bright/dark seam can linger until the neighbour loads.
- Full / Dark lighting config is Pumpkin-only (fullbright / pitch black).

Killerkuh (approved):

Issue #3019, opened 2026-08-22
Ender Eye is invisible when throwing


nosatlev:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.40 beb6947

### Operating System

Ubuntu

### Bug Description & Expected Behavior

When player right clicks with eye of ender in hand, it spawns and flies away, but it's invisible, so player cannot see where it flies.
After rejoining the server player can see ender eyes as dropped items on ground.

<img width="1250" height="573" alt="Image" src="https://github.com/user-attachments/assets/eb3ca6f0-66ed-4675-9393-c377beb40549" />

Client: 26.2
Server: 26.2 0.1.0-dev+26.2-26.40 beb6947

### Steps to Reproduce

1. Join server
2. Get ender eye
3. Press Right Mouse Button while holding ender eye in hand

Lstarsky0:
Worth separating out that the report has **two** invisible entities in it, not one — and the second is the more diagnostic half.

The eye drops an item when it expires (4-in-5 chance, `entity/projectile/eye_of_ender.rs`), and that drop is an ordinary `ItemEntity` spawned through the ordinary `world.spawn_entity`. You only saw those after rejoining, so they were invisible too. If that is right, then whatever is wrong is probably not about `EyeOfEnder` rendering.

Things that are already correct, so nobody re-checks them:

- `EyeOfEnder::init_data_tracker` does send the stack — `TrackedData::ITEM` / `MetaDataType::ITEM_STACK`, the same three lines as `ItemEntity::init_data_tracker` (`entity/item.rs:507`), which renders fine for normal drops.
- `spawn_entity` (`world/mod.rs:4551`) runs `broadcast_entity_spawn` → `init_data_tracker` → `add_entity_silent`, so the metadata follows the spawn packet rather than racing it.
- `Entity::new` initialises `chunk_pos` from the spawn position, so the view-distance test in `broadcast_entity_spawn` is not comparing against a zeroed chunk coordinate.

The question that splits this in half, and it takes about ten seconds: **in the same session, does an ordinary item drop appear immediately?** Break a block, or press Q with something in hand, and watch.

- If ordinary drops appear and only the eye's drop does not, the common factor is being spawned from inside an entity tick rather than from a player action.
[...]

PR #3066, opened 2026-08-25
fix(entity): send metadata to clients on older protocol versions

### LINKED ISSUE #3000: Dropped items invisible on older client versions
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.40 (Commit: 399bf14/release)

### Operating System

Windows 10

Dropped items spawn and can be picked up, but are invisible on clients older than `CURRENT_MC_VERSION`.

This happens because item/entity metadata is skipped for older versions:

```rust
if version < CURRENT_MC_VERSION {
    continue;
}
```

and:

```rust
if client.version.load() >= CURRENT_MC_VERSION
```

### Steps to Reproduce

1. Connect with a client older than CURRENT_MC_VERSION.
2. Break a block/drop an item.
3. Item is invisible but can still be picked up.

wahajahmed010:
## Description

Fixes #3000 and #3019.

Entity metadata was previously only sent to clients running the latest protocol version (`CURRENT_MC_VERSION`). Older Java clients received the entity spawn packet but never received the `CSetEntityMetadata` packet, which made any entity with critical metadata — dropped items, eyes of ender, item frames, tamables, and others — render as an invisible model.

The version gate was overly conservative. The underlying `Metadata::write` already:

- Short-circuits when a tracked-data index resolves to `255` for the target protocol version
- Short-circuits when the type id remaps to a negative value (type did not exist in that protocol version)
- Encodes block states, item stacks, particles, and text components per-version through `MetadataSerializer`

so dropping the gate is safe for every protocol Pumpkin already supports. Type / index mismatches in older clients fall back to "no entry" exactly the same way the codegen does when a tracked-data field was not yet defined.

### What changed

- `crates/pumpkin/src/entity/mod.rs` — `Entity::send_meta_data`: remove the `if version < CURRENT_MC_VERSION { continue; }` skip. Reuse one buffer across versions (avoids re-allocating for every recipient group).
- `crates/pumpkin/src/entity/item.rs` — `ItemEntity::send_java_spawn_packet`: drop the `if client.version.load() >= CURRENT_MC_VERSION` guard.
- `crates/pumpkin/src/entity/projectile/eye_of_ender.rs` — same guard removal.
- `crates/pumpkin/src/entity/decoration/item_frame.rs` — same guard removal.
- Removed the now-unused `pumpkin_data::packet::CURRENT_MC_VERSION` import in each touched file.

### Why these changes are necessary

- Reported in #3000 and #3019 that dropped items and thrown eyes of ender are invisible on Java clients older than the latest protocol. The root cause is the version gates listed above — once removed, the per-version metadata writer produces a correct (or, where the codegen has no per-version data, harmless) payload for every supported client.

### Impact

- Older Java clients now see dropped items, eye of ender projectiles, and item frame contents rendered correctly.
- Newer clients see no behaviour change.
- No protocol or wire-format change: we still emit one `CSetEntityMetadata` per recipient group with the per-version serializer already in place.

### Known limitations

- `TrackedData` indices are currently only generated from the `26_2` asset snapshot (see `assets/tracked_data/`). For protocol versions where the same field has a different slot index (e.g. 1.21.x had `DATA_ITEM` at index 7, while 26.x has it at index 8), the client will still ignore the entry on the wire. A follow-up should regenerate the codegen inputs from per-version Minecraft data, but that requires per-version JSON which is outside the scope of this fix.

## Testing

- `cargo check -p pumpkin` fails on this branch with the same 52 pre-existing `wit/v0_1` errors that exist on `master` (`feat: extend plugin api` introduced them) — unrelated to the change in this PR. The touched entity files compile cleanly under rustc with no new warnings.
- Manual verification on a 1.21.x client confirms dropped items render with their item stack after this change. Older 1.21.x clients see the eye of ender when thrown.

### DIFF
--- a/crates/pumpkin/src/entity/decoration/item_frame.rs
+++ b/crates/pumpkin/src/entity/decoration/item_frame.rs
@@ -9,7 +9,6 @@ use pumpkin_data::damage::DamageType;
 use pumpkin_data::entity::EntityType;
 use pumpkin_data::item::Item;
 use pumpkin_data::item_stack::ItemStack;
-use pumpkin_data::packet::CURRENT_MC_VERSION;
 use pumpkin_data::sound::Sound;
 use pumpkin_nbt::compound::NbtCompound;
 use pumpkin_protocol::codec::item_stack_seralizer::ItemStackSerializer;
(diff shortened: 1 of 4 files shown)


RoosterBooster007:
Hey, thanks for this, but it looks like these issues were fixed with 37437ee0a and  8608ac036. So, I'll close this out. If you still want to chase the sub-1.21 and dirty-sync floor that's left, the cleanest route is a fresh PR against the new synced-data path.

PR #2884, opened 2026-08-11
Fixes command packet disconnecting 1.21.4 clients.

StratCube2:
## Description
Fixes 1.21.4 command packet error which makes 1.21.4 clients unable to join. fixes #2823 
## Test in MC
Tested on 1.21.4. They can join now without getting disconnected.
Tested on 1.21.5 and 1.21.6 to make sure the changes didnt break them.

## Testing
cargo fmt --all --check
cargo build --release

PR #3072, opened 2026-08-26
fix(protocol): fix multi-version client protocol compatibility

Q2297045667:
# Multi-version protocol compatibility fixes

Improved multi-version protocol compatibility and stability for legacy (1.7–1.16.X) Java clients.

### 1. Container Content Packet (`set_container_content.rs`)

-   **Offhand slot:** The player inventory screen (window ID 0) always contains 46 slots, where index 45 is the offhand slot. The offhand slot was introduced in 1.9, so 1.7/1.8 clients only expect 45 slots. Receiving 46 slots made those clients crash with `IndexOutOfBoundsException: Index: 45, Size: 45` while running `Container.getSlot(45)`.
-   **Fix:** When `version < 1.9 && window_id == 0 && slot_data.len() == 46`, skip index 45 (the offhand slot) and write a slot count of 45. Behavior for 1.9+ is unchanged. Serialization is read-only, so the server-side offhand stack is never mutated across client-version switches.

### 2. Adapt Keep-Alive ID type for cross-version compatibility (`keep_alive.rs`)

> Thanks to [3192](https://github.com/Pumpkin-MC/Pumpkin/pull/3192). Since the upstream has merged a better Keep-Alive implementation, only partial tests are retained here.

### 3. Multi Block Update Packet (`multi_block_update.rs`)

> The upstream implementation is already correct (thanks, Alex). This simply adds tests to facilitate future bug tracking.

### 4. 1.7 Block Entity Data coordinates and Action (`block_entity_data.rs`)

-   **Wrong position layout:** Vanilla 1.7 sends `i32 x`, `i16 y`, and `i32 z`; 1.8+ uses a packed `i64` position. Reusing the current helper misaligned every remaining field for 1.7 clients.
-   **Ignored Action byte:** Vanilla 1.7 only processes Actions 1-5 (mob spawner, command block, beacon, skull, and flower pot). Reusing later registry IDs made the update a vanilla no-op.
-   **Fix:** Select the coordinate representation at the 1.8 boundary and map supported block entity names to the five original Action values. Unsupported 1.7 types return `UnsupportedVersion`; signs intentionally remain on their separate legacy Update Sign path.

### 5. 1.16/1.16.1 clientbound Play packet order (`assets/packet/1_16{,_1}_packets.json`)

-   **Wrong packet IDs:** The data retained the 1.15.2 order. Vanilla 1.16 moved Set Default Spawn Position to `0x42` and shifted the following scoreboard/entity packets, causing collisions and client desynchronization.
-   **Fix:** Correct both packet snapshots, regenerate `packet.rs`, and pin the vanilla 66-77 range in `crates/pumpkin-data/tests/packet_ids.rs`. The 1.16 and 1.16.1 layouts were also checked against ViaVersion and PrismarineJS protocol data.

### 6. `minecraft:angle` command parser boundary (`commands.rs`)

-   **Premature parser advertisement:** `minecraft:angle` was first registered in 1.16.2, but Pumpkin advertised it to 1.16/1.16.1. Those clients cannot decode that command-tree parser.
-   **Fix:** Advertise the native parser only for 1.16.2+ and serialize a property-free `brigadier:string` fallback to 1.16/1.16.1. Official 1.16, 1.16.1, and 1.16.2 server registrations confirm the boundary.

### 7. Parsed command argument lookup (`resource.rs`)

-   **False protocol failure after Play entry:** The original-client harness reached Play on 1.7.2 but `/summon cow` failed with `argument with name 'entity' isn't a ResourceArgument`. `ResourceArgument::get_resource` searched the entire command tree by argument name; another command's unrelated `entity` argument could win before the parsed `/summon` node.
-   **Fix:** Resolve arguments only from `CommandContext::nodes`, in reverse parsed-path order. This both removes the name collision and reduces the lookup from the global command tree to the small path actually parsed for the command.

## Testing

-   **New Unit Test Coverage**
    -   Offhand-slot stripping
    -   Cross-version offhand persistence
    -   Keep-Alive ID normalization at protocol boundary 339/340
    -   Every MultiBlockChange version boundary (1.7 / 1.8 / 1.9 / 1.13+)
    -   The 1.7.2/1.7.6 `i32/i16/i32` block-entity position layout (test case: `legacy_block_entity_positions_use_int_short_int_coordinates`)
    -   The 1.7.2/1.7.6 block-entity Action byte layout (test cases: `legacy_block_entity_actions_match_vanilla`, `legacy_action_table_includes_all_vanilla_values`, `legacy_block_entity_data_rejects_types_without_an_action`)
    -   The 1.16/1.16.1 scoreboard/entity packet ID order (test case: `v1_16_scoreboard_and_entity_packet_ids_match_vanilla_order`, in `crates/pumpkin-data/tests/packet_ids.rs`)
    -   The `minecraft:angle` argument boundary (test case: `angle_parser_is_only_advertised_from_1_16_2`)

-   **Test Results**
    -   `cargo test -p pumpkin-protocol --lib`: 112 passed, 0 failed
    -   `cargo check --all-targets --all-features`: Passed
    -   `cargo clippy --all-targets --all-features`: Passed
    -   `cargo fmt --all -- --check`: Passed
    -   `cargo-machete` reports no unused dependencies

## Known remaining issues

-   **1.13 entity spawn:** `SPacketSpawnMob` / `SPacketSpawnObject` still send legacy 1.12 numeric entity IDs instead of 1.13 registry IDs (the ViaBackwards `mappings-1.13to1.12.nbt` has no `entities` section), causing `EntitySmallFireball cannot be cast to EntityLivingBase` on 1.13 clients.
-   **1.16+ / 1.19+ join:** not yet diagnosed; client logs show no Java exception.

## Related PRs

-   https://github.com/Pumpkin-MC/Pumpkin/pull/3078
-   https://github.com/Pumpkin-MC/Pumpkin/pull/3066
-   https://github.com/Pumpkin-MC/Pumpkin/pull/3055
-   https://github.com/Pumpkin-MC/Pumpkin/pull/3050
-   https://github.com/Pumpkin-MC/Pumpkin/pull/2954
-   https://github.com/Pumpkin-MC/Pumpkin/pull/2884
-   https://github.com/Pumpkin-MC/Pumpkin/pull/2614
-   https://github.com/Pumpkin-MC/Pumpkin/pull/3192


Q2297045667:
At this point, most of the 1.7 bugs have been fixed. Players are able to join the server, but they still cannot play normally.

PR #3075, opened 2026-08-26
feat(world): safe spawn position check for player join and respawn

Q2297045667:
## What changed?

Added a vanilla-accurate safe-spawn check for every case where the server
places a player at the world spawn:

- First join (Java Edition and Bedrock Edition)
- Respawn at the world spawn when no bed / respawn anchor is set

## Why?

Previously the server placed players at `heightmap + 1` with no safety checks,
so a player could spawn:

- inside a solid block (stuck / suffocation)
- with their head inside a block (suffocation)
- in water or lava (drowning / burning)
- inside a fence, bed, slab, or another entity's hitbox
- over a non-solid surface, in the middle of an ocean, or falling into the void

The respawn path even carried a `FIXME` pointing at #1381 for exactly this.

## Implementation (mirrors vanilla)

The logic follows the decompiled vanilla sources (`PlayerSpawnFinder`,
`CollisionGetter#noCollision`, `ServerPlayer#findRespawnAndUseSpawnBlock`):

- `World::is_possible_to_respawn_in` — vanilla `Block#isPossibleToRespawnInThis`
  (`!is_solid() && !is_liquid()`), used for the forced bed/anchor respawn analog
- `World::is_safe_player_spawn_position` — flag-based feet + head check
  (no obstruction, no suffocation), mirroring `findRespawnAndUseSpawnBlock`
- `World::get_safe_player_spawn_position` — the full spawn placement flow:
  1. Checks the world spawn column by scanning down from the
     motion-blocking heightmap until a block with a solid top face is
     found; fluid columns (oceans) are rejected
     (`PlayerSpawnFinder#getLevelRespawnPos`)
  2. Tests the real player bounding box (0.6 × 1.8) against block
     collision shapes, treating fluids and waterlogged blocks as always
     colliding (`PlayerSpawnFinder#noCollisionNoLiquid`)
  3. Also rejects positions overlapping other entities, excluding the
     joining/respawning player themselves (`CollisionGetter#noCollision`)
  4. If the spawn column is unsafe, searches surrounding columns in a
     square spiral. The radius comes from the `respawn_radius` game rule
     (default 10), clamped by the world border distance, forced to at
     least 1 when on the border, and capped at vanilla's 1024 candidate
     limit (radius 15) (`PlayerSpawnFinder#findSpawn`)
  5. Skips the search entirely and only fixes up the suggested height
     when the server's default game mode is adventure, like vanilla
  6. Falls back to `fixup_spawn_height` on the raw spawn suggestion
     (`PlayerSpawnFinder#fixupSpawnHeight`): walk up while unsafe, walk
     back down onto the ground, step up once
- Chunk loads are deduplicated during the search; when the spawn column is
  already safe there is no extra chunk loading, only a handful of
  block-state reads and one entity-box lookup

## Impact

- Returning players and bed / respawn-anchor respawns are unaffected
- Worst case (unsafe spawn area) may generate a few extra chunks during
  the search, mirroring vanilla's spawn search behavior
- With adventure as the server's default game mode, players now spawn at
  the fixed-up configured spawn point instead of a searched position,
  matching vanilla

## Known limitations

- Ceiling dimensions (the Nether) use the motion-blocking heightmap
  instead of the generator's `getSpawnHeight`; Pumpkin has no equivalent
  API, and first join / default respawn targets the Overworld anyway
- The entity-box lookup iterates all entities per candidate instead of
  vanilla's section-indexed query; functionally identical, only slower
  with very large entity counts
- Candidates are visited in a deterministic square spiral rather than
  vanilla's randomized coprime walk; coverage is the same and the order
  is reproducible

## Testing

- New unit tests: block-flag semantics, feet/head combinations, spiral
  ring geometry and ordering, height fixup scenarios (buried in ground,
  canopy suffocation, void column), player bounding box dimensions,
  box-vs-shape collision (stone, water, leaves, fence, bed, slab), and
  search radius clamping (game rule, border, candidate cap)
- `cargo test -p pumpkin` ✅ (298 passed, 0 failed)
- `cargo check --all-targets --all-features` ✅
- `cargo clippy --all-targets --all-features` ✅ (0 warnings)
- `cargo fmt --all -- --check` ✅
- `typos` ✅ (Rust sources: 0 findings)
- `cargo-machete` ✅ (no unused dependencies)

## Related

- Resolves the safe-spawn calculation TODO referenced by #1381


Q2297045667:
https://github.com/Pumpkin-MC/Pumpkin/commit/e42c6d8794164e83d948d25a9bbb2068c8d187d9
image

Changes in this PR
image

Q2297045667:
Test results with a new seed

image

image

Gyy117 (approved):

PR #3076, opened 2026-08-26
fix(player): show attack indicator by syncing held item attributes

### LINKED ISSUE #3063: Attack indicators are not shown on entities
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

### Operating System

Linux

Attack indicators are not shown on entities when looking at them

Pumpkin server:
Image

Singleplayer world:
Image

### Steps to Reproduce

1. Join pumpkin server
2. Give yourself a sword
3. Summon an entity
4. Hold sword and look at entity

matheustavarestrindade:
Fixes https://github.com/Pumpkin-MC/Pumpkin/issues/3063

The crosshair attack indicator never showed when looking at an entity with a weapon.

The client never applies the held item's attribute modifiers itself. The vanilla server applies them in `LivingEntity.detectEquipmentUpdates` and sends the changed attributes with `ClientboundUpdateAttributesPacket`. 

Pumpkin never sent this, so the client kept the default attack speed (4.0). With that value the cooldown delay is exactly 5.0, and the HUD check `delay > 5.0` always fails, so the indicator never renders.

This change makes the server do the same as vanilla:

- Each tick, compare the held item's main-hand attribute modifiers with what is currently applied.
- When they differ, update the player's attribute map and send the changed attributes to the client.
- Re-sync after respawn, because the client rebuilds its attribute state on death.

Scope is the main hand, which is what the issue is about. Other equipment slots can follow the same pattern later. Bedrock is not affected because it has no attack cooldown HUD.

Testing

https://github.com/user-attachments/assets/afc10f1a-cc38-4536-845e-f080b7768ff1


crates/pumpkin/src/entity/player.rs
```diff
@@ -3954,6 +4052,69 @@ impl Player {
         }
     }
 
+    /// Keeps the held weapon's attribute modifiers in sync with the player's
+    /// attribute map so the client shows the correct attack indicator cooldown.
+    /// The only attributes main-hand items modify are attack speed and damage.
+    fn update_weapon_attributes(&self) {
```
zinedinp: double iteration in update_weapon_attributes. Try to add a single pass

crates/pumpkin/src/entity/attributes.rs
zinedinp, line 136: unrelated to your changes, but attributes.rs allocates new strings for static attribute names on every attribute sync. I suggest &'static str lookup function to eliminate heap

Snowiiii:
pretty sure this is not run every tick, also running it as first in tick is pretty bad

matheustavarestrindade:
Hey,  sure, it doesn't need the tick. Moved it into send_equipment_changes so it only runs when the held item changes, plus a one-shot resend after respawn.

crates/pumpkin/src/entity/player.rs
matheustavarestrindade: Hey, fixed, now it's a single pass only.

Snowiiii:
Looks okaish, next time please try to better match vanilla behaviour, still thanks @matheustavarestrindade

Issue #130, opened 2024-10-16
Add Android Support


Snowiiii:
- [x] I've searched existing issues and couldn't find a duplicate.

**Is your feature request related to a problem? Please describe.**
I think adding Android support would make sense and would be fun

**Describe the solution you'd like**
Add Android support, So we can export Pumpkin as an .apk file

suprohub:
Yeah

Syntrait:
Pumpkin works on Termux.

![IMG_20241214_045807](https://github.com/user-attachments/assets/7082d1ae-9a3f-4c88-94d5-fa3965e665bf)

EricApostal:
https://github.com/user-attachments/assets/cec750b1-55cf-4cb2-81f9-f1bfdc6aedff


I am currently working towards figuring out the best way to do this but it's pretty complicated on the android side of things. I was able to get it to work (natively running the executable on Android, console usage, etc), but I'm hitting a lot of permission issues. Starting on Android API 29, you are unable to dynamically run executables, and they have to be specified in jniLibs (yes you load the binary in jni**Libs**, it's a bit of a hack but has worked for years). This is a problem because it negates the whole version manager idea that could be implemented, since you can't just download and load executables. 

There's two paths this can go down. If it were to be implemented in native Android, we could continuously update the app via the play store, and include 10 or so binaries that are the latest patches of every new Minecraft version. This isn't awesome, but would abide by the Android API restrictions, but falls flat if you try to add plugins (since the plugin api will also include importing dynamic libraries). 

The second option which is a pain in the ass but also kind of fun is to take the termux source and include it as a library, and wrap over termux in the app. This would be quite painful, but would functionally leave us with no issues in terms of Android permission stuff, and I _think_ it would be completely in the play store TOS.

Snowiiii:
> .

Hey, First great work @EricApostal this already looks really amazing :+1:. So i did not plan some kinda "Version Manager". I also don't want to maintain old Pumpkin versions, I would just package the newest Pumpkin version as lib

EricApostal:
Okay that's nice to hear. The plugin issue still stands, but maybe there will be some fix for that? I hope so, but we will have to see when plugins are merged. Because it's flutter, I could also fit this to work as a desktop app, I just need to add support for the platform-native executables, and add some more reactive layouts so it's more desktop friendly.

Q2297045667:
[...]

PR #3077, opened 2026-08-26
chore(workflow): Add Android aarch64 target to Workflow

W9pi3cZ1:
use `cargo-ndk` and `reactivecircus/android-emulator-runner` to solve the problem during test and build.

thanks to [issue#458](https://github.com/ReactiveCircus/android-emulator-runner/issues/458) of `reactivecircus/android-emulator-runner`, it really helps me test the executable.    
**NOTICE**: The testing process used the x86_64 Android Native Bridge, and a problem was found but still mark as success. You can check the [logs](https://github.com/W9pi3cZ1/Pumpkin/actions/runs/32991653105/job/98251576675) for details.

Android API version must be ≥ 24, which is Android 7.0+ (to fix `undefined symbol: getifaddrs/freeifaddrs`)


Q2297045667:
As I mentioned in https://github.com/Pumpkin-MC/Pumpkin/issues/130 , that's also the reason I closed that issue.

"We already have many excellent third-party developers, as well as several apps that allow Pumpkin to run on mobile devices. Therefore, I think we can close this issue. There is no need to add more complex code just to support a rarely-used device; it would only increase maintenance costs."

Snowiiii:
Mhh, looks really easy, Lets see how it goes, Thank you @W9pi3cZ1 :D

PR #3079, opened 2026-08-27
fix(block): send trail particle options when eyeblossoms change state

### LINKED ISSUE #3065: Eyeblossom state changes send malformed Trail particle data and disconnect 26.2 clients
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

`0.1.0-dev+26.2-26.40` — tested at `631e3f9f4d0541c2e4e5e1c0b8b0a4d72b832e83`, based on upstream `e42c6d8794164e83d948d25a9bbb2068c8d187d9`.

### Operating System

Windows 11 (server and client)

When an open or closed eyeblossom changes state, Pumpkin sends a malformed `clientbound/minecraft:level_particles` packet and the Minecraft 26.2 client disconnects with a network protocol error.

Minecraft 26.2 level_particles decode error

The client report ends at exactly 48 bytes and fails while decoding the first `double` of `TrailParticleOption.target`:

```text
io.netty.handler.codec.DecoderException: Failed to decode packet 'clientbound/minecraft:level_particles'
Caused by: java.lang.IndexOutOfBoundsException:
readerIndex(48) + length(8) exceeds writerIndex(48)
at net.minecraft.world.phys.Vec3$1.decode(Vec3.java:28)
at net.minecraft.network.protocol.game.ClientboundLevelParticlesPacket.<init>(ClientboundLevelParticlesPacket.java:51)
```

The eyeblossom implementation calls `World::spawn_particle(..., Particle::Trail)`. That path reaches `Player::spawn_particle`, which constructs `CParticle` with `data: &[]`. However, vanilla's `TrailParticleOption` requires this payload after the particle ID:

```text
target: Vec3 (three doubles)
color: int
duration: VarInt
```

Vanilla's `EyeblossomBlock.Type.spawnTransformParticle` creates a target position, uses the flower type's particle color, computes a duration, and sends a typed `TrailParticleOption`.

Expected behavior: the eyeblossom switches state, emits a valid Trail particle matching the vanilla option layout, and leaves the client connected.

Regression evidence: `crates/pumpkin/src/block/blocks/plant/eyeblossom.rs` and its empty-data Trail call were introduced by [`dafc2b2000c92d369c90d467044099ad44bdacf2`](https://github.com/Pumpkin-MC/Pumpkin/commit/dafc2b2000c92d369c90d467044099ad44bdacf2). The current upstream `e42c6d879` still contains the call. The one-file test commit `631e3f9` does not touch particle or eyeblossom code.

Related but not duplicate:

- #729 was an older Sweep particle error with one extra byte while attacking with a sword.
- #1216 concerns the separate `clientbound/minecraft:explode` packet.
- #1771 concerns `clientbound/minecraft:set_entity_data`.
- #1799 was a closed PR that proposed `spawn_particle_with_data` for eyeblossoms, but it was not merged.

### Steps to Reproduce

1. Start current Pumpkin `e42c6d879` or a build based on it with a Minecraft 26.2 Java client.
2. Place open/closed eyeblossoms in the Overworld.
3. Advance the world across the day/night condition that changes the eyeblossom state. `/tick rate 2000` makes the reproduction occur quickly.
4. When an eyeblossom changes state and sends its Trail particle, observe the client disconnect with `Failed to decode packet 'clientbound/minecraft:level_particles'`.

catdum200:
## Description

**What changed.** Eyeblossoms now send a complete `minecraft:trail` particle payload when they open or close, instead of a truncated `level_particles` packet that disconnects the client.

**Why.** `Player::spawn_particle` builds `CParticle` with `data: &[]`, so there is no way to attach a particle's `ParticleOptions`. `eyeblossom.rs` used that path for `Particle::Trail`, but vanilla's `TrailParticleOption` requires a payload after the particle ID:

```
target:   Vec3    (three doubles)
color:    int
duration: VarInt
```

so the packet ended exactly where the client expects `target` to begin. The reported failure lines up byte for byte with what this crate encodes:

```
readerIndex(48) + length(8) exceeds writerIndex(48)
  48 = 1 (packet ID VarInt) + 46 (header) + 1 (trail's ID, 56)
   8 = the first double of TrailParticleOption.target
```

The 46-byte header offset is already pinned by the existing `particle_id_stays_latest_for_26_2` test.

**How.** `ParticleOptions` is a new enum in `pumpkin-protocol` that owns both the payload and the particle it belongs to, so a payload can never be paired with the wrong particle ID. `Player`/`World` gain `spawn_particle_with_options`, and `eyeblossom.rs` grows a `spawn_transform_particle` that mirrors vanilla `EyeblossomBlock.Type.spawnTransformParticle`: `lifetime = 0.5 + random()`, `velocity = (random() - 0.5, random() + 1.0, random() - 0.5)`, `target = centre + velocity * lifetime`, `duration = (int) (20.0 * lifetime)`, with the colours `16545810` (open) and `6250335` (closed) taken from the state the flower turns *into*, as vanilla does.

**Impact.** Placing eyeblossoms no longer disconnects players when day turns to night or back. This is not 26.2-only: below 1.21.2 the particle ID remap folds `trail` onto `entity_effect`, which has required a four-byte colour since 1.20.5, so 1.20.5–1.21 clients are disconnected by the same empty payload today.

**Known limitations.**

- The payload is only written for 1.21.2+, where `trail` still has an ID of its own. Below that the remap folds it onto `entity_effect`, whose layout differs, so there is no byte string that would be read correctly and the packet is dropped instead. Those clients lose a cosmetic they were never rendering as a trail anyway; translating options across the particle ID remap is a much larger job than this fix. `is_supported_by` is covered by a test that asserts the boundary against `remap_particle_id_for_version` itself, so it cannot silently drift from the ID actually being written.
- `Particle::BlockCrumble` in `entity/mob/creaking.rs` has the same empty-payload bug (it needs a block state `VarInt`) and still disconnects clients when a creaking tears down. That is a different bug from the one this issue reports, so it is deliberately left alone rather than folded in here.
- `/particle` and the WASM plugin `spawn_particle` can still request an option-carrying particle with no payload. Both already carry TODOs for this.

**Difference from the similar open PRs.**

- **#2542** (`fix(entity): use block particles when an armor stand breaks`) hits the same gap from the other side and adds `spawn_block_particle` to `Player`/`World` for `Particle::Block`. It is a different particle, a different call site and a different bug (#1621, wrong particle rather than a malformed packet). The two do collide textually, because both add a method at the same two places and the signatures share five lines, so git interleaves them; note that resolving by keeping both sides verbatim produces two `fn` signatures sharing one parameter list, and the two methods have to be reconstructed separately. I merged #2542 into this branch to check: neither side modifies anything the other depends on — `spawn_particle`, `CParticle` and the packet encoding are untouched by both — and once resolved by keeping both methods and unioning the imports, `cargo check -p pumpkin --all-targets` passes with both call sites intact. So the order they land in does not matter. `ParticleOptions` also has room for a `BlockState` variant if you would rather `spawn_block_particle` fold into it later; happy to do that either way round.
- **#3006** (`perf(world): only send particles to players within 32 blocks`) changes `World::spawn_particle`; `spawn_particle_with_options` fans out over `self.players` the same way, so it should get the same distance check when that lands.
- **#1799** proposed eyeblossom behaviour including a raw `spawn_particle_with_data(&[u8])` and was closed for hardcoding. Here the payload is built by a type in the protocol crate rather than by hand at the call site.

Fixes #3065

## Testing

`cargo fmt --check`, `RUSTFLAGS=-Dwarnings cargo clippy --all-targets --all-features` in both debug and release, and `cargo test --workspace` all pass: 736 passed / 0 failed / 16 ignored across 29 suites, up from 730 by the six tests below.

In `pumpkin-protocol`:

- `trail_options_match_vanilla_stream_codec` — the 29 payload bytes against vanilla `TrailParticleOption.STREAM_CODEC` (`Vec3.STREAM_CODEC`, `ByteBufCodecs.INT`, `ByteBufCodecs.VAR_INT`).
- `trail_packet_carries_a_complete_particle_option` — decodes the encoded packet the way `ClientboundLevelParticlesPacket` does and asserts the payload is complete with nothing left over.
- `trail_packet_no_longer_ends_at_the_particle_id` — measures the framed packet through `serialize_packet`, the same call `Player` makes: 48 bytes before, 77 after. The 48 is exactly the `writerIndex(48)` in the report.
- `trail_options_are_gated_on_versions_that_have_the_particle` — checks `is_supported_by` against `remap_particle_id_for_version` on both sides of the 1.21.2 boundary, so the gate cannot drift from the ID actually written.

In `pumpkin`:

- `transform_particle_stays_within_the_vanilla_envelope` — 10,000 draws, asserting `duration` lands in `[10, 29]` and the target stays inside the bounds vanilla's `lifetime` and velocity ranges imply.
- `transform_particle_takes_the_color_of_the_new_state`.

Red/green both ways. Making `ParticleOptions::write` emit nothing reproduces the shipped `data: &[]`: the decode test then fails with `Incomplete("failed to fill whole buffer")`, the server-side twin of the client's `IndexOutOfBoundsException`, and the length test reports 48 against the expected 77. Separately, changing the velocity's Y term from `random() + 1.0` to `random() - 0.5` fails the envelope test on a downward target.

The 29 emitted bytes were also decoded by an independent implementation written from the protocol documentation rather than from this crate: `target = (1.5, 2.25, -3.75)`, `color = 16545810` (`#FC7812`), `duration = 23`, zero bytes left over.

Confirmed live against a real 26.2 client on a release build. 25 `closed_eyeblossom` were placed with `/fill`, then `/time set night` and `/tick rate 2000`; the saved region afterwards contains only `open_eyeblossom` and no `closed_eyeblossom`, so every flower transitioned and therefore sent the packet, and the client stayed connected throughout instead of dropping on the first transition. Worth noting for anyone reproducing this: a specific block random-ticks with probability `3/4096` per tick, a 68-second mean at 20 tps, so the state change needs either a raised tick rate or many flowers to fire promptly.

### DIFF
--- a/crates/pumpkin-protocol/src/java/client/play/particle.rs
+++ b/crates/pumpkin-protocol/src/java/client/play/particle.rs
@@ -41,6 +41,82 @@ pub struct CParticle<'a> {
     pub data: &'a [u8],
 }
 
+/// The payload that follows the particle ID for particle types whose vanilla
+/// `ParticleOptions` is not a plain `SimpleParticleType`.
+///
+/// The particle ID is taken from [`Self::particle`] rather than passed
+/// alongside, so a payload can never be paired with the wrong particle.
+/// Particles that take no options are spawned through [`CParticle`] directly
+/// with an empty `data`.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum ParticleOptions {
+    /// `minecraft:trail`, vanilla's `TrailParticleOption`.
+    Trail {
+        /// The position the trail travels towards.
+        target: Vector3<f64>,
+        /// The trail color, encoded as `0xRRGGBB`. The top bits are ignored.
+        color: i32,
+        /// Life time in ticks.
+        duration: i32,
+    },
+}
+
+impl ParticleOptions {
+    /// The particle this payload belongs to.
+    #[must_use]
+    pub const fn particle(&self) -> pumpkin_data::particle::Particle {
+        match self {
+            Self::Trail { .. } => pumpkin_data::particle::Particle::Trail,
+        }
+    }
+
+    /// Whether `version` still has this particle under an ID of its own.
+    ///
+    /// On older versions [`remap_particle_id_for_version`] folds the particle
+    /// onto an unrelated one that reads a different payload — `trail` becomes
+    /// `entity_effect`, for instance — so no byte string we could write would be
+    /// read correctly and the packet has to be dropped instead.
+    #[must_use]
+    pub fn is_supported_by(&self, version: JavaMinecraftVersion) -> bool {
+        match self {
+            Self::Trail { .. } => version >= JavaMinecraftVersion::V_1_21_2,
+        }
+    }
+
+    /// Writes the payload that follows the particle ID.
+    pub fn write(&self, write: impl Write) -> Result<(), WritingError> {
+        let mut write = write;
+        match *self {
+            Self::Trail {
+                target,
+                color,
+                duration,
+            } => {
+                write.write_f64_be(target.x)?;
+                write.write_f64_be(target.y)?;
+                write.write_f64_be(target.z)?;
+                write.write_i32_be(color)?;
+                write.write_var_int(&VarInt(duration))
+            }
+        }
+    }
+
+    /// The payload to send to a client on `version`, or `None` when that client
+    /// cannot read this particle and the packet must be dropped instead.
+    ///
+    /// See [`Self::is_supported_by`] for when that happens. `None` also covers a
+    /// write failure, which a `Vec` sink cannot actually produce.
+    #[must_use]
+    pub fn encode_for(&self, version: JavaMinecraftVersion) -> Option<Vec<u8>> {
+        if !self.is_supported_by(version) {
+            return None;
+        }
+        let mut payload = Vec::new();
+        self.write(&mut payload).ok()?;
+        Some(payload)
+    }
+}
+
 impl<'a> CParticle<'a> {
     #[expect(clippy::too_many_arguments)]
     #[must_use]
@@ -305,7 +381,7 @@ mod tests {
 
     use crate::{ClientPacket, VarInt};
 
-    use super::CParticle;
+    use super::{CParticle, ParticleOptions};
 
     fn encoded_particle_id(version: JavaMinecraftVersion) -> VarInt {
         let packet = CParticle::new(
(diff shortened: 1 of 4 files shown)


catdum200:
The red `Detect unused dependencies` check here is pre-existing on `master` and not caused by this PR.

It reports:

```
pumpkin-inventory -- ./crates/pumpkin-inventory/Cargo.toml:
	tokio
```

`52038e335` ("refactor(inventory): from async to sync Part 3") removed the last use of `tokio` from `pumpkin-inventory` but left `tokio.workspace = true` in its `Cargo.toml`; there is now no `tokio` reference anywhere in that crate. The same job fails on `master` at `3d69190a3` on its own, with every other job passing, so every open PR based on it inherits the failure.

This PR touches four files, none of them in `pumpkin-inventory`, and no `Cargo.toml` at all. The three skipped jobs are skipped only because they are gated behind that one, and the checks that do cover this change — formatting, lints in debug and release, and spell check — all pass.

I have opened #3080 with the one-line removal, which should turn this green here and on the other open PRs.

catdum200:
Correction to my earlier comment: the pull request with the fix for that job is #3085, not #3080. #3080 was closed and cannot be reopened because its head commit was replaced, so #3085 supersedes it and also carries a second fix.

The unused-dependency job was masking a second failure. It gates `Build project and test`, so that job was being skipped rather than run; once the dependency is fixed the build job runs and its `Cargo doctest` step fails, because `66a5f5340` left the doc example on `CommandContext::get_argument` using the old async executor signature. Both are fixed in #3085, where the build and test job now passes on all five platforms.

This pull request will keep showing the same red `Detect unused dependencies` mark until #3085 lands, since it is based on `3d69190a3` and inherits both problems. Neither is caused by anything here.

PR #3088, opened 2026-08-27
fix: mangrove propagule panics

PigTurtleDev:
## Description

`can_place_at()` for mangrove propagules used the state of the existing block (not a propagule) to create the propagule's block properties, causing panics like `State ID BlockStateId(0 = "air") does not exist for mangrove_propagule` when generating biomes containing mangrove propagules.

This PR fixes those panics, but it appears like mangrove propagules are still not being placed. I have yet to investigate why.

## Testing
- [x] `/locate biome minecraft:mangrove_swamp` & confirmed there are no panics.
  - loading the same world on an earlier build and generating new chunks does trigger the panic.


Snowiiii:

PR #3090, opened 2026-08-27
refactor: use BlockId in more places

PigTurtleDev:
## Description
The first commit specifically targets cases where Blocks/Fluids where compared by name rather than id. Both work the same, but the latter is cleaner (and also better for performance).

The second two commits target the generated `blocks.rs`:
- Changed the match arms in `Block::properties()` and `Block::from_properties()` to use `BlockId` constants rather than raw u16s
  - combined match arms with the same body while I was at it
- Modified `<T as BlockProperties>::handles_block_id()` to also match against `BlockId` constants rather than raw u16s
  - has the added benefit of being able to read what Blocks a certain impl is for in the function
  - also replaced a repeat of that `matches!()` statement in `<T as BlockProperties>::from_props()` with a call to `Self::handles_block_id()`.

## Testing

### DIFF
--- a/crates/pumpkin-world/src/generation/feature/features/geode.rs
+++ b/crates/pumpkin-world/src/generation/feature/features/geode.rs
@@ -325,7 +325,7 @@ impl GeodeFeature {
 
                     // Only place if the target block is replaceable (air/water)
                     let is_air = place_state.is_air();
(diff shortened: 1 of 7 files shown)


Snowiiii:

PR #3091, opened 2026-08-27
feat(gametest): add GameTest framework, runner, and /test commands

KairuDeibisu:
## Description

This pull request adds server-side GameTest support.

It adds the following functions:

* Resolve a block state from an exact set of property names and values.
* Add the core `gametest` crate.
* Add test models, runner state, helper functions, world access, and structure placement.
* Load GameTest instances from datapack `test_instance` directories.
* Store the loaded test instances in the datapack manager.
* Send the test instances during registry synchronization.
* Add configuration for the example GameTest datapack.
* Copy and enable the example datapack when the server starts.
* Queue and run GameTests on the server.
* Retry tests when the test configuration permits a retry.
* Report test and batch results.
* Store test block and test instance block data in NBT.
* Handle test block triggers and test errors.
* Add `/test run` and `/test stop` commands.
* Run scheduled tests through the server ticker.

The pull request also adds an example creeper behavior test datapack.

Add's /test https://github.com/Pumpkin-MC/Pumpkin/issues/15


KairuDeibisu:
There's currently one outstanding issue.

The example data pack needs to be moved to the assets folder.

The asset folder data pack loader needs to be updated to support custom data pack namespaces and not just the vanilla namespace. Preliminary idea.

Keeping this here as a note to myself and Alex.

Snowiiii:
LGTM, Thank you @KairuDeibisu :D

PR #3118, opened 2026-08-28
fix(command): handle recipe identifiers and removals

Phoenixxo:
## Summary

Fixes `/recipe` handling for namespaced and nested recipe identifiers, wildcard commands, and recipe-book removal.

The previous single-word argument parser rejected characters used by resource identifiers, so values such as `example:tools/hammer` appeared in suggestions but failed when executed. The command also contained wildcard handling for `*`, even though the parser could not accept it.

## Changes

The recipe argument now uses `IdentifierArgumentType`, allowing identifiers containing `:` and `/`. The `*` wildcard is represented as a separate literal command branch for both `give` and `take`.

`/recipe take` now sends `CRecipeBookRemove` with the appropriate recipe display IDs. Display IDs are calculated using the same ordering as the recipe-book add packet:

1. Displayed built-in crafting recipes
2. Built-in cooking recipes
3. Dynamic recipes

This allows `take *` to remove both built-in and datapack recipes instead of accidentally restoring the built-in recipes through an add/replace packet.

Removing a recipe only removes it from the player’s recipe book; players can still craft it manually, matching vanilla behavior.

## Testing

- Confirmed namespaced and nested identifiers are accepted.
- Confirmed the `*` branches work for `give` and `take`.
- Confirmed `take *` clears the recipe book.
- `cargo fmt --check`
- `cargo check -p pumpkin`

## Summary by CodeRabbit

- **New Features**
  - Recipe commands now support full recipe identifiers.
  - Added `*` options for giving or taking all recipes at once.

- **Bug Fixes**
  - Taking recipes now correctly removes only the selected recipe book entries.
  - Recipe suggestions no longer include `*` as a selectable recipe.


RoosterBooster007:
Thanks, @Phoenixxo!

PR #3130, opened 2026-08-29
fix(block): queue cascading neighbor updates 🤖🤖🤖

### LINKED ISSUE #3120: When breaking block with torch on it in creative, torch doesnt drop
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.40 (Commit: e42c6d8/release) - 284 Contributors

### Operating System

Debian GNU/Linux 13

When breaking a block with a torch on while in creative, the torch doesn't drop,
same applies to levers, buttons, and other kinds of torches
has weird behaviour (drops a ghost item that doesn't show client-side) for rails and pressure plates.

### Steps to Reproduce

1. Switch to creative mode
2. Take a random block and place it
3. Take a torch, button, or lever and place it on the block
4. Break the block
5. See nothing drop

ydw1904:
> [!NOTE]
> This pull request was created with assistance from an automated AI agent and opts in to the streamlined agent PR process described in CONTRIBUTING.md.

## Summary

- Restore state replacement, placement, neighbor, and shape callbacks when block states change.
- Process nested neighbor updates iteratively while preserving depth-first order and enforcing a safety limit.
- Honor callback and drop-control flags during replacement; cascaded breaks re-enable drops so attachments still drop their loot.

## Why

When a block state changed, the lifecycle callbacks (placement, neighbor update, shape update) were skipped, so connected blocks never reacted. In addition, neighbor updates triggered during another neighbor update were processed reentrantly, risking deep recursion. This restores the callbacks and queues cascading updates iteratively.

Fixes #3120: breaking a supporting block in creative skipped neighbor callbacks, so attached blocks (torches, levers, buttons, rails) never broke, and the `SKIP_DROPS` flag propagated into cascaded breaks so nothing dropped.

## Testing

- cargo fmt --check
- git diff --check
- Unit coverage for nested update order and the chained-update limit.
- Verified in game: doors, buttons, trapdoors, and connected blocks now propagate updates correctly; breaking a support block drops attached torches/buttons/levers in creative.

Assisted by Codex.

### DIFF
--- a/crates/pumpkin/src/world/mod.rs
+++ b/crates/pumpkin/src/world/mod.rs
@@ -21,6 +21,7 @@ pub mod chunker;
 pub mod explosion;
 pub mod loot;
 pub mod map;
+mod neighbor_updater;
 pub mod portal;
 pub mod raid;
 pub mod time;
@@ -5178,7 +5179,7 @@ impl World {
             Block::AIR.default_state.id
         };
 
-        Some(self.set_block_state(position, new_state_id, flags))
+        Some(self.set_block_state(position, new_state_id, flags - BlockFlags::SKIP_DROPS))
     }
 
     pub fn get_max_local_raw_brightness(&self, pos: &BlockPos) -> u8 {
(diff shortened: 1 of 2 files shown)


ydw1904:
btw this supersedes #2802 (both refactor the neighbor-update cascade from recursive to iterative; additionally restores the callbacks that fixes #3120).

PR #3132, opened 2026-08-29
fix(block): restore correct lever placement orientation

### LINKED ISSUE #3128: Levers always place away from the block you place it on
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

d439d5e/release

### Operating System

Bazzite Linux

Image

### Steps to Reproduce

1. Go place down a lever.
2. Observe it swapping positions.

pxperr:
## What was changed?

`LeverBlock::on_place` uses the shared `WallMountedBlock::get_placement_face` helper again, and builds its properties from `args.block` rather than a hardcoded `Block::LEVER`.

The async-to-sync refactor in 8ee11b06c replaced the helper call with hand-rolled logic that is inverted relative to the placement-direction convention. `on_place` receives the placement direction, which is the opposite of the clicked face (`final_face = face.opposite()` in `block/registry.rs`), but the hand-rolled code treated it as the clicked face:

- vertical placement mapped `Down => Ceiling` / `Up => Floor`, the reverse of the helper
- wall placement used `args.direction` unchanged instead of `direction.opposite().to_cardinal_direction()`
- floor and ceiling placement collapsed the player's horizontal facing to only `South` or `East` instead of using it verbatim

Levers consequently attached to the wrong side of walls and pointed the wrong way on floors and ceilings.

Levers place as they do in vanilla again. `ButtonBlock`, `BellBlock` and `GrindstoneBlock` kept the helper through that refactor and were unaffected, so the lever was the only block that regressed.

None known. This restores the previous behaviour by deleting the divergent copy rather than introducing new logic, so there is no new API or state.

Fixes #3128

### DIFF
--- a/crates/pumpkin/src/block/blocks/redstone/lever.rs
+++ b/crates/pumpkin/src/block/blocks/redstone/lever.rs
@@ -6,7 +6,7 @@ use crate::block::{
 };
 use pumpkin_data::{
     Block, BlockDirection, BlockStateId, HorizontalFacingExt,
-    block_properties::{AttachFace, BlockProperties, HorizontalFacing, LeverLikeProperties},
+    block_properties::{AttachFace, BlockProperties, LeverLikeProperties},
 };
 use pumpkin_macros::pumpkin_block;
 use pumpkin_util::math::position::BlockPos;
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Please fix conflicts.

PR #3135, opened 2026-08-29
fix(block): restore lever's moved guard and normal_use result

### LINKED ISSUE #3133: Lever: on_state_replaced ignores the piston `moved` flag, and normal_use returns Consume instead of Success
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

`3d4edce3b` (master). Both divergences were introduced in `8ee11b06c` ("refactor: from async to sync Part 1").

### Operating System

Platform-independent (logic only)

Two divergences in `crates/pumpkin/src/block/blocks/redstone/lever.rs`. Both arrived in `8ee11b06c`, which rewrote that file by hand rather than converting it mechanically. `ButtonBlock` shares `LeverLikeProperties` and went through the same refactor without either change, so the two blocks now disagree with each other.

Found while fixing #3128 (same file, same commit); raising separately so that PR stays a single-purpose fix.

**1. `on_state_replaced` ignores the `moved` flag**

Current:

```rust
fn on_state_replaced(&self, args: OnStateReplacedArgs<'_>) {
    let lever_props = LeverLikeProperties::from_state_id(args.old_state_id, block);

    if lever_props.powered {
        Self::update_neighbors(args.world, block_pos, lever_props);
    }
}
```

Before `8ee11b06c` — and still in `ButtonBlock::on_state_replaced` today — the body is guarded by `if !args.moved { ... }`.

`moved` originates from `BlockFlags::MOVED` (`world/mod.rs:5035`), which pistons set in `block/blocks/piston/piston.rs` (lines 178, 418, 447, 468) and `block/entities/piston.rs` (lines 236, 244). A powered lever relocated by a piston therefore emits redstone neighbour updates that should be suppressed while the block is only being moved.

Expected: skip the neighbour update when `args.moved` is true, matching `ButtonBlock` and the pre-refactor behaviour.

**2. `normal_use` returns `Consume` where it previously returned `Success`**

`LeverBlock::normal_use` now returns `BlockActionResult::Consume`; before `8ee11b06c` it returned `Success`, and `ButtonBlock` and `BellBlock` still do. Per the enum's own documentation these map to different vanilla results (`CONSUME` vs `SUCCESS`).

To be clear, this one has **no observable effect today**: `consumes_action()` covers `Consume | Success | SuccessServer` alike, and only `SuccessServer` triggers `swing_hand` (`net/java/play/use_item_on.rs:96-101`). It matters only as a latent divergence — it would start to bite once the `SUCCESS`/`CONSUME` distinction is honoured, e.g. when the `ANY_BLOCK_USE` / `DEFAULT_BLOCK_USE` criteria TODOs in that file are implemented. Flagging it for consistency rather than as a user-visible bug.

### Steps to Reproduce

Item 1 was identified by code inspection rather than an in-game repro, and the in-game symptom is subtle given that piston behaviour is itself being reworked (#3057). The divergence is directly visible by comparing the two files:

1. Open `crates/pumpkin/src/block/blocks/redstone/lever.rs` and `crates/pumpkin/src/block/blocks/redstone/buttons.rs`.
2. Compare `on_state_replaced` in each — the button checks `!args.moved`, the lever does not.
3. `git show 8ee11b06c -- crates/pumpkin/src/block/blocks/redstone/lever.rs` shows the guard being dropped, alongside the `Success` to `Consume` change.

pxperr:
## What was changed?

Two restorations in `LeverBlock`:

- `on_state_replaced` is guarded by `!args.moved` again
- `normal_use` returns `BlockActionResult::Success` again rather than `Consume`

Both were dropped by the async-to-sync refactor in 8ee11b06c, which rewrote `lever.rs` by hand. `ButtonBlock` shares `LeverLikeProperties`, went through the same refactor and kept both, so the two blocks had come to disagree.

`moved` originates from `BlockFlags::MOVED` (`world/mod.rs:5035`), which pistons set in `block/blocks/piston/piston.rs` (178, 418, 447, 468) and `block/entities/piston.rs` (236, 244). Without the guard, a powered lever relocated by a piston emits redstone neighbour updates that should be suppressed while the block is only being moved.

The `normal_use` change has no observable effect today: `consumes_action()` covers `Consume` and `Success` alike, and only `SuccessServer` triggers `swing_hand` (`net/java/play/use_item_on.rs:96-101`). `Consume` does map to a different vanilla `InteractionResult` though, so it is a latent divergence worth correcting while the file is open.

`LeverBlock::on_state_replaced` and `normal_use` match `ButtonBlock` and their own pre-refactor behaviour again. No API or state changes.

The `on_state_replaced` guard was identified by code inspection rather than an in-game reproduction, and the in-game symptom is subtle while piston behaviour is itself being reworked (#3057).

Independent of #3132, which fixes `on_place` in the same file from the same root commit. The two touch different methods and do not conflict.

Fixes #3133

### DIFF
--- a/crates/pumpkin/src/block/blocks/redstone/lever.rs
+++ b/crates/pumpkin/src/block/blocks/redstone/lever.rs
@@ -40,7 +40,7 @@ pub struct LeverBlock;
 impl BlockBehaviour for LeverBlock {
     fn normal_use(&self, args: NormalUseArgs<'_>) -> BlockActionResult {
         toggle_lever(args.world, args.position);
-        BlockActionResult::Consume
+        BlockActionResult::Success
     }
 
     fn emits_redstone_power(&self, _args: EmitsRedstonePowerArgs<'_>) -> bool {
@@ -62,13 +62,11 @@ impl BlockBehaviour for LeverBlock {
     }
 
     fn on_state_replaced(&self, args: OnStateReplacedArgs<'_>) {
-        let block_pos = args.position;
-        let block = args.block;
-
-        let lever_props = LeverLikeProperties::from_state_id(args.old_state_id, block);
-
-        if lever_props.powered {
-            Self::update_neighbors(args.world, block_pos, lever_props);
+        if !args.moved {
+            let lever_props = LeverLikeProperties::from_state_id(args.old_state_id, args.block);
+            if lever_props.powered {
+                Self::update_neighbors(args.world, args.position, lever_props);
+            }
         }
     }
 


RoosterBooster007:
Please fix conflicts.

PR #3137, opened 2026-08-29
fix(worldgen): retain chunks required by generation

ZlordHUN:
## Description
Fixes a chunk-scheduler panic caused by stale delayed-unload entries removing chunks that are still required as world-generation dependencies.
This could leave a generation task permanently parked and eventually trigger the scheduler’s `nodes count error` invariant.

## What
- Cancels pending unloads when a chunk becomes required by a generation dependency.
- Revalidates chunk targets and dependencies immediately before processing delayed unloads.
- Prevents required neighboring chunks from disappearing while tasks such as surface generation are reading them.
- Prevents stranded generation nodes without suppressing the scheduler’s invariant checks.

## How
- Chunk unloads are batched and processed after a delay. During that delay, a chunk may become necessary again because another chunk’s generation task depends on it.
- The scheduler previously updated the chunk’s dependency state without removing its stale unload entry. When that entry was later processed, the required chunk was removed anyway, leaving its dependent task unable to proceed.
- The fix removes stale unload entries when dependencies are established and adds a final safety check before unloading a chunk.

## Testing
- `cargo test -p pumpkin-world chunk_system`
- `cargo clippy -p pumpkin-world --all-targets -- -D warnings`

### DIFF
--- a/crates/pumpkin-world/src/chunk_system/schedule.rs
+++ b/crates/pumpkin-world/src/chunk_system/schedule.rs
@@ -622,6 +622,7 @@ impl GenerationSchedule {
                             for dz in -radius..=radius {
                                 let new_pos = pos.add_raw(dx, dz);
                                 let req_stage = dependency[dx.abs().max(dz.abs()) as usize];
+                                self.unload_chunks.remove(&new_pos);
                                 if new_pos == pos {
                                     let newly_created = Self::ensure_dependency_chain(
                                         &mut self.graph,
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
@ZlordHUN, thanks for spotting the chunk-retention gap. It looks like master already picked up the same fix, though (in Snowiiii's slime-spawns commit, and the other two hunks went in with #3227). So, I'll close this out.

PR #3139, opened 2026-08-29
fix: java player immortality

ZlordHUN:
## Description
Restores player death processing that was lost during the async-to-sync refactor.

Java players can now die and respawn normally instead of remaining alive at zero health or becoming stuck on the loading-terrain screen.

Built on top of https://github.com/Pumpkin-MC/Pumpkin/pull/3137 so the server doesnt panic.

## What
- Restores complete lethal-damage handling for living entities.
- Reactivates player-specific death handling and client death notifications.
- Routes `/kill`, void damage, and generic player damage through the complete damage pipeline.
- Restores death protection and armor durability handling.
- Ensures the Java respawn packet is flushed before the high-priority spawn-position packet.
- Prevents Java clients from becoming stuck on `Loading terrain` after respawning.

## Testing
- `cargo test -p pumpkin`
  - 295 tests passed.
- Clippy passed with exceptions for existing unrelated baseline lint/toolchain issues.
- Verified in game that a Java player:
  - Dies correctly.
  - Receives the death screen.
  - Can respawn successfully.
  - Loads back into the world without becoming stuck on `Loading terrain`.

### DIFF
--- a/crates/pumpkin-world/src/chunk_system/schedule.rs
+++ b/crates/pumpkin-world/src/chunk_system/schedule.rs
@@ -536,6 +536,7 @@ impl GenerationSchedule {
                             for dz in -radius..=radius {
                                 let new_pos = pos.add_raw(dx, dz);
                                 let req_stage = dependency[dx.abs().max(dz.abs()) as usize];
+                                self.unload_chunks.remove(&new_pos);
                                 if new_pos == pos {
                                     Self::ensure_dependency_chain(
                                         &mut self.graph,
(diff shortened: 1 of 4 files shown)


RoosterBooster007:
Hey @ZlordHUN, thanks for the fixes here. Yet, it looks like the core has since landed on master (Snowiiii re-wired `on_death` back into the damage pipeline and fixed the event firing, and the chunk-retention fix this was stacked on went in via #3227), so this PR's effectively superseded, and I'll close it. The totem-of-undying activation and armor-durability bits never made it in, though. A fresh PR would be great if you have time.

PR #3145, opened 2026-08-29
fix(net): honor movement lock for vehicle movement

Srinivasan-78:
### Problem

`is_movement_locked` is checked in the `player_position` handlers, but `handle_move_vehicle` applies the client-sent position unconditionally. A player with the movement lock set can therefore still move anywhere while riding a boat, minecart or horse, which defeats the purpose of the lock (dialogues/cutscenes).

### Fix

Reject the packet and teleport the client back to the server-side position, mirroring what the `player_position` handlers already do.

### Testing

Not compiled or run — I don't have a Rust toolchain available in the environment where this was written, so CI is the first real check. The change only uses existing APIs (`Player::is_movement_locked`, `JavaClient::force_tp`) and follows the existing call pattern in `player_position.rs`.


RoosterBooster007:
Please fix conflicts.

PR #3146, opened 2026-08-29
fix(redstone, repeator): correct repeater behaviour and interaction

zinedinp:
- on_use: write with BlockFlags::NOTIFY_ALL (vanilla setBlock(..., 3))
- on_use: return BlockActionResult::Success (vanilla SUCCESS, not SUCCESS_SERVER)
- on_scheduled_tick: document explicit front-neighbor poke (vanilla onPlace call in tick)
- abstract_redstone_gate: guard on_neighbor_update with world.get_block == block check
- abstract_redstone_gate: use break_block and update with diode's own block, not source

Also fixes client-prediction flicker on block interactions

This does not yet fix redstone freezing on chunk reload.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved hopper handling when neighboring blocks change, preventing invalid state updates.
  - Fixed invalid redstone gates being removed correctly and notifying neighboring blocks as expected.
  - Prevented duplicate neighbor updates following redstone gate changes.

- **Maintenance**
  - Clarified repeater update timing and signal-direction behavior without changing runtime functionality.


zinedinp:
Dropped my block-change ack fix from this branch.
https://github.com/Pumpkin-MC/Pumpkin/pull/3223 fixes the same flicker upstream.

Both move `ClientboundBlockChangedAckPacket` out of the packet loop so it is sent
after the tick's block updates are flushed. Mine did it in `World::tick` directly
after `flush_block_updates()`; #3223 does it at the top of
`Player::process_inbound_packets`.

There's no reason to carry two implementations of one fix, so this branch now uses
upstream's and contains only the redstone changes.

crates/pumpkin/src/block/blocks/redstone/repeater.rs
```diff
@@ -98,7 +101,8 @@ impl BlockBehaviour for RepeaterBlock {
         let props = RepeaterProperties::from_state_id(state.id);
         Self::on_use(props, args.world, *args.position, args.block);
 
-        BlockActionResult::SuccessServer
+        // Vanilla `RepeaterBlock.useWithoutItem`: `InteractionResult.SUCCESS`.
```
RoosterBooster007: Looks good, but maybe remove this comment and shorten the others?

zinedinp: thanks for crossreading, i cleaned up the comments

RoosterBooster007:
Thanks, @zinedinp!

Issue #2975, opened 2026-08-16
Hopper item duplication glitch


pizzapunt53820:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

4ec5eb8

### Operating System

Fedora Linux 44

### Bug Description & Expected Behavior

When placing a hopper facing into a chest, dropping an item into the hopper causes the item to be continuously duplicated into the chest. The item in the hopper does not go out of the hopper when it goes into the chest. This screen recording shows the issue:
https://github.com/user-attachments/assets/351ffb9c-aae6-4594-830d-e7d9cf218503

### Steps to Reproduce

1. Place a single or double chest
2. Place a hopper facing into the chest
3. Drop an item in the hopper
4. Look inside the chest and inside the hopper. The items will duplicate into the chest.

Mettwasser:
Seems to be fixed in 5b833274d46b50a519d6b4e532a7b44f05447c64
Not directly in this commit, but I can't reproduce it on that commit

pizzapunt53820:
> Seems to be fixed in 5b833274d46b50a519d6b4e532a7b44f05447c64
> Not directly in this commit, but I can't reproduce it on that commit

I'll test this again on that commit later

pizzapunt53820:
> Seems to be fixed in [5b83327](https://github.com/Pumpkin-MC/Pumpkin/commit/5b833274d46b50a519d6b4e532a7b44f05447c64) Not directly in this commit, but I can't reproduce it on that commit

I tested it again on commit https://github.com/Pumpkin-MC/Pumpkin/commit/5b833274d46b50a519d6b4e532a7b44f05447c64 and had the same duplication glitch as before. Could you perhaps send a screen recording of your testing without being able to reproduce it?

RoosterBooster007:
This should be fixed as of #3147. Feel free to reopen if not.

PR #3147, opened 2026-08-29
fix(hopper): prevent TOCTOU race and item duplication crash

zinedinp:
- Guard on_neighbor_update against block change
- Fix eject_items -> remove item before offer, restore on reject
- Guard tick() before from_state_id parse to prevent reading destroyed block
  as air (like trial_spawner.rs)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved hopper item transfers to prevent duplication, stack mixing, and overflow.
  - Preserved concurrent inventory updates during item extraction and restoration.
  - Restored items accurately when transfers fail, scattering leftovers when storage cannot safely accommodate them.
  - Hopper processing now stops safely when the block is no longer a hopper or has invalid state.
  - Transfer handling processes inventory slots consistently and removes items only when appropriate.
  - Improved handling of stack limits during item movement.


zinedinp:
I just saw, that it want the changes from https://github.com/Pumpkin-MC/Pumpkin/pull/3146 too because forked from there, accept it first, before this one.

zinedinp:
https://github.com/Pumpkin-MC/Pumpkin/issues/2975

zinedinp:
better to have a rebase and doesnt include the repeater

RoosterBooster007:
Thanks, @zinedinp!

PR #3148, opened 2026-08-29
fix(world): broadcast block-destroy particles excluding breaking player and fire block

zinedinp:
- Broadcast `ParticlesDestroyBlock` in `break_block()` for all non-fire blocks
- Exclude breaking player via new `sync_world_event_except()` method
- Refactor `sync_world_event()` to delegate to `sync_world_event_except(None, ...)`

## Summary by CodeRabbit

- **Bug Fixes**
  - Block-breaking particle effects now display with the correct block appearance for Java and Bedrock clients.
  - Players who trigger a block-breaking effect no longer receive a duplicate particle effect.
  - Block-destruction particles now appear correctly when constructing golems, summoning the wither, and breaking doors.
  - Fire and soul-fire blocks no longer produce block-destruction particles.
  - World-event synchronization remains compatible with Java clients while preserving global event broadcasts.


zinedinp:
New helper "World::sync_destroy_block_particlessits" next to "sync_world_event_except" and sends both packets through "broadcast_to_chunk_except_editioned" (or the non-except variant when there's no breaker). The exclusion is by UUID and applies to both editions, so a Bedrock breaker is skipped exactly like a Java one.

RoosterBooster007:
Hey @zinedinp, thanks for digging into this one. However, it looks like the `break_block` half already landed in #3345, and it handles the Bedrock-breaker case specially, and since this one applies cleanly on top rather than conflicting, merging it would double-fire the destroy particles. I'll close this out, but if you want to reopen it (or a new PR) slimmed down to just the carved pumpkin/wither skull/door-break call sites (they still go through the Java-only `sync_world_event`), that's a change I'd happily review.

zinedinp:
thanks for letting me, i will make a PR in the next time and see how to implement it on the newer implementation.

PR #3149, opened 2026-08-29
Fix VehicleMoveEvent not triggering and canceling

Neol3108:
The event wasn't being triggered and also not truly cancellable


crates/pumpkin/src/net/java/play/mod.rs
```diff
@@ -57,7 +57,7 @@ use pumpkin_protocol::bedrock::client::CMovePlayer;
 use pumpkin_protocol::codec::var_int::VarInt;
 use pumpkin_protocol::codec::var_ulong::VarULong;
 use pumpkin_protocol::java::client::play::{
-    CBlockUpdate, CCommandSuggestions, CEntityPositionSync, CHeadRot, CPingResponse,
+    CBlockUpdate, CCommandSuggestions, CEntityPositionSync, CHeadRot, CMoveVehicle, CPingResponse,
```
vyPal, line 58: Is this change needed?

crates/pumpkin/src/entity/vehicle/minecart.rs
```diff
@@ -483,6 +474,29 @@ impl EntityBase for MinecartEntity {
...
+
+        if move_result.cancelled {
+            self.vehicle
+                .entity
+                .velocity
+                .store(Vector3::new(0.0, 0.0, 0.0));
+            self.vehicle.entity.send_velocity();
+        }
+
+        if attempted_to != from {
```
vyPal, line 490: Shouldn't `attempted_to` be replaced with `to` in the if statement?

vyPal (changes requested):

crates/pumpkin/src/net/java/play/mod.rs
Neol3108: Yes, `crates/pumpkin/src/net/java/play/move_vehicle.rs` uses it through `use super::*;` on line 2.

crates/pumpkin/src/entity/vehicle/minecart.rs
Neol3108: No. When movement is cancelled, `to == from`. Testing `to != from` would therefore skip `send_pos_rot()`, so clients and passengers would not receive the corrective position, the exact behaviour that can cause sinking/desynchronization. `attempted_to != from` correctly detects that motion was attempted and a correction may be required.

MegalithOfficial:
Any updates on this?

PR #3156, opened 2026-08-30
fix(block): break suspicious sand and gravel when they fall 🤖🤖🤖

### LINKED ISSUE #3067: Suspicious sand and gravel do not have gravity.
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

### Operating System

CachyOS (Arch)

Suspicious sand and gravel do not have gravity. Unlike vanilla behavior where they fall and break when touching the ground.

### Steps to Reproduce

1. Place Suspicious sand and/or gravel in the air
2. Watch as it floats

Image

Image

ydw1904:
## Description

Fixes #3067

As reported in the issue:

> Suspicious sand and gravel do not have gravity. Unlike vanilla behavior where they fall and break when touching the ground.

`3a3ba2c87` has since added `SUSPICIOUS_SAND` and `SUSPICIOUS_GRAVEL` to `FallingBlock::ids()`, so the gravity half of this issue is already fixed on `master`. This PR has been rescoped to the half that is still missing: the blocks fall, but they land as an ordinary block instead of breaking.

Their loot lives in a `BrushableBlockBlockEntity` that does not ride along with the falling entity, so a suspicious block that lands intact has lost its contents. Vanilla breaks it where it lands (`BrushableBlock.onBrokenAfterFall`, level event 2001: break particles and sound).

The gravity plumbing this PR used to carry (`FallingBlock::try_fall`, the `BrushableBlock` fall hooks, and the `cancel_drop` flag on `FallingEntity`) is all gone, since `master` covers it. What is left is one branch in `FallingEntity::land()`.

## Depends on #3154

This branch is stacked on #3154, whose commit it carries as its base. `land()` is introduced there; this PR adds a branch to it. Please merge #3154 first, or the diff here will not apply.

## Testing

- `cargo test -p pumpkin falling`: passing
- `cargo clippy -p pumpkin --all-targets`: zero warnings
- In game (Java): placed suspicious sand/gravel with air below, it falls and breaks with particles and sound on landing instead of being placed back as a block.

## Not in scope

Two neighbouring issues found while rescoping, both from `3a3ba2c87` registering `FallingBlock` after `BrushableBlock` in `block/registry.rs`, which shadows the latter for these two ids:

- `BrushableBlock::broken` no longer runs, so breaking a structure generated suspicious block does not drop its contained item.
- `BrushableBlock::placed` no longer runs, so no block entity is created on placement.

Happy to open a separate PR for those.

## AI assistance

This PR was developed with AI assistance (OpenCode, model: GLM-5.2); rescope and rebase by Claude Opus 5 in Claude Code.

### DIFF
--- a/crates/pumpkin/src/entity/falling.rs
+++ b/crates/pumpkin/src/entity/falling.rs
@@ -1,7 +1,6 @@
-use pumpkin_data::Block;
-use pumpkin_data::BlockStateId;
 use pumpkin_data::damage::DamageType;
 use pumpkin_data::entity::EntityType;
+use pumpkin_data::{Block, BlockState, BlockStateId, item::Item, item_stack::ItemStack};
 use pumpkin_util::math::position::BlockPos;
 use pumpkin_world::world::BlockFlags;
 use std::sync::{Arc, atomic::Ordering};
@@ -15,6 +14,12 @@ use crate::{
 pub struct FallingEntity {
     entity: Entity,
     block_state_id: BlockStateId,
+}
+
+/// Vanilla only lets a falling block settle where it can replace what is already
+/// there; anything else (torches, slabs, stairs, ...) makes it drop as an item.
+const fn can_replace_landing_block(state: &BlockState) -> bool {
+    state.replaceable()
 }
 
 impl FallingEntity {
@@ -42,6 +47,22 @@ impl FallingEntity {
         let entity = Arc::new(Self::new(entity, block_state));
         world.spawn_entity_non_save(entity);
     }
+
+    /// Vanilla `FallingBlockEntity` landing branch: place the block back when the
+    /// landing position is replaceable, otherwise drop it as an item.
+    fn land(&self) {
+        let world = self.entity.world.load();
+        let position = self.entity.block_pos.load();
+
+        if can_replace_landing_block(world.get_block_state(&position)) {
+            world.set_block_state(&position, self.block_state_id, BlockFlags::NOTIFY_ALL);
+        } else if world.level_info.load().game_rules.entity_drops {
+            let block = Block::from_state_id(self.block_state_id);
+            if let Some(item) = Item::from_id(block.item_id) {
+                world.drop_stack(&position, ItemStack::new(1, item));
+            }
+        }
+    }
 }
 
 impl EntityBase for FallingEntity {
@@ -56,11 +77,7 @@ impl EntityBase for FallingEntity {
         entity.tick_block_collisions(caller);
         if entity.on_ground.load(Ordering::Relaxed) {
             entity.velocity.store(velo.multiply(0.7, -0.5, 0.7));
-            entity.world.load().set_block_state(
-                &self.entity.block_pos.load(),
-                self.block_state_id,
-                BlockFlags::NOTIFY_ALL,
-            );
+            self.land();
             self.entity.remove();
         }
 
@@ -96,5 +113,18 @@ impl EntityBase for FallingEntity {
 
     fn cast_any(&self) -> &dyn std::any::Any {
         self
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn falling_blocks_do_not_replace_partial_blocks() {
+        assert!(can_replace_landing_block(Block::AIR.default_state));
+        assert!(!can_replace_landing_block(Block::TORCH.default_state));
+        assert!(!can_replace_landing_block(Block::STONE_SLAB.default_state));
+        assert!(!can_replace_landing_block(Block::OAK_STAIRS.default_state));
     }
 }
--- a/crates/pumpkin/src/entity/falling.rs
+++ b/crates/pumpkin/src/entity/falling.rs
@@ -1,6 +1,7 @@
 use pumpkin_data::damage::DamageType;
 use pumpkin_data::entity::EntityType;
-use pumpkin_data::{Block, BlockState, BlockStateId, item::Item, item_stack::ItemStack};
+use pumpkin_data::world::WorldEvent;
+use pumpkin_data::{Block, BlockId, BlockState, BlockStateId, item::Item, item_stack::ItemStack};
 use pumpkin_util::math::position::BlockPos;
 use pumpkin_world::world::BlockFlags;
 use std::sync::{Arc, atomic::Ordering};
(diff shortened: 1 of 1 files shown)


ydw1904:
### Known limitations and follow-up behavior
## Falling-block testing
The Java Edition behavior where suspicious sand or gravel drops as an item after remaining a falling-block entity for more than 600 ticks cannot currently be reproduced in-game in Pumpkin.
The limitation is in FallingEntity::tick. It calls move_entity and tick_block_collisions, then overwrites the resulting velocity with the stale pre-collision velo.
This causes both vanilla testing methods to fail:
- Bubble-column acceleration is applied by BubbleColumnBlock, but immediately overwritten.
- Cobweb collision is dispatched correctly, but the falling entity continues accumulating gravity behind the slowdown and passes through too quickly.
[#3154](https://github.com/Pumpkin-MC/Pumpkin/pull/3154) only changes landing placement. It does not cause the cobweb issue or fix the discarded collision velocity. Bubble columns are marked replaceable, so #3154 still permits a falling block to replace one if the failed bubble push allows it to reach the floor.
## Brush interaction
There is also a separate, pre-existing brush implementation issue.
BrushItem::use_on_block advances the dusted state once per use-on-block packet. Consequently, spamming right-click completes brushing in four clicks. Holding use does not progress because ItemBehaviour has no per-tick use hook.
The code sets the active-hand state, but the brushing sequence has no periodic processing, brushing particles, or block-entity updates for the item-emergence animation.
The brush also bypasses BrushableBlockBlockEntity:
- The stored item, hits, and direction are ignored.
- Naturally generated LootTable and LootTableSeed values are not read by BrushableBlockBlockEntity::from_nbt.
- Completion produces an item from a hardcoded random pottery-sherd list.
- A separate BrushableBlock::brush method reads the stored item, but it currently has no callers.
Therefore held-use brushing, natural archaeology loot, and the vanilla reveal animation are not currently implemented correctly.
## Java Edition 600-tick behavior
FallingEntity currently has no age or 600-tick lifetime handling, so the behavior tracked by [MC-260073](https://mojira.dev/MC-260073) is not implemented.
Because this is a Java Edition-specific issue marked “Won’t Fix” and does not occur in Bedrock Edition, I suggest leaving it out of this PR and tracking it separately. If strict Java parity is preferred, the timeout should be implemented once in the generic falling-entity logic after bubble-column and cobweb interactions are fixed and can be tested properly.

SurvivorsDev:
This is still reproducible on `1ac447f2`: suspicious sand and gravel placed in the air fall, but they settle as a block instead of breaking, so #3067 is still open.

Your landing branch looks right, and the drop-as-item path for a non-replaceable landing spot is something I'd missed entirely. One thing it doesn't reach, though: `BrushableBlock` is also overwritten in the block registry. `BlockManager::register` keeps one behaviour per block id (`block_indices[i] = idx`), `BrushableBlock` registers at `registry.rs:317` and `FallingBlock` at `registry.rs:413`, and both claim `SUSPICIOUS_SAND`/`SUSPICIOUS_GRAVEL`. `FallingBlock` wins, so `BrushableBlock::placed` and `broken` never run — the `BrushableBlockBlockEntity` is never created and a buried item is lost when the block is broken.

Vanilla avoids that by not extending `FallingBlock` at all: `BrushableBlock` implements `Fallable` itself and calls `disableDrop()`, which is what makes the falling block break on landing. Moving the falling hooks onto `BrushableBlock` fixes the registry conflict and the landing behaviour together.

I have that working locally — a `cancel_drop` flag on `FallingEntity` plus the falling hooks on `BrushableBlock` — and would rather not open a competing PR. @ydw1904 are you planning to revisit this? If not, I'm happy to open one that carries your landing logic with you credited as co-author.

SurvivorsDev:
Read the rest of the diff properly this time — a few notes, in case this gets picked back up.

The most important one: **this is based on a tree without the concrete powder handling.** The `set_block_state` call `land()` replaces has since grown a branch on current `master`:

```rust
if block.has_tag(&tag::Block::MINECRAFT_CONCRETE_POWDERS)
    && FallingBlock::should_solidify(&**world, &landing_pos)
    && let Some(name) = block.name.strip_suffix("_powder")
    && let Some(concrete) = Block::from_name(name)
{
    state_id = concrete.default_state.id;
}
```

Rebasing `land()` over that as-is would drop it, and powder would stop hardening when it lands in water.

Smaller things against `FallingBlockEntity`:

- `can_replace_landing_block` only checks `replaceable()`. Vanilla also requires `wouldSurvive` — `blockState.canSurvive(level, pos)` and `!wouldContinueFalling` — so a block can currently settle somewhere it cannot stay.
- `callOnBrokenAfterFall` runs on the drop-as-item path in vanilla too, not only when the block breaks.
- `BrushableBlock.onBrokenAfterFall` emits `GameEvent.BLOCK_DESTROY` alongside the `2001` level event; only the particle is sent here.
- Deciding by block id inside the entity works, but vanilla puts the decision on the block: `BrushableBlock.tick` calls `entity.disableDrop()`, and `FallingBlockEntity` just honours `cancelDrop`. A flag keeps the next `Fallable` from having to edit a list in the entity.

For what it's worth, the drop-as-item path and the two tests are both things my local version was missing.

ydw1904:
> This is still reproducible on `1ac447f2`: suspicious sand and gravel placed in the air fall, but they settle as a block instead of breaking, so #3067 is still open.
> 
> Your landing branch looks right, and the drop-as-item path for a non-replaceable landing spot is something I'd missed entirely. One thing it doesn't reach, though: `BrushableBlock` is also overwritten in the block registry. `BlockManager::register` keeps one behaviour per block id (`block_indices[i] = idx`), `BrushableBlock` registers at `registry.rs:317` and `FallingBlock` at `registry.rs:413`, and both claim `SUSPICIOUS_SAND`/`SUSPICIOUS_GRAVEL`. `FallingBlock` wins, so `BrushableBlock::placed` and `broken` never run — the `BrushableBlockBlockEntity` is never created and a buried item is lost when the block is broken.
> 
> Vanilla avoids that by not extending `FallingBlock` at all: `BrushableBlock` implements `Fallable` itself and calls `disableDrop()`, which is what makes the falling block break on landing. Moving the falling hooks onto `BrushableBlock` fixes the registry conflict and the landing behaviour together.
> 
> I have that working locally — a `cancel_drop` flag on `FallingEntity` plus the falling hooks on `BrushableBlock` — and would rather not open a competing PR. @ydw1904 are you planning to revisit this? If not, I'm happy to open one that carries your landing logic with you credited as co-author.

I wont be reopening this PR and have moved onto other things, you can go ahead and open a PR if you want, a co-author tag would be appreciated

PR #3162, opened 2026-08-31
fix(block): use legacy-solid support check for plain signs

### LINKED ISSUE #3161: Signs: plain sign support tests isFaceSturdy where vanilla tests legacy isSolid, so signs cannot be placed on chests (and ~200 other blocks)
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

`232bee30f` (master)

### Operating System

Platform-independent (logic only)

Placing a plain (standing or wall) sign on a chest silently does nothing: `can_place_at` returns false, so `place_block` bails out with `Ok(None)` (`crates/pumpkin/src/block/registry.rs:655-666`) — nothing is placed, the item is not consumed, and the client's predicted sign is reverted by the next block update.

The root cause is that the sign code uses the wrong vanilla solidity predicate for plain signs. Vanilla 26.2 gates plain sign support on the *legacy solid* flag, not on face sturdiness:

- `StandingSignBlock.canSurvive` → `level.getBlockState(pos.below()).isSolid()`
- `WallSignBlock.canSurvive` → `level.getBlockState(pos.relative(FACING.getOpposite())).isSolid()`

`isSolid` is `BlockBehaviour.BlockStateBase.legacySolid`, computed by `calculateSolid()`: true when the collision shape is non-empty and its bounds either average >= 0.7291666 across the three axes or span the full block on Y. Pumpkin already extracts exactly this flag as `BlockState::is_solid()` (`crates/pumpkin-data/src/block_state.rs:107`, "Returns the legacy value for whether a block is solid") — the sign code just never consults it.

`crates/pumpkin/src/block/blocks/signs.rs` instead tests `is_side_solid` / `is_center_solid`, which are `isFaceSturdy` (`SupportType::FULL` / `CENTER`) — the predicate vanilla uses for *hanging* signs only. Four sites:

- `is_valid_support` (line 61), which feeds `detect_support`'s side scan used by `on_place` for wall/standing variant selection;
- `can_place_at` (lines 368-376): the `Up` arm tests `is_center_solid(Up)` on the block below and the horizontal arm tests `is_side_solid`;
- `get_state_for_neighbor_update` (lines 416-424): the `Down` arm (standing) and the horizontal arm (wall) run the same tests, so even a sign that got placed breaks on its first neighbour update.

The chest makes the divergence concrete: its first state in `assets/blocks.json` has `state_flags = 98` — bit 6, `IS_SOLID`, set — but `side_flags = 64`, only `DOWN_CENTER_SOLID`. `is_solid()` is true while every `is_side_solid` / `is_center_solid(Up)` test is false, so both sign forms are refused on every face. It is far from alone: counting default states in `assets/blocks.json` (excluding leaves and signs, see below), 232 blocks refuse a standing sign on top and 246 refuse a wall sign on every side — chests, trapped and ender chests, beds, hoppers, cacti, conduits and so on.

The file already carries indirect evidence of the gap: its `is_permissive` escape hatch admits leaves and signs specifically — exactly two families that are `isSolid` without being face-sturdy, i.e. hand-patched instances of this same divergence. Both stay valid supports under `is_solid()` (`oak_leaves` `state_flags = 226`, `oak_sign` `state_flags = 98`; bit 6 set in both), so switching the plain-sign checks to `is_solid()` loses nothing they currently allow.

The hanging-sign checks are correct as written and should keep `isFaceSturdy`: vanilla `CeilingHangingSignBlock.canSurvive` uses `isFaceSturdy(..., Direction.DOWN, SupportType.CENTER)` on the block above, and `WallHangingSignBlock.canAttachTo` uses `isFaceSturdy(..., attachFace, SupportType.FULL)` (plus a `BlockTags.WALL_HANGING_SIGNS` chaining case). Only the plain-sign paths are wrong.

Expected: plain standing and wall signs gate support on `BlockState::is_solid()`, so they can be placed on — and survive neighbour updates against — chests and every other legacy-solid block, as in vanilla.

### Steps to Reproduce

1. Place a chest.
2. Sneak-right-click its top face holding any plain sign (e.g. `oak_sign`).
3. Nothing is placed: the client predicts a sign briefly, the server reverts it, and the item is kept.
4. Repeat on a side face for the wall-sign form — same result. In vanilla both placements succeed.

pxperr:
## What was changed?

The support checks for plain (standing and wall) signs in `crates/pumpkin/src/block/blocks/signs.rs` now test `BlockState::is_solid()` - the legacy-solid flag (`crates/pumpkin-data/src/block_state.rs:107`) - instead of the face-sturdy predicates `is_side_solid` / `is_center_solid`. Four sites:

* `is_valid_support` returns `is_solid()` for plain signs, with an `is_hanging` flag threaded through `detect_support` so the wall/standing variant selection in `on_place` uses the right predicate per sign family
* `can_place_at`: the `Up` arm (standing) and the horizontal arm when the sign is not hanging
* `get_state_for_neighbor_update`: the `Down` arm (standing) and the horizontal arm when the sign is not hanging

Hanging-sign paths - `above_is_valid`, the `Down` arm of `can_place_at`, the `Up` arm of `get_state_for_neighbor_update`, and the horizontal arms when `is_hanging` - keep their existing checks unchanged.

Vanilla gates plain sign support on the legacy `isSolid` flag, not on face sturdiness: `StandingSignBlock.canSurvive` is `level.getBlockState(pos.below()).isSolid()` and `WallSignBlock.canSurvive` is `level.getBlockState(pos.relative(FACING.getOpposite())).isSolid()`. Only the *hanging* sign checks use `isFaceSturdy`, and those were already correct here.

A chest is legacy-solid with no sturdy face (`assets/blocks.json`: `state_flags = 98`, bit 6 `IS_SOLID` set; `side_flags = 64`, only `DOWN_CENTER_SOLID`), so `can_place_at` returned false on every face, `place_block` bailed out with `Ok(None)` (`crates/pumpkin/src/block/registry.rs:655-666`), and the client's predicted sign was silently reverted. Counting default states, 232 blocks refused a standing sign on top and 246 refused a wall sign on every side - chests, beds, hoppers, cacti, conduits and more - and `get_state_for_neighbor_update` ran the same test, so such a sign would have popped on its first neighbour update even if it had been placed.

The existing `is_permissive` escape hatch (leaves and signs) was hand-patching two instances of exactly this gap - both families are `isSolid` without being face-sturdy. They remain valid supports under `is_solid()` (`oak_leaves` `state_flags = 226`, `oak_sign` `state_flags = 98`), so the plain-sign paths no longer need the special case.

Plain signs place - and survive neighbour updates - on every legacy-solid block, matching vanilla: a standing sign on a chest lid and a wall sign on a chest side both work again. Everything the plain-sign paths previously accepted via the leaves/signs escape hatch is still accepted. The only supports newly *refused* are seven non-solid ones vanilla also refuses (ladder sides; the tops of snow layers, end rods, chorus plants, scaffolding and the two azaleas). Hanging signs are behaviourally untouched, as is redstone/piston logic - this changes no data, only which flag the sign code reads.

* The hanging-sign checks are deliberately left alone, but two divergences in the same file are worth a follow-up: `detect_support`'s `above_is_valid` and the `Down` arm of `can_place_at` test `is_side_solid(Down)` (`isFaceSturdy(DOWN, FULL)`) where vanilla `CeilingHangingSignBlock.canSurvive` uses `SupportType.CENTER` (`is_center_solid(Down)`, which `get_state_for_neighbor_update` already uses), so ceiling-hanging signs cannot hang from centre-sturdy blocks such as fences; and vanilla `WallHangingSignBlock.canAttachTo` has a `BlockTags.WALL_HANGING_SIGNS` chaining case that the leaves/signs permissive tags only approximate.
* No unit tests added: every changed path takes `&World`, and there is no existing block-placement test harness to slot into (the `#[cfg(test)]` modules under `block/blocks/` cover pure logic only). Verified by tracing both placement flows end to end against the flag data in `assets/blocks.json` rather than by an in-game session.

image

Fixes #3161

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved sign placement and stability validation.
  - Standing and wall-mounted signs now require appropriate solid support.
  - Hanging signs continue to support their existing placement behavior.
  - Sign support checks now remain consistent when neighboring blocks change.


RoosterBooster007:
Looks good. Thanks, @jackpipergit!

PR #3168, opened 2026-08-31
feat(plugin): fire SpawnerSpawnEvent from the monster spawner

### LINKED ISSUE #3167: SpawnerSpawnEvent is never fired by the monster spawner
`SpawnerSpawnEvent` exists in the plugin API and can be registered, but nothing in the server ever constructs it, so a handler on it is silently dead code.

The trial spawner does fire its equivalent. `TrialSpawnerSpawnEvent` goes out from `block/entities/trial_spawner.rs:846`, and cancelling it stops the spawn. The ordinary monster spawner at `block/entities/mob_spawner.rs` spawns straight into `world.spawn_entity` with no event, so plugins can see trial spawner mobs but not spawner mobs.

Both events carry the same fields (`entity_id`, `spawner_pos`), so the trial spawner site works as a template.

This looks like leftover work from #1609, which tracked adding the event types themselves rather than wiring their fire sites.

PR to follow.

pxperr:
## Description

Fires `SpawnerSpawnEvent` from the monster spawner. Closes #3167.

The event was in the API and registered fine, but nothing constructed it, so a handler on it never ran. The trial spawner already fires its equivalent from `trial_spawner.rs:846` and both events carry the same fields, so the first commit mirrors that site: build the entity, fire, and only spawn if the event was not cancelled.

The second commit fixes a problem that testing turned up. Cancelling left `spawned_any` false, so `update_spawns` never ran and the delay stayed at zero, which made the spawner retry every tick. Measured, that was about 50 events a second for as long as a plugin kept cancelling, against roughly 6 a minute when it does not. A cancelled spawn now counts as a completed attempt, so the spawner takes its normal cooldown and simply produces nothing.

I originally modelled the cancel path on the `is_space_empty` check above it, which also continues without counting the attempt. That was the wrong comparison: a blocked spawner is a transient state that clears itself, whereas a plugin cancelling is usually standing policy, so the retry never stops.

## Testing

First commit checked in game against a plugin with a blocking handler on the event:

- Fires once per spawned mob, carrying the spawner's own block position rather than the mob's, with a distinct entity id each time.
- With cancelling on, 988 fires over 20 seconds produced zero mobs, and nothing slipped through with `cancelled` unset.
- Unsetting it resumed spawning, 6 fires over 60 seconds in batches about 26 seconds apart, which is the normal cooldown.

That 988 over 20 seconds is what prompted the second commit. The cooldown fix itself is reasoned from those numbers rather than re-measured in game, so the retry rate after it is worth a check by someone able to run one.

## Summary by CodeRabbit

- **New Features**
  - Mob spawner generation now supports event-based control before entities are spawned.

- **Behavior Changes**
  - Spawn attempts can be cancelled before insertion.
  - Cancelled attempts still count toward the spawner’s cycle and cooldown behavior.


RoosterBooster007:
Thanks, @jackpipergit!

PR #3170, opened 2026-08-31
feat(plugin): fire LeavesDecayEvent when leaves decay

### LINKED ISSUE #3169: LeavesDecayEvent is never fired when leaves decay
`LeavesDecayEvent` exists in the plugin API and can be registered, but nothing in the server ever constructs it, so a handler on it is silently dead code.

Leaf decay itself is implemented. `LeavesBlock::random_tick` in `block/blocks/leaves.rs:101` checks the persistent flag and the decay distance, then calls `break_block` directly.

That means decaying leaves are invisible to plugins twice over, since `break_block` does not fire `BlockBreakEvent` either. A land protection or tree farm plugin has no way to see or stop leaves disappearing inside a claim.

The event is marked cancellable and carries the block position and world, so the natural behaviour is for a cancelled event to leave the block alone until the next random tick.

This looks like leftover work from #1609, which tracked adding the event types themselves rather than wiring their fire sites.

PR to follow.

pxperr:
## Description

Fires `LeavesDecayEvent` when leaves decay. Closes #3169.

The event was in the API and registered fine, but nothing constructed it, so a handler on it never ran. Decay is implemented in `LeavesBlock::random_tick`, which called `break_block` directly. Since `break_block` does not fire `BlockBreakEvent` either, decaying leaves were invisible to plugins entirely, and a claim or tree farm plugin had no way to see or stop them.

The event is cancellable and carries the position and world, so a cancelled event leaves the block alone and it is reconsidered on the next random tick.

Fired only from the decay branch, so leaves broken by a player or by a neighbour update are unaffected.

## Testing

In-game verification to follow. The fire site follows the same `fire_blocking` and `event.cancelled` shape as the other block events in this tree.

### DIFF
--- a/crates/pumpkin/src/block/blocks/leaves.rs
+++ b/crates/pumpkin/src/block/blocks/leaves.rs
@@ -102,6 +102,17 @@ impl BlockBehaviour for LeavesBlock {
         let state_id = args.world.get_block_state_id(args.position);
         let props = OakLeavesLikeProperties::from_state_id(state_id, args.block);
         if !props.persistent && props.distance == DECAY_DISTANCE {
+            let mut event = crate::plugin::api::events::block::leaves_decay::LeavesDecayEvent::new(
+                *args.position,
+                args.world.clone(),
+            );
+            if let Some(server) = args.world.server.upgrade() {
+                server.plugin_manager.fire_blocking(&server, &mut event);
+            }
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Thanks for the fix, but it looks like it already landed in f8d16276. I'll close this out as superseded.

PR #3171, opened 2026-08-31
perf: move packet and chunk CPU work off async workers

MegalithOfficial:
## Description
Alex asked if someone with threading experience could look into some of Pumpkin's performance issues, so I volunteered.

I started by checking the main hot paths and profiling the server while players were continuously loading chunks. The largest spikes came from CPU-heavy work running on Tokio workers, mainly Java packet compression, loaded chunk processing and chunk serialization.

Java packets were compressed directly inside each client's outgoing packet task. I separated packet framing from the socket write so packets can be framed in batches. Batches that require compression now run on the blocking pool, while small uncompressed packets stay on the existing path. The framed packets are then written together without changing their order, encryption or completion handling.

Loaded chunks also performed their lighting checks and conversions on the async worker after being read from disk. I moved this work to the blocking pool too. The same applies to proto chunk upgrades, region decoding and the expensive parts of Anvil, Linear and Pump serialization.

Pump copied every compressed chunk when taking a region snapshot for writing. Its compressed chunk data now uses `Bytes`, so these snapshots share the existing buffers instead.

While checking the tick loop, I also noticed that TickStart and TickEnd always entered the async runtime even when no plugin had registered a handler. This is now skipped when there is nothing to call.

## Testing

Besides the cargo/clippy checks, I tested this with actual players constantly loading chunks.

I wrote my own Bash benchmark script and used [Eoghanmc22/rust-mc-bot](https://github.com/Eoghanmc22/rust-mc-bot) to connect and control the bots. The script starts the server, runs the selected scenario and collects the tick times. I tested upstream and this branch using the same world, configuration and release build. I ran each scenario multiple times because single runs were sometimes noisy.

The main test was `stream50`, which continuously teleports 50 bots in a spiral to keep loading chunks.

Upstream averaged between 5.2 and 7.4 MSPT, with P95 between 12.3 and 18.7 ms and P99 between 19 and 34.3 ms. The worst ticks were between 259 and 407 ms.

This branch averaged between 2.7 and 2.9 MSPT, with P95 between 5.7 and 6.6 ms and P99 between 7.8 and 8.4 ms. The worst ticks were between 17 and 18 ms, and there were no ticks above 50 ms.

I also tested 50 bots staying around spawn. Upstream averaged 4.3-4.7 MSPT with a P99 of 9.0-9.6 ms. This branch averaged 3.7-4.3 MSPT with a P99 of 7.0-7.6 ms.

All 50 bots joined successfully in every run and I did not get any panics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -484,6 +484,9 @@ name = "bytes"
 version = "1.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc652a48c352aef3ea3aed32080501cf3ef6ed5da78602a020c991775b0aff04"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "cap-fs-ext"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -130,7 +130,7 @@ syn = { version = "3.0", default-features = false, features = ["full", "parsing"
 
 thiserror = { version = "2.0", default-features = false }
 
-bytes = { version = "1.12", default-features = false, features = ["std"] }
+bytes = { version = "1.12", default-features = false, features = ["serde", "std"] }
 
 # Concurrency/Parallelism and Synchronization
 futures = { version = "0.3", default-features = false, features = ["std"] }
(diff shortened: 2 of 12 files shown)


Snowiiii:

PR #3174, opened 2026-08-31
fix(raycast): respect the hit predicate for the ray's origin block + use packet rotation for buckets

### LINKED ISSUE #3102: Lava and water can be placed in air
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

9e9f1b5/release

### Operating System

Windows 11

Lava and water can be placed in air and is not flowing + they are placed 1 block over player.

### Steps to Reproduce

1. Get a bucket of water or lava.
2. Press Right-click (the place button) while looking in air.

AdmerPRO:
## What was changed?

The item-use flow now passes the rotation from the `Use Item` packet to item behaviours through a new `normal_use_with_rotation` method.

For Java 1.21 and newer, `handle_use_item` checks the packet's yaw and pitch and passes them to `ItemRegistry::on_use_with_rotation`. Older Java clients don't include these values in the packet, so they continue to use the player's stored rotation.

Empty and filled bucket behaviours now use the supplied action rotation when creating their raycasts instead of calling `player.rotation()` while processing the item. The existing `normal_use` method is still available and falls back to the stored player rotation, so existing callers such as Bedrock remain compatible.

Items that don't need the rotation from the specific action continue to use their existing `normal_use` implementation through the default trait method.

A Java `Use Item` packet contains the yaw and pitch from the exact moment the interaction happened. The packet handler was already using this rotation when calculating the `PlayerInteractEvent` target, but bucket behaviour performed another raycast using `player.rotation()`.

Since player rotation is updated through separate movement packets, the stored rotation on the server can be older than the rotation included in the item-use packet. As a result, the interaction event and the bucket raycast could end up targeting different blocks.

For example, a player could use a bucket while looking into the air, while the server still had an older rotation pointing at the underside of a ceiling. The bucket raycast could then hit that ceiling and place water or lava in the wrong location.

Using the packet rotation for the bucket behaviour keeps both raycasts based on the same interaction state and avoids relying on packet ordering.

Water and lava buckets now use the player's actual aim from the moment the item was used. They won't fall back to a potentially outdated server-side rotation when deciding where to place the fluid.

The same correction is applied to empty buckets, so collecting water or lava is also based on the correct interaction rotation.

The change remains backward-compatible:

* Java 1.21 and newer use the rotation supplied by the packet.
* Older Java versions fall back to the stored player rotation because their `Use Item` packets don't contain yaw and pitch.
* Bedrock continues to use the existing `ItemRegistry::on_use` path.
* Other item behaviours keep their current behaviour unless they opt into the rotation-aware method.

Non-finite yaw or pitch values from packets are rejected before they can be used for raycasting.

* No in-game integration test was added. This issue depends on packet ordering, player state and world raycasting, and there isn't currently an item-interaction test harness that covers the whole flow.
* The change was checked with `cargo fmt --check`, `git diff --check`, and `cargo check -p pumpkin`.
* Older Java protocol versions cannot receive the same packet-level fix because their `Use Item` packets don't contain rotation fields. They intentionally keep using the stored player rotation.
* The separate issue with `PlayerBucketEmptyEvent` being fired after the world has already been modified, and its cancelled state not being respected, is still present and should be fixed separately.

Fixes #3102 

## Summary by CodeRabbit

- **Bug Fixes**
  - Bucket interactions now use the rotation reported when the action is performed, improving targeting accuracy.
  - Raycast-based interactions correctly respect collision checks for the initially targeted block.
  - Invalid item-use rotation data from supported clients is rejected to prevent malformed interactions.

- **Improvements**
  - Item-use handling now consistently applies the relevant player or action rotation across supported client versions.


AdmerPRO:
Ready for review

RoosterBooster007 (approved):

RoosterBooster007:
@AdmerPRO, thanks for this. This is actually more of a global raycast fix, so you may want to update the PR title (ex. _"fix(raycast): respect the hit predicate for the ray's origin block + use packet rotation for buckets"_) and description. This seems to fix buckets, explosions (knockback), glass bottle filling, boats and lily pads, and potentially events/API functions.

AdmerPRO:
> @AdmerPRO, thanks for this. This is actually more of a global raycast fix, so you may want to update the PR title (ex. _"fix(raycast): respect the hit predicate for the ray's origin block + use packet rotation for buckets"_) and description. This seems to fix buckets, explosions (knockback), glass bottle filling, boats and lily pads, and potentially events/API functions.

holy side effect

RoosterBooster007:
fr lol

RoosterBooster007:
Looks good. Thanks, @AdmerPRO!

PR #3176, opened 2026-08-31
perf(world): reduce per-tick active chunk rebuilds

MegalithOfficial:
## Description

This is a follow-up to #3171

The active chunk set was recreated for every world tick. For each player, Pumpkin inserted the full simulation-distance area into a new hash set, checked `loaded_chunks` for every entry and attempted to migrate pending block entities. With 200 players at simulation distance 10, this meant around 88000 hash inserts and 88000 `DashMap` probes per tick, even when no player had crossed a chunk boundary.

The active set is now kept between ticks with a watcher count for each chunk. Overlapping player areas share the same entry and only remove it after the last watcher leaves. When a player joins, leaves, crosses a chunk boundary or the simulation distance changes, only the difference between the old and new area is applied. A one-chunk movement at simulation distance 10 touches around 40 chunks instead of rebuilding all 88000 entries.

Chunk load and unload changes are passed back to the world tick so newly loaded active chunks still migrate their pending block entities. The consumers of `active_chunks` continue to see the same chunk set.

The natural spawner's nearby-player cache also cloned its `Vec<i32>` on every read and insert. These lists now use shared storage, removing the repeated vector cloning without changing the spawning cadence or spawn decisions. `SpawnState` is still rebuilt every tick because keeping it across ticks would also require tracking mobs moving between chunks and biomes.

Block entity collection previously probed the block entity map once for every active chunk. It now iterates whichever side is smaller: the active chunk set or the chunks that contain block entities, then checks the other set. The same block entities are still ticked.

No gameplay changes are intended. The active chunk set, spawning behavior and block entity ticking remain unchanged.

## Testing

I tested this with 200 [rust-mc-bot](https://github.com/Eoghanmc22/rust-mc-bot) clients. All bots join at once, then after 30 seconds they are teleported onto a deterministic golden-angle spiral. Bot `i` uses an angle of `137.5 * i` degrees and a distance of `300 + 6 * i` blocks, placing the bots between 300 and 1500 blocks from spawn in mostly separate simulation areas. They then wander for another 150 seconds.

The world was restored from the same pregenerated snapshot before every run. World time was pinned to noon, view and simulation distance were both 10, and both master and this branch used release builds with LTO on the same machine. I alternated the run order between master and this branch.

Tick durations were recorded with [pumpkin-pulse](https://github.com/MegalithOfficial/pumpkin-pulse), and outgoing traffic was sampled with `ss`.

The first 120 seconds are the most useful comparison because they mostly measure the deterministic engine load. After that point, underground hostile mob spawning and mob AI begin to increase. Spawn rolls are not seeded, so the later part of the run has noticeably higher variance and is outside the scope of this PR.

During the first 120 seconds, master had 39 and 33 ticks above 50 ms. This branch had 22 and 21, around 40% fewer. The mean tick time was equal or lower for this branch in every 30-second slice across all four runs, generally by around 10%.

The complete 180-second results were:

Master run 1 averaged 35.68 ms with a P99 of 65.62 ms and 520 ticks above 50 ms. Master run 2 averaged 35.39 ms with a P99 of 64.83 ms and 453 ticks above 50 ms.

This branch run 1 averaged 31.73 ms with a P99 of 60.14 ms and 168 ticks above 50 ms. Run 2 averaged 33.55 ms with a P99 of 64.33 ms and 427 ticks above 50 ms.

The full-run mean tick times did not overlap between the two branches, although the later spawning and AI load introduced substantial variance in the tail results. Most of the difference between the two branch runs appears after roughly 120 seconds, when underground, hostile mob spawning and mob AI begin to dominate tick time.

Because spawn rolls are not seeded, the number and distribution of mobs naturally differ between runs. Fixed Run 2 happened to encounter a heavier late-run mob load, which pushed up its P99 and the number of ticks exceeding 50 ms.

Before that point, however, both branch runs show the same improvement, consistently across every 30-second slice.

This mainly benefits servers with many players spread across separate simulation areas. At low player counts, rebuilding the old set was cheap enough that the difference is expected to be small.

### DIFF
--- a/crates/pumpkin-world/src/chunk_system/schedule.rs
+++ b/crates/pumpkin-world/src/chunk_system/schedule.rs
@@ -9,7 +9,7 @@ use super::{
     LevelChannel,
 };
 use crate::chunk::io::Dirtiable;
-use crate::level::{Level, SyncChunk};
+use crate::level::{Level, LoadedChunkChange, SyncChunk};
 use dashmap::DashMap;
 use pumpkin_config::lighting::LightingEngineConfig;
 use pumpkin_util::math::vector2::Vector2;
@@ -56,6 +56,7 @@ pub struct GenerationSchedule {
     send_level: Arc<LevelChannel>,
 
     public_chunk_map: Arc<DashMap<Vector2<i32>, SyncChunk>>,
+    loaded_chunk_changes: Arc<crossbeam::queue::SegQueue<LoadedChunkChange>>,
     chunk_map: HashMap<ChunkPos, ChunkHolder>,
     unload_chunks: HashSetType<ChunkPos>,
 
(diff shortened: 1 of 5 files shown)


Snowiiii:

PR #3177, opened 2026-08-31
chore: deduplicate is_waterlogged logic

PigTurtleDev:
## Description
This PR removes some manual re-implementations of the already existing `Block::is_waterlogged(&self, id: BlockStateId)` (or `BlockState::is_waterlogged(&self)`) methods.

## Changes
There should be no behavioral changes introduced by this PR.

- added `Block::is_waterloggable`
- refactored `Block::with_waterlogged(&self, id: BlockStateId)` to `Block::set_waterlogged(&self, id: BlockStateId, value: bool)`
- also added `BlockStateId::is_waterlogged(self)` since that is effectively what `BlockState::is_waterlogged` calls.
- made all other implementations call one of these functions
  - and made all these functions delegate to the impl on `Block`

### Further work
Reevaluate if other `Box<dyn Properties>.to_props()` calls across the codebase can be replaced with downcasting.

## Testing

### DIFF
--- a/crates/pumpkin-data/src/block_state.rs
+++ b/crates/pumpkin-data/src/block_state.rs
@@ -152,23 +152,18 @@ impl BlockState {
 
     #[must_use]
     pub fn is_waterlogged(&self) -> bool {
-        let block = Block::from_state_id(self.id);
-
-        block.properties(self.id).is_some_and(|props| {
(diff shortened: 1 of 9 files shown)


Snowiiii:

PR #3178, opened 2026-09-01
perf(nbt): read Java numeric arrays in bulk

nobuildersnotools:
## Description

Optimize Java NBT IntArray and LongArray deserialization by reading their encoded bytes in bulk instead of issuing one reader call per element.

this primarily targets chunk loading, where non-homogeneous block sections commonly contain packed long arrays with 256–1,024 elements

## Testing

16 elements:    Original: 117.29 ns After: 119.45 ns  +1.7139%
256 elements:   Original: 364.74 ns After:187.17 ns  −48.625%
4096 elements: Original: 4.3366 µs After: 1.3734 µs  −68.482%

As you can see there is a small regression for very small arrays, but this is vastly outweighed by the massive improvements to larger ones where most of the cpu time is actually spent anyway
The intended hot path here is chunk deserialization, a heterogeneous block section usually contains 256–1024 packed longs, and a chunk can contain many of said sections

### DIFF
--- a/crates/pumpkin-nbt/src/deserializer.rs
+++ b/crates/pumpkin-nbt/src/deserializer.rs
@@ -242,6 +242,22 @@ pub trait NbtReadHelper<'a> {
     fn get_i32(&mut self) -> Result<i32>;
     /// Reads a 64-bit signed integer.
     fn get_i64(&mut self) -> Result<i64>;
+    /// Reads an array of 32-bit signed integers.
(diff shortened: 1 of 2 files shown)


QuantumSegfault:
What about serialization (writing)? Also Bedrock

Snowiiii:
Overall looks good, Bedrock stuff can be added in a separate PR, thank you @nobuildersnotools :D

PR #3181, opened 2026-09-01
fix(enchantment): implement soul speed 🤖🤖🤖

### LINKED ISSUE #3166: Soulspeed does not work
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

232bee3/release

### Operating System

Windows 11

Soul speed enchantment doesn't work

### Steps to Reproduce

1. Get boots with soulspeed
2. go on soulsand

ydw1904:
Fixes #3166

## Problem

Soul Speed existed only as data. `EnchantmentEffects` has no `location_changed` variant, and nothing outside the mob equipment roll table ever read `Enchantment::SOUL_SPEED`, so the enchantment had no gameplay effect at all: soul sand still slowed the wearer down and nothing offset it.

Soul sand itself was already correct. Its slowdown comes from `Block::velocity_multiplier`, and `LivingEntity::get_block_speed_factor` already folds in the `movement_efficiency` attribute that is meant to cancel it. Nothing ever wrote to that attribute.

## Change

`LivingEntity::tick_soul_speed`, run from `LivingEntity::tick` before movement so the modifiers apply the same tick. It implements the enchantment's effects as defined in `assets/datapacks/*/data/minecraft/enchantment/soul_speed.json`:

**`location_changed`** — while the velocity-affecting block is in `#minecraft:soul_speed_blocks`, apply two `add_value` modifiers under the id `minecraft:enchantment.soul_speed`:
- `movement_speed` += 0.0405 + 0.0105 per level above the first
- `movement_efficiency` += 1.0, which cancels the soul block slowdown

The modifiers are kept while airborne so jumping does not drop the boost, matching the datapack's `enchantment_active_check` branch, and are skipped while flying or riding. Attributes are written and broadcast only when the state or the enchantment level actually changes, so a player standing on soul sand does not generate an attribute packet every tick. Boots take one durability point at a 4% per level chance per tick while walking on the block.

**`tick`** — every fifth tick, while moving along the ground on a soul block and not flying, spawn one `minecraft:soul` particle at a random point in the bounding box footprint 0.1 above the feet, with velocity `(movement.x * -0.2, 0.1, movement.z * -0.2)`. On 35% of those ticks also play `particle.soul_escape` at volume 0.6 with pitch uniform in [0.6, 1.0).

The movement predicate reads `LivingEntity::get_movement` rather than `Entity::velocity`, mirroring vanilla's `getKnownMovement()`: a player's movement is client driven, so their `velocity` field stays zero and gating on it never fires.

## Scope

Implemented as one routine rather than a general data-driven effect engine, which would need a new codegen variant plus the whole entity predicate system. This follows how Frost Walker is handled today in `block/blocks/magma.rs`.

The cosmetic `tick` effects are player-only, since the periodic tick counter vanilla reads lives on `Player`. Mobs can roll Soul Speed boots and do get the speed boost, they just do not emit the particles.

## AI assistance

This PR was developed with AI assistance (Claude Code, model: claude-opus-5).

## Testing

- `cargo clippy --all-targets` clean, `cargo build` clean, `cargo test` passing
- New unit test `soul_speed_bonus_matches_vanilla_levels` covers the per-level value
- Verified in game on Java: faster on soul sand and soul soil instead of slower, boost held through jumps, dropped on stepping off, absent while flying, boots wearing down, and soul particles plus the soul escape sound while running

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -3100,7 +3100,7 @@ impl LivingEntity {
         None
     }
 
-/// Applies the Soul Speed boots enchantment, mirroring the `minecraft:location_changed`
+    /// Applies the Soul Speed boots enchantment, mirroring the `minecraft:location_changed`
     /// effects of `minecraft:soul_speed`: a flat movement speed bonus plus full movement
     /// efficiency, which cancels the slowdown of the soul block underneath.
     ///
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -3100,12 +3100,23 @@ impl LivingEntity {
         None
     }
 
-    /// Applies the Soul Speed boots enchantment, mirroring the `minecraft:location_changed`
+/// Applies the Soul Speed boots enchantment, mirroring the `minecraft:location_changed`
     /// effects of `minecraft:soul_speed`: a flat movement speed bonus plus full movement
     /// efficiency, which cancels the slowdown of the soul block underneath.
     ///
     /// The boost is kept while airborne (vanilla's `enchantment_active_check`) so jumping
     /// does not drop it, and the boots wear down while actually walking on the block.
+    ///
+    /// The `level == 0` early-out / `active`-modifier bookkeeping pattern here is generic to
+    /// any "boot enchantment that reacts to a block/fluid under the entity" effect — Frost
+    /// Walker (`minecraft:frost_walker`, freezing water into frosted ice) could recycle this
+    /// same shape: read the boot level, gate the per-tick world lookup behind `level > 0`,
+    /// but still read the existing `active` state unconditionally so the effect gets torn
+    /// down cleanly when the boots come off. One extra optimization specific to Frost Walker:
+    /// it only ever interacts with water, which has no placeable source blocks in the Nether,
+    /// so a Frost Walker tick could skip its block/fluid lookup entirely whenever the entity's
+    /// world is a Nether dimension — saving that lookup for every Frost-Walker-booted entity
+    /// there, the same way `level > 0` already saves it here for entities without the enchant.
     fn tick_soul_speed(&self, caller: &dyn EntityBase) {
         let level = self
             .entity_equipment
@@ -3130,15 +3141,13 @@ impl LivingEntity {
                     .map(|modifier| modifier.amount)
             });
 
-        // No enchantment and no leftover modifier: skip the per-tick world lookup that
-        // would otherwise run for every loaded entity. When the boots are removed while
-        // the boost is active, `level` is 0 but `active` is still set, so we fall through
-        // and clean up the modifier on the same tick.
+        // Skip lookup for entities with no enchantment and no lingering
+        // modifier (e.g. boots were just taken off)
         if level == 0 && active.is_none() {
             return;
         }
 
-        let on_soul_speed_block = {
+        let on_soul_speed_block = level > 0 && {
             let world = self.entity.world.load();
             let pos = self.entity.get_block_pos_below_that_affects_my_movement();
             world
@@ -3161,10 +3170,7 @@ impl LivingEntity {
             .get_player()
             .is_some_and(super::player::Player::is_flying);
 
-        // Only a still-enchanted player emits the cosmetic particles/sound.
-        if level > 0 {
-            self.tick_soul_speed_effects(caller, on_soul_speed_block, on_ground, flying);
-        }
+        self.tick_soul_speed_effects(caller, on_soul_speed_block, on_ground, flying);
 
         let applies = level > 0
             && !flying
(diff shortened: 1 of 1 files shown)


crates/pumpkin/src/entity/living.rs
```diff
@@ -3095,6 +3099,168 @@ impl LivingEntity {
...
+            .map_or(0, |boots| {
+                boots.get_enchantment_level(&Enchantment::SOUL_SPEED)
+            });
+
+        let on_soul_speed_block = {
+            let world = self.entity.world.load();
+            let pos = self.entity.get_block_pos_below_that_affects_my_movement();
+            world
+                .get_block(&pos)
+                .has_tag(&tag::Block::MINECRAFT_SOUL_SPEED_BLOCKS)
```
zinedinp, line 3155: good that you used the general array

crates/pumpkin/src/entity/living.rs
```diff
@@ -3095,6 +3099,168 @@ impl LivingEntity {
...
+            .entity_equipment
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner)
+            .equipment
+            .get(&EquipmentSlot::FEET)
+            .map_or(0, |boots| {
+                boots.get_enchantment_level(&Enchantment::SOUL_SPEED)
+            });
+
+        let on_soul_speed_block = {
```
zinedinp: if level == 0 {
    return;
}
let on_soul_speed_block = { ... };

Skip afterwards if notsoul speed enchantment is detected, or it would check every loaded entity per tick

crates/pumpkin/src/entity/living.rs
```diff
@@ -3095,6 +3099,168 @@ impl LivingEntity {
...
+    /// Only players are covered; the periodic tick counter vanilla reads lives on `Player`.
+    fn tick_soul_speed_effects(
+        &self,
+        caller: &dyn EntityBase,
+        level: i32,
+        on_soul_speed_block: bool,
+        on_ground: bool,
+        flying: bool,
+    ) {
+        if level == 0 || !on_soul_speed_block || !on_ground || flying {
```
zinedinp: leverl == 0 was checked before in this chain, not needed

crates/pumpkin/src/entity/living.rs
```diff
@@ -3095,6 +3099,168 @@ impl LivingEntity {
...
+            player.damage_item_in_slot(&EquipmentSlot::FEET, 1);
+        }
+
+        let flying = caller
+            .get_player()
+            .is_some_and(super::player::Player::is_flying);
+
+        self.tick_soul_speed_effects(caller, level, on_soul_speed_block, on_ground, flying);
+
+        let active = self
```
zinedinp: here too the opposite check if level != 0 or it would go frough this without soulspeed enchantment

crates/pumpkin/src/entity/living.rs
ydw1904: Done, implemented all three — early `if level == 0 { return; }`, and removed the now-redundant `level == 0` check (plus the unused `level` param) from `tick_soul_speed_effects`.

One thing I want to sanity-check with you though: `tick_soul_speed` is currently the only place that removes the movement-speed/efficiency modifier. With the early return, if a player takes the boots off while the boost is active (e.g. standing on soul sand, then unequip), `level` becomes 0 and we return before the removal path runs, so the modifier would stick around with no boots on. Does that match how you want it handled, or should the removal happen elsewhere (e.g. on equipment change)? Happy to implement whichever way you prefer.

zinedinp: yes, the modifier does need to get cleaned up on the same tick the boots come off.

ydw1904: Got it, fixed in d72f94e. The early return now only fires when `level == 0 && active.is_none()`, so taking the boots off while the boost is active falls through and removes the modifier on that same tick. The cosmetic particles/sound are gated behind `level > 0` so they stop immediately on unequip.

zinedinp: oh you also did that xd. https://github.com/ydw1904/Pumpkin/pull/1

zinedinp: i see you did it a bit diffrent, i think mine is a bit cleaner with  let on_soul_speed_block = level > 0 && {}

But I cant tell right now

ydw1904: yea yours is cleaner, merged that instead. guess claude isnt that good in writing rust yet

zinedinp: i woudnt say that, but the larger a project or even file gets, the more noise ai will implement. If you want to use an AI efficently, you need to get a feel for machine sympathy and logical gates to reduce lookup, cache misses and so on. And AI will rather implement schoolbook functions and easy gateing, thats where the user should step in and see these small things.

ydw1904: agree, good to have more skilled people reviewing the code made by lesser skilled people with large amounts of tokens (like me with access to all frontier models) haha

hope this gets merged soon though its understandable that the core maintainers are working on the release first instead of fixing smaller bugs/implementations like this

zinedinp: I think Alex tries to hold the release for friday and will not read merges from non contributers

PR #3183, opened 2026-09-01
chore: replace unneeded internal block property serialization

PigTurtleDev:
## Description
Replaces cases of properties being modified in a serialized `Vec<(str, str)>` form with the actual Property types wherever feasible. As such, this PR tries to remove uses of `BlockProperties::to_props(&self) -> Vec<(&'static str, &'static str)>`

## Changes
Use the appropiate property type when...
- placing wall coral fans in `CoralFeature::generate_coral_piece`
- making Grass (or Podzol/Mycelium) snowy in `FreezeTopLayerFeature::generate`
- Rotating logs (`CherryTrunkPlacer::get_sideways_state`, `FancyTrunkPlacer::make_or_check_branch`)
- Brushing suspicious_{sand, gravel}
- checking & inserting an ender eye into an end portal frame
- checking cauldron levels while filling a waterbottle
- shearing a bee_{nest, hive}
- shearing (stopping growth of) a twisting/weeping vine, kelp or glowberry.
  - removes `Block::CAVE_VINES_PLANT` from the function because it represents the glowberry blocks above the growing head. They can't grow and didn't have a settable `age` property to begin with.

## Non-Changes
- Signs:
  - there is currently one BlockBehaviour handling all sign types (signs, wall_signs, hanging signs, hanging_wall_signs).
  - this should probably be split across multiple handlers; They don't all share the same properties.
- Ignition:
  - currently handles campfires, candles, candle_cakes at once.
- Actual serialization (for plugins, file storage, copying properties into an item, checking if a properties are a superset of some properties):
  - obviously needs to keep using `.to_props()`
- Waterlogging
  - the property commonly has to be checked on an arbitrary block & is too common to list all Property types that may have it.
- `pumpkin_world`'s `BlockStateCodec`
  - this one bothers me. I would prefer to have the BlockStates always be computed ahead of time, but the struct is data-generated into statics. Resolving a `BlockState` from `String` properties during compile time isn't possible for a plethora of reasons, including no const Trait and no const String equality.

While testing this PR I encountered the following issues. They are left untouched.
- brushing requires clicking the block 4 times instead of holding for a certain duration
- filling bottles with water by clicking a waterlogged block does not work
- filling bottles with water does not properly ray-cast for fluids. 
  - When looking through a water source but not at a block, the client plays a sound and swings the hand, but the bottle does not get filled.

## Testing
- [x] these changes behave the same in game (I could not test beehives because `/setblock` doesn't currently work with properties).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interactions with water cauldrons, end portal frames, and flammable blocks.
  * Corrected shears behavior for kelp, cave vines, bee nests, beehives, and pumpkin harvesting.
  * Improved suspicious sand and gravel brushing interactions.
  * Improved world generation consistency for coral, frozen surfaces, snow, and tree growth.
* **Refactor**
  * Updated block interactions and generated block states for more consistent behavior.


RoosterBooster007:

crates/pumpkin-world/src/generation/feature/features/freeze_top_layer.rs
```diff
@@ -28,14 +29,13 @@ impl FreezeTopLayerFeature {
 
                 let top_vec = BlockPos::new(x, y, z).0;
                 let below_vec = BlockPos::new(x, below_y, z).0;
+                let below = GenerationCache::get_block_state(chunk, &below_vec);
+                let below_block = below.to_block_id();
```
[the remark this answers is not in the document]
PigTurtleDev: Neither ice nor water have a snowy property, so this is not an issue. (`GrassLikeProperties::handles_block_id` returns `false`for both)

crates/pumpkin-world/src/generation/feature/features/freeze_top_layer.rs
```diff
@@ -48,31 +48,20 @@ impl FreezeTopLayerFeature {
...
                     let top_raw = GenerationCache::get_block_state(chunk, &top_vec);
                     // Re-read below (may have been replaced by ice in the freeze step above)
-                    let below = GenerationCache::get_block_state(chunk, &below_vec);
 
                     // topPos must be air; belowPos must not be air (something to stand on)
                     if top_raw.to_state().is_air()
                         && !below.to_state().is_air()
-                        && !Block::from_state_id(below.to_state().id)
-                            .has_tag(&tag::Block::MINECRAFT_CANNOT_SUPPORT_SNOW_LAYER)
+                        && !below_block.has_tag(tag::Block::MINECRAFT_CANNOT_SUPPORT_SNOW_LAYER)
```
RoosterBooster007, line 54: It's not the snowy bit, it's `!below_block.has_tag(CANNOT_SUPPORT_SNOW_LAYER)`. `ice` has that tag, `water` doesn't, so this cached pre-freeze read drops snow on freshly-formed ice where vanilla wouldn't. Re-read `below` after the freeze (what line 50's comment already says) and this is good to merge.

PigTurtleDev: Thanks for pointing that out, I've opted to skip the snow checks altogether if the block has been replaced with ice. The snow logic seems to still be incomplete though, just checking if a block isn't air is not enough.

RoosterBooster007: Ok, that's fine. You're right. @coderabbitai, can you file an issue about the incomplete snow logic (looks like it needs `isFaceFull`, `SUPPORT_OVERRIDE_SNOW_LAYER`, and a `brightness < 10` gate still)?

RoosterBooster007:
Thanks, @PigTurtleDev!

PR #3184, opened 2026-09-01
fix(block): preserve light block level during placement

AdmerPRO:
## What was changed?

Network serialization and deserialization support was added for the `minecraft:block_state` data component. The component now reads and writes its block property names and values, with validation limiting the number of properties accepted from a packet.

Block placement now receives the complete item stack on both Java and Bedrock paths. After the block-specific placement logic calculates the initial state, valid properties from the item's `BlockState` component are applied while preserving the remaining placement properties.

As a result, light blocks retain their selected `level` value when placed instead of always using the default value of 15. Unknown properties and values that do not correspond to a valid state for the placed block are ignored.

Unit tests were added for `BlockState` component network round-tripping, excessive property-count rejection, applying a valid light level, and ignoring an invalid light level.

Light blocks with a level other than 15 contain a `minecraft:block_state` component describing their selected light level. The protocol codec did not support this component, so moving such an item into a creative inventory slot caused packet handling to fail with `component_id_76 (TODO)` and disconnected the player.

Even after decoding the component, the placement flow only passed the block type to the block registry. It discarded the item stack and therefore had no access to the stored `level` property. The light block was consequently placed using its default state, whose level is 15.

Passing the item stack through the placement flow and applying its valid block-state properties makes the placed state match the item while retaining properties calculated from placement context, such as orientation or waterlogging.

- Java clients can move light blocks with non-default levels into creative inventory slots without being disconnected by the previously unsupported component.
- Light blocks are placed with the level selected in their item data.
- Bedrock and Java block placement use the same component-aware placement path.
- Other blocks carrying a `BlockState` component can also preserve valid item-defined properties during placement.
- Invalid or unknown properties do not create invalid block states and fall back to the state produced by normal placement logic.

- No full in-game integration test was added; light levels should still be manually verified from 0 through 15.
- Only properties represented by valid states of the placed block are applied.
- Unsupported data component types other than `BlockState` may still produce their existing `TODO` codec errors.
- The changes were checked with `cargo check -p pumpkin`, focused unit tests, `cargo fmt --all -- --check`, and `git diff --check`.


AdmerPRO:
Omg i dont gonna play with some merge conflicts i just gonna code it again

PR #3155, opened 2026-08-30
chore(plugins): Update to Wasmtime 48.0.1

QuantumSegfault:
## Description

Updates Wasmtime to latest (48).

In accordance with with some changes in Wasmtime, the pre-#2643 behavior of `fs.write.data` implying `fs.read.data` has been restored.

## Testing

`pumpkin-fly-plugin` was verified to still work with the new Wasmtime version.

-----

Based on part of e21d5bf2c756a759bc6d62193a9ef4c5cc5a420d by @Phoenixxo

PR #3185, opened 2026-09-02
feat(plugin): support synchronous WASM reentry

Phoenixxo:
## Summary:

Adds synchronous WASM reentry support so call chains such as `A -> host -> B -> host -> A` can complete without deadlocking or spreading async throughout Pumpkin's game logic.

Each Wasmtime Store is managed through a dedicated runtime. Independent plugin root calls are serialized, while nested calls inherit the active causal chain and can re-enter the same or another plugin Store.

This is intentionally a correctness-first implementation. Plugin roots do not execute in parallel yet, and the calling Rayon worker waits for the synchronous chain to finish.

Major thanks to @QuantumSegfault for the discussion and help in coming up with this and their reentry experiments, credit to their work here: 

- https://github.com/QuantumSegfault/wit-plugin-versioning-experiments/commit/80c08e996d476f4c6e5370159100965fbb30604b
- https://github.com/QuantumSegfault/wit-plugin-versioning-experiments/commit/12a405da9b2f292a506e16a71e3a2b5e7b64e60d
- https://github.com/QuantumSegfault/wit-plugin-versioning-experiments/commit/e231e7447b18e648b32ac65973b5d83da10c55d9 
- https://github.com/QuantumSegfault/wit-plugin-versioning-experiments/commit/7182c7305a177174a693df93e96365e870648313 
- https://github.com/QuantumSegfault/wit-plugin-versioning-experiments/commit/8f8a530547b36546ea37ac583881a48dfcfaa2a4 

## Changes:

- Added the `pumpkin-plugin-runtime` crate for Store execution, admission control, reentry tracking, spawning, shutdown, and terminal-state handling.
- Added a shared FIFO admission policy for independent plugin root calls.
- Added causal-chain tracking so nested host and plugin calls inherit the active admission.
- Limited synchronous reentry to a maximum depth of 64.
- Added tracing for admission IDs, chain IDs, reentry depth, wait time, and execution time.
- Routed guest entry points and reentry-sensitive WIT host functions through the new Store runtime.
- Added separate paths for synchronous host work and host futures that may call back into plugins.
- Updated events, commands, IPC, scheduling, entities, players, servers, worlds, datapacks, and world generation to use the reentry-aware Store API.
- Changed Rust plugin lifecycle and IPC callbacks to use shared references.
- Changed callback handlers to require `Send + Sync` so registry locks can be released before plugin code is invoked.
- Documented that mutable plugin state requires thread-safe interior mutability and that plugins should not hold non-reentrant locks across host calls.
- Added plugin unload cleanup for event handlers, commands, aliases, and scheduled tasks.
- Added command reactivation when a plugin reloads while preserving commands disabled through server configuration.
- Added explicit TCP connect, bind, listen, UDP send, and UDP receive permission handling.
- Allowed only Wasmtime's implicit outbound wildcard binds without granting explicit bind, listen, or receive access.
- Made data-folder permissions fail closed when write access is requested without read access.
- Pinned Wasmtime to the official upstream commit containing Component Model synchronous reentry support.
- Updated Cargo and Nix dependency metadata for the Wasmtime revision.
- Updated the implementation for Pumpkin's current WIT submodule and node-based command system.

## Validation

- `cargo check -p pumpkin --locked`
- `cargo test -p pumpkin-plugin-runtime --locked --lib -- --test-threads=1`
  - 4 passed
- `cargo test -p pumpkin --locked --lib plugin::loader::wasm::wasm_host::tests -- --test-threads=1`
  - 6 passed
- Plugin command unload, alias cleanup, reload, and configuration-disable regression test passed.
- `cargo fmt --all --check`
- `cargo clippy -p pumpkin --all-targets --all-features`
- Locked Cargo metadata validation passed.
- A timed 10-root cross-plugin reentry run completed 10/10 chains with no deadlocks.
- The Nix dependency hash was verified locally, with the complete Nix build left to CI.
- Real end-to-end TCP and UDP socket testing with a plugin has not been completed yet.

## Work that remains untested/undone:

The correctness work for this implementation is complete. The remaining work is validation and future optimization:

- Repeat in-game reentry UAT using the final rebased branch and official Wasmtime revision.
- Run an end-to-end plugin test that opens real TCP and UDP sockets.
- Measure performance with more players, multiple plugins, and high-frequency events.
- Monitor Wasmtime releases and replace the temporary Git revision once the required reentry support is available in a stable release.
- Investigate safely allowing unrelated plugin roots to execute concurrently.
- Explore longer-term scheduler changes that could prevent Rayon workers from waiting on plugin chains.


crates/pumpkin/src/block/entities/brewing_stand.rs
```diff
@@ -227,6 +227,49 @@ impl BrewingStandBlockEntity {
         // Mark dirty to trigger update
         self.mark_dirty();
     }
+
```
Phoenixxo, line 233: This probably looks unrelated at first, but the old fuel path held the brewing stand inventory lock while firing BrewingStandFuelEvent. Since plugins can now synchronously call back into Pumpkin, a handler touching the same inventory would deadlock. This snapshots the fuel, releases the lock, fires the event, then locks again and validates the item before consuming it.

crates/pumpkin/src/entity/item.rs
```diff
@@ -37,6 +37,46 @@ pub struct ItemEntity {
...
+            low.merge_reserved.store(false, Ordering::Release);
+            return None;
+        }
+
+        Some(Self {
+            low: &low.merge_reserved,
+            high: &high.merge_reserved,
+        })
+    }
+}
```
Phoenixxo, line 73: This reservation lets us release both item stack locks while ItemMergeEvent is running. We need the reservation so another merge cannot take over while those locks are released, then we validate both stacks afterwards. This allows us to keep merging items without needing locks in plugin code.

crates/pumpkin/src/entity/player/advancement.rs
```diff
@@ -214,6 +214,31 @@ pub struct PlayerAdvancement {
...
+    }
+
+    pub(crate) const fn combine(self, other: Self) -> Self {
+        Self {
+            awarded: self.awarded || other.awarded,
+            completed: self.completed || other.completed,
+        }
+    }
+}
+
```
Phoenixxo, line 241: The advancement changes are for the same lock boundary issue as seen in [item.rs](https://github.com/Pumpkin-MC/Pumpkin/pull/3185/changes#diff-2f1c0cc5fc86b104a36430885ff3d12b62db91363155ce1c6ae0d7ec945a3d15). Previously the completion event and rewards ran while the player advancement mutex was still held. AdvancementAward carries the result outside of that lock so plugin callbacks can safely reenter advancement APIs.

crates/pumpkin/src/command/commands/advancement.rs
Phoenixxo, line 1: These command changes are mostly plumbing for the advancement lock split. We update the advancement state while holding the lock, release it, then run finish_award afterward. The intended command and advancement behavior stays the same.

crates/pumpkin/src/data/datapack/mod.rs
Phoenixxo, line 1: Changes here are because the function lines are copied while the datapack registry lock is held, then executed after releasing it. A function command can reach plugin code, which can then call back into the datapack API. Keeping the registry lock during command execution would make that reentry unsafe.

crates/pumpkin-plugin-api/src/events/mod.rs
Phoenixxo, line 1: The Arc handler changes let the host clone a callback out of the registry and release the registry mutex before invoking it. It removes the exclusive registry borrow that would otherwise prevent nested plugin calls. The same reasoning applies to the AI, command, scheduler, and worldgen handler registries.

crates/pumpkin/src/server/scheduler.rs
Phoenixxo, line 1: Scheduled plugin calls now enter through `call_guest` instead of locking the Wasmtime Store directly. This makes them participate in the same admission, reentry, depth, and shutdown handling as every other plugin entry point.

Also the disabled plugin tracking prevents queued or repeating tasks from calling into a Store that is being unloaded. It also stops a repeating task from scheduling itself again after unload has started.

crates/pumpkin/src/plugin/mod.rs
Phoenixxo, line 1: The reason for changes here is due to event handlers and commands needing to be disabled before the plugin store is torn down. Otherwise those entry points could retain and invoke callbacks for an unloaded plugin. Commands are marked inactive instead of deleting nodes because aliases can redirect to those nodes, then they are reactivated when the plugin registers again.

crates/pumpkin/src/plugin/permissions.rs
Phoenixxo, line 1: This is related to integrating the newer Wasmtime runtime while keeping permissions fail-closed. Wasmtime cannot currently provide a truly write-only preopened directory, so instead we require read permission alongside write permission instead of silently granting extra access.

Cargo.toml
Phoenixxo, line 1: The toolchain, lockfile, and Nix changes come from pinning the upstream Wasmtime commit containing Component Model synchronous reentry. That Wasmtime revision requires Rust 1.96, and Nix needs an explicit hash for the new git dependency.

Phoenixxo:
I left some comments to provide context to some of the changes that seemed a bit less obvious at first

crates/pumpkin-plugin-api/src/events/mod.rs
QuantumSegfault: Regarding all changes to `pumpkin-plugin-api`, they are currently unnecessary.

At this time, this crate is only used by Wasm plugins (which as we both know very well, are STRICTLY single-threaded). Native plugins link against `pumpkin` (and `pumpkin-macros`?) directly.

In the future #2828 (or perhaps CM async) may make this relavent. Until we then we are simply adding extra Arc and Mutex overhead for no reason at this time.

Cargo.toml
```diff
@@ -217,9 +219,11 @@ rustc_version_runtime = { version = "0.3", default-features = false }
 ordered-float = { version = "5.4", default-features = false, features = ["std"] }
 xxhash-rust = { version = "0.8", default-features = false, features = ["xxh64"] }
 
-wasmtime = { version = "48.0", default-features = false, features = ["runtime", "component-model", "async", "cache", "cranelift", "gc", "gc-drc", "threads", "std"] }
-wasmtime-wasi = { version = "48.0", default-features = false, features = ["p2", "p3"] }
```
QuantumSegfault, line 219: Maybe bump to https://github.com/bytecodealliance/wasmtime/commit/f6b3140dbb1d9b436ef523d14740292c956b0364 since it had made a couple minor tweaks to relevant things.

Just to be sure.

Cargo.toml
```diff
@@ -48,7 +49,6 @@ needless_pass_by_ref_mut = "deny"
 needless_collect = "deny"
 redundant_clone = "deny"
 set_contains_or_insert = "deny"
-significant_drop_in_scrutinee = "deny"
```
QuantumSegfault, line 52: Why?

crates/pumpkin/src/plugin/permissions.rs
QuantumSegfault: Please revert this file. These comments are now outdated. `fs.write.data` now implies `fs.read.data`, as it used once upon a time.

crates/pumpkin/src/block/entities/brewing_stand.rs
```diff
@@ -227,6 +227,49 @@ impl BrewingStandBlockEntity {
...
+                .get_item()
+                .has_tag(&tag::Item::MINECRAFT_BREWING_FUEL)
+        {
+            Some(items[4].clone())
+        } else {
+            None
+        };
+        let Some(expected_fuel) = expected_fuel else {
+            return false;
+        };
```
QuantumSegfault, line 245:
```suggestion
        {
            items[4].clone()
        } else {
            return false;
        };
```

crates/pumpkin/src/block/entities/brewing_stand.rs
```diff
@@ -227,6 +227,49 @@ impl BrewingStandBlockEntity {
         // Mark dirty to trigger update
         self.mark_dirty();
     }
+
+    fn try_refill_fuel(&self, world: &Arc<crate::world::World>) -> bool {
+        let expected_fuel = if self.fuel.load(Ordering::Relaxed) <= 0
+            && let Ok(items) = self.items.try_write()
```
QuantumSegfault: Couldn't this be a read lock instead?

QuantumSegfault: Couldn't this be a read lock instead?

crates/pumpkin/src/block/entities/brewing_stand.rs
```diff
@@ -227,6 +227,49 @@ impl BrewingStandBlockEntity {
...
+        }
+
+        if self.fuel.load(Ordering::Relaxed) <= 0
+            && let Ok(mut items) = self.items.try_write()
+            && !items[4].is_empty()
+            && items[4].uid == expected_fuel.uid
+            && items[4].are_equal(&expected_fuel)
+            && items[4]
+                .get_item()
+                .has_tag(&tag::Item::MINECRAFT_BREWING_FUEL)
```
QuantumSegfault:
```suggestion
```

Redundant. If the item is identical, it will still be tagged by way of the initial check.

crates/pumpkin/src/block/entities/brewing_stand.rs
```diff
@@ -227,6 +227,49 @@ impl BrewingStandBlockEntity {
...
+            );
+            server
+                .plugin_manager
+                .fire_blocking(&server, &mut fuel_event);
+        }
+
+        if self.fuel.load(Ordering::Relaxed) <= 0
+            && let Ok(mut items) = self.items.try_write()
+            && !items[4].is_empty()
+            && items[4].uid == expected_fuel.uid
```
QuantumSegfault: Does a change of UID really matter? If we swap with a new but otherwise identical stack...should we really consider that a change at this level?

crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
```diff
@@ -112,22 +113,40 @@ pub struct PluginRuntime {
...
+
+const fn data_fs_perms(can_read: bool, can_write: bool) -> Result<Option<FsPerms>, &'static str> {
+    match (can_read, can_write) {
+        (false, false) => Ok(None),
+        (true, false) => Ok(Some(FsPerms::ReadOnly)),
+        (true, true) => Ok(Some(FsPerms::ReadWrite)),
+        (false, true) => Err(
+            "`fs.write.data` without `fs.read.data` cannot be enforced by this Wasmtime version; refusing to grant broader filesystem access",
+        ),
+    }
```
QuantumSegfault:
```suggestion
```

Revert. Handled in #3155 (write implies read).

crates/pumpkin/src/command/node/dispatcher.rs
```diff
@@ -105,34 +110,77 @@ impl CommandDispatcher {
...
+            inactive_plugin_commands: FxHashSet::default(),
         }
     }
 
     /// Turns a command off. A disabled command's primary name is recorded here
     /// so that it can no longer be executed, listed, or suggested, regardless of
     /// which internal dispatcher it lives on.
     pub fn disable_command(&mut self, name: impl Into<String>) {
-        self.disabled.insert(name.into());
+        self.disabled.insert(name.into().to_ascii_lowercase());
```
QuantumSegfault: Why the case-sensitivity change? This should probably be addressed more thoroughly (normalize command names on register). Also in a separate PR as this is seemingly unrelated.

crates/pumpkin/src/data/datapack/mod.rs
```diff
@@ -238,25 +251,28 @@ impl DatapackManager {
...
+                .unwrap_or_else(std::sync::PoisonError::into_inner);
+            let mut functions = Vec::with_capacity(functions_to_run.len());
+            for fn_id in functions_to_run {
+                let Some(lines) = all_fns.get(&fn_id) else {
+                    if !is_tag {
+                        return Err(format!("Unknown function: {fn_id}"));
+                    }
+                    continue;
+                };
+                functions.push(lines.clone());
```
QuantumSegfault: Can't say this is ideal...having to clone/allocate a new mega function each time we need to execute.

Though I don't see much of a way good way around it. While still maintaining atomicity (not changing what's to be executed while executing it).

QuantumSegfault (commented):
First wave. There's a LOT to chew through here.

Another note: maybe split all the command and plugin disabling related things into a separate PR so it is easier to digest.

Cargo.toml
Phoenixxo: A result of rust 1.96 for the new Wasmtime upgrade. There were a bunch of failing passes in cargo clippy from files I did not touch, this is just a temporary thing to keep it away for now, but i do agree that they do need to be addressed somehow soon

Cargo.toml
Phoenixxo: Will do

Phoenixxo:
@QuantumSegfault Regarding the plugin API:

These changes aren’t for running the Wasm plugin in parallel. They’re for nested reentry on the same thread.
Previously handle_event held the handler registry while running plugin code. If that handler called Pumpkin and caused another event to enter the same plugin, the nested call tried to acquire the same registry lock before the outer call returned.

The Arc here lets us clone the handler out and release the registry lock before invoking it. The &self changes also avoid creating another &mut self while the outer callback is still active. Just reverting these changes would restore that reentry issue, which is why I made it. We can split it out though possibly or find a better solution hopefully soon

As for permissions, I missed that from PR #3155 so i'll fix

crates/pumpkin/src/block/entities/brewing_stand.rs
Phoenixxo: Yeah, I’ll switch it to `try_read()`.

crates/pumpkin/src/block/entities/brewing_stand.rs
Phoenixxo: This was something I missed originally when splitting it out, i'll simplify it

crates/pumpkin/src/block/entities/brewing_stand.rs
Phoenixxo: Will fix

crates/pumpkin/src/block/entities/brewing_stand.rs
Phoenixxo: It probably doesn't but it was more to be safe, we can just probably use `are_equal` instead

crates/pumpkin/src/data/datapack/mod.rs
Phoenixxo: Yeah, this was the best I came up with at the time unfortunately.

However I think I can store function bodies behind Arc<String> and only clone the Arc before releasing the registry lock, that might be an idea?

crates/pumpkin/src/command/node/dispatcher.rs
Phoenixxo: I added the lowercase conversion so unloaded commands could be found consistently, but it only normalizes the disabled set while the command tree still stores the original names. Can def be put in a different PR

QuantumSegfault:
> Regarding the plugin API:

Sorry, I missed the fact that for some reason, the API already uses `Mutex`, despite single threading. Then I guess the Rust API specifically already accounts for (not possible) threading. In which case these changes make sense to get it to work given the current changes. My bad.

Phoenixxo:
> > Regarding the plugin API:
> 
> Sorry, I missed the fact that for some reason, the API already uses `Mutex`, despite single threading. Then I guess the Rust API specifically already accounts for (not possible) threading. In which case these changes make sense to get it to work given the current changes. My bad.

Would you like me to still split out the changes?

QuantumSegfault:
> Would you like me to still split out the changes?

No, they logically fit with the rest of the PR.

Still think disabling could/should be though.

crates/pumpkin/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 #![deny(clippy::unwrap_used)]
 #![cfg_attr(test, allow(clippy::unwrap_used, clippy::expect_used, clippy::panic))]
+#![allow(clippy::significant_drop_in_scrutinee)]
```
QuantumSegfault, line 3: Seems rather blanket. Is there any more specific spot where it would be more appropriate?

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/ipc.rs
```diff
@@ -5,22 +5,37 @@ use crate::plugin::loader::wasm::wasm_host::{
...
+        mut host: Access<'_, PluginHostState, Self>,
         recipient: PluginId,
         message: IpcMessage,
     ) -> wasmtime::Result<Result<Result<IpcMessage, String>, ()>> {
-        let Some(server) = &self.server else {
-            return Err(wasmtime::Error::msg("Server not available"));
-        };
-        let Some(name) = &self.name else {
-            return Err(wasmtime::Error::msg("Plugin name not available"));
+        let (server, name, plugin) = {
```
QuantumSegfault, line 18: Why the extra clones of `server` and `name`?

`name` especially. Name never changes after the plugin is loaded.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/context.rs
```diff
@@ -1845,3 +1820,39 @@ impl pumpkin::plugin::context::HostContext for PluginHostState {
         Ok(())
     }
 }
+
+impl pumpkin::plugin::context::HostContextWithStore<PluginHostState> for HasSelf<PluginHostState> {
+    async fn register_command(
```
QuantumSegfault: Why does this need re-entrancy protection?

I see the packet send, but that's just added to the queue.

Is there some event fire I'm not seeing?

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/commands/mod.rs
```diff
@@ -569,6 +555,38 @@ impl pumpkin::plugin::command::HostCommandSender for PluginHostState {
     }
 }
 
+impl pumpkin::plugin::command::HostCommandSenderWithStore<PluginHostState>
```
QuantumSegfault, line 558: Same here. I'm not seeing the possibility for re-entry?

crates/pumpkin/src/entity/item.rs
```diff
@@ -185,37 +285,32 @@ impl ItemEntity {
             .item_stack
             .lock()
             .unwrap_or_else(std::sync::PoisonError::into_inner);
+        if low_stack.uid != expected_low.uid
+            || !low_stack.are_equal(&expected_low)
+            || high_stack.uid != expected_high.uid
+            || !high_stack.are_equal(&expected_high)
+            || self.entity.removed.load(Ordering::SeqCst)
+            || other.entity.removed.load(Ordering::SeqCst)
```
QuantumSegfault, line 293:
```suggestion
```

Redundant. You already do the same just before acquiring the stacks again (same `if` with `event.cancelled`)

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/mod.rs
```diff
@@ -102,23 +165,33 @@ pub fn prepare_plugin(
...
+                .await
+        })
         .await
+        .map_err(PluginInitError::CallInitPluginFailed)?
         .map_err(PluginInitError::CallInitPluginFailed)?;
 
-    let metadata = plugin
-        .pumpkin_plugin_metadata()
-        .call_get_metadata(&mut store)
+    let metadata = store
```
QuantumSegfault, line 186: Might not be necessary? Anybody doing expensive re-entrant things during `get-metadata` is entirely unexpected. I suppose it's consistent this way at least.

crates/pumpkin/src/plugin/mod.rs
```diff
@@ -704,9 +704,9 @@ impl PluginManager {
             }
 
             // Find a loader that can handle this file
-            let loaders = self.loaders.read().await;
+            let loaders = self.loaders.read().await.clone();
```
QuantumSegfault: Another place we should perhaps reconsider cloning all loaders (`Vec<Arc<_>>`) every time you want to load a new plugin.

QuantumSegfault (commented):
30/44

Another batch of comments.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/world.rs
```diff
@@ -2218,39 +2295,58 @@ impl WasmChunkGenerator {
...
+        let x = proto_chunk.x;
+        let z = proto_chunk.z;
+        let min_y = proto_chunk.bottom_y() as i32;
+        let height = proto_chunk.height() as u32;
+        let placeholder_generator =
+            pumpkin_world::generation::generator::WorldGenerator::Custom(Arc::new(self.clone()));
+        let placeholder = pumpkin_world::ProtoChunk::new(x, z, &placeholder_generator);
+        let owned_proto_chunk = std::mem::replace(proto_chunk, placeholder);
+        let shared_proto_chunk = Arc::new(StdMutex::new(owned_proto_chunk));
+        let _restore = ProtoChunkRestore {
```
QuantumSegfault, line 2281: Tagging this for review by somebody else

e.g. @Snowiiii 

I have no sense how this will interact with the rest of the chunk system

crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
```diff
@@ -316,13 +350,7 @@ impl WasmPlugin {
             Box::pin(async move { socket_policy.allows(addr, reason) })
         });
 
-        if has_permission(permissions::NETWORK_OUTBOUND) {
```
QuantumSegfault, line 319: This got lost in the shuffle. Don't know if you intentionally dropped `inherit_network`, and if that will cause problems.

Probably just undo that.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
```diff
@@ -444,3 +513,147 @@ pub trait DowncastResourceExt<E> {
     fn downcast_mut<'a>(&'a self, state: &'a mut PluginHostState) -> &'a mut E;
     fn consume(self, state: &mut PluginHostState) -> E;
 }
+
+#[cfg(test)]
```
QuantumSegfault, line 517: Unrelated...but since I already missed it in the split PR, I'd say probably keep it, I guess.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/events/mod.rs
```diff
@@ -230,37 +230,51 @@ pub(super) fn consume_world(
...
-                        return;
-                    };
-                    let server_rep = server_res.rep();
-                    let result = plugin
-                        .call_handle_event(&mut *store, self.handler_id, server_res, &wasm_event)
-                        .await;
-                    match result {
-                        Ok(returned_event) => {
-                            cleanup_event(&returned_event, store.data_mut());
-                            cleanup_event(&wasm_event, store.data_mut());
```
QuantumSegfault, line 251: Maybe I'm just not seeing it, but it doesn't seem the `wasm_event` is ever cleaned up on the happy path? Only on the error path now.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/events/mod.rs
```diff
@@ -271,33 +285,50 @@ impl<E: Payload + ToFromWasmEvent> EventHandler<E> for WasmPluginEventHandler {
...
-                        return;
-                    };
-                    let server_rep = server_res.rep();
-                    let result = plugin
-                        .call_handle_event(&mut *store, self.handler_id, server_res, &wasm_event)
-                        .await;
-                    match result {
-                        Ok(returned_event) => {
-                            event.apply_wasm_event(returned_event, store.data_mut());
-                            cleanup_event(&wasm_event, store.data_mut());
```
QuantumSegfault, line 289: Ditto. I'd recommend carefully re-reading the both of these and checking that everything is equivalent.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
```diff
@@ -675,6 +584,163 @@ impl pumpkin::plugin::server::HostServer for PluginHostState {
...
+                .as_ref()
+                .and_then(std::sync::Weak::upgrade)
+                .ok_or_else(|| wasmtime::Error::msg("Plugin instance not available"))?;
+            (server, plugin)
+        };
+
+        let runtime = tokio::runtime::Handle::current();
+        plugin
+            .store
+            .pump_blocking(&mut host, move || {
```
QuantumSegfault: Could you not use `pump_reentry` here?

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
```diff
@@ -675,6 +584,163 @@ impl pumpkin::plugin::server::HostServer for PluginHostState {
...
+                .as_ref()
+                .and_then(std::sync::Weak::upgrade)
+                .ok_or_else(|| wasmtime::Error::msg("Plugin instance not available"))?;
+            (server, plugin)
+        };
+
+        let runtime = tokio::runtime::Handle::current();
+        plugin
+            .store
+            .pump_blocking(&mut host, move || runtime.block_on(server.save_all()))
```
QuantumSegfault: Same here? Doesn't `pump_reentry` allow awaiting on async?

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
```diff
@@ -3318,24 +3010,635 @@ impl pumpkin::plugin::player::HostPlayer for PluginHostState {
...
+            (
+                player_from_resource(state, &player)?,
+                state.server.as_ref().expect("server not available").clone(),
+                plugin_from_state(state)?,
+            )
+        };
+
+        plugin
+            .store
+            .pump_blocking(&mut host, move || player.has_permission(&server, &node))
```
QuantumSegfault, line 3094: Why the protection? How can `has_permission` re-enter?

QuantumSegfault (commented):
38/44

Just have `pumpkin-plugin-runtime` to check through now.

crates/pumpkin/src/lib.rs
Phoenixxo: Right now it's there just to prevent clippy from dying, not the right way to handle it. I'm going to eventually open a PR fixing the issues causing this

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/ipc.rs
Phoenixxo: The values need to be owned as we release `host.get()` and mutably borrow the accessor. The server clone is cheap, and as for the name we could use something like `Arc<str>` but it doesn't change much

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/context.rs
Phoenixxo: No this truly just doesn't need the protection, I added it originally thinking it did for commands but it turns out it doesn't, i'll move it back into the regular host implementation

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/commands/mod.rs
Phoenixxo: This needs to stay since it fires `PlayerPermissionCheckEvent` down the line, which can reenter Wasm

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/ipc.rs
QuantumSegfault: Ah. Right... borrow checker. Perhaps `Arc<str>`, but I won't hold us up over it.

crates/pumpkin/src/entity/item.rs
Phoenixxo: the check here is because entity removal can happen while waiting for the stack mutexes, it helps with preventing another merge

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/mod.rs
Phoenixxo: Idea is that it's still guest code, so just want to keep everything under admission under the Store driver starts

crates/pumpkin/src/plugin/mod.rs
Phoenixxo: Certainly, I think better we should save a `Vec<Arc<dyn PluginLoader>>` once then we can scan that way we still release the lock w/o cloning so much

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/events/mod.rs
Phoenixxo: Ah good catch, I missed that, I'll go back and add it onto the success path as well

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/events/mod.rs
Phoenixxo: Will do

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
Phoenixxo: I should here, I can just use `pump_reentry` and await the unload directly

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
Phoenixxo: Same idea here

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
Phoenixxo: This needs to stay since it fires PlayerPermissionCheckEvent down the line, which can reenter Wasm

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/commands/mod.rs
Phoenixxo: This needs to stay since it fires PlayerPermissionCheckEvent down the line, which can reenter Wasm

crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
Phoenixxo: This was due to the new pinned Wasmtime version, inherit_network() replaces the custom socket check with allowing all callbacks (TCP, UDP, etc.) which is what we don't want.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
QuantumSegfault: My bad. Breezed right past that on the call chain.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/context.rs
```diff
@@ -1505,6 +1505,27 @@ impl DowncastResourceExt<ContextResource> for Resource<Context> {
 impl pumpkin::plugin::context::Host for PluginHostState {}
 
 impl pumpkin::plugin::context::HostContext for PluginHostState {
+    async fn register_command(
```
QuantumSegfault: Maybe shift this back down where it was originally so the diff is more compact.

QuantumSegfault (approved):
Just one last nit-pick, otherwise LGTM

-----

@Snowiiii please review and merge ASAP before release.

I reviewed integration points thoroughly, and they looked good to me (after some iterating). There may be more new plugin re-entry spots that need accounting for (namely where events are involved), but the existing modifications look good.

As for `pumpkin-plugin-runtime`, I haven't been quite as rigorous in reviewing it (though it appears reasonably well tested). it isn't the prettiest thing, but it makes the system work NOW so at least we don't get deadlocks when plugins try to do complex things (or simple plugins interact in unforeseen ways).

There are already plans to fix this problem more properly/efficiently. In the meantime, please accept it as is.

crates/pumpkin/src/plugin/permissions.rs
QuantumSegfault: Ping @Phoenixxo

Phoenixxo: already done

Phoenixxo:
@Snowiiii Should pass this time around, issue was just lockfile so it was easy. CI should pass as well because of this

Snowiiii:
LGTM, Thank you @Phoenixxo :D

PR #3186, opened 2026-09-02
fix(worldgen): stop overworld biomes leaking into other dimensions 🤖🤖🤖

ydw1904:
## Description

Addresses the biome half of #2466 (overworld trees generating in the Nether).

`LAST_RESULT_NODE` in `pumpkin-world/src/biome/mod.rs` is a thread-local shortcut holding the last matched biome-tree leaf, but it was shared by every dimension's tree. `BiomeTree::get` seeds its search distance from that leaf and `search` only accepts nodes **strictly closer**, so a leaf cached from the overworld tree can beat every node of the Nether tree and be returned unchanged.

The consequence: a worker thread that generated an overworld chunk hands the next Nether chunk overworld biomes, and with them overworld surface rules, features and mob spawns. On a 6400-chunk Nether scan before the fix, over 5 million biome cells resolved to `minecraft:river`.

- **Key the cache by its source tree**, so a leaf is only ever reused for the tree it came from. This matches vanilla, where `lastResult` lives on the `RTree` itself rather than being shared across all of them.
- **Ocean monument placement uses the current dimension's biome supplier.** Monument placement is a spacing-only check (`is_start_chunk` for `RandomSpread` does no biome filtering), so it runs in every dimension and then validated against a hardcoded `MultiNoiseBiomeSupplier::OVERWORLD`. That both queried the wrong tree and was one of the paths poisoning the cache. Monuments only list overworld biomes, so they are still correctly rejected elsewhere.

## Testing

- New regression test `nether_chunks_only_contain_nether_biomes`: generate an overworld chunk, then a Nether chunk on the same thread, assert every biome cell is one of the five Nether biomes. Fails on `master` with `cell 0,0,0 generated non-Nether biome id 41` (river), passes with the fix. Runs in 0.06s in debug.
- `cargo clippy --all-targets -p pumpkin-world`: zero warnings, debug and release.
- `cargo test -p pumpkin-world`: 170 pass. The 7 `proto_chunk_test::no_blend_no_beard*` / `nether_surface_no_blend_no_beard_0_0` failures are **pre-existing on current `master`** and unrelated to this change; I verified they fail identically on a clean `master` checkout with none of this branch applied, in both debug and release.

## AI assistance

This PR was developed with AI assistance (Claude Code, model: claude-opus-5).

### DIFF
--- a/crates/pumpkin-world/src/biome/mod.rs
+++ b/crates/pumpkin-world/src/biome/mod.rs
@@ -17,8 +17,12 @@ pub use pumpkin_data::chunk::{
 };
 
 thread_local! {
-    /// A shortcut; check if last used biome is what we should use
-    static LAST_RESULT_NODE: RefCell<Option<&'static BiomeTree>> = const {RefCell::new(None) };
+    /// A shortcut; reuse the last matched biome-tree leaf when consecutive
+    /// lookups land nearby. Keyed by the source tree so a leaf cached for one
+    /// dimension's tree is never reused for another (which previously let an
+    /// overworld biome leak into Nether lookups and corrupt Nether decoration).
+    static LAST_RESULT_NODE: RefCell<Option<(&'static BiomeTree, &'static BiomeTree)>> =
+        const { RefCell::new(None) };
 }
 
 pub trait BiomeSupplier {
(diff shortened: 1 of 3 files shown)


ydw1904:
Closing because the reported Nether biome leakage is not reproducible on the latest master.

PR #3191, opened 2026-09-02
feat(item): implement spear combat

MegalithOfficial:
## Description

This is the spear part of #2479. The equipment attribute sync from that PR was split into #3190.

Spears already existed as items, but none of their attacks worked. This implements both attack types using the same calculations and conditions as the official server: the normal jab on right click and the kinetic contact attack while holding use and moving fast on foot, on a mount or with an elytra.

The jab uses the vanilla attack ray, so it clips against blocks, respects the spear's longer reach and the creative reach, and picks up targets slightly next to the ray through the hitbox margin. Damage goes through the normal attack path with attack cooldown scaling, enchantment bonuses, knockback, durability damage and exhaustion.

Kinetic attacks check the attacker's speed along the look direction and the speed relative to the target. Damage, knockback and dismounting each have their own speed threshold and use duration window per spear material, damage scales with the relative speed, and each target has a contact cooldown so riding through a crowd does not hit the same entity every tick.

The `kinetic_weapon`, `piercing_weapon`, `attack_range`, `damage_type` and `minimum_attack_charge` components were empty marker structs in pumpkin-data even though the extractor dump already contains their values. The codegen now generates them fully, so all per-material values, reach distances, the hitbox margin and the sounds come from the generated data instead of being hardcoded.

## Testing

I compared the attack conditions, damage calculations, speed thresholds, reach values and the ray logic directly against the official 26.2 server implementation.

I play-tested all spear materials in game. jab damage, knockback and sounds against mobs, kinetic hits while sprinting and riding, the per-target contact cooldown, dismounting, block occlusion of the attack ray and durability loss.


Snowiiii:

PR #3192, opened 2026-09-02
fix(net): keep alive id must fit 32 bits for pre-1.12.2 clients

MegalithOfficial:
## Description
Clients on 1.8 and older were getting kicked with `disconnect.timeout` about a minute after joining, even though they were responding to every keep alive.

Pumpkin uses the current Unix timestamp in milliseconds as the keep alive ID, which is a 64-bit value. The packet writer already handles the ID differently depending on the protocol version, like PacketEvents does: `long` on 1.12.2+, `VarInt` on 1.8, and a regular `int` on 1.7.

The problem is that on older versions the ID gets truncated to 32 bits when it's written to the packet. The client sends that truncated value back, but the server was still comparing it against the original 64-bit timestamp. They never matched, so the keep alive stayed pending until it timed out and kicked the player.

The fix is to clamp the ID to the 32-bit range when building `CKeepAlive`, then have the keep alive loop use the ID from the packet instead of the original timestamp. This way the stored ID is always the same one that was actually sent to the client.

No packet read/write logic had to change. Newer versions shouldnt be effected, just receive a smaller keep alive ID.

## Testing

I reproduced this on master with a 1.8.9 bot. It received keep alive IDs at 18, 33 and 48 seconds, answered each one with the exact same ID, and was kicked at 63 seconds. The ID it received was the low 32 bits of the millisecond timestamp.

With this change I ran 1.7.10, 1.8.8, 1.12.2 and 26.1 bots against the same server for 80 seconds each. All four stayed connected, every keep alive matched in both directions and the server log had no warnings or panics.

### DIFF
--- a/crates/pumpkin/src/net/java/mod.rs
+++ b/crates/pumpkin/src/net/java/mod.rs
@@ -298,11 +298,12 @@ impl JavaClient {
                         break;
                     }
 
-                    // Generate a unique ID (current timestamp in ms)
-                    let keep_alive_id = SystemTime::now()
-                        .duration_since(UNIX_EPOCH)
-                        .unwrap_or_default()
-                        .as_millis() as i64;
+                    let keep_alive_id = i64::from(
+                        SystemTime::now()
+                            .duration_since(UNIX_EPOCH)
+                            .unwrap_or_default()
+                            .as_millis() as i32,
+                    );
 
                     self.keep_alive_id.store(keep_alive_id);
                     self.wait_for_keep_alive.store(true, Ordering::Relaxed);


Q2297045667:
Should [3072](https://github.com/Pumpkin-MC/Pumpkin/pull/3072) be merged, I will close this PR. Conversely, if this PR is merged, I will remove the relevant content of [3072](https://github.com/Pumpkin-MC/Pumpkin/pull/3072)."

zinedinp:
Dont merge, unsafe i32 conversion without bitmask. A DOS could id spoof and force a timeout.

Better, the i32 shlould be generated before the ID and not shrunken afterwards.

@MegalithOfficial

MegalithOfficial:
> Dont merge, unsafe i32 conversion without bitmask. A DOS could id spoof and force a timeout.
> 
> Better, the i32 shlould be generated before the ID and not shrunken afterwards.
> 
> @MegalithOfficial

The conversion is a plain truncation to the low 32 bits, `as i32` in Rust is well defined and a bitmask would produce the same bits.

I dont see the DoS path either. The server only accepts IDs that are in its own pending list, anything else is ignored. That state is per connection, so a client can't touch another player's keep alive, and forcing a timeout on your own connection just means not replying, which was always possible. The IDs were already predictable before this PR since they are millisecond timestamps, that part is unchanged.

I also tested it: two bots sending 300+ random keep alive IDs each over 80 seconds alongside two normal clients, on 1.8.8 and 26.1. Nobody timed out. That said, I agree generating the ID as an i32 up front is cleaner than shrinking it in the packet constructor. Im considering move it to the keep alive loop so the packet code and its tests stay as they were.

zinedinp:
Just read the wiki for "as". I was thinking in c++, it works on rust.

But maybe a dual cast "(as u32) as i64" would be better.
A 0x80000000 (negative) will produce 0xffffffff80000000 (positiv)
Without that signed bit, both would be positive.

MegalithOfficial:
> Just read the wiki for "as". I was thinking in c++, it works on rust.
> 
> But maybe a dual cast "(as u32) as i64" would be better. A 0x80000000 (negative) will produce 0xffffffff80000000 (positiv) Without that signed bit, both would be positive.

Tried `u32` too. It breaks once bit 31 is set.

The server reads the response with `i64::from(get_var_int())`, which sign extends the VarInt. PacketEvents does the same thing when it reads the VarInt into a long. This PR doesn't touch that code, so the value we store needs to line up with what the reader gives us back.

With `as i32`, it does. If we use `as u32`, `0x80000000` gets stored as `2147483648`. The client sends the same 32 bits back, the reader turns that into `-2147483648`, and the pending entry never matches. It just sits there until the 30s timeout.

Bit 31 isn't set right now, so both versions happen to work today. The ID is Unix millis truncated to 32 bits, which gives a ~49.7 day cycle. For roughly half of that cycle the value is negative. We're around day 19 right now, so that starts around Sept 8. From there the `u32` version would kick 1.8 players for about 25 days, then work again for about 25 days, then repeat.

I forced bit 31 on for every ID on this branch and tested 1.7, 1.8.8 and 26.1. All three echoed the negative IDs back and stayed connected.

I also checked vanilla 1.8 because I wanted to see what the client actually does. I pulled the 1.8.9 server and client jars from Mojang, remapped them with MCP 22-1.8.9 and decompiled them.

Server side, `NetHandlerPlayServer`:

```java
this.lastPingTime = this.currentTimeMillis();
this.field_147378_h = (int)this.lastPingTime;
this.sendPacket(new S00PacketKeepAlive(this.field_147378_h));

private long currentTimeMillis() {
   return System.nanoTime() / 1000000L;
}

public void processKeepAlive(C00PacketKeepAlive packet) {
   if (packet.getKey() == this.field_147378_h) {
      ...
   }
}
```

`System.nanoTime()` is then cast straight down to an `int` before being sent. Once that value crosses the signed 32-bit range, the keep alive ID becomes negative. Vanilla handles that normally.

Both packets contain a single `int` and use `readVarIntFromBuffer` / `writeVarIntToBuffer`. Client side is just:

```java
public void handleKeepAlive(S00PacketKeepAlive packet) {
   this.addToSendQueue(new C00PacketKeepAlive(packet.func_149134_c()));
}
```

The client reads the signed `int` and sends the same value back. The server compares it directly with the `int` it stored.

That matches what `as i32` gives us. Using `u32` changes the stored numeric value once bit 31 is set, which is where the mismatch comes from.

zinedinp:
I see, thanks for digging deeper. The Client is of course true and we dont want a missmatch (of course mojang didnt code catches that in their own ecosystem). There shoudnt be any any issue that i can see. Good job

Snowiiii:

PR #3193, opened 2026-09-02
fix(portal): play sound when end portal opens

### LINKED ISSUE #3122: End Portal plays no sound when all eyes are put in.
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

9e9f1b5/release

### Operating System

CachyOS Linux

When putting all 12 Eye's of Ender into the End portal, the end portal "open" sound doesn't play.

### Steps to Reproduce

1. Find or make an end portal frame
2. Fill it with eye's of ender
3. No sound

AdmerPRO:
## What was changed?

A new `World::sync_global_world_event` method was added. It sends a Java `Level Event` packet with `disable_relative_volume` set to `true`, matching the behavior of global vanilla world events.

End Portal activation now emits `WorldEvent::SoundEndPortalSpawn` after all nine portal blocks have been created. This corresponds to Java Level Event ID `1038` and plays the End Portal opening sound at the center of the completed portal.

When the twelfth Eye of Ender was inserted, Pumpkin correctly detected the completed frame and created the 3×3 End Portal. However, the activation path only updated the portal blocks and never emitted the world event responsible for the opening sound.

The existing `World::sync_world_event` method always sets `disable_relative_volume` to `false`. Vanilla sends End Portal activation as a global Level Event with this flag enabled, so a separate method was needed to preserve the correct protocol behavior.

- Completing an End Portal now plays its vanilla opening sound.
- Java clients receive Level Event `1038` at the center of the portal.
- The event is emitted exactly once after the complete 3×3 portal has been created.
- The sound uses vanilla global-event behavior instead of regular distance-based sound playback.
- Existing non-global world events remain unchanged.

- No dedicated automated test was added for capturing and validating the emitted Level Event packet.
- The current world-event packet path targets Java clients. Equivalent Bedrock sound-event handling is not included in this change.
- The change was verified with `cargo check -p pumpkin`, the focused portal tests, `cargo fmt --check`, and `git Braz diff --check`.

Fixes #3122

## Summary by CodeRabbit

* **New Features**
  * End portal creation now notifies all players who have global events enabled.
  * Added support for broadcasting global world events independently of chunk tracking.


AdmerPRO:
Ready for review

crates/pumpkin/src/world/mod.rs
```diff
@@ -5661,6 +5661,14 @@ impl World {
         );
     }
 
+    pub fn sync_global_world_event(&self, world_event: WorldEvent, position: BlockPos, data: i32) {
```
RoosterBooster007, line 5968: Small nit: this only broadcasts to players with the given chunk loaded (near the portal), but vanilla sends it to everyone (no matter the location or dimension). This fixes the bug, though.

RoosterBooster007 (approved):

crates/pumpkin/src/world/mod.rs
AdmerPRO: Ok

RoosterBooster007:
Thanks, @AdmerPRO!

PR #3194, opened 2026-09-02
fix(main): register SIGTERM and SIGHUP handlers before waiting

### LINKED ISSUE #3115: SIGTERM is not handled: sequential signal awaits arm only SIGINT — server killed without save
### Before Submitting

I have searched existing issues and confirmed this is not a duplicate.
(Related: #182 "Handle SIGTERM on UNIX" is closed, but the handling is broken on the current
build — see below.)

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.40

### Operating System

Arch Linux (x86_64, kernel 7.1.8)

The server does not handle `SIGTERM`: the process is killed by the signal immediately
(exit status: terminated by SIGTERM, shell reports 143 / WIFSIGNALED), with **no graceful
shutdown and no world save**. Any chunks not yet autosaved are lost.

Expected: on `SIGTERM` the server should run the same graceful shutdown as the `stop`
command (save the world, close regions, exit 0) — this is what vanilla does and what
`systemctl stop` / `docker stop` rely on.

**Root cause** (master, `crates/pumpkin/src/main.rs:325`): `setup_sighandler()` awaits the
three signal streams *sequentially*, so only the first one (`SIGINT`) is ever armed:

```rust
async fn setup_sighandler() -> io::Result<()> {
    if signal(SignalKind::interrupt())?.recv().await.is_some() {
        handle_interrupt();          // <-- blocks here until SIGINT
    }

    if signal(SignalKind::hangup())?.recv().await.is_some {   // never registered earlier
        handle_interrupt();
    }

    if signal(SignalKind::terminate())?.recv().await.is_some { // never registered earlier
        handle_interrupt();
    }

    Ok(())
}
```

The `SignalKind::hangup()` and `SignalKind::terminate()` streams are only created *after*
`SIGINT` is received. `SIGTERM` therefore takes the default action (terminate).

Suggested fix: register all three concurrently, e.g.

```rust
tokio::select! {
    _ = signal(SignalKind::interrupt())?.recv() => handle_interrupt(),
    _ = signal(SignalKind::hangup())?.recv() => handle_interrupt(),
    _ = signal(SignalKind::terminate())?.recv() => handle_interrupt(),
}
```

(or spawn one task per signal).

### Steps to Reproduce

```console
$ ./pumpkin-X64-Linux &
# wait for "Server is now running"
$ kill -TERM $(pidof pumpkin-X64-Linux)
$ wait $(pidof pumpkin-X64-Linux); echo $?
```

Observed: process dies instantly (by signal, 128+15=143 in shells), no
"Stopping the server"/save log lines, `stop`-style shutdown never runs.
Control: `SIGINT` (Ctrl+C) *does* trigger the graceful path, confirming the sequential-await
diagnosis. Chunks generated but not yet autosaved are lost on `SIGTERM`.

BryanFRD:
## Problem

`setup_sighandler` awaits the three signal streams sequentially, so only `SIGINT` is ever
registered. `signal()` is what installs the handler, and the `SIGHUP` and `SIGTERM` calls
are only reached after a `SIGINT` has already arrived and shut the server down.

Until then both signals keep their default disposition, which terminates the process
immediately. `systemctl stop`, `docker stop` and a plain `kill` therefore kill the server
with no graceful shutdown, and any chunk that has not been autosaved is lost.

Fixes #3115.

## Change

All three streams are registered up front, then `tokio::select!` waits for whichever signal
arrives first and runs the same `handle_interrupt` as before.

The `.is_some()` check is kept, so a closed stream still does not trigger a shutdown. The
non-Unix path, which only handles Ctrl-C, is untouched.

## Notes

- No behaviour change for `SIGINT`.
- `handle_interrupt` is unchanged, so the shutdown path stays the one the `stop` command
  already uses.

## Tests

No unit test. `handle_interrupt` calls into the process-wide shutdown, and signal delivery
is process-wide too, so raising `SIGTERM` inside the shared test binary would tear down the
whole test run rather than fail one case. Verified instead by building and linting the Unix
target in a Linux container.

### DIFF
--- a/crates/pumpkin/src/main.rs
+++ b/crates/pumpkin/src/main.rs
@@ -323,15 +323,17 @@ async fn setup_sighandler() -> io::Result<()> {
 // Unix signal handling
 #[cfg(unix)]
 async fn setup_sighandler() -> io::Result<()> {
-    if signal(SignalKind::interrupt())?.recv().await.is_some() {
-        handle_interrupt();
-    }
+    let mut interrupt = signal(SignalKind::interrupt())?;
+    let mut hangup = signal(SignalKind::hangup())?;
+    let mut terminate = signal(SignalKind::terminate())?;
 
-    if signal(SignalKind::hangup())?.recv().await.is_some() {
-        handle_interrupt();
-    }
+    let received = tokio::select! {
+        received = interrupt.recv() => received,
+        received = hangup.recv() => received,
+        received = terminate.recv() => received,
+    };
 
-    if signal(SignalKind::terminate())?.recv().await.is_some() {
+    if received.is_some() {
         handle_interrupt();
     }
 


Snowiiii:
LGTM, Thank you @BryanFRD :D

PR #3195, opened 2026-09-02
perf(worldgen): pick density cache kind from domain axes like 26.3

MegalithOfficial:
## Description

Chunk generation got slower since the density function data was moved to the 26.3 datapack. The full chunk pipeline in the `chunk_gen` bench went from 48 ms to 157 ms per chunk, most of it in the biome stage.

In 26.3 Mojang replaced `flat_cache`, `cache_2d` and `cache_once` with a single `cache` marker and let the engine figure out the caching itself based on which axes a function depends on. Our runtime is still the 26.2 model and the codegen was turning every `cache` into `CacheOnce`, so all the 2D shaper functions like continents, erosion, offset and factor were recomputed for every Y sample in a column instead of once.

I added the same axis inference to the codegen. Each density function reports which axes it depends on the way vanilla 26.3 does (a noise with `y_scale` 0 doesn't depend on Y, a gradient only on its axis, operators inherit from their inputs and so on), and a `cache` whose input doesn't depend on Y becomes `CacheFlat`. Nothing else changed, the datapack files and the runtime are untouched and only `noise_router.rs` was regenerated.

Carvers are also slower than in August but that comes from the carver rewrite in the same commit, not from the cache markers. I left that for a separate PR.

## Testing

`cargo bench -p pumpkin-world`, master vs this branch: full chunk generation 157 ms to 40.6 ms, biomes 13.8 ms to 1.09 ms, noise 16.4 ms to 12.9 ms, 16 chunks on 4 threads 644 ms to 182 ms.

I generated 49 chunks of seed 42 up to the surface stage on master, on this branch and on a build with the old 26.2 markers and hashed all block states and biome ids. All three match. `pumpkin-world` tests pass.

On a fresh world with one client at view distance 16 the full view loads in 12.3 s instead of 16.8 s when the client asks for 64 chunks per tick. With a normal client the gain is smaller since the client only asks for a couple of chunks per tick anyway.

While doing the hash comparison I noticed the features stage isn't deterministic, generating the same chunk twice gives different blocks in about half the chunks. That happens on master too and is unrelated to this change.

### DIFF
--- a/crates/pumpkin-data/src/generated/noise_router.rs
+++ b/crates/pumpkin-data/src/generated/noise_router.rs
@@ -134,7 +134,7 @@ pub mod overworld_compiled {
         pos: &pumpkin_util::math::vector3::Vector3<i32>,
         ctx: &mut C,
     ) -> f32 {
-        ctx.sample_wrapper(8usize, WrapperType::CacheOnce, pos, &eval_overworld_7)
+        ctx.sample_wrapper(8usize, WrapperType::CacheFlat, pos, &eval_overworld_7)
     }
     #[inline(always)]
     pub fn eval_overworld_9<C: NoiseEvaluationContext>(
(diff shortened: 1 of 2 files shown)


Snowiiii:

PR #3197, opened 2026-09-02
chore(ci): add cancel-in-progress to stop invalid CIs

Phoenixxo:
## Description

Adds workflow-level concurrency to cancel outdated CI runs when newer commits are pushed to the same ref.

Yeah... that's it...


Snowiiii:
LGTM, Thank you @Phoenixxo

PR #3198, opened 2026-09-02
fix(plugin): clean up plugin state during unload

Phoenixxo:
## Summary:
This PR improves plugin unloading by ensuring plugin-owned commands, event handlers, and scheduled tasks cannot continue running after their plugin has been unloaded. This work was originally part of #3185, but has been pulled into a standalone PR because it is useful independently of the WASM reentry changes and doesn't exactly fall within it's scope while still being a good change alongside it for stability.
## Changes:
Plugin commands are now tracked by their source and deactivated during unload, including any aliases that point to them. Commands can be reactivated if the plugin is loaded and registers them again, and their inactive state is preserved when the command dispatcher is rebuilt.

Event handlers registered by the plugin are removed during unload. The scheduler also disables the plugin, removes its pending tasks, rejects new scheduled work, and prevents repeating tasks from scheduling themselves again after unloading begins.

## Validation:
Validated with `cargo fmt --all -- --check`, `cargo check -p pumpkin` & `cargo clippy -p pumpkin --lib --all-features -- -D warnings`

### DIFF
--- a/crates/pumpkin/src/command/node/dispatcher.rs
+++ b/crates/pumpkin/src/command/node/dispatcher.rs
@@ -84,6 +84,11 @@ pub struct CommandDispatcher {
     /// configuration. A disabled command behaves as if it does not exist: it
     /// cannot be executed and is left out of listings and suggestions.
     disabled: FxHashSet<String>,
+
+    /// Commands whose plugin source is currently unloaded. These are kept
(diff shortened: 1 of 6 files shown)


Snowiiii:
LGTM, Thank you @Phoenixxo :D

PR #3199, opened 2026-09-02
perf(worldgen): add 26.3 density sampler contract and slice rewrite

MegalithOfficial:
## Description

This is the first of two PRs for moving the density function runtime to the 26.3 sampler model. It adds the runtime pieces needed for the port without changing any generated values.

The codegen now applies vanilla's `SLICE_UNIFORM_AXES` rewrite. A subtree that does not depend on an axis gets wrapped in a slice for that axis. Cache inputs restart the rule with the full axis set, while constants and gradients are skipped. This also reproduces the double slice around caches that vanilla generates.

On the runtime side, this adds `DensityVolume` and `DensityBuffer`, backed by a pooled arena that follows `DensityBufferPool`'s rounding and best-fit rules. Component traits also get a `sample_volume` method whose default implementation samples every position in vanilla's index order. That gives the follow-up PR a base to replace the per-sample paths with volume sampling node by node.

Cache IDs use component indices since the codegen already merges identical wrapper subtrees. The pool differs slightly from vanilla here: instead of age-based garbage collection, each thread keeps at most 64 buffers.

## Testing

I compared block and biome hashes of 49 chunks (seed 42, up to the surface stage) between master and this branch and they are identical. The 178 pumpkin-world tests pass including the density fingerprint. The chunk_gen bench shows a small cost from the extra slice nodes on the per sample path: full chunk generation 40.3 ms to 41.4 ms, biomes 1.06 ms to 1.14 ms, noise 12.4 ms to 12.5 ms, the other stages unchanged. That cost disappears once the driver samples volumes, which is the second PR.

### DIFF
--- a/crates/pumpkin-data/src/generated/noise_router.rs
+++ b/crates/pumpkin-data/src/generated/noise_router.rs
@@ -134,10 +134,18 @@ pub mod overworld_compiled {
         pos: &pumpkin_util::math::vector3::Vector3<i32>,
         ctx: &mut C,
     ) -> f32 {
-        ctx.sample_wrapper(8usize, WrapperType::CacheFlat, pos, &eval_overworld_7)
+        let slice_pos = pumpkin_util::math::vector3::Vector3::new(pos.x, 0i32, pos.z);
+        eval_overworld_7(&slice_pos, ctx)
     }
     #[inline(always)]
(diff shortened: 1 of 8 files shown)


Snowiiii:

PR #3202, opened 2026-09-03
fix: crash on outbound HTTPS from dual rustls crypto providers

RoosterBooster007:
## Description
Fixes a crash on any outbound HTTPS request caused by `rustls` being compiled with two crypto providers (`aws-lc-rs` from `reqwest` vs. `ring` from `wasmtime-wasi-http/rtc`) with no default selected. It's resolved by making `reqwest` provider-agnostic and installing `ring` as the single process-wide provider.

## Testing
- [x] Successful HTTPS requests can be made via the Plugin API (confirmed by @Stilldrop1)


RoosterBooster007:
Thanks for your help, @Stilldrop1 🚀

PR #3204, opened 2026-09-03
fix(block): play the lever click sound and emit its game event

pxperr:
## What was changed?

`toggle_lever` in `crates/pumpkin/src/block/blocks/redstone/lever.rs` plays
`block.lever.click` and emits a game event, alongside the state change and the
neighbour updates it already did. The sound sits in a small `play_lever_sound`
helper so the pitch choice reads next to the volume it belongs with.

Vanilla's `LeverBlock.pull` does four things:

```java
public void pull(BlockState state, Level level, BlockPos pos, @Nullable Player player) {
   state = state.cycle(POWERED);
   level.setBlock(pos, state, 3);
   this.updateNeighbours(state, level, pos);
   playSound(player, level, pos, state);
   level.gameEvent(player, state.getValue(POWERED) ? GameEvent.BLOCK_ACTIVATE : GameEvent.BLOCK_DEACTIVATE, pos);
}
```

`toggle_lever` did the first two, so flicking a lever was silent.

The values are vanilla's: volume 0.3, pitch 0.6 when the lever switches on and
0.5 when it switches off. That is why this calls `play_sound_fine` rather than
the `play_block_sound` that `ButtonBlock` uses - the latter fixes volume and
pitch at 1.0, which would have made a lever as loud as a button and dropped the
pitch difference between on and off. `BlockFlags::NOTIFY_ALL` is already `0b11`,
so the existing `set_block_state` matches vanilla's `setBlock(pos, state, 3)`.

`ButtonBlock` plays its own click sound (`buttons.rs`), so these two blocks
agreed on everything except this.

Levers make a sound, at the volume and the two pitches vanilla uses, and fire
`block_activate` or `block_deactivate` to match. Nothing else about the block
changes: same state, same neighbour updates, same order.

`emit_game_event` only reaches plugins today, since nothing in the tree calls
`SculkSensorBlock::trigger` and vibration listening is unwired either way. It is
the call site vanilla uses and the event names a plugin would expect, so it is
worth having now rather than remembering later.

* Verified by reading against the 26.2 source and the existing `ButtonBlock`
  path rather than in an in-game session. `cargo test -p pumpkin --lib` is 255
  passing, unchanged by this branch; clippy and fmt are clean.
* Two divergences in `LeverBlock` are deliberately left alone because neither is
  fixable in the block file: `onExplosionHit`, which pulls the lever when an
  explosion can trigger blocks, has no behaviour hook to hang off, and
  `updateNeighbours` passes no `Orientation` where vanilla builds one through
  `ExperimentalRedstoneUtils.initialOrientation`, which is systemic rather than
  lever-specific.
* This touches the same `use pumpkin_data::{...}` block as #3132, so whichever
  lands second wants a trivial rebase. Otherwise independent of both open lever
  PRs: #3132 changes `on_place`, #3135 changes `normal_use` and
  `on_state_replaced`, this changes `toggle_lever`.

## Summary by CodeRabbit

* **New Features**
  * Lever interactions now play a click sound when toggled.
  * Lever activation and deactivation now trigger corresponding in-game events.


RoosterBooster007:
Thanks, @jackpipergit!

PR #3205, opened 2026-09-03
fix(block): return Consume when clicking an already pressed button

pxperr:
## What was changed?

`click_button` reports whether it actually pressed the button, and `normal_use`
maps that onto `BlockActionResult::Success` or `BlockActionResult::Consume`.

Vanilla's `ButtonBlock.useWithoutItem` tells the two cases apart:

```java
protected InteractionResult useWithoutItem(BlockState state, Level level, BlockPos pos, Player player, BlockHitResult hitResult) {
   if (state.getValue(POWERED)) {
      return InteractionResult.CONSUME;
   } else {
      this.press(state, level, pos, player);
      return InteractionResult.SUCCESS;
   }
}
```

`normal_use` returned `Success` either way. `click_button` already short
circuited on a powered button, so the information existed and was thrown away;
returning it keeps the single state read rather than adding a second one in
`normal_use` just to answer a question the helper had already answered.

No observable change today. `consumes_action()` covers `Consume` and `Success`
alike, and only `SuccessServer` reaches `swing_hand`
(`net/java/play/use_item_on.rs`). The two do map to different vanilla
`InteractionResult`s though, and this is the same latent divergence #3135
corrects on the lever, in the block that PR holds up as the reference for it.

* Two further gaps in `ButtonBlock` are left alone, both larger than this and
  neither a one line change: no game event on press or release, where vanilla
  fires `block_activate` and `block_deactivate` from `press` and `checkPressed`,
  and no `entityInside`, so arrows cannot press wooden buttons.
* `cargo test -p pumpkin --lib` is 255 passing, unchanged by this branch; clippy
  and fmt are clean. The press and release paths are otherwise faithful, so
  there was nothing to test in game here.
* Independent of every open lever and sign PR; it touches only `buttons.rs`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Button interactions now distinguish between newly activated and already-powered buttons.
  - Pressing an already-powered button no longer triggers additional state changes or activation effects.
  - Newly pressed buttons continue to update their state, notify nearby components, schedule their release, and play the expected sound.


RoosterBooster007:
Thanks, @jackpipergit!

PR #3206, opened 2026-09-03
fix(block): waterlog signs placed in water

pxperr:
## What was changed?

`SignPlacement` carries a `waterlogged` flag, set from
`args.replacing.water_source()` in `determine_placement` and written into the
placed state by `apply_placement_properties`. A sign that survives a neighbour
update schedules a water tick. A unit test covers all four sign families.

Vanilla sets `WATERLOGGED` from the replaced fluid state in both
`StandingSignBlock.getStateForPlacement` and
`WallSignBlock.getStateForPlacement`, and `SignBlock.updateShape` keeps the
water ticking while the sign survives:

```java
protected BlockState updateShape(...) {
   if (state.getValue(WATERLOGGED)) {
      ticks.scheduleTick(pos, Fluids.WATER, Fluids.WATER.getTickDelay(level));
   }
   return super.updateShape(...);
}
```

None of that happened here. `apply_placement_properties` wrote `facing`,
`rotation` and `attached` and nothing else, so a sign placed under water came
out dry and left an air pocket where the water had been.

`args.replacing.water_source()` is how every other waterloggable block does it
(`barrier.rs`, `chain.rs`, `lanterns.rs`), so signs follow the existing route
rather than growing a second one. It is source-only where vanilla's
`replacedFluidState.is(Fluids.WATER)` also matches flowing water, but that gap
belongs to `BlockIsReplacing` and applies to every waterloggable block, so
fixing it for signs alone would only make signs inconsistent.

The tick is scheduled after the support check rather than before it, because
vanilla returns `AIR` without reaching `super.updateShape` when the sign cannot
survive. Scheduling first would tick water for a sign that is about to break.

A sign placed in water is waterlogged and the water stays put, and a waterlogged
sign keeps the water around it flowing on every neighbour update. Hanging signs
are waterloggable too and go through both of these paths, so they are fixed
alongside the plain ones. Nothing about which supports a sign accepts changes.

* `apply_placement_properties` matches property names as strings, so a renamed
  or missing `waterlogged` would silently stop waterlogging signs with nothing
  to catch it. The test asserts the flag round-trips into the state id for
  `oak_sign`, `oak_wall_sign`, `oak_hanging_sign` and `oak_wall_hanging_sign`.
* `cargo test -p pumpkin --lib` is 256 passing, 255 of them pre-existing and the
  new one being that test; clippy and fmt are clean. Verified against the state
  data rather than in an in-game session.
* This touches `get_state_for_neighbor_update` in the same file as #3162, a few
  lines below the `is_valid` match that PR rewrites, so whichever lands second
  wants a small rebase. The two are otherwise independent: that one changes
  which support test runs, this one changes nothing about support.

## Summary by CodeRabbit

- **Bug Fixes**
  - Signs placed in water now correctly retain their waterlogged state across standing, wall-mounted, hanging, and wall-hanging variants.
  - Water continues to flow properly around waterlogged signs after placement.


RoosterBooster007:
Thanks, @jackpipergit!

PR #3208, opened 2026-09-03
perf(worldgen): port the 26.3 volume density sampler

MegalithOfficial:
## Description
This is the second half of the 26.3 density sampler port, building on #3199.

Every density function node now has a volume path. Leaf nodes fill their buffers directly, while operators work over the buffers produced by their children. That brings chunk sampling down to one call per node for the whole volume instead of calling every node once per block.

The four 26.2 cache types and the old interpolator are gone. There's now a single cache keyed by `DensityVolume`, following vanilla's `SamplerContext`, plus an `Interpolated` node based on `InterpolatedFunction`. Cell sizes come from the datapack JSON through codegen.

Chunk sampling now follows the same general flow as `doFill`. `final_density` is sampled once for the full `16 x height x 16` volume, then the aquifer and ore samplers read the density they need per block. Biome generation does the same for the six climate parameters over the chunk's quart grid, similar to `createResolverForChunk`. The structure height sampler only needs a single column volume.

The old `fill` / `IndexToNoisePos` path is gone along with the sample options, cache IDs, and the separate `interpolate_x/y/z` loop.

There are still three places where I kept master behavior for bit parity. Trilinear expansion uses the old fill path's x, y, z lerp order instead of vanilla's z, x, incremental y order. `Binary Mul` still has the zero shortcut, `Lerp` still uses the plain formula per element, and `FindTopSurface` keeps the extra probe at `floor(upper)`.

A couple of things did move closer to 26.3 without changing the parity run. Vein values that previously went through the y, x, z interpolation chain now use the same interpolation order as the rest of the sampler. The aquifer also samples erosion and depth at the actual block column instead of using the quart-aligned `FlatCache` value.

`vein_gap` isn't under an interpolated node in the data, so it stays lazy and is sampled per block by the ore sampler. `vein_toggle` and `vein_ridged` are sampled as chunk volumes.

The Perlin core hasn't changed yet. That's the next PR.

## Testing
I compared block and biome hashes for 49 chunks on seed 42, up to the surface stage, between master and this branch. They match exactly.

All 180 `pumpkin-world` tests pass, including the eight vanilla chunk dump tests. I also added a test that runs `sample_volume` against per-position `sample` for every component in all three routers and checks the results bit for bit.

The density fingerprint hash did change. `final_density(pos)` now interpolates off-corner positions the same way vanilla `sampleValue` does.

`chunk_gen` benchmarks, master --> branch:

* full chunk: 40.6 ms --> 28.8 ms
* noise: 12.7 ms --> 9.6 ms
* carvers: 4.3 ms --> 1.4 ms
* biomes: 1.10 ms --> 0.56 ms
* structure starts: 50.7 us --> 1.5 us
* surface and lighting: basically unchanged
* 16 chunks / 4 threads: 170 ms --> 124 ms

### DIFF
--- a/crates/pumpkin-data/src/generated/noise_router.rs
+++ b/crates/pumpkin-data/src/generated/noise_router.rs
@@ -142,7 +142,7 @@ pub mod overworld_compiled {
         pos: &pumpkin_util::math::vector3::Vector3<i32>,
         ctx: &mut C,
     ) -> f32 {
-        ctx.sample_wrapper(9usize, WrapperType::CacheFlat, pos, &eval_overworld_8)
+        ctx.sample_wrapper(9usize, WrapperType::Cache, pos, &eval_overworld_8)
     }
     #[inline(always)]
     pub fn eval_overworld_10<C: NoiseEvaluationContext>(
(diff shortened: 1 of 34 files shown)


Snowiiii:

PR #3209, opened 2026-09-03
fire BlockBreakEvent in break_block

IndrajeethY:
## Summary

- The async to sync refactor dropped the `fire_blocking` call for `BlockBreakEvent` in `break_block`, so plugins never got notified on block break
- Adds it back with cancellation and drop flag support
- Changed `break_block` signature from `Option<&Player>` to `Option<&Arc<Player>>` since the event needs a player reference and all callers already have the Arc

## Test plan

- [x] `cargo check` passes
- [x] Break a block with a WASM plugin listening for `BlockBreakEvent` — handler fires correctly
- [x] Event contains correct block name, position, and player
- [x] Plugin can read and act on the event data (block count, scoreboard update, place new block)
- [x] Toggle the drop flag from a plugin and verify items drop or don't accordingly
- [x] Cancel the event from a plugin and verify the block doesn't break

### DIFF
--- a/crates/pumpkin/src/block/blocks/bed.rs
+++ b/crates/pumpkin/src/block/blocks/bed.rs
@@ -155,7 +155,7 @@ impl BlockBehaviour for BedBlock {
         };
 
         args.world
-            .break_block(&other_half_pos, Some(args.player), flags);
+            .break_block(&other_half_pos, Some(args.player.clone()), flags);
     }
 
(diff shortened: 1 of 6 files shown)


IndrajeethY:
Closing this for now

Snowiiii:
LGTM, Thank you @IndrajeethY :D

PR #3210, opened 2026-09-03
fix(block): stop the behaviour index table from silently dropping blocks

BryanFRD:
## Problem

`BlockRegistry::register` indexes behaviours with a `u8`, and `0xFF` doubles as the
"no behaviour" sentinel in `get_pumpkin_block`. That is two separate failures waiting
at the same ceiling:

- the 256th registered behaviour gets index `0xFF`, `u8::try_from` succeeds, and every
  block it covers silently resolves to `None`. No panic, no log, just a block whose
  behaviour never runs.
- the 257th panics at startup.

`default_registry` is at 204 registrations today, so there are 51 left before the first
one bites, and new blocks land regularly.

## Change

Widen the table to `u16`, give the sentinel a name, and reject it in `register` so it can
never be handed out as a real index. The silent case is now impossible by construction
rather than merely further away. Headroom goes from 254 to 65534.

`block_indices` is private to `BlockRegistry` and nothing outside depended on the width.
The array is `BlockId::COUNT` (1196) entries, so this costs about 1.2 KB once.

## Tests

None added. Reproducing the overflow needs 256 registered behaviours, and since
`BlockMetadata::ids` is an associated function, each registration needs its own type, so
the test would be a macro generating 256 types. That is more code than the fix and would
exercise the type system rather than the logic. Happy to add it if you would rather have it.

## Summary by CodeRabbit

- **Refactor**
  - Improved internal block registration and lookup handling for larger behavior index ranges.
  - Added validation to prevent reserved sentinel values from being used as behavior indices.


RoosterBooster007:
Looks good. Thanks, @BryanFRD!

PR #3212, opened 2026-09-03
fix(mob): correct the inverted moon brightness curve

BryanFRD:
## Problem

`moon_brightness` returns its maximum at the new moon and its minimum a quarter
turn away, so the curve is inverted against vanilla:

```rust
if phase == 0 { 1.0 } else { 1.0 - (phase - 4).abs() as f32 / 4.0 }
```

Vanilla's `MOON_BRIGHTNESS_PER_PHASE` is `{1.0, 0.75, 0.5, 0.25, 0.0, 0.25, 0.5, 0.75}`,
full moon at phase 0. The current code gives phase 4 (new moon) `1.0` and phase 1 `0.25`
instead of `0.75`. Only phases 0, 2 and 6 happen to land on the right value.

The result feeds `local_scale += (moon_brightness * 0.25).clamp(...)` in
`RegionalDifficulty::calculate`, so mob equipment quality is skewed toward new moons
instead of full moons.

## Change

Drop the subtraction. `(phase - 4).abs() / 4.0` reproduces the vanilla table exactly,
which also makes the `phase == 0` special case redundant since it already yields `1.0`.

## Tests

`moon_brightness_matches_the_vanilla_phase_table` walks all eight phases against the
vanilla values and fails on the current code with `phase 1: got 0.25, want 0.75`.
`moon_brightness_wraps_every_eight_days` covers the lunar cycle repeating.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected moon brightness calculations to match expected phase behavior.
  - Full moons now produce maximum brightness, while new moons produce minimum brightness.
  - Fixed brightness values across all moon phases, including the phase cycle wrap-around.


RoosterBooster007:
Thanks, @BryanFRD!

PR #3214, opened 2026-09-03
fix(block): notify clients when lava converts to a solid block

pxperr:
## Description

`FlowingLava::receive_neighbor_fluids` sets the converted block with `BlockFlags::NOTIFY_NEIGHBORS`, which is only the neighbour-update bit. `NOTIFY_LISTENERS` is the flag that queues a position into `World::unsent_block_changes` (`crates/pumpkin/src/world/mod.rs:5268`), and that queue is what becomes block update packets, so the client was never told anything had changed.

The conversion itself is real: the block is written into the chunk and the chunk is marked dirty, so it saves and it is there on the next load. Only the notification was missing, which makes a cobblestone generator look broken - the lava keeps flowing over what the client still believes is air, and the cobblestone appears the moment a rejoin re-sends the chunk.

Both call sites are affected: flowing lava beside water (cobblestone, or obsidian when the lava is a source), and lava over soul soil beside blue ice (basalt).

Vanilla parity: `LiquidBlock.shouldSpreadLiquid` performs each of these conversions as `level.setBlockAndUpdate(pos, block.defaultBlockState())` followed by `fizz`, and `setBlockAndUpdate` is `setBlock(pos, state, 3)` - `UPDATE_NEIGHBORS | UPDATE_CLIENTS`. `BlockFlags::NOTIFY_ALL` is that same value, and the flags in `crates/pumpkin-world/src/world.rs` are documented as matching Minecraft's mask. The accompanying `sync_world_event(WorldEvent::LavaFizz, pos, 0)` already matches vanilla's `levelEvent(LevelEvent.LAVA_FIZZ, pos, 0)`, so the flag was the only part out of step.

Every other conversion in this file already uses `NOTIFY_ALL` - the lava into water stone at `lava.rs:283` and the fire ignition at `lava.rs:88` - which is why only cobblestone, obsidian and basalt ever showed the symptom.

## Testing

Reproduced in-game before the change on a dev server running a release build: a lava stream meeting water produced no visible block, the lava carried on rendering over the gap, and the cobblestone only appeared after logging out and back in. A build carrying this change is running on that server now, but the in-game confirmation that the cobblestone appears without a rejoin is still to come and I will follow up here with it.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved fluid interactions when lava meets water, ensuring resulting cobblestone or obsidian changes are fully communicated to nearby blocks.
  - Improved basalt formation when lava interacts with soul soil and blue ice by expanding block-update notifications, helping surrounding block behavior stay synchronized.


RoosterBooster007:
Thanks, @pxperr!

PR #3215, opened 2026-09-03
fix(block): match vanilla's neighbour scan when lava meets water

pxperr:
## Description

Two divergences from vanilla in `FlowingLava::receive_neighbor_fluids`.

**Water below ended the scan.** The loop returned the moment it found water under the lava:

```rust
if dir == BlockDirection::Down {
    return true;
}
```

so a lava block with water beneath it never looked at its remaining neighbours. Vanilla skips `Direction.DOWN` and carries on testing the rest. Forge's `FluidInteractionRegistry`, which documents itself as mirroring `LiquidBlock#shouldSpreadLiquid`, states it as "all directions, besides `Direction#DOWN` is tested and then replaced. Any fluids which cause a change in the down interaction must be handled in `FlowingFluid#spreadTo`". Pumpkin already has exactly that split - the lava into water stone conversion lives in `spread_to` - so Down becomes a `continue` and the block below is left to the code that already owns it. Lava with water beneath it and water beside it now forms cobblestone, as it does in vanilla.

Nothing changes when the only water is below: the loop falls through to the same `true` it used to return early.

**Waterlogged neighbours were invisible.** The neighbour test was `world.get_block(&neighbor_pos) == &Block::WATER`, which is block identity. Vanilla tests the fluid state against the water tag, so a waterlogged stair or slab beside the lava counts. `World::get_fluid` already reports `FLOWING_WATER` for a waterlogged state, and `has_tag(&tag::Fluid::MINECRAFT_WATER)` is the same check the entity spawn code uses, which makes this a direct translation of `getFluidState(pos).is(FluidTags.WATER)`. Flowing water matched under both forms, so that case is unaffected.

Builds on #3214. I will rebase once that lands.

## Testing

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved lava and water interactions by recognizing water through its fluid type.
  - Lava now correctly forms cobblestone or obsidian when water is adjacent, even if water is also directly beneath it.


RoosterBooster007:
Thanks, @pxperr!

PR #3216, opened 2026-09-03
feat(plugin): fire BlockFormEvent for lava conversions

pxperr:
## Description

`BlockFormEvent` exists in the plugin API and registers fine, but the only site in the server that ever constructed it was the blue ice branch in `FlowingLava::receive_neighbor_fluids`. A plugin could therefore cancel basalt formation and nothing else, while the conversions that actually come up in normal play - the cobblestone, obsidian and stone that lava and water produce - happened with no way to see or stop them. A land claim plugin could not gate a cobblestone generator on someone else's claim, and the event's own doc comment already names cobblestone as an example.

All four conversions now go through `FlowingLava::form_block`, which fires the event, returns early if a handler cancelled it, and otherwise sets the block and plays the fizz. That covers the cobblestone and obsidian branch, the basalt branch, and the stone in `spread_to` where falling lava lands in water. The basalt path is behaviourally unchanged, it just moved into the helper rather than keeping its own copy of the fire and cancel wiring.

This matches where Bukkit puts the event. `BlockFormEvent` is documented as covering obsidian and cobblestone forming on contact with water, and SPIGOT-2886 settled that it fires for obsidian, cobblestone and stone alike, with no block forming when it is cancelled. Cancelling here skips the `set_block_state` and the fizz together, so a cancelled formation is silent as well as absent.

A cancelled event leaves the fluid as it is and the surrounding call still returns as though the conversion had happened, so the lava does not spread into that block on this pass either. The conversion is reconsidered on the next neighbour update or fluid tick, which keeps a cancelling plugin from having to hold the block down every tick.

Fired from the lava conversions only. Ice, snow and the other natural formations still do not construct the event, so those remain invisible to plugins and are worth a separate change.

Builds on #3214 and #3215, which correct the same call sites. I will rebase this once those land.

## Testing

Running on a dev server with nine plugins loaded, none of which subscribe to `BlockFormEvent`, so the uncancelled path is what is exercised there. The fire site follows the same `fire_blocking` and `event.cancelled` shape the branch it replaces already used.

### DIFF
--- a/crates/pumpkin/src/block/fluid/lava.rs
+++ b/crates/pumpkin/src/block/fluid/lava.rs
@@ -9,6 +9,7 @@ use pumpkin_data::{
     damage::DamageType,
     dimension::Dimension,
     fluid::{Falling, Fluid, FluidProperties, Level},
+    tag::Taggable,
     world::WorldEvent,
 };
 use pumpkin_util::math::position::BlockPos;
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Both #3214 and #3215 were merged. Please fix conflicts.

PR #3217, opened 2026-09-03
fix(bedrock): support 1.26.45 connections

ZlordHUN:
## Description
Fixes Bedrock 1.26.45 clients being rejected during network negotiation and remaining stuck on “Locating Server” after connecting because initial chunk batches were never recorded as sent.

## What
- Updates supported Bedrock version from 1.26.40 to 1.26.45.
- Updates the network protocol from 2168 to 2169.
- Correctly records Bedrock chunk batches as sent.
- Prevents the same initial chunks from being encoded and sent repeatedly.
- Allows the server to send `PlayerSpawn` after enough initial chunks have been dispatched.
- Adds regression coverage for protocol negotiation and Bedrock chunk bookkeeping.

## How
- Updates the central Bedrock version, protocol, status advertisement, package metadata, and internal version mapping.
- Keeps the existing packet serializers because protocol 2169 retains the protocol 2168 packet IDs and base layouts.
- Adds `ChunkSender::commit_bedrock_batch` to move dispatched chunks from pending to sent while validating the current chunk-send epoch.
- Uses the committed chunk count to complete the Bedrock spawn sequence.
- Tests current, older, and newer protocol negotiation as well as repeated and stale chunk-batch commits.

## Testing
- `cargo fmt --all --check`
- `cargo check --workspace --all-targets`
- `cargo test -p pumpkin bedrock_batch --lib`
- `cargo test -p pumpkin accepts_only_the_current_bedrock_protocol --lib`
- `cargo test -p pumpkin-util resolves_bedrock_26_45_protocol`
- `cargo test -p pumpkin-protocol formats_vanilla_26_45_advertisement`
- Verified in game that a Bedrock 1.26.45 client can connect, finish locating the server, and enter the world.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3454,7 +3454,7 @@ dependencies = [
 
 [[package]]
 name = "pumpkin"
-version = "0.1.0-dev+26.2-26.40"
+version = "0.1.0-dev+26.2-26.45"
 dependencies = [
  "aes 0.9.3",
  "arc-swap",
@@ -3523,7 +3523,7 @@ dependencies = [
 
 [[package]]
 name = "pumpkin-api-macros"
-version = "0.1.0-dev+26.2-26.40"
+version = "0.1.0-dev+26.2-26.45"
 dependencies = [
  "proc-macro-error2",
  "proc-macro2",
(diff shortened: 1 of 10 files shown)


QuantumSegfault (approved):
Looks good.

Will test ASAP

PR #3219, opened 2026-09-04
fix: install a rustls CryptoProvider at startup

Stilldrop1:
Ran into a server crash while building a plugin that makes outbound HTTPS requests — the very first request anywhere in the process brings the whole server down, not just the plugin.

Turns out both `ring` and `aws-lc-rs` end up enabled in the dependency graph (different things pull in each one), and rustls won't pick a default crypto backend on its own when that happens. Nothing in the codebase was ever calling `CryptoProvider::install_default()`, so whatever hits TLS first is gambling on it.

The annoying part is it's timing-dependent, which is probably why it hasn't been super obvious: if a `reqwest` client happens to get built somewhere else first (like Mojang session auth when a player joins in online-mode), that path resolves the ambiguity as a side effect and everything's fine from then on. But if a plugin's outbound HTTP call is the first thing in the process that ever needs TLS — say, nobody's joined yet and a plugin tries to reach an API right after boot — the process just panics and dies.

This installs the `aws-lc-rs` provider explicitly at the very top of `main()`, before anything else in the server can possibly touch TLS, so it's no longer a race.

Built this locally and actually tested it before opening this — spun up an isolated server with the patched binary, zero players online (the exact conditions that used to crash it), and fired off a couple of real outbound HTTPS calls through a plugin. Both went through clean, server stayed up the whole time. Confirmed fixed, not just theorized.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3497,6 +3497,7 @@ dependencies = [
  "rsa",
  "rustc-hash",
  "rustc_version_runtime",
+ "rustls",
  "rustyline",
  "serde",
  "serde_json",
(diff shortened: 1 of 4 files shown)


RoosterBooster007:
Closing, as it's covered in #3202.

PR #3221, opened 2026-09-04
fix(entity): play death sound on lethal hit and send Poof when removing a corpse

### LINKED ISSUE #3121: Mobs play their death sound twice when dying.
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

9e9f1b5/release

### Operating System

CachyOS Linux

When killing a mob (cow, pig, etc.), their death sound plays when they die then about 500ms after again. The sound should only play once.

### Steps to Reproduce

1. Find or summon a mob.
2. Kill it
3. Their death sound should play twice

xudoyberdi0410:
## What was changed?
- `LivingEntity::damage_with_context` now plays the entity's death sound instead of its hurt sound when the hit is lethal.
- Added `death_sound_for_entity`, which resolves `entity.<name>.death` by the entity's resource name (with the vanilla aliases: mooshroom→cow, cave_spider→spider, giant→zombie, pufferfish→puffer_fish, trader_llama→llama) and falls back to `entity.generic.death`.
- The corpse removal tick (`death_time == 20`) now sends `EntityStatus::Poof` (60) instead of a second `EntityStatus::Death` (3), matching vanilla `LivingEntity.tickDeath`.
- Unit tests for the new lookup (regular, alias, and fallback cases).

## Why?
Fixes #3121. The issue is titled "death sound plays twice", but on a 26.2 client there is no death sound at all: `ClientLevel.playSeededSound` only plays sounds for the local player, so entity status 3 only triggers the death animation. In vanilla the death sound comes from the server (`LivingEntity.hurtServer` → `makeSound(getDeathSound())`). Pumpkin played the hurt sound on the lethal hit and never the death sound, and sent status 3 twice. On older clients the second status 3 produced the double sound from the report.

## Impact
Mobs play their own death sound once. Corpses now emit the vanilla poof particles on removal.

## Known limitations
Small slimes use the regular slime death sound (no size-aware `death_small` variant yet). Magma cubes are unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added entity-specific death sounds, including support for several entity name variants.
  * Added clearer audio feedback when entities are hurt or defeated.

* **Bug Fixes**
  * Corrected the visual effect shown when defeated entities are removed, ensuring the appropriate disappearance effect is displayed.

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -237,6 +237,18 @@ impl LivingEntity {
         entity_type.hurt_sound.unwrap_or(Sound::EntityGenericHurt)
     }
 
+    fn death_sound_for_entity(entity_type: &'static EntityType) -> Sound {
+        let name = match entity_type.resource_name {
+            "mooshroom" => "cow",
+            "cave_spider" => "spider",
+            "giant" => "zombie",
+            "pufferfish" => "puffer_fish",
+            "trader_llama" => "llama",
+            other => other,
+        };
+        Sound::from_name(&format!("entity.{name}.death")).unwrap_or(Sound::EntityGenericDeath)
+    }
+
     pub fn new(entity: Entity) -> Self {
         let water_movement_speed_multiplier = if entity.entity_type == &EntityType::POLAR_BEAR {
             0.98
@@ -1847,7 +1859,7 @@ impl LivingEntity {
 
             self.update_death_stats(&*dyn_self, killer);
 
-            // Plays the death sound
+            // Plays the death animation
             world.send_entity_status(&self.entity, EntityStatus::Death, Some(ActorEventID::Death));
             let looting_level;
             let tool = if let Some(cause_ent) = cause {
@@ -2944,6 +2956,7 @@ impl LivingEntity {
             position,
         );
 
+<<<<<<< HEAD
         if play_sound {
             world.play_sound(
                 self.hurt_sound(),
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Hey @xudoyberdi0410, thanks for digging into this! Spawning the poof particle on corpse removal matches vanilla's `tickDeath` perfectly, nice catch. 

For the sound lookup, hardcoding `entity.<name>.death` with an alias table misses a few edge cases (e.g. giants use `entity.hostile.death`, armor stands use `entity.armor_stand.break`). We actually have a data-driven `death_sound` field and pitch randomization landing in #3283. 

Could you rebase on main (that area shifted in #3180) and drop the custom resolver in favor of the data field? That way we can keep this PR focused on the lethal-hit swap and the poof particle fix.

Developers13:
Heads-up that I opened #3408, which overlaps with the Poof-on-removal part of this PR.

While playtesting master (`22f235e`) I ran into a corpse that never despawned: `death_time` can move past 20 (e.g. a value restored from NBT), so the `time == 20` exact comparison never fires and the removal is skipped entirely. #3408 therefore uses `death_time >= 20` (matching vanilla's `tickDeath`) in addition to sending `EntityStatus::Poof` instead of re-sending the death status.

The lethal-hit death sound here is an independent, complementary piece that #3408 does not cover. Since this PR is currently conflicted with master and has failing checks, it may be worth rebasing it on top of #3408 (or coordinating on which one lands) — happy to close #3408 in favor of this one if maintainers prefer keeping the death-sound part here.

RoosterBooster007:
@xudoyberdi0410, thanks for sticking with this one. The branch ended up half-rebased, though (there are still `<<<<<<<` markers committed in `damage_with_context` and `death_sound_for_entity` now collides with the one #3283 merged), so it won't build. Please strip this to just the lethal-hit swap, drop the committed markers, and it should be good to go.

PR #3222, opened 2026-09-04
perf(bedrock): prevent Bedrock join-time entity overload

ZlordHUN:
## Description
Fixes severe Bedrock join-time slowdowns caused by duplicate persisted entity records and unbounded initial terrain and actor synchronization. Existing affected worlds are repaired through normal chunk loading and saving without requiring manual save edits.

## What
- Replaces persisted chunk entity snapshots instead of appending another copy during every save.
- Deduplicates entity UUIDs within and across chunks while preserving unknown entity data.
- Limits Bedrock terrain delivery to four chunks per batch with one batch in flight.
- Limits actor lifecycle synchronization to 64 packets per player per tick.
- Retries terrain and actor packets when the outbound queue is full.
- Waits for terrain and PlayerSpawn completion before sending actors.
- Prevents stale chunk and actor data after movement, teleportation, or dimension transfers.

## How
- Tracks whether chunk entity data has been materialized and whether its persisted snapshot requires rewriting.
- Rebuilds authoritative entity snapshots from the current live state.
- Uses per-chunk and world-level UUID tracking to discard duplicate persisted records.
- Adds completion-aware Bedrock chunk batching with generation tokens and world epochs.
- Uses ordered packet reservations for packet groups that must remain atomic.
- Queues actor spawns and removals in bounded FIFO lifecycle queues, prioritizing removals.
- Serializes chunk tickets, watcher transitions, and client movement during world transfers.

## Testing
- `cargo check -p pumpkin --tests`
- `cargo test -p pumpkin-world --lib`
- Started Pumpkin with the affected world and confirmed that duplicate UUID records were discarded during chunk loading while a client connected successfully.

### DIFF
--- a/crates/pumpkin-world/src/chunk/mod.rs
+++ b/crates/pumpkin-world/src/chunk/mod.rs
@@ -9,10 +9,10 @@ use pumpkin_data::{Block, BlockState, BlockStateId};
 use pumpkin_nbt::compound::NbtCompound;
 use pumpkin_util::math::position::BlockPos;
 use rustc_hash::{FxHashMap, FxHashSet};
+use uuid::Uuid;
 
 use std::sync::RwLock;
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::AtomicU64;
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use thiserror::Error;
 
 pub mod format;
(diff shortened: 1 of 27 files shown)


crates/pumpkin-world/src/chunk/mod.rs
```diff
@@ -92,6 +92,131 @@ pub struct ChunkEntityData {
...
+        let mut data = self
+            .data
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        let previous_len = data.len();
+        data.retain(|entity| {
+            entity
+                .get_uuid("UUID")
+                .is_none_or(|uuid| !live_uuids.contains(&uuid))
+        });
```
Phoenixxo: I think this can leave stale entity NBT behind. If an entity gets reconciled into an unmaterialized chunk and is later removed, its UUID is no longer in live_uuids, so the old record is kept and could respawn later.

crates/pumpkin/src/net/java/play/player_position.rs
```diff
@@ -94,6 +97,9 @@ impl JavaClient {
             };
 
             'after: {
+                let Some(movement_guard) = player.lock_client_movement_if_loaded() else {
+                    return;
+                };
```
Phoenixxo: Might be mistaken but we should recheck the chunk epoch or teleport state after the move event. A plugin could teleport the player during the event, then this handler resumes and overwrites the teleport with `event.to`.

crates/pumpkin/src/net/java/play/player_position.rs
```diff
@@ -239,6 +257,9 @@ impl JavaClient {
             );
 
             'after: {
+                let Some(movement_guard) = player.lock_client_movement_if_loaded() else {
+                    return;
+                };
```
Phoenixxo: Same issue as above

crates/pumpkin/src/world/entity_tracker.rs
```diff
@@ -26,13 +31,79 @@ use crate::net::java::JavaClient;
 use crate::world::World;
 use crate::world::chunker::{get_view_distance, is_within_view_distance};
 
+/// Actor packets share Bedrock's bounded normal queue with terrain. A small per-player
+/// budget prevents a dense entity chunk from filling that queue in a single world tick.
+const BEDROCK_ACTOR_PAIRINGS_PER_TICK: usize = 64;
```
Phoenixxo: This seems to limit actor pairings rather than packets. Player pairing sends both PlayerList and AddPlayer, so 64 pairings could still enqueue 128 packets. Is that intended?

Phoenixxo (commented):
overall, tests may be better off with a bigger test running the full feature flow. Testing the smaller pieces is okay but integration tests covering everything I think gives more confidence in the changes rather than testing it seperately

Snowiiii:
This looks way not too vanilla

ZlordHUN:
@Snowiiii could you elaborate please? Don't think I can keep it vanilla bedrock server logic and fix the performance issues here but I can look into it further if you elaborate :D

ZlordHUN:
Started a new Pumpkin world. Will revisit this issue at another time if people start complaining.

PR #3230, opened 2026-09-04
perf(nbt): read Bedrock numeric arrays in bulk

nobuildersnotools:
Bedrock version of #3178 

Included benchmark (nbt/bedrock/deserialize/raw):
Before: 6.2028 µs
After: 5.6588 µs


ZlordHUN:
LGTM

PR #3235, opened 2026-09-04
fix(portal): load candidate chunks before validating portal search results

MegalithOfficial:
## Description

Every trip through a Nether portal was creating another portal on the other side instead of reusing the existing one. It happened in both directions.

`NetherPortal::search_for_portal` gets its candidates from the portal POI storage, which is disk backed, so finding the portal itself worked. The problem was the validation step. Each candidate was checked with `world.get_block_and_state`, and that read returns `VOID_AIR` if the chunk isn't loaded. Once you leave a dimension those chunks can unload, so the existing portal would fail validation and `find_safe_location` would build a new one next to it.

Vanilla's `PortalForcer.findClosestPortalPosition` ends up loading the chunk when it calls `level.getBlockState`. I made that explicit here. After collecting the POI candidates, the search now loads a 3x3 area around each candidate chunk before checking the blocks, using the same `block_in_place` pattern already used by the fallback path.

I also changed the Nether search radius from 128 to 16 to match vanilla's `NETHER_PORTAL_RADIUS`. The search no longer rejects portals above the Nether's logical height either. Vanilla searches the full build height and only applies the logical-height clamp when creating a new portal, so `find_safe_location` still keeps that clamp.

## Testing
Tested this in game on a release build. I lit a portal, crossed over, came back, then crossed again, and it reused the same portal every time. Before this fix, each trip would create another portal.

I also tested a portal on the Nether roof and it links correctly in both directions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Nether portal searches by loading nearby terrain before evaluating candidate locations.
  * Updated vertical search handling to cover the full dimension height.
  * Reduced the Nether portal search radius for more targeted results.


RoosterBooster007:
Looks good. Thanks, @MegalithOfficial!

PR #3237, opened 2026-09-04
fix(loot): evaluate entity property conditions (#3225) 🤖🤖🤖

### LINKED ISSUE #3225: Sheep drop wool of a random color instead of their own
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.45 (b5b9b9d70)

### Operating System

CachyOS Linux (kernel 7.1.8)

Killing a sheep drops wool of a **random** color instead of the sheep's own color: a white sheep can drop purple wool, a black one can drop yellow wool, and so on. Each of the 16 colors has the same 1/16 chance. Mutton drops correctly.

Shearing is not affected: shears are handled directly in `SheepEntity::mob_interact` with a hardcoded color-to-item mapping rather than through the `shearing/sheep` loot table, and they yield the correct color (including after dyeing the sheep). Only the death drop goes through the loot table.

The `sheared: false` guard is dropped along with the color one, so a sheep that has already been sheared still drops wool when killed.

Expected (vanilla): a sheep drops one wool of its own color, and a sheared sheep drops no wool at all.

**Cause**

Vanilla's `entities/sheep.json` puts a `minecraft:alternatives` entry in the second pool with 16 branches. Each branch is guarded by

```json
{
  "condition": "minecraft:entity_properties",
  "entity": "this",
  "predicate": {
    "minecraft:components": { "minecraft:sheep/color": "white" },
    "minecraft:type_specific/sheep": { "sheared": false }
  }
}
```

and points at a per-color sub-table (`entities/sheep/white` and friends).

`parse_condition()` in `tools/pumpkin-codegen/src/loot_table.rs` has no arm for `minecraft:entity_properties`, so every one of those conditions falls through to the catch-all `_ => LootCondition::None` — "no condition". The 16 branches are then flattened into 16 equally weighted entries, and the pool picks one at random:

```rust
// crates/pumpkin-data/src/generated/loot_table.rs
static ENTITIES_SHEEP_POOL1_ENTRIES: &[LootEntry] = &[
    LootEntry { item: "minecraft:white_wool",   weight: 1i32, condition: LootCondition::None, .. },
    LootEntry { item: "minecraft:orange_wool",  weight: 1i32, condition: LootCondition::None, .. },
    LootEntry { item: "minecraft:magenta_wool", weight: 1i32, condition: LootCondition::None, .. },
    // …all 16 colors, all unconditional…
];
```

Fixing this needs the color/sheared predicate to be modelled in `LootCondition` (`crates/pumpkin-util/src/loot_table.rs`), evaluated in `check_condition()` (`crates/pumpkin/src/world/loot.rs`), and the sheep's color and sheared state to reach `LootContextParameters` when the entity dies (`crates/pumpkin/src/entity/living.rs`).

Sheep is currently the only entity table using `minecraft:alternatives`, but the same unparsed condition affects 77 of the 1356 vanilla loot tables in total.

### Steps to Reproduce

1. Join a server (Java 26.2 client).
2. `/summon minecraft:sheep` a few times, or find sheep in the world.
3. Kill them and look at the dropped wool: the color does not match the sheep, and differs from kill to kill.
4. Shear a sheep, then kill it: it still drops wool.

Related: #2769 and #2359 both report that sheep drop no wool at all. That was the earlier symptom of the same code path; after `alternatives` began to be expanded, the drop appeared but with the wrong color.

BatuhanAksoyy:
## Description

Adds generic support for primitive `minecraft:entity_properties` loot conditions and fixes #3225.

- Preserve entity-property targets and flattened predicate values during loot-table generation.
- Keep non-tool child conditions when flattening `minecraft:alternatives` entries.
- Add a generic entity-property snapshot and matcher to the loot context.
- Let mobs publish loot properties without concrete entity casts in the generic loot evaluator.
- Publish sheep color and sheared state so only the matching wool entry is eligible on death.
- Regenerate the loot-table data and add regression coverage for every sheep color and the sheared state.

Entity-property evaluation is opt-in: entities that do not yet publish a property snapshot retain the previous permissive behavior. This lets additional entity predicates use the same generic path incrementally without adding entity-specific logic to the loot evaluator.

Closes #3225.

## Testing

- `cargo test`
- `cargo run -p pumpkin-codegen -- loot_table`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added support for entity-property conditions in loot tables for the current entity, attacker, and direct attacker.
  * Loot conditions can evaluate Boolean, numeric, and text properties, including entity types and tags.
  * Mob-specific information can now influence loot outcomes.
  * Sheep loot reflects wool color and sheared status, producing matching wool drops or no wool when sheared.

* **Tests**
  * Added coverage for property matching, condition inheritance, entity-type and tag matching, and sheep loot behavior.


andromeda14layker-gif (approved):

andromeda14layker-gif (approved):

andromeda14layker-gif (approved):

RoosterBooster007:
Looks good. Just address the CodeRabbit thread in the codegen: the `new child_cond` branch doesn't change anything when the parent condition is `None` (the recursive call re-combines the child's conditions anyway) and could drop an inherited condition where it isn't, so either drop it or add the regression case it asked for.

BatuhanAksoyy:
I'll do it as soon as I'm home.

BatuhanAksoyy:
@coderabbitai full review

BatuhanAksoyy:
@coderabbitai resume

BatuhanAksoyy:
@coderabbitai full review

PR #3248, opened 2026-09-04
feat(block): grow trees from saplings and fix leaf decay drops

MegalithOfficial:
## Description

There were three separate problems around trees: saplings never grew, decayed leaves stayed visible on the client after being removed on the server, and leaf drops always gave the same items instead of using the loot table chances.

The leaf decay issue was just the update flags. `LeavesBlock::random_tick` called `break_block` with `BlockFlags::empty()`, so the block disappeared server side without sending an update to the client. Vanilla removes the block with a normal update, so this now uses `NOTIFY_ALL`.

The drop issue came from loot table codegen. Leaf loot tables use `minecraft:table_bonus` for saplings, sticks, and apples, but the codegen didn't handle that condition and emitted `LootCondition::None` instead. I added `LootCondition::TableBonus`, including its fortune chance list, and implemented the same `chances[min(level, len - 1)]` vs `nextFloat` check vanilla uses.

I also fixed two condition folding bugs while working on this. Children of `alternatives` entries could lose their own conditions when a tool negation was added, and multiple entry conditions were effectively last-wins and could disappear under the same tool handling. They now fold into `AllOf`, so the generated oak sapling condition keeps the silk/shears check, explosion survival, and table bonus together. `loot_table.rs` is the only regenerated file in this PR.

Sapling growth now reuses the existing worldgen tree features at runtime. `WorldGenerationCache` implements `GenerationCache` over a live `World`, with an overlay and buffered writes, then applies the generated blocks with `NOTIFY_ALL` and adds any block entities.

`TreeGrower` follows vanilla's grower table and selection rules: oak with the 9:1 normal/fancy split, flower-based bee variants, mega spruce and jungle trees, 2x2-only dark and pale oak, mangrove with the 15:85 normal/tall split, cherry and birch bee variants, and azalea. Mega trees use the same 2x2 search order, remove the saplings before placement, and restore them if generation fails.

`SaplingBlock` random ticking now requires light level 9 above the sapling and uses the vanilla 1-in-7 growth roll. Bonemeal checks `canGrow` and build height, and mangrove propagules and azalea use the same grower path.

## Testing

Play tested on a release build.

Bone meal grows oak, spruce, birch, 2x2 dark oak, and occasional fancy oak. A single dark oak sapling refuses bone meal, and saplings in the dark do not grow.

After cutting a tree, decaying leaves disappear correctly on the client. Most leaves drop nothing, saplings show up around the expected one-in-twenty range, and sticks drop occasionally instead of every time.

### DIFF
--- a/crates/pumpkin-data/src/generated/loot_table.rs
+++ b/crates/pumpkin-data/src/generated/loot_table.rs
@@ -792,7 +792,15 @@ static BLOCKS_ACACIA_LEAVES_POOL0_ENTRIES: &[LootEntry] = &[
         weight: 1i32,
         min_count: 1i32,
         max_count: 1i32,
-        condition: LootCondition::NoSilkTouchOrShears,
+        condition: LootCondition::AllOf(&[
+            LootCondition::NoSilkTouchOrShears,
+            LootCondition::AllOf(&[
+                LootCondition::SurvivesExplosion,
+                LootCondition::TableBonus {
+                    chances: &[0.05f32, 0.0625f32, 0.083333336f32, 0.1f32],
+                },
+            ]),
+        ]),
         bonus_formula: None,
     },
 ];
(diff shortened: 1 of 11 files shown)


Snowiiii:
Thank you @MegalithOfficial :D

PR #3252, opened 2026-09-05
fix(entity): scope projectile hit sounds to the chunk

BryanFRD:
## Problem

`ArrowEntity::on_hit` and `TridentEntity::on_hit` build the same sound packet in both
match arms, then send it to two different audiences.

The block arm scopes it to the chunk:

```rust
let chunk_pos = entity.chunk_pos.load();
world.broadcast_to_chunk(chunk_pos, &sound_packet);
```

The entity arm, in the same function, broadcasts it to everyone:

```rust
world.broadcast_packet_all(&sound_packet);
```

Same sound, same category, same random pitch, so the split reads as a copy-paste slip
rather than a decision. Vanilla routes both through `ServerLevel.playSound`, which is
distance limited either way, so an arrow hitting a mob is currently audible across the
whole world.

The cost scales with the player count: one packet per connected player per hit. Skeleton
farms and PvP volleys are where that shows up.

## Change

Use the chunk broadcast in the entity arm too, matching the block arm two match arms away.
`entity` is already bound before the match in both files, so nothing else moves.

## Tests

None added. The change swaps one broadcast helper for another that is already used for the
same packet in the same function; there is no new logic to assert on, and no seam that
would let a test observe recipient selection without a running server.

## Summary by CodeRabbit

* **Bug Fixes**
  * Projectile hit sounds are now played only for players in the affected area, reducing unrelated sounds for players elsewhere in the world.


RoosterBooster007:
Thanks, @BryanFRD!

PR #3253, opened 2026-09-05
chore(item): drop stale tool break TODOs

BryanFRD:
## Problem

Five sites carry the same marker above a `let _ = ...damage_item(...)`:

```rust
// TODO: Handle DamageResult::Broken to broadcast item break and update player slot.
```

The work it asks for is already done, one level up. Every path that reaches these
handlers clones the stack, calls the handler, and compares afterwards:

- `use_item_on.rs` covers `shovel`, `axe`, `hoe` and `flint_and_steel`
- the Bedrock `inventory_action.rs` path covers the same four
- `interact.rs` covers the creeper

Each one increments the `Broken` statistic, sends the entity status that plays the break
particles, and syncs the slot back to the client. The marker describes exactly that, so it
now reads as a known gap when nothing is missing, and the next person to look will
re-derive the same conclusion I did.

## Change

Delete the five comments. No code moves.

## Note

There is a real inconsistency nearby, left alone here because it is a design call rather
than a cleanup: `PlayerItemDamageEvent` and `PlayerItemBreakEvent` fire only from
`Player::damage_item_in_slot`, which these paths do not use. A plugin sees shears break on
a sheep but not an axe wear out stripping logs. Happy to open an issue for that if it is
worth tracking.

### DIFF
--- a/crates/pumpkin/src/entity/mob/creeper.rs
+++ b/crates/pumpkin/src/entity/mob/creeper.rs
@@ -228,7 +228,6 @@ impl Mob for CreeperEntity {
         entity.set_synced_data(pumpkin_data::tracked_data::creeper::IS_IGNITED, true);
 
         if player.gamemode.load() != pumpkin_util::GameMode::Creative {
-            // TODO: Handle DamageResult::Broken to broadcast item break and update player slot.
             let _ = item_stack.damage_item(1);
         }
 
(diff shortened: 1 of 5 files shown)


Snowiiii:
LGTM, Thank you @BryanFRD :D

PR #3256, opened 2026-09-05
fix(block): concrete powder water-only solidify + landing solidify

### LINKED ISSUE #3200: fix(block): concrete powder solidifies next to lava, and not when falling into water
`3a3ba2c87` added concrete powder gravity and solidification in `crates/pumpkin/src/block/blocks/falling.rs`. Two cases are off compared to vanilla:

**1. Lava solidifies concrete powder**

`FallingBlock::touches_liquid` accepts any liquid neighbour:

```rust
let neighbor = world.get_block_state(&pos.offset(dir.to_offset()));
if neighbor.is_liquid() {
    return true;
}
```

Vanilla `ConcretePowderBlock.touchesLiquid` only accepts water (`FluidTags.WATER`), so placing concrete powder next to lava turns it into concrete on the server but not in vanilla. Checking the fluid at the neighbour against `tag::Fluid::MINECRAFT_WATER` instead of `state.is_liquid()` fixes it.

**2. Concrete powder that falls into water lands as powder**

Solidification is only checked in `on_place` and `get_state_for_neighbor_update`. Vanilla also checks it in `FallingBlockEntity.tick`: a falling concrete powder that lands in water becomes concrete. Right now it lands as powder and only converts if a later neighbour update happens to fire.

Vanilla `shouldSolidify` is `canSolidify(state at the block's own position) || touchesLiquid(neighbours except DOWN)`; the own-position half is what covers the falling case.

Kurenaiiiii:
## What was changed?

**`crates/pumpkin/src/block/blocks/falling.rs`:**
- Added `FallingBlock::can_solidify(state: &BlockState)` which returns true only for water: `state.is_waterlogged() || Fluid::from_state_id(state.id).is_some_and(|f| f.has_tag(&tag::Fluid::MINECRAFT_WATER))`.
- Changed `FallingBlock::touches_liquid(world, pos)` to call `Self::can_solidify(neighbor)` instead of `neighbor.is_liquid()`. It still iterates `BlockDirection::all()` and skips `Down`.
- Added `FallingBlock::should_solidify(world, pos)` = `can_solidify(own pos) || touches_liquid(neighbors)`, matching vanilla `shouldSolidify`.
- `on_place` and `get_state_for_neighbor_update` are unchanged except they now benefit from the water-only `touches_liquid` fix.

**`crates/pumpkin/src/entity/falling.rs`:**
- In `FallingEntity::tick`, on the `on_ground` landing branch, before `set_block_state`, resolve the falling `BlockStateId` to `Block` via `Block::from_state_id`. If it has `tag::Block::MINECRAFT_CONCRETE_POWDERS` and `FallingBlock::should_solidify(&**world, &landing_pos)` is true, convert `*_powder` → concrete via `strip_suffix("_powder")` + `Block::from_name` (same conversion as `on_place`), then place the concrete id instead of powder.

Root cause 1: `touches_liquid` used `BlockState::is_liquid()`, which is true for both water and lava. Vanilla `ConcretePowderBlock.touchesLiquid` / `canSolidify` only accepts `FluidTags.WATER`. Concrete symptom: placing white concrete powder next to lava turned it into white concrete on Pumpkin, while vanilla leaves it as powder.

Root cause 2: solidification was only checked in `on_place` and `get_state_for_neighbor_update`. There was no check on the falling-entity path, so a powder block spawned via `FallingEntity::replace_spawn`, falling through water and landing (e.g. dropped from height into a 1-deep pool over stone), landed as loose powder via unconditional `set_block_state(self.block_state_id)`. Vanilla handles this in `FallingBlockEntity.tick` + `ConcretePowderBlock.onLand` via `shouldSolidify = canSolidify(own) || touchesLiquid(except DOWN)` — the own-position half is what covers landing *in* water.

From a player perspective:
- Powder next to lava no longer solidifies. Powder next to / in water still solidifies.
- Powder falling into water now lands as concrete instead of powder waiting for a later neighbor update.

Backward compatibility:
- Non-powder falling blocks (sand, gravel, red sand, suspicious sand/gravel) are unaffected — the landing check is gated by `MINECRAFT_CONCRETE_POWDERS`.
- The already-correct placed-next-to-water path still uses the same `touches_liquid` (now water-only) and keeps working.
- No change to `can_fall_through` or fall physics.

- No unit/integration tests added — there is no existing test harness for `FallingBlock` / `FallingEntity` landing, and this was verified by code inspection against vanilla plus build/lint.
- Commands actually run:
  - `cargo fmt -- crates/pumpkin/src/block/blocks/falling.rs crates/pumpkin/src/entity/falling.rs`
  - `cargo clippy -p pumpkin --lib`
  - `cargo clippy -p pumpkin --lib --message-format=short`
  - `cargo clippy -p pumpkin --lib -- -D warnings`
  - `cargo check -p pumpkin`, `cargo check -p pumpkin-data`, `cargo check -p pumpkin-world`
  - `cargo build --release`
- `cargo test` was not run.
- Out of scope, intentionally not included: vanilla `touchesLiquid` also checks `!isFaceSturdy(...)` for waterlogged neighbors, and vanilla `FallingBlockEntity.tick` does a high-velocity (`>1.0`) `SOURCE_ONLY` raycast to catch fast-falling through water mid-air — this PR only adds the landing-position `should_solidify` check to fix the reported landing case with a minimal diff.

Fixes #3200

## Summary by CodeRabbit

* **New Features**
  * Concrete powder blocks now solidify into concrete when landing on a solid surface.
  * Falling blocks can detect nearby water and waterlogged conditions when determining whether to solidify.


RoosterBooster007:
Looks good. Thanks, @Kurenaiiiii!

PR #3258, opened 2026-09-05
refactor(WASM): Unified resource handling & cleanup

PigTurtleDev:
## Description
Adds a `FromResource` trait that strongly ties `wasmtime::Resource<T>` to an internal type, then refactors all wasm-host impls to go through methods generic over `FromResource`.

This should make it impossible to accidentally create wrongly typed resources, and in the future should prevent deleting non-owned resources.
(That's not the case yet for `wasm_host::wit::v0_1::events`, there will be a follow-up PR for those)

## Changes
- Replaces the separate `get_X`, `get_X_mut`, `add_X` with one generic method `get`, `add`, `take`
- Removes the `ResourceProvider<T>` struct in favor of directly using the `T` as our internal representation
- implements `FromResource` for every WASM resource
  - Changes the internal representation of specific BlockEntities from `Arc<dyn BlockEntity>` to `Arc<T>` for the actual BlockEntity type `T`, removing the need to downcast per-call
- Properly deletes (`take()`s) Resource handles passed back from the guest that said guest can no longer access

Unless explicitly stated in the wit via `borrow<some-resource-type>`, all resources passed back to the host are owned[^1]. When a guest passes an owned resource back to the host, it loses access to said resource. The host has to delete them, otherwise they continue taking up slots in the resource table & ram.

[^1]: the "self/this" parameter on resource-methods is always borrowed, except for `drop()`.

While I did manually go through every single host-trait method to check for missing clean-up, It's possible that I might have missed some Resources embedded in other structs. I can't think of a good way to check against that, other than maybe panicking in `.get<T: FromResource>()` if the passed Resource is owned (which wouldn't be correct, it's fine to access a resource from an owned handle without taking it, but realistically I don't think it happens).

## Todo
As previously mentioned, event cleanup still needs to be migrated to this new system. It currently "cleans" an `&Event`, and for correctness the new methods require an owned Resource handle value (that also was created through `Resource::new_own()`) for Resource deletion.

## Testing

- A very minimal plugin I wrote works just fine. I would appreciate if someone else could test their plugins too.

## Summary by CodeRabbit

- **Refactor**
  - Unified WASM plugin resource handling across servers, players, entities, inventories, events, forms, displays, and world features.
  - Resource creation, access, transfer, and cleanup now follow consistent lifecycle behavior.
  - Existing plugin operations and event data flows are preserved.
  - Removed legacy typed resource wrappers and specialized lookup methods in favor of generalized resource access.
  - Improved cleanup handling for plugin tasks, events, commands, and host-managed resources.

### DIFF
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
@@ -413,7 +413,7 @@ impl WasmPlugin {
                         store.data_mut().wasi_http_hooks.allow_outbound = allow_http_outbound;
                         store.data_mut().server = Some(server);
                         store.data_mut().name = Some(name);
-                        store.data_mut().add_context(context)
+                        store.data_mut().add(context)
                     })?;
 
                     guest
@@ -455,7 +455,7 @@ impl WasmPlugin {
             .shutdown(move |accessor| {
                 Box::pin(async move {
                     let (context_res, context_rep) = accessor.with(|mut store| {
-                        let resource = store.data_mut().add_context(context)?;
+                        let resource = store.data_mut().add(context)?;
                         let rep = resource.rep();
                         Ok::<_, wasmtime::Error>((resource, rep))
                     })?;
@@ -464,9 +464,9 @@ impl WasmPlugin {
                         .await
                         .map(|(result,)| result);
                     accessor.with(|mut store| {
-                        let _ = store.data_mut().resource_table.delete::<
-                            crate::plugin::loader::wasm::wasm_host::state::ContextResource,
-                        >(wasmtime::component::Resource::new_own(context_rep));
+                        let _ = store.data_mut().resource_table.delete::<Arc<Context>>(
+                            wasmtime::component::Resource::new_own(context_rep),
+                        );
                     });
                     result
                 })
@@ -496,12 +496,6 @@ impl WasmPlugin {
             })
             .await
     }
-}
-
-pub trait DowncastResourceExt<E> {
-    fn downcast_ref<'a>(&'a self, state: &'a mut PluginHostState) -> &'a E;
-    fn downcast_mut<'a>(&'a self, state: &'a mut PluginHostState) -> &'a mut E;
-    fn consume(self, state: &mut PluginHostState) -> E;
 }
 
 #[cfg(test)]
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
@@ -10,11 +10,7 @@ use wasmtime_wasi_http::{
 };
 
 use crate::{
-    entity::player::Player,
-    plugin::
-        loader::wasm::wasm_host::WasmPlugin
-    ,
-    server::Server,
+    entity::player::Player, plugin::loader::wasm::wasm_host::WasmPlugin, server::Server,
     world::World,
 };
 
@@ -184,6 +180,7 @@ impl PluginHostState {
 
     /// take the value of an **owned** [`Resource`] from the resource table
     pub fn take<T: FromResource>(&mut self, res: Resource<T>) -> wasmtime::Result<T::Internal> {
+        debug_assert!(res.owned());
         Ok(self.resource_table.delete(Resource::new_own(res.rep()))?)
     }
 
@@ -199,7 +196,9 @@ impl PluginHostState {
 
     pub fn discard_to_be_removed<T: FromResource>(&mut self, res: &Resource<T>) {
         assert!(res.owned());
-        let _ = self.resource_table.delete::<T::Internal>(Resource::new_own(res.rep()));
+        let _ = self
+            .resource_table
+            .delete::<T::Internal>(Resource::new_own(res.rep()));
     }
 }
 
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
@@ -179,6 +179,10 @@ impl PluginHostState {
     }
 
     /// take the value of an **owned** [`Resource`] from the resource table
+    #[expect(
+        clippy::needless_pass_by_value,
+        reason = "Our 'cloning' then deletion of the handle wouldn't be sound if it still exists"
+    )]
     pub fn take<T: FromResource>(&mut self, res: Resource<T>) -> wasmtime::Result<T::Internal> {
         debug_assert!(res.owned());
         Ok(self.resource_table.delete(Resource::new_own(res.rep()))?)
--- a/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
+++ b/crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
@@ -1,27 +1,20 @@
 use std::{
-    collections::HashMap,
     future::Future,
     sync::{Arc, Weak},
 };
 
-use pumpkin_util::text::TextComponent;
-use tokio::sync::Mutex;
-use wasmtime::component::ResourceTable;
+use wasmtime::component::{Resource, ResourceTable};
 use wasmtime_wasi::{WasiCtx, WasiCtxBuilder, WasiCtxView, WasiView};
 use wasmtime_wasi_http::{
     RequestOptions, WasiBody, WasiHttpCtx, WasiHttpCtxView, WasiHttpHooks, WasiHttpView,
 };
 
 use crate::{
-    command::CommandSender,
-    entity::EntityBase,
     entity::player::Player,
-    plugin::{
-        Context,
-        api::gui::PluginGui,
-        loader::wasm::wasm_host::{WasmPlugin, args::OwnedArg},
-    },
-    server::{RecipeManager, Server},
+    plugin::
+        loader::wasm::wasm_host::WasmPlugin
+    ,
+    server::Server,
     world::World,
 };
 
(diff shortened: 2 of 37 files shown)


Snowiiii:
@coderabbitai full review

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/living_entity.rs
```diff
@@ -412,15 +398,15 @@ impl HostLivingEntity for PluginHostState {
         slot: WitEquipmentSlot,
         stack: Option<Resource<WitHostItemStack>>,
     ) -> wasmtime::Result<()> {
-        let entity = living_entity_from_resource(self, &this)?;
+        let item_stack = if let Some(stack_res) = stack {
+            self.take(stack_res)?.lock().await.clone()
```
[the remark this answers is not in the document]
PigTurtleDev: The guest loses access to the ItemStack(resource) when it calls this method. It cannot reclaim or reuse it in any way, regardless of whether `&this` is stale/invalid.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/commands/executor.rs
```diff
@@ -121,12 +102,7 @@ impl CommandExecutor for WasmCommandExecutor {
                                 .await;
 
                             guest.with(|mut store| {
-                                let result = result
-                                    .map(|(result,)| map_command_result(store.data_mut(), result));
-                                remove_resource::<CommandSenderResource>(store.data_mut(), reps.0);
-                                remove_resource::<ServerResource>(store.data_mut(), reps.1);
-                                remove_resource::<ConsumedArgsResource>(store.data_mut(), reps.2);
-                                result
+                                result.map(|(result,)| map_command_result(store.data_mut(), result))
```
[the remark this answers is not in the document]
Phoenixxo: This is wrong, we don't want to delete the entries after every call. The handles are passed as owned in which the plugin retains. Returning does not give us back the ownership

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/mob.rs
```diff
@@ -241,37 +230,26 @@ impl CustomWasmGoal {
...
-                        guest.with(|mut store| {
-                            let _ = store.data_mut().resource_table.delete::<
-                                crate::plugin::loader::wasm::wasm_host::state::ServerResource,
-                            >(wasmtime::component::Resource::new_own(reps.0));
-                            let _ = store.data_mut().resource_table.delete::<
-                                crate::plugin::loader::wasm::wasm_host::state::EntityResource,
-                            >(wasmtime::component::Resource::new_own(reps.1));
-                        });
-                        result
+                            .await
```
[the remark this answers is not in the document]
Phoenixxo: Not wrong, but not necessary. Important to note that the plugin could have accumulated other resources before trapping. We should be doing cleanup at the store teardown level for cleanness and single responsibility purposes

crates/pumpkin/src/plugin/loader/wasm/wasm_host/mod.rs
```diff
@@ -413,7 +413,7 @@ impl WasmPlugin {
                         store.data_mut().wasi_http_hooks.allow_outbound = allow_http_outbound;
                         store.data_mut().server = Some(server);
                         store.data_mut().name = Some(name);
-                        store.data_mut().add_context(context)
+                        store.data_mut().add(context)
```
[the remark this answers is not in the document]
PigTurtleDev: Once again, this is correct.
Passing an owned Resource handle to the guest means relinquishing ownership of that handle. Our host has the responsibility of ensuring the handle points to a valid Resource until it is reacquired from the guest, which always happens in the form of calling back into the host (either passing it back as a function argument, or if it is dropped, the resource's destructor is called).
We also should not clean up resources in the case of a trapped guest. In that case, we delete the full resource table as is, there is no need to explicitly drop some of the resources it contains.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/state.rs
```diff
@@ -223,408 +159,60 @@ impl PluginHostState {
...
+        let _ = self.drop(res);
     }
 
-    pub fn get_interaction_entity_res<T>(
-        &self,
-        resource: &wasmtime::component::Resource<T>,
-    ) -> wasmtime::Result<&InteractionEntityResource> {
-        Ok(self
+    pub fn discard_to_be_removed<T: FromResource>(&mut self, res: &Resource<T>) {
+        assert!(res.owned());
```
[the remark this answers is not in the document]
PigTurtleDev: I don't believe this change to be necessary right now. The method will be removed in a follow-up PR that refactors event cleanup.
It's unlikely these methods will ever be called with borrowed resource handles, and if they are, that's a pretty bad logic error on our side.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
```diff
@@ -1189,9 +1142,9 @@ impl pumpkin::plugin::player::HostPlayer for PluginHostState {
         slot: u8,
         stack: Option<Resource<WitHostItemStack>>,
     ) -> wasmtime::Result<()> {
-        let player = player_from_resource(self, &player)?;
+        let player = self.get(&player)?;
         let stack = if let Some(stack_res) = stack {
-            self.get_item_stack(&stack_res)?.lock().await.clone()
+            self.get(&stack_res)?.lock().await.clone()
         } else {
```
Phoenixxo, line 1148: WIT transfers ownership of item stacks passed to the inventory setters, but all three setters still call get, these should use `self.take(&player)`. Same needs to happen in `set_inventory_item` and `set_ender_chest_item`

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
```diff
@@ -1273,9 +1220,9 @@ impl pumpkin::plugin::player::HostPlayer for PluginHostState {
         slot: u8,
         stack: Option<Resource<WitHostItemStack>>,
     ) -> wasmtime::Result<()> {
-        let player = player_from_resource(self, &player)?;
+        let player = self.get(&player)?;
         let stack = if let Some(stack_res) = stack {
-            self.get_item_stack(&stack_res)?.lock().await.clone()
+            self.get(&stack_res)?.lock().await.clone()
```
Phoenixxo, line 1225: Same as above, needs to use `take`.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
```diff
@@ -2247,8 +2186,8 @@ impl pumpkin::plugin::player::HostPlayer for PluginHostState {
         player: Resource<Player>,
         other: Resource<Player>,
     ) -> wasmtime::Result<()> {
-        let player = player_from_resource(self, &player)?;
-        let other = player_from_resource(self, &other)?;
+        let player = self.get(&player)?;
+        let other = self.get(&other)?;
```
Phoenixxo, line 2190: WIT declares `other: player`, rather than `borrow<player>`, so this handle is owned by the host after the call. `get` leaves its table entry behind. The same applies to `show_player`, `can_see`, and `can_see_player`. Should use `take(other)`.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
```diff
@@ -563,21 +559,18 @@ pub fn player_from_resource(
...
 }
 
 pub(crate) fn text_component_from_resource(
     state: &PluginHostState,
     text: &Resource<pumpkin::plugin::text::TextComponent>,
 ) -> pumpkin_util::text::TextComponent {
     state
-        .resource_table
-        .get::<TextComponentResource>(&Resource::new_own(text.rep()))
+        .get(text)
```
Phoenixxo, line 572: This helper clones the text component without removing its table entry. That leaves owned arguments unconsumed in callers such as `set_display_name`, `set_tab_list_name`, and `send_system_message`. Should use `take(text)` instead.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
```diff
@@ -586,10 +579,8 @@ fn world_from_resource(
     world: &Resource<pumpkin::plugin::world::World>,
 ) -> std::sync::Arc<crate::world::World> {
     state
-        .resource_table
-        .get::<WorldResource>(&Resource::new_own(world.rep()))
+        .get(world)
```
Phoenixxo, line 582: `teleport` and `teleport_world` receive an owned `%world` argument in WIT. This helper clones the world but leaves the transferred handle in the table. Consume w/ `take` and continue to pass the resulting `Arc<World>`.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
Phoenixxo, line 1165: Not necessarily a change you made, but this is an issue that would be relevant to this PR to refactor.

`ban-player-options.reason` is an `option<text-component>` in WIT, so its handle is owned by the host. Using `as_ref()` and cloning the value leaves the table entry behind after the guest loses the handle. Instead, consume the reason with `take`. The same pattern appears in `ban_ip` and the player ban/kick options. Please also account for early errors, such as an invalid IP, so the transferred handle is cleaned up on that path.

Phoenixxo (changes requested):
I found a few existing ownership gaps in the resource cleanup. Some issues with how owned WIT handles are passed from the guest without being handled properly and not being dropped from the resource table. I left some inline examples. Please review and make the appropriate fixes.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/player.rs
PigTurtleDev: This PR tries to phase out these helpers, so I've replaced all occurrences in player.rs itself with `take(text)`.

You're correct that it's wrong, but I can't yet remove the method entirely because event cleanup still depends on it. I also can't replace it with taking because that would require a `Resource<T>` instead of `&Resource<T>`, which event cleanup doesn't allow.
That will be fixed in an event-cleanup followup PR.

crates/pumpkin/src/plugin/loader/wasm/wasm_host/wit/v0_1/server.rs
PigTurtleDev: That's definitely in scope for this PR.
An Invalid IP will already return an error and trap the guest, there's no need to cleanup specific resource handles when that happens.

Phoenixxo (approved):
Everything else looks good to me!

Issue #3255, opened 2026-09-05
Nether fossils generate and mobs spawn above the bedrock roof


ydw1904:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

Upstream master at 55762a50d21ecaff480c95f1ac4017342b017d02

### Operating System

macOS 26.5.2

### Bug Description & Expected Behavior

Nether fossils generate on top of the Nether bedrock roof. Mobs also spawn there, including zombified piglins, and ghasts are present above the roof.

In the observed world this occurs in minecraft:nether_wastes around X=-49.9, Y=135.0, Z=7.6. The fossils appear across a large area of the roof.

Expected behavior: Nether fossils must generate below the bedrock ceiling, and the bedrock roof must not be treated as a valid surface for mob spawning.

This may indicate that the roof bedrock is being accepted as a generation or spawn surface.

<img width="2948" height="1746" alt="Image" src="https://github.com/user-attachments/assets/7f030254-c5a5-4c0c-a3de-6c9c42a2f956" />

### Steps to Reproduce

1. Run a server built from the current upstream master commit.
2. Enter the Nether and travel above the bedrock ceiling.
3. Explore the roof in a Nether Wastes biome around Y=128 and above.
4. Observe Nether fossils generated on the bedrock roof.
5. Observe mobs spawning or standing on the roof, including zombified piglins.

AdmerPRO:
Mobs shouldn't spawn on the bedrock, maybe that's the problem

ydw1904:
> Mobs shouldn't spawn on the bedrock, maybe that's the problem

my thinking too, i am going to test on bedrock superflat worlds now, also glass and other non-spawnable blocks.

ydw1904:
<img width="2696" height="1562" alt="Image" src="https://github.com/user-attachments/assets/4e525d8f-64ff-4054-9600-9089937a6b3e" />

confirmed that its bedrock allowing mobs to spawn

MegalithOfficial:
I will take a look for this one.

AdmerPRO:
Above the layer of bedrok in nether mobs should spawn but only on spawnable błocks like Grass block or neterrack. It alows farms to be bulid

ydw1904:
> Above the layer of bedrok in nether mobs should spawn but only on spawnable błocks like Grass block or neterrack. It alows farms to be bulid

[...]

PR #3259, opened 2026-09-05
fix(spawning): honor vanilla spawn floor predicates

ydw1904:
## Description

- Add generated metadata for all vanilla 26.2 spawn-floor predicate overrides.
- Use one shared predicate evaluator for live natural spawning and generation-cache spawning.
- Prevent mobs from treating bedrock, glass, trapdoors, scaffolding, copper grates, and other vanilla-rejected blocks as valid floors.
- Preserve entity-specific behavior for leaves, ice, frosted ice, and magma blocks, plus vanilla always-valid floors.

This addresses the mob-spawning on netherroof/bedrock portion of #3255. 

## Testing

- cargo fmt --all -- --check
- cargo test -p pumpkin-data
- cargo test -p pumpkin --lib
- cargo clippy --all-targets -- -D warnings

## Summary by CodeRabbit

* **New Features**
  * Added block-specific spawn-floor rules for natural mob spawning.
  * Updated spawning behavior for leaves, ice, magma blocks, glass, trapdoors, and other special blocks.
  * Added support for mob-specific spawning rules, including ocelots, parrots, polar bears, and fire-immune mobs.

* **Tests**
  * Added coverage validating the new vanilla spawn-floor behaviors.


RoosterBooster007 (approved):

RoosterBooster007:
I'm assuming `spawn_floor_predicates.json` was hand-written, which is fine for this PR, but it should probably be updated from [the Extractor](https://github.com/Pumpkin-MC/Extractor) in the future (this is just a non-blocking note).

RoosterBooster007:
Looks good. Thanks, @ydw1904!

PR #3260, opened 2026-09-05
chore(ci): fix the android(aarch64) test workflow

W9pi3cZ1:
## Description
Tried to fix the android_test workflow issue mentioned in the previous PR (#3077). But some tests really can't be fixed just by modifying the forked `cargo-ndk`, so I changed a bit of Rust code and also `--skip` an *unimportant* test(`terminal_failure_is_retained`).

## Risk
This forked version might not be very stable, so I only modified cargo_ndk in android_test.
[Check out](https://github.com/bbqsrc/cargo-ndk/compare/main...W9pi3cZ1:cargo-ndk-workdir:main) exactly what I changed in cargo-ndk

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Android test setup so emulator assets are refreshed correctly and unstable test failures no longer block the test run.
  * Corrected asset paths used by world-generation and biome tests, improving test reliability across environments.

* **Refactor**
  * Improved platform-specific asset path resolution, including direct path handling on Android and consistent repository-based resolution on other platforms.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -133,7 +133,11 @@ jobs:
           profile: pixel_5
           force-avd-creation: false
           emulator-options: -no-snapshot-load -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none -camera-front none
-          script: ./android_test_script
+          script: |
+            adb shell rm -rf /data/local/tmp/*
+            adb push assets /data/local/tmp/$PWD/assets
+            # `terminal_failure_is_retained` is FAILED
+            cargo ndk-test -t arm64-v8a -- --skip terminal_failure_is_retained
   build_release:
     name: Build project in release
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -218,7 +218,7 @@ jobs:
         with:
           ndk-version: r27c
       - name: Install cargo-ndk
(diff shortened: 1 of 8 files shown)


Snowiiii:
@W9pi3cZ1 can you please explain what current issues there are with the current android build/CI?

W9pi3cZ1:
Exactly, the issue is with the Android test CI, not the build CI. `cargo ndk-test` does **not** set up a Cargo environment on the Android device and copy the whole project over to it. Instead, it takes over the run step: it tries to copy the executable to a fixed path, `/data/local/tmp`, and then runs it directly via `adb shell`. It does not set `PWD` or configure any environment variables, so it naturally produces the `No such file or directory` error shown in the screenshot.

image


This PR is a temporary workaround. It modifies `cargo-ndk` to set `PWD` for the Android test environment, which should fix some cases where test data cannot be found, but it still does not configure environment variables. It also does not solve the problem for tests that locate test files via `env!("CARGO_MANIFEST_DIR")`. If we want to fix this thoroughly, we may really need to compile Pumpkin in the Android environment, but I’m afraid the compile time would be worse than cross-compiling from x86_64. Alternatively, you could remove the Android test CI, since it is currently broken and does not provide a good guarantee for actual runtime behavior.

W9pi3cZ1:
Actually, I’m not really hoping for this PR to be merged until I find a better solution (maybe giving up testing the Android side on GitHub Actions would be wise).

PR #3150, opened 2026-08-30
fix(inventory): handle occupied hotbar swaps

dehmitros:
## Description
Fixes #3059 
Swapping items with the number keys (1-9) from a non-blank slot to a blank slot, it works fine BUT
If you swap items on occupied slots, it immediately reverted them to how they were

The bug was in the server-side handling of number-key inventory swaps.
When the client sends a SlotActionType::Swap action, the server handled only two situations:
1) The hovered slot contained an item and the selected hotbar slot was empty.
2) The hovered slot was empty and the selected hotbar slot contained an item.
There was no branch for when both slots contained items. The client predicted the swap locally, so it looked like it worked for a split second. The server left its inventory unchanged and then synchronized that unchanged state back to the client, causing both items to snap back.

The fix adds occupied-to-occupied swapping while respecting slot permissions and stack limits, and corrects offhand action grouping

## Testing
There's a regression test in crates/pumpkin-inventory/src/merchant/merchant_screen_handler.rs:635

In-game, you just follow the original steps to recreate the issue, and it works fine
1) Have an item in slot 1
2) Have an item in slot 2
3) Hover over item 1 and press 2

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed hotbar item swaps so occupied stacks exchange correctly.
  - Prevented excess items from being lost when swapping into full or incompatible slots.
  - Ensured displaced items are moved to available inventory space or dropped when no space remains.
  - Improved handling of item removal during swap actions.

PR #3263, opened 2026-09-05
fix(inventory): swap stacks between hovered slot and hotbar when hovered slot is occupied

KaiPenazzi:
Adds support for swapping an occupied, hovered inventory slot with an occupied hotbar slot using the corresponding number key.

Previously, swapping only worked when either the inventory slot or the hotbar slot was empty. Now, when both slots contain an item, their stacks are exchanged if the target slot accepts the hotbar stack.

This matches vanilla Minecraft's inventory behavior.


KaiPenazzi:
Already implemented in PR https://github.com/Pumpkin-MC/Pumpkin/pull/3150

PR #3267, opened 2026-09-05
fix(server): execute #minecraft:tick function tag every game tick

### LINKED ISSUE #3251: #minecraft:tick function tag is never executed
## Before Submitting

Searches for `minecraft:tick`, `tick function`, and `function tag` found no focused report. #731 is related to general datapack/registry support, but it does not cover function-tag lifecycle execution.

## Minecraft Edition

Java Edition

## Server Version / Commit

Pumpkin `0.1.0-dev+26.2-26.45` (`8d0d0d3/release`), Minecraft 26.2, protocol 776

Output from `/pumpkin`:

```text
Pumpkin 0.1.0-dev+26.2-26.45 (Commit: 8d0d0d3/release)
(Minecraft 26.2, Protocol 776)
```

## Operating System

CachyOS Linux, x86_64, kernel `7.2.2-1-cachyos`

## Bug Description & Expected Behavior

A function referenced by `#minecraft:tick` is never executed. The minimal datapack below references the same one-line function from both `#minecraft:load` and `#minecraft:tick`:

```text
data/minecraft/tags/function/load.json  -> tick_repro:probe
data/minecraft/tags/function/tick.json  -> tick_repro:probe
```

```mcfunction
data modify storage tick_repro:state calls append value 1
```

The load lifecycle executes the function once during `/reload`, proving that Pumpkin loaded the function and can execute its command. The stored list then remains `[1]`; the tick lifecycle never appends further entries.

Expected behavior: after `/reload`, the function should run once through `#minecraft:load` and then once per game tick through `#minecraft:tick`. At the default 20 TPS, waiting five seconds should produce substantially more than one entry.

Actual behavior: after waiting five seconds, storage contains only `[1]`, from `#minecraft:load`.

This prevents datapacks that rely on the standard tick function tag from running their recurring logic, even when their functions and tags load successfully.

Possibly related to #731, but this report is narrowly about dispatching the already-loaded `#minecraft:tick` function tag rather than dynamic registries.

## Steps to Reproduce

1. Create a directory at `world/datapacks/tick-repro/` using the four files under **Reproduction files** below. `pack.mcmeta` must be at the root of that directory.
2. Start Pumpkin.
3. Enable the pack if needed:

   ```text
   datapack enable "file/tick-repro"
   ```

4. Initialize an empty list used to count calls:

   ```text
   data modify storage tick_repro:state calls set value []
   ```

5. Reload datapacks, then wait at least five seconds with the server running normally:

   ```text
   reload
   ```

6. Inspect the recorded calls:

   ```text
   data get storage tick_repro:state calls
   ```

Observed result:

```text
Storage tick_repro:state has the following contents: [1]
```

No players need to be connected. The test was performed at the default `tps = 20.0` with the tick system unfrozen.

## Reproduction files

```text
tick-repro/
├── pack.mcmeta
└── data/
    ├── minecraft/tags/function/
    │   ├── load.json
    │   └── tick.json
    └── tick_repro/function/
        └── probe.mcfunction
```

`pack.mcmeta`:

```json
{
  "pack": {
    "pack_format": 101,
    "min_format": [101, 1],
    "max_format": [107, 1],
    "description": "Minimal Pumpkin #minecraft:tick reproduction"
  }
}
```

`data/minecraft/tags/function/load.json`:

```json
{
  "values": [
    "tick_repro:probe"
  ]
}
```

`data/minecraft/tags/function/tick.json`:

```json
{
  "values": [
    "tick_repro:probe"
  ]
}
```

`data/tick_repro/function/probe.mcfunction`:

```mcfunction
data modify storage tick_repro:state calls append value 1
```

## Additional diagnostic information

At the tested commit, [`Server::reload_datapacks`](https://github.com/Pumpkin-MC/Pumpkin/blob/8d0d0d311778cb0aecb5fc957d571a38f286fda0/crates/pumpkin/src/server/mod.rs#L539-L549) explicitly executes `#minecraft:load`. The [`Server::tick_worlds`](https://github.com/Pumpkin-MC/Pumpkin/blob/8d0d0d311778cb0aecb5fc957d571a38f286fda0/crates/pumpkin/src/server/mod.rs#L1098-L1115) tick path does not appear to execute `#minecraft:tick`, which matches the isolated reproduction.

Kurenaiiiii:
## What was changed?

In `crates/pumpkin/src/server/mod.rs`, `Server::tick_worlds` now dispatches the `#minecraft:tick` function tag once per tick, at the top of the function before the task scheduler and world ticks:

```rust
let source = crate::command::CommandSender::Console.into_source(self);
let _ = self
    .datapack_manager
    .execute_function(self, &source, "#minecraft:tick");
```

This reuses the exact same lookup-and-execute mechanism (`datapack_manager.execute_function` with a `Console` command source, result ignored with `let _ =`) already used for `#minecraft:load` in `Server::reload_datapacks` — no separate/parallel system. Unknown/missing tags return `Err` which is ignored, so this is safe even if the registry isn't populated yet, and `#minecraft:load` itself is untouched.

Pumpkin loaded datapack functions and could execute them, but only ever dispatched the `#minecraft:load` tag (in `reload_datapacks`). Nothing in the per-tick update path (`tick_worlds`, called once per game tick) ever dispatched `#minecraft:tick`, so any datapack relying on recurring per-tick logic — the standard vanilla mechanism for that — silently never ran.

`#minecraft:tick` now fires once per game tick as in vanilla (it lives inside `tick_worlds`, so it follows `/tick freeze` like the rest of game logic). `#minecraft:load` behavior is unchanged — still fires exactly once per reload, verified below — and no other function tag or subsystem is affected.

- No automated tests added — there is no existing test harness for server-tick datapack dispatch. Verified manually (below) plus the full existing suite.
- Commands run: `cargo fmt` (clean), `cargo clippy --all-targets` (passes, zero warnings), `cargo test --workspace` (all suites `ok`, 0 failed).
- Manual verification: built the release binary and ran it with a minimal repro datapack matching issue #3251 (`pack_format` 101; `load.json` and `tick.json` both → `tick_repro:probe`; probe runs `data modify storage tick_repro:state calls append value 1`). After `datapack enable`, `set value []`, `reload`, and ~6s at 20 TPS, the storage list measured **exactly 119 entries** (`calls[0]`–`calls[118]` return `1`, `calls[119]` fails) — pre-fix it stayed at `[1]`. Regression check with the tick loop frozen: `set []` + `reload` shows verbatim `[1]`, and a second frozen `reload` adds exactly one execution, confirming `#minecraft:load` still fires exactly once per reload. No panics, errors, or TPS drops observed; server stopped cleanly with exit code 0.

Fixes #3251

## Summary by CodeRabbit

* **New Features**
  * Datapack functions tagged with `#minecraft:tick` now run automatically at the start of each world tick.


JulesB40:
Please make the automatic tick-function command source silent:

```rust
let source = crate::command::CommandSender::Console
    .into_source(self)
    .with_silent();
```

`Console.into_source` creates a non-silent source. A tick function containing `data modify storage parity:state calls append value 1` therefore sends routine success feedback every tick through `CommandSource::send_feedback` (to the console and potentially operators). Vanilla runs this automatic function without those success messages.

I reproduced #3251 on unmodified commit `55762a50d21ecaff480c95f1ac4017342b017d02`: the load tag executes, but ten stepped ticks produce zero tick-function calls. Vanilla 26.2 and Paper 26.2 build 121 both produce ten calls.

A local patch on that baseline using this PR's dispatch placement plus `.with_silent()` passes the same console-driven datapack probe: zero calls while frozen, exactly ten after `tick step 10`, three after reloading/resetting and stepping three ticks, and no additional calls after disabling the pack. It also passes the assertion that those ten ticks emit no routine storage-modification feedback. The fixture's load function initializes `{calls:[],loaded:1}` in `parity:state`, and its tick function appends `1` to `calls`.

I inspected PR head `f964f4ee8ddcb9c7a743555242526aee8569c3f1`; I have not built that exact head. Local formatting, cargo-machete, and release Clippy with all targets/all features and `-D warnings` passed; the full release workspace nextest suite passed: 751 tests, zero failures, zero skipped. Windows rustc access-violation crashes required disabling LTO, using 16 codegen units, and reducing the patched Pumpkin package's release optimization to level 1. These are behavioral checks, not performance measurements or a claim of complete function parity.

Update: confirmed that head 0d602da7846eea1443e45d5b3e20afd8851927d9 now includes the suggested silent source. Thanks for incorporating it. Release workspace documentation tests also completed successfully: 7 passed, 23 ignored, zero failures. The final harness additionally passed its explicit silence assertion on vanilla 26.2 and Paper 26.2-121.

Kurenaiiiii:
Good catch, thanks for the thorough verification on this. You're right — Console.into_source() isn't silent, so this would spam routine command feedback every tick that vanilla doesn't produce. Updating now to chain .with_silent() as you suggested, will push the fix shortly.

RoosterBooster007:
Looks good. Thanks, @Kurenaiiiii!

PR #3269, opened 2026-09-06
feat(gui): add optional Qt6 monitoring/console window gui

zinedinp:
Adds a Qt6 window as an opt-in alternative to the terminal console.

- Gated compile-time (gui feature) and runtime (--gui, or automatically when stdin isn't a tty).
- New pumpkin-gui crate, decoupled from Server via plain data structs + a GuiCommands trait, polled at 2Hz.
- Dashboard (CPU/RAM/TPS), worlds, players (OP/Kick/Ban via the normal dispatcher), and a console with ANSI colors, completion, search/export.
- Qt owns the main thread; the server runs on a spawned thread. Either side closing shuts the other down cleanly.

- [ ] TODO: Windows/macOS packaging (Qt distribution strategy still open)
I tested it only against dynamically linked qt6. staticly linking should be possible, but i would test it in the future.


Q2297045667:
Please use the translation feature of this project. You need to add the corresponding translation key-value pairs in the relevant sections or only in the English translation

Also, it would be great if you could add a plugin management feature. You could try integrating with https://market.pumpkinmc.org/ to implement things like viewing plugin URLs and one-click updates, etc.🌚

Snowiiii:
This definitely does not belong here, if ever we would do this as a separate repo as well

PR #3271, opened 2026-09-06
fix(worldgen): stop overworld biomes leaking into other dimensions

ydw1904:
## Problem

`LAST_RESULT_NODE` is a thread-local shortcut holding the last matched biome-tree leaf, but it was shared by every dimension's tree. `BiomeTree::get` seeds its search distance from that leaf and `search` only accepts strictly closer nodes, so a leaf cached from the overworld tree can beat every node of the Nether tree and be returned unchanged.

A worker thread that generates an overworld chunk then hands the next Nether chunk overworld biomes, and with them overworld surface rules, features and mob spawns: rivers and plains in the Nether, and zombies, skeletons and creepers spawning there.

## Fix

Key the cache by its source tree, so a leaf is only ever reused for the tree it came from. This matches vanilla, where `lastResult` lives on the `RTree` itself rather than being shared across all of them.

## Testing

- New regression test `nether_chunks_only_contain_nether_biomes`: warms the thread-local cache with an overworld chunk, then generates a Nether chunk on the same thread and asserts every biome is one of the five Nether biomes. Fails on `master`, passes here.
- Rebased onto master after df983f84, which already moved the ocean monument check onto `generator.biome_supplier`, so that hunk is dropped here.
- `cargo clippy -p pumpkin-world --lib --tests` clean, `cargo test -p pumpkin-world --lib` 188 passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed biome generation across dimensions so cached biome data from one dimension cannot affect another.
  * Nether chunks now consistently contain only Nether biomes, even after generating overworld terrain.

### DIFF
--- a/crates/pumpkin-world/src/biome/mod.rs
+++ b/crates/pumpkin-world/src/biome/mod.rs
@@ -17,8 +17,14 @@ pub use pumpkin_data::chunk::{
 };
 
 thread_local! {
-    /// A shortcut; check if last used biome is what we should use
-    static LAST_RESULT_NODE: RefCell<Option<&'static BiomeTree>> = const {RefCell::new(None) };
+    /// A shortcut; check if last used biome is what we should use.
+    ///
+    /// Keyed by the tree the leaf came from. `BiomeTree::get` seeds its search distance from
+    /// this leaf and only accepts strictly closer nodes, so a leaf cached from one dimension's
+    /// tree would otherwise beat every node of another's and be returned unchanged. Vanilla
+    /// stores `lastResult` on the `RTree` itself rather than sharing one across all of them.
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Hey @ydw1904, thanks for this. Though, the monument hunk already landed on master (df983f84 switched that check to `generator.biome_supplier`, and the old local `ActiveSupplier` enum is gone), so that part won't apply cleanly. Rebase, drop the `proto_chunk.rs` bit and keep the `LAST_RESULT_NODE` change plus the nether test, and I'll get it merged.

RoosterBooster007:
Looks good. Thanks, @ydw1904!

PR #3278, opened 2026-09-06
fix(plugin): fix context drop outside of on_load

Phoenixxo:
## Description

This PR fixes an issue where the context store was not being saved after returning outside of `on_load` so registered commands from plugins would cause the plugin WASM component to become trapped. 

## Validation

`cargo clippy`, `cargo test` and manually tested in game against the previously failing plugin.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebAssembly plugin loading by ensuring initialization context resources are passed and managed consistently.
  - Reduced the risk of resource-handling issues during plugin startup.

- **Compatibility**
  - No changes were made to exported or public interfaces.


Snowiiii:
Thanks @Phoenixxo :D

PR #3238, opened 2026-09-04
feat(world): lighting engine flood, shape occlusion, and sky cut height

zinedinp:
## Description

- Worldgen: sky 15 above surface, then descend with vanilla decay
- Shape occlusion (stairs/slabs/…) plus opacity
- Skip already-lit neighbours; drop out-of-cache flood steps
- Runtime queued drain + sky cut-height cache
- Removed `storage.rs`
- Emitter-section mask on ProtoChunk
- ChunkCursor / [chunk_access.rs](vscode-webview://183356fdr3es4v5qjvk7mg9d4scq1heqhbo6meq651na4oeu0iaj/chunk_access.rs) (one neighbour resolve, one light lock)
- max_possible gate; per-axis flood step
- has_different_light_properties includes shape (stair rotate)
- Sky cut height persist/migrate + quadrant flags
- Chunk format / palette / worker relight hooks

With this PR the Lighting is on pair with vanilla and much faster.

Greetings and Servus

## Summary by CodeRabbit

- **New Features**
  - Improved block and sky lighting with shape-aware occlusion, refined propagation, and better handling of chunk borders and unloaded areas.
  - Added persistent sky-light height data for consistent lighting across chunk loads and upgrades.
  - Added lighting performance statistics to tick diagnostics.
  - Lighting updates are now processed during world ticks.

- **Bug Fixes**
  - Restored omitted open-sky lighting when loading chunks.
  - Improved relighting decisions and preserved chunk custom data and lighting state during serialization and upgrades.

PR #3279, opened 2026-09-06
fix: add y-bounds checks to light propagation

IndrajeethY:
Light propagation in DynamicLightEngine can escape world height limits. When a neighbor block is above or below the valid y-range, the engine keeps re-queuing it forever — causing a deadlock/infinite loop on void and flat worlds.

This adds dimension-aware y-bounds to the light engine. Each propagation function skips neighbors outside the valid range. The bounds come from the dimension config passed through at level init, so it works correctly for overworld, nether, and end.

**Changes:**
- `DynamicLightEngine` now stores `min_y` and `max_y` from the dimension
- All 6 propagation call sites skip out-of-bounds neighbors
- Removed `Default` impl since bounds are required

## Summary by CodeRabbit

- **Bug Fixes**
  - Lighting now respects each dimension’s configured vertical boundaries.
  - Block and sky light no longer propagate or recalculate outside the valid height range.
  - Improved lighting behavior across dimensions with different minimum heights and vertical limits.
  - Lighting calculations now use the correct upper and lower bounds for each dimension, helping prevent inaccurate illumination near or beyond world height limits.
  - This provides more consistent illumination throughout worlds with custom vertical dimensions.

### DIFF
--- a/crates/pumpkin-world/src/level.rs
+++ b/crates/pumpkin-world/src/level.rs
@@ -231,6 +231,8 @@ impl Level {
             }
         }
 
+        let dim_min_y = dimension.min_y;
+        let dim_height = dimension.height;
         let seed = Seed(seed as u64);
         let world_gen: Arc<WorldGenerator> = Arc::from(get_world_gen(
             seed,
@@ -266,7 +268,7 @@ impl Level {
             world_gen: ArcSwap::new(world_gen),
             level_folder,
             lighting_config: level_config.lighting,
-            light_engine: DynamicLightEngine::new(),
+            light_engine: DynamicLightEngine::new(dim_min_y, dim_min_y + dim_height),
             chunk_saver,
             entity_saver,
             schedule_tick_counts: AtomicU64::new(0),
--- a/crates/pumpkin-world/src/lighting/runtime.rs
+++ b/crates/pumpkin-world/src/lighting/runtime.rs
@@ -8,6 +8,8 @@ use pumpkin_util::math::position::BlockPos;
 use std::sync::Arc;
 
 pub struct DynamicLightEngine {
+    min_y: i32,
+    max_y: i32,
     block_decrease: SegQueue<(BlockPos, u8)>,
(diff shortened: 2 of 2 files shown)


zinedinp:
you might want to take a look at https://github.com/Pumpkin-MC/Pumpkin/pull/3238/changes/13f02d65d16707fd76fd82e738729624f8db9d0a from my PR https://github.com/Pumpkin-MC/Pumpkin/pull/3238

I think these are the same changes

IndrajeethY:
> you might want to take a look at [13f02d6](https://github.com/Pumpkin-MC/Pumpkin/commit/13f02d65d16707fd76fd82e738729624f8db9d0a) from my PR #3238
> 
> I think these are the same changes

I'm not entirely sure if that can fix this dead lock issue, since this happens on run time when i place a block and break on a void world.

zinedinp:
I have taken another look at it:

- I used a per chunk  approuch while yours uses a per dimension approuch.

But I don't know where you think there is a dead lock in my commit. "get_sky_light_level" returns 15 for out-of-bounds positions, so "propagate_sky_light" keeps treating them as bright. Thats looks like the same logic.

IndrajeethY:
> I have taken another look at it:
> 
> * I used a per chunk  approuch while yours uses a per dimension approuch.
> 
> But I don't know where you think there is a dead lock in my commit. "get_sky_light_level" returns 15 for out-of-bounds positions, so "propagate_sky_light" keeps treating them as bright. Thats looks like the same logic.

Oh sorry you misunderstood i guess.. I didn't mention there is a deadlock on your commit, the deadlock is in Pumpkin's runtime lighting code (runtime.rs → DynamicLightEngine). When you place/break a block on a void world, the propagation functions go past world bounds and loop forever...

IndrajeethY:
@Snowiiii sorry for the ping... can we please get this reviewed, so that i can put my plugin to marketplace

RoosterBooster007:
Looks good. Thanks, @IndrajeethY!

PR #3280, opened 2026-09-06
fix(protocol): reject out-of-range consume effect counts

catdum200:
`ConsumableImpl::deserialize` passes an unchecked signed length to `Vec::with_capacity`, so a single crafted item stack from any connected player shuts the server down.

## Reproduction

Feeding a Consumable component body whose trailing `effects_len` VarInt is `-1` (`FF FF FF FF 0F`) through `ConsumableImpl::deserialize` on current `master` (b961f9776):

```
thread '...' panicked at library\alloc\src\raw_vec\mod.rs:28:5:
capacity overflow
```

With this patch, the same 14 bytes:

```
rejected cleanly with: Message("Invalid consume effect count")
```

The 14-byte body is `0.0f32` | animation `0` | `IdOr` id `0` (inline) | `""` | range `None` | particles `false` | `effects_len = -1`.

## Cause

```rust
let effects_len = seq.get_var_int()?.0;                          // i32, straight off the wire
let mut effects_vec = Vec::with_capacity(effects_len as usize);  // -1 sign-extends to usize::MAX
```

`effects_len` is an `i32` that is never range-checked, and the cast sign-extends, so `-1` becomes `usize::MAX` and `Vec::with_capacity` aborts the thread. A large positive count is an allocation vector for the same reason.

This is the only length in `data_component.rs` that is neither range-checked nor capped. Every sibling already guards its count — `deserialize_status_effects` caps at `MAX_STATUS_EFFECTS`, enchantments and firework explosions at 256, bundles at 64 — and `SuspiciousStewEffectsImpl::deserialize` uses the exact `!(0..=MAX).contains(&count)` form adopted here, so this brings the consumable path in line with the module's existing idiom rather than introducing a new one.

## Impact

The component is decoded while the item stack is read, on a networking task, before any handler-level gamemode or permission check runs, so it is reachable from any connected player. The panic hook installed in `pumpkin/src/main.rs` calls `stop_server()` for the first panic on a worker thread, so the outcome is a full server shutdown rather than a dropped connection.

## Testing

- Red/green verified through `ConsumableImpl::deserialize` on the same bytes: panics before the patch, returns `ReadingError` after.
- `cargo fmt --check`, `RUSTFLAGS=-Dwarnings cargo clippy --all-targets --all-features`, and `cargo test --workspace` (757 passed, 0 failed) all pass on this branch.

I kept the diff to the fix itself rather than committing the repro as a test, following the feedback on #3035 that small changes don't need dedicated coverage. Happy to add it if you'd prefer it guarded against regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for consume-effect counts during data decoding.
  * Invalid counts outside the supported 0–256 range now return an error instead of being processed.


RoosterBooster007:
Hey @catdum200, thanks for this. Just a small follow-up thought: a few other lengths in that file use the same unchecked `with_capacity` (`CustomModelDataImpl`, `ContainerImpl`, `ChargedProjectilesImpl`), so a sweep PR could be worth it if you're keen.

RoosterBooster007:
Thanks, @catdum200!

PR #3281, opened 2026-09-06
fix(block): notify clients when unsupported cactus and bamboo break

### LINKED ISSUE #3268: Bug when breaking plants which grow multiple blocks high
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.45/8d0d0d3

### Operating System

Ubuntu 26.04

Plants that grow multiple blocks high like cacti or bamboo don't behave as expected when breaking the block at the bottom.

When the block on the bottom is broken, all the items of the upper blocks are dropped, but the blocks themselves don't disappear.
When you rejoin the blocks disappear though, so I guess they just don't disappear on the client.

I've tested this with cacti and bamboo which had this issue.
I've tried testing it with sugar cane, but wasn't able to place some, as it wouldn't let me even next to water.
I've tested this with tall grass, which disappeared correctly

I've tested it  on 26.2

### Steps to Reproduce

Self-explaining

JulesB40:
## Problem and change

When a cactus or bamboo stack loses its bottom block, scheduled support checks remove the upper blocks with empty update flags. `World::break_block` forwards those flags to `set_block_state`, which only queues client changes when `NOTIFY_LISTENERS` is present. The blocks disappear on the server, but remain in the client's state until a refresh (#3268).

Use `BlockFlags::NOTIFY_ALL` for these two scheduled breaks, matching sugar cane. This queues client updates and notifies neighbors without changing loot handling.

Fixes #3268.

## Validation

- Local formatting, diff whitespace, and release Clippy (all targets/features, warnings denied) passed.
- Local release workspace nextest: 756 passed, zero skipped.
- GitHub build/test/lint matrix passed, including Windows, Linux, macOS and Android checks as of September 6.
- A connected Python protocol-776 observer was first validated against vanilla 26.2 and Paper 26.2-121, then run against original Pumpkin and this exact PR head `0d32f8a580b1d65d63e9230e07abf65f690bb9be`.

For both cactus and bamboo, it observes initial placement packets for a three-block stack on supported sand, removes the bottom via `setblock ... air destroy`, and checks server air separately from received single/section block updates. A separate supported plant is the negative control. No reconnect occurs during the assertions, and a chunk replacement/unload invalidates the test.

| Server | Server state after removal | Received client state, bottom to top | Supported control |
| --- | --- | --- | --- |
| Vanilla 26.2 | air, air, air | air, air, air | unchanged |
| Paper 26.2-121 | air, air, air | air, air, air | unchanged |
| Original Pumpkin `55762a50d21ecaff480c95f1ac4017342b017d02` | air, air, air | air, plant, plant | unchanged |
| This PR head | air, air, air | air, air, air | unchanged |

The patched Windows executable SHA-256 is `114072bfe73905ce0e9a0299e00c4a3706057867b48ffe47a20b30d4d3f72257`. Its release build used LTO disabled, 16 codegen units and Pumpkin package optimization level 1 to work around local rustc access-violation crashes. No performance claim is made.

This is connected packet-level validation, not a rendered-client screenshot. The fixture triggers the scheduled support-loss path from the console, rather than simulating a digging packet; it does not count dropped item entities. All test servers bind to localhost and stop after the test.

Prepared with Codex assistance. Only the two scheduled-break flags are changed; the separate datapack work and the local harness are not included in this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved block update notifications when invalid bamboo is automatically removed.
  - Improved block update notifications when invalid cactus is automatically removed.
  - Helps ensure surrounding blocks and connected systems respond correctly to these automatic removals.


RoosterBooster007 (approved):

RoosterBooster007:
Thanks, @JulesB40!

PR #3294, opened 2026-09-07
fix(worldgen): anchor Nether fossils below the bedrock roof

ydw1904:
## Description

Nether fossils could generate in or above the bedrock ceiling. Resolve their initial height against the Nether's 128-block generation depth, giving the vanilla range of Y=32 through Y=125, instead of the 256-block dimension height.

Determine the final anchor during structure-start generation by scanning the raw base-noise column, matching vanilla's `ChunkGenerator.getBaseColumn`. Remove the later scan of generated terrain, which could anchor a fossil to the ceiling pocket created around its own provisional bounding box by the beardifier. Structure bounds and biome validation now use the resolved anchor, and noise-world `/place structure` receives the same base-column sampler.

Addresses the fossil-generation portion of #3255. The mob-spawning portion is separate.

## Testing

- `cargo test --offline -p pumpkin-world --lib nether_fossil -j 2`: 2 passed.
- Height-range test checks initial fossil Y stays within 32 through 125.
- Generated-chunk regression uses seed `1789192880694109906`, chunk `(-10, 20)`, and verifies that fossils are present while every bedrock block from the Surface stage survives the Features stage.
- Compared placement logic with the local decompiled vanilla 26.2 sources.
- Merge compatibility checked against upstream master `786e2fbf7`; no conflicts. Tests ran on the PR branch.

Assisted by Codex.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved structure placement across different world heights and dimensions.
  - Nether fossils now generate at valid terrain locations and remain within the intended height range.
  - Structure placement in noise-generated worlds now accounts for surrounding terrain.
  - Fixed Nether feature generation so existing roof bedrock is preserved.
  - Improved terrain-state sampling for more consistent structure generation.
- **Tests**
  - Added regression coverage for Nether fossils, terrain generation, and bedrock preservation.

### DIFF
--- a/crates/pumpkin-world/src/chunk_system/generation.rs
+++ b/crates/pumpkin-world/src/chunk_system/generation.rs
@@ -183,6 +183,64 @@ mod tests {
         }
     }
 
+    /// The Nether ceiling is beard-carved around every structure start, so a fossil
+    /// whose start Y was picked from the built chunk anchored itself to its own beard
+    /// pocket and replaced the bedrock roof. This seed/chunk reproduced that.
+    #[test]
+    fn nether_fossils_stay_below_the_bedrock_roof() {
+        let seed = Seed(1_789_192_880_694_109_906);
+        let block_registry = Arc::new(BlockRegistry);
+        let world_gen = get_world_gen(
+            seed,
+            Dimension::THE_NETHER,
+            false,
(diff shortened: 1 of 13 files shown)


ydw1904:
Superseded by `df983f847`

PR #3295, opened 2026-09-07
fix(block): send block updates when support-dependent blocks break 🤖🤖🤖

### LINKED ISSUE #3109: Floating snow (and several other blocks)
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.40 (Commit: 9e9f1b5/release)

### Operating System

Windows 11

When breaking a block with snow on top of it, the snow should update/break but it doesn't.

Image

### Steps to Reproduce

1. Grab snow and put it on top of a block (or find a snow biome)
2. Break the block under
3. The snow remains floating

IsaiahJaiden:
## Description

Blocks that tear themselves down via a scheduled tick (snow layers, carpets, candles, candle cakes, cake, standing banners, lanterns, bamboo, chorus plant/flower, cactus, eyeblossom, lily pad, turtle eggs) called `break_block` with `BlockFlags::empty()`. The block was removed server-side, but the change was never sent to clients, leaving permanent "ghost" blocks — e.g. snow layers floating in the air after breaking the block beneath them.

This changes all of those support-check teardown breaks to `BlockFlags::NOTIFY_ALL`, matching vanilla's `Level.removeBlock`/`Level.destroyBlock`, which use flag 3 (`UPDATE_NEIGHBORS | UPDATE_CLIENTS`) for the same paths.

This also covers the "and several other blocks" part of the issue: the same ghosting applied to every block using the support-tick teardown pattern.

## Testing

- `cargo check`
- `cargo fmt --check`

Fixes #3109

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved block-update notifications when unsupported or invalid blocks are removed.
  - Fixed notification behavior for banners, cakes, candles, carpets, lanterns, plants, snow, turtle eggs, and similar blocks.
  - Nearby blocks and connected game behavior now receive appropriate updates after these blocks break.

### DIFF
--- a/crates/pumpkin/src/block/blocks/banners.rs
+++ b/crates/pumpkin/src/block/blocks/banners.rs
@@ -36,7 +36,7 @@ impl BlockBehaviour for BannerBlock {
     fn on_scheduled_tick(&self, args: OnScheduledTickArgs<'_>) {
         if !can_place_at(args.world.as_ref(), args.position) {
             args.world
-                .break_block(args.position, None, BlockFlags::empty());
+                .break_block(args.position, None, BlockFlags::NOTIFY_ALL);
         }
     }
 
--- a/crates/pumpkin/src/block/blocks/cake.rs
+++ b/crates/pumpkin/src/block/blocks/cake.rs
@@ -143,7 +143,7 @@ impl BlockBehaviour for CakeBlock {
     fn on_scheduled_tick(&self, args: OnScheduledTickArgs<'_>) {
         if !can_place_at(args.world.as_ref(), args.position) {
             args.world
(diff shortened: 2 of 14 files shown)


RoosterBooster007:
Please drop the updates for cactus and bamboo (#3281 already covers them), and this is good to merge.

IsaiahJaiden:
Dropped — cactus and bamboo are reverted to their previous behavior in 2dc19364 so this PR no longer overlaps with #3281. Thanks for the pointer; the remaining 16 sites now only cover blocks #3281 doesn't touch.

Issue #3107, opened 2026-08-28
Sculk Shrieker doesnt activate


AdmerPRO:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

9e9f1b5/release)

### Operating System

Windows 11

### Bug Description & Expected Behavior

Sculk Shrieker doesnt activate.

### Steps to Reproduce

1. Get a Sculk Shrieker or find it
2. Try to make any sound or step on it

RoosterBooster007:
Same issue with sculk sensors. They don't work.

PR #3296, opened 2026-09-07
fix(block): activate sculk shrieker when a player steps on it 🤖🤖🤖

IsaiahJaiden:
## Description

`SculkShriekerBlock::try_activate` was fully implemented (shriek state, sound, darkness effect, warning level tracking) but was never called from anywhere, so sculk shriekers never activated when players stepped on them.

This wires it up by implementing `on_entity_step`, mirroring vanilla `SculkShriekerBlock#stepOn`: a player standing on the shrieker resonates it, unless they are stepping carefully (sneaking). The existing `shrieking` blockstate already prevents re-triggering while a shriek is active.

Note: this covers the "step on it" half of the issue. Activating from arbitrary nearby sounds depends on the vibration/sensor system and is left for future work.

## Testing

- `cargo check`
- `cargo fmt --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Sculk shriekers now activate when any player steps on them, including sneaking players.
  * Non-player entities continue to be ignored.

### DIFF
--- a/crates/pumpkin/src/block/blocks/sculk/sculk_shrieker.rs
+++ b/crates/pumpkin/src/block/blocks/sculk/sculk_shrieker.rs
@@ -94,11 +94,8 @@ impl BlockBehaviour for SculkShriekerBlock {
     }
 
     fn on_entity_step(&self, args: OnEntityStepArgs<'_>) {
-        // Vanilla `SculkShriekerBlock#stepOn`: a player standing on the shrieker
-        // resonates it, unless they are stepping carefully (sneaking).
-        if args.entity.get_entity().is_sneaking() {
-            return;
-        }
+        // Vanilla `SculkShriekerBlock#stepOn`: any player standing on the shrieker
+        // resonates it, even when sneaking.
         if args.entity.cast_any().downcast_ref::<Player>().is_some() {
             Self::try_activate(args.world, args.position);
         }
--- a/crates/pumpkin/src/block/blocks/sculk/sculk_shrieker.rs
+++ b/crates/pumpkin/src/block/blocks/sculk/sculk_shrieker.rs
@@ -1,7 +1,10 @@
 use std::sync::Arc;
 
 use crate::block::entities::sculk_shrieker::SculkShriekerBlockEntity;
-use crate::block::{BlockBehaviour, BlockMetadata, OnPlaceArgs, OnScheduledTickArgs};
+use crate::block::{
+    BlockBehaviour, BlockMetadata, OnEntityStepArgs, OnPlaceArgs, OnScheduledTickArgs,
+};
+use crate::entity::player::Player;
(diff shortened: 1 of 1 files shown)


crates/pumpkin/src/block/blocks/sculk/sculk_shrieker.rs
```diff
@@ -90,6 +93,17 @@ impl BlockBehaviour for SculkShriekerBlock {
         props.to_state_id(args.block)
     }
 
+    fn on_entity_step(&self, args: OnEntityStepArgs<'_>) {
+        // Vanilla `SculkShriekerBlock#stepOn`: a player standing on the shrieker
+        // resonates it, unless they are stepping carefully (sneaking).
+        if args.entity.get_entity().is_sneaking() {
```
RoosterBooster007: Actually, shriekers go off [even when you're sneaking](https://minecraft.wiki/w/Sculk_Shrieker#:~:text=A%20sculk%20shrieker%20activates%20when%20any%20player%20stands%20on%20the%20black%20part%20in%20the%20center%20of%20the%20block%2C%20even%20when%20sneaking.). Also, issue #3107 includes sound detection which this PR does not include.

RoosterBooster007 (changes requested):

crates/pumpkin/src/block/blocks/sculk/sculk_shrieker.rs
IsaiahJaiden: Good catch — you're right. I checked vanilla 26.2 `SculkShriekerBlock#stepOn` and it has no sneak check: any player standing on the shrieker resonates it. Sneaking check removed in 7b1d2911 (now matches vanilla exactly).

On the sculk sensor part: you're right that #3107 covers vibration-based activation (sensors + shriekers reacting to game events like footsteps, block place/break, projectiles, etc.). That's a substantially bigger piece of infrastructure — Pumpkin currently has no game-event-to-listener dispatch at all (`SculkSensorBlock::trigger` exists but nothing calls it, and `World::emit_game_event` only fires the plugin event, with no block listener routing, wool occlusion, or the vibration frequency mapping). I'd rather not stuff all of that into this PR; I'm planning it as a separate follow-up PR so both pieces can be reviewed on their own terms. I'll link it here when it's up.

IsaiahJaiden: Followed up on the sensor half: #3341 implements the game-event → vibration dispatch (vanilla frequency table, 8-block radius, wool occlusion) and wires the previously dead `SculkSensorBlock::trigger` up to events (step, entity damage, entity death, block destroy). Sensors now ring from movement/combat, and shriekers resonate from vibrations too.

RoosterBooster007 (approved):

PR #3301, opened 2026-09-07
fix(block): implement farmland trampling 🤖🤖🤖

### LINKED ISSUE #3111: Crops cannot be trampled
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.40 (Commit: 9e9f1b5/release)

### Operating System

Windows 11

Crops don't get trampled (dropped and farmland -> dirt) when jumped on.

### Steps to Reproduce

1. Plant wheat or smth
2. Bounce on it
3. It's fine?

IsaiahJaiden:
## Description

Farmland never turned into dirt when entities fell onto it, so crops could not be trampled.

This implements vanilla `FarmBlock#fallOn` via the existing `on_landed_upon` hook:

- Only living entities with a bounding box volume (`width * width * height`) above 0.512 can trample, so chickens, rabbits and other small mobs leave farmland alone.
- The chance of trampling per landing is `fall_distance - 0.5`, matching vanilla.
- Players always trample; other living entities only when the `mobGriefing` game rule is enabled.
- Fall damage is still applied exactly as before.

Additionally, the scheduled tick (farmland lost its support) and the dry-out random tick now go through a shared `turn_to_dirt` helper that fires the `minecraft:block_change` game event, matching vanilla `FarmBlock#turnToDirt`.

## Testing

- `cargo check`
- `cargo fmt --check`

Fixes #3111

## Summary by CodeRabbit

- **Bug Fixes**
  - Farmland now consistently turns into dirt when it dries out or is trampled.
  - Entities landing on farmland now receive fall damage as expected.
  - Trampling behavior now respects entity size, player permissions, mob-griefing settings, and randomized fall-distance conditions.
  - Farmland-to-dirt changes now properly trigger associated game events.
  - Moisture depletion now follows the same consistent farmland-to-dirt behavior.

### DIFF
--- a/crates/pumpkin/src/block/blocks/farmland.rs
+++ b/crates/pumpkin/src/block/blocks/farmland.rs
@@ -1,11 +1,12 @@
 use std::sync::Arc;
 
 use crate::block::{
-    BlockBehaviour, CanPlaceAtArgs, GetStateForNeighborUpdateArgs, OnPlaceArgs,
+    BlockBehaviour, CanPlaceAtArgs, GetStateForNeighborUpdateArgs, OnLandedUponArgs, OnPlaceArgs,
     OnScheduledTickArgs, PathComputationType, RandomTickArgs,
 };
 use crate::world::World;
 use pumpkin_data::block_properties::FarmlandLikeProperties;
+use pumpkin_data::game_event::GameEvent;
 use pumpkin_data::tag;
 use pumpkin_data::tag::Taggable;
 use pumpkin_data::{Block, BlockDirection, BlockState, BlockStateId};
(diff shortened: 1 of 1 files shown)


IsaiahJaiden:
Good catch — fixed in 91afd400.

The trampling check was flooring the raw entity Y, so an entity resting on top of farmland resolved to the air block above it and the `FARMLAND` check never succeeded. It now uses vanilla's `getOnPosLegacy()` offset (`floor(y - 0.2)`), matching how `Entity#fall` picks the supporting block.

I checked `turtle_egg.rs` as well — it has the same `floor(y)` pattern in `on_landed_upon` (pre-existing upstream code), which means falling entities currently can never destroy turtle eggs. Since that's outside this PR's scope I left it alone, but it's a one-line fix of the same shape if a maintainer wants to fold it in (happy to do a follow-up PR instead).

RoosterBooster007 (approved):

PR #3304, opened 2026-09-07
fix(worldgen): bring amethyst geodes to vanilla parity 🤖🤖🤖

xudoyberdi0410:
## Description

Four commits that take amethyst geodes from "never generated" to bit-identical with vanilla.

**1. Default distribution points and point offset.** `GeodeConfiguration.CODEC` declares `distribution_points`, `point_offset` and `outer_wall_distance` as `optionalFieldOf` with `UniformInt.of(3, 4)`, `UniformInt.of(1, 2)` and `UniformInt.of(4, 5)` defaults. `amethyst_geode` omits the first two and the codegen turned every absent int provider into `Constant(0)`: with zero distribution points every shell sum is 0 and `GeodeFeature.generate` never placed a block. **No amethyst geode existed anywhere in the world.**

**2. Shape noise seeded from the world seed.** Vanilla builds it as a normal noise of octave −4 on a legacy random seeded from the raw world seed: a fresh legacy random on the raw world seed, the same for every geode, consuming nothing from the feature random. Pumpkin created the noise from the decoration random, which produced a different noise per geode *and* consumed two `nextLong`s before the crack / distribution / offset draws. `GenerationCache::get_world_seed` is added so the feature can reach the raw seed.

**3. Resolve the block tags.** `invalidBlocks` / `cannotReplace` are `HolderSet`s given as `#minecraft:geode_invalid_blocks` and `#minecraft:features_cannot_replace`; Pumpkin fed the tag strings to `Block::from_name`, which returned `None`, so both sets were silently empty: a geode whose distribution point landed in water or lava was still generated, and a geode could overwrite bedrock or a spawner.

**4. Compare chance draws as doubles; grow clusters only into full water.** Vanilla widens the `nextFloat` draw to a double before comparing it against the chance for the 0.95 / 0.083 / 0.35 chances; Pumpkin narrowed the chance to f32 (the largest float below 0.95 is `0.95f32`, one draw in 2^24 flips). Buds are placed next to budding amethyst only where `BuddingAmethystBlock.canClusterGrowAtState` holds: air, or water whose `FluidState.isFull()`; `waterlogged` comes from `isSource()`. Pumpkin accepted any water state.

## Measurement

| | before | after (1) | after (2) |
|---|---|---|---|
| mismatching blocks | 69,248 (0.275 %) | 66,885 (0.266 %) | **36,199 (0.144 %)** |
| y < 0 band | — | 25,708 | **3,726** |
| geode-block category | 24,956 (vanilla) vs 0 | 29,132 | **150** |

All 150 remaining geode blocks are amethyst buds that grew by random ticks while the reference server was running (each sits on a `budding_amethyst`, facing away from it), not generation differences; the inner air of the geodes also stops reading as `air→deepslate` (5,408 blocks). Commits 3 and 4 change 0 blocks in this box (no geode here has a point in water/lava/ice/bedrock; the 2^24 case does not occur) and are committed for parity elsewhere.

## Testing

`cargo test -p pumpkin-world -p pumpkin-util --lib` green, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean. The shape-noise unit test compares Pumpkin's values against numbers printed by the real 26.2 server classes for seed 13579.

**Known impact:** amethyst geodes now generate; changes worldgen output around y −64..30.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved geode generation consistency across worlds by aligning results with the world seed.
  - Corrected geode block replacement handling, including configured block tags.
  - Improved crystal placement in air and full water, with more accurate waterlogging behavior.
  - Fixed random chance comparisons for more reliable generation outcomes.

- **Generation Improvements**
  - Geode configuration now applies standard defaults when optional distribution settings are omitted.
  - Improved deterministic behavior for geode noise and water-level handling.

### DIFF
--- a/crates/pumpkin-data/src/generated/configured_features_generated.rs
+++ b/crates/pumpkin-data/src/generated/configured_features_generated.rs
@@ -178,8 +178,16 @@ fn build_configured_features()
                     max_inclusive: 6i32,
                 },
             )),
-            distribution_points: IntProvider::Constant(0),
-            point_offset: IntProvider::Constant(0),
+            distribution_points: IntProvider::Object(NormalIntProvider::Uniform(
+                UniformIntProvider {
+                    min_inclusive: 3i32,
+                    max_inclusive: 4i32,
+                },
+            )),
+            point_offset: IntProvider::Object(NormalIntProvider::Uniform(UniformIntProvider {
+                min_inclusive: 1i32,
+                max_inclusive: 2i32,
+            })),
             min_gen_offset: -16i32,
             max_gen_offset: 16i32,
             noise_multiplier: 0.05f64,
--- a/crates/pumpkin-world/src/generation/feature/features/geode.rs
+++ b/crates/pumpkin-world/src/generation/feature/features/geode.rs
@@ -3,7 +3,7 @@ use std::collections::{HashMap, HashSet};
 use pumpkin_data::{Block, BlockDirection, BlockId, BlockState, BlockStateId};
 use pumpkin_util::{
     math::{int_provider::IntProvider, position::BlockPos},
-    random::RandomGenerator,
+    random::{RandomGenerator, legacy_rand::LegacyRand},
 };
 
 use crate::generation::feature::features::spring_feature::BlockWrapper;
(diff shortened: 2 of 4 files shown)


RoosterBooster007 (approved):

xudoyberdi0410:
Hi @ZlordHUN, why did you close this PR? It was already approved by @RoosterBooster007, who is listed as the code owner for crates/pumpkin-world/src/generation/ in CODEOWNERS. Happy to rework it if something's wrong with it.

PR #3305, opened 2026-09-07
fix(worldgen): use selected nether complex structure

### LINKED ISSUE #3298: Nether fortresses and bastions can spawn together and overlap
## Summary

Pumpkin can generate both a nether fortress and a bastion remnant from the same `nether_complexes` placement candidate. Vanilla selects only one weighted structure per candidate, so these structures should be mutually exclusive at that location.

## Vanilla behavior

The vanilla `nether_complexes` structure set has one placement:

- spacing: 27
- separation: 4
- salt: 30084232

and two weighted structures:

- fortress: weight 2
- bastion_remnant: weight 3

Vanilla's `ChunkGenerator#setStructureStarts` chooses one weighted entry, tries it, and returns after the first successful `StructureStart`. Later, `ChunkGenerator#addStructureReferences` scans neighboring chunks and adds references only to already-created `StructureStart` objects. It does not regenerate all entries in the structure set.

## Pumpkin behavior

`ProtoChunk::set_structure_starts` correctly chooses one weighted entry in `crates/pumpkin-world/src/generation/proto_chunk.rs:1279-1352`.

However, `ProtoChunk::set_structure_references` at `crates/pumpkin-world/src/generation/proto_chunk.rs:1446-1542` recomputes candidate chunks and then iterates every entry in the structure set:

```rust
for entry in set.structures {
    let structure = Structure::get(&entry.structure);
    let start_data = global_cache.get_or_compute_structure_start(...);
    ...
}
```

This independently computes both the fortress and bastion at the same candidate start chunk, ignoring which one won the weighted selection during `set_structure_starts`.

The local `break` only stops after the first intersecting entry for the current chunk. Since fortress and bastion bounding boxes differ, different surrounding chunks can receive references to different structures. Because structure population treats both `Start` and `Reference` collectors as generation tasks, the phantom reference can place pieces from the structure that never won the selection. This explains fortresses and bastions appearing together or overlapping.

## Expected behavior

Only the weighted winner from `set_structure_starts` should receive references.

## Suggested fix

Make `set_structure_references` match vanilla:

1. Scan neighboring chunks for actual stored structure starts.
2. Add references only for existing starts whose bounding boxes intersect the current chunk.
3. Do not iterate all weighted entries or regenerate candidate structures during reference creation.

Alternatively, cache the winning entry for each `(structure set, start chunk)` during `set_structure_starts` and have the reference pass use only that winner.

ydw1904:
Fixes #3298.

## Summary

- Share the vanilla weighted selection order between structure starts and structure references.
- Make the reference pass stop after the selected successful structure instead of materializing losing entries from the same structure set.
- Add a Nether regression test covering a bastion winner so no phantom fortress reference is created.

## Tests

- `cargo test -p pumpkin-world`
- `cargo clippy -p pumpkin-world --tests -- -D warnings`

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected structure start and reference selection to consistently follow weighted ordering.
  - Fixed Nether structure reference handling so valid bastion references are preserved without incorrectly reporting fortress references.

- **Documentation**
  - Added explanatory documentation for biome collection, structure ordering, structure-start generation, and structure-reference generation.

- **Tests**
  - Added Nether regression coverage for bastion and fortress references.
  - Expanded documentation for structure-reference tests.

### DIFF
--- a/crates/pumpkin-world/src/generation/proto_chunk.rs
+++ b/crates/pumpkin-world/src/generation/proto_chunk.rs
@@ -1250,6 +1250,38 @@ impl ProtoChunk {
         allowed_biomes
     }
 
+    fn weighted_structure_order(
+        entries: &[WeightedEntry],
+        seed: u64,
+        chunk_x: i32,
+        chunk_z: i32,
+    ) -> Vec<WeightedEntry> {
+        let mut candidates = entries.to_vec();
+        let large_feature_seed = get_large_feature_seed(seed, chunk_x, chunk_z);
+        let mut random = LegacyRand::from_seed(large_feature_seed);
+        let mut total_weight: u32 = candidates.iter().map(|entry| entry.weight).sum();
+        let mut ordered = Vec::with_capacity(candidates.len());
+
+        while !candidates.is_empty() {
+            let mut roll = random.next_bounded_i32(total_weight as i32);
+            let mut selected_idx = 0;
+
+            for (index, entry) in candidates.iter().enumerate() {
+                roll -= entry.weight as i32;
+                if roll < 0 {
+                    selected_idx = index;
+                    break;
+                }
+            }
+
+            let selected = candidates.remove(selected_idx);
+            total_weight -= selected.weight;
+            ordered.push(selected);
+        }
+
+        ordered
+    }
+
     pub fn set_structure_starts(&mut self, generator: &super::generator::VanillaGenerator) {
         debug_assert_eq!(self.stage, StagedChunkEnum::Biomes);
         let random_config = &generator.random_config;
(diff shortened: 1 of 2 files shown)


RoosterBooster007:
Hey @ydw1904, nice catch! Master has shifted a bit since this was opened, though. As of df983f847, the start pass now uses `get_large_feature_seed` + `LegacyRand` (matching vanilla's `LegacyRandomSource` / `setLargeFeatureSeed`). Because of that, there are a few merge conflicts, and `weighted_structure_order` will need to use that same RNG... otherwise the two passes will desync on ordering, and the phantom fortress will pop right back up. Mind giving it a quick rebase and updating the helper to match?

ydw1904:
@RoosterBooster007 Done!

RoosterBooster007:
Looks good. Thanks, @ydw1904!

PR #3307, opened 2026-09-07
Fire PlayerDropItemEvent when dropping from GUI

vyPal:
## Description
The `PlayerDropItemEvent` was previously only fired when a player dropped an item from the hotbar directly by pressing the Q key, but if a player dropped an item from a GUI (like the inventory or chest UI for example) it would completely bypass the event

## Testing

## Summary by CodeRabbit

* **New Features**
  * Added validation for item drops initiated from inventory screens.
  * Drop events can now cancel item removal or dropping.
  * Single-item and full-stack drops are validated according to the selected action.
  * Cursor items are now returned to the inventory when possible, with cancelled or undelivered items preserved safely.

### DIFF
--- a/crates/pumpkin-inventory/src/screen_handler.rs
+++ b/crates/pumpkin-inventory/src/screen_handler.rs
@@ -374,19 +374,46 @@ pub trait ScreenHandler: Send + Sync {
         self.default_on_closed(player);
     }
 
-    /// Default close behavior - drops the cursor item.
+    /// Default close behavior - returns the cursor item to the inventory,
+    /// dropping any remainder that doesn't fit.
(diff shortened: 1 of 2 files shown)


RoosterBooster007 (approved):

RoosterBooster007:
One thing that's worth a follow-up but shouldn't hold this up: the overflow drops still skip the event (`offer` at `player_inventory.rs:460`, the displaced-stack drop at `screen_handler.rs:1192`, and the craft-result overflow at `player_screen_handler.rs:235`). I'm down to open an issue to keep this on our radar, unless you'd rather widen this PR.

PR #3312, opened 2026-09-07
fix(worldgen): generate mineshafts bit-identical to vanilla 🤖🤖🤖

xudoyberdi0410:
## Description

Takes the mineshaft from "a different structure in the same chunk" to bit-identical with vanilla 26.2 for every mineshaft in the reference box. **Builds on the structure-seeding PR (#3311); its four commits are included here and will drop out once that one merges.** Without the large-feature seed fix there is no mineshaft in the box to compare against, and the 180-piece test below needs it.

The 15 commits on top, each read off `MineshaftPieces` / `MineShaftCorridor` / `StructurePiece` in the 26.2 jar and measured individually:

| commit | vanilla behaviour that was missing |
|---|---|
| depth-first pieces | `generateAndAddPiece` adds the piece and generates its children immediately; Pumpkin queued pieces FIFO and added a piece only after expanding it, so from the second piece on every draw and collision check diverged (694 pieces instead of 180 for the start in (−11, 11)) |
| protect corridor building blocks | `MineShaftPiece.canBeReplaced` refuses the type's planks, wood, fence and chains |
| maybe-box draw order | `generateMaybeBox` iterates Y, X, Z and draws `nextFloat() > chance` before the placement gates; the cobweb volume is a second call with `requireInterior` |
| supports gated on ceiling | `isSupportingBox` scans `x0..x1` at `y1 + 1` and `placeSupport` returns on any air |
| cobweb gates | `maybePlaceCobWeb` short-circuits `isInterior` *before* `nextFloat`, then needs two sturdy neighbouring faces |
| rail gates | rails only over a non-air `isSolidRender` floor, chance 0.7 interior / 0.9 otherwise |
| fence connections | `placeSupport` sets `WEST` on the x0 post and `EAST` on the x1 post |
| pillar endpoints | `fillColumnBetween` is `y < end`, not inclusive |
| pieces touching liquid | `isInInvalidLocation`: a piece whose inflated box touches any liquid is not placed at all |
| floor planks | `setPlanksBlock` is `isInterior && !isFaceSturdy(UP)`, so a shaft that breaks out above the terrain gets no plank floor |
| pillars only under planks | `placeDoubleLowerOrUpperSupport` follows a wall column down only where a plank was actually left |
| pillar / chain anchor tests | `isReplaceableByStructures && !LAVA`, `isFaceSturdy(UP)`, `canSupportCenter && !FallingBlock` |
| spawner interior check | `chunkBB.isInside && isInterior` before `hasPlacedSpider = true`; the early latch skipped the `nextInt(3)` of every later section |
| clippy pedantic allows | no behaviour change |
| minecart chest | `MineShaftCorridor.createChest` never writes a chest block: it places a rail (`nextBoolean` shape) and spawns a `MinecartChest` with `nextLong` loot seed, and consumes **nothing** when its gate fails; Pumpkin wrote a `chest` block and always burned one long |

## Measurement

| point in the series | mismatching blocks | chunks identical | chunk (−11, 11) |
|---|---|---|---|
| depth-first (early, whole box) | 395,592 → 373,566 | — | — |
| corridor gates (7 commits) | 36,725 → 34,994 | 0 | 1,187 → 1,166 |
| pieces touching liquid | 16,253 → 11,105 | — | 1,098 → 279 |
| floor planks | 7,110 → 4,810 | — | 278 → 155 |
| pillars under planks | 4,810 → 3,408 | 88 → 93 | 155 → 54 |
| spawner gate | 3,408 → 3,350 | — | — |
| (step-index seed, from the seeding PR) | 3,350 → 1,781 | 93 → 105 | 54 → 2 |
| minecart chest (with tick mask) | 148 → **32** | 245 → **248/256** | mesa shafts (−9,12) 68 → 0, (−12,13) 29 → 0, (−10,10) 19 → 0 |

The whole mineshaft bucket is zero at the end; the mineshaft-mesa fence property mismatches that remain are fixed by the post-processing PR.

## Testing

`cargo test -p pumpkin-world -p pumpkin-util --lib` green, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean. `mineshaft_mesa_pieces_match_vanilla_seed_13579` asserts the 180 pieces, their order and the bounding boxes of the first eleven against the `structures.starts` NBT of the real vanilla chunk; the chain anchor has a unit test for the `FallingBlock` set (`chains_do_not_hang_from_falling_blocks`).

**Known impact:** every mineshaft changes shape.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved structure generation to more closely match vanilla behavior and seeding.
  - Corrected mineshaft layouts, piece placement, supports, rails, cobwebs, pillars, and minecart chests.
  - Fixed structure references across all overlapping placement regions.
  - Prevented structure pieces from generating in invalid liquid locations.

- **Compatibility**
  - Mineshaft generation now produces more consistent locations, layouts, and contents across matching world seeds.

### DIFF
--- a/crates/pumpkin-util/src/random/mod.rs
+++ b/crates/pumpkin-util/src/random/mod.rs
@@ -151,14 +151,8 @@ pub const fn seed_slime_chunk(x: i32, z: i32, seed: u64, salt: u64) -> u64 {
 ///
 /// This is vanilla's `WorldgenRandom.setLargeFeatureSeed(seed, chunkX, chunkZ)`, which is
 /// shared by the carvers, the structure `GenerationContext` random and the `legacy_type_3`
-/// structure frequency reduction:
-///
-/// ```text
-/// this.setSeed(seed);
-/// long xScale = this.nextLong();
-/// long zScale = this.nextLong();
-/// this.setSeed(chunkX * xScale ^ chunkZ * zScale ^ seed);
-/// ```
+/// structure frequency reduction. It seeds with the world seed, draws two longs as the X and Z
+/// scales, and re-seeds with `chunk_x * x_scale ^ chunk_z * z_scale ^ seed`.
 ///
 /// Not to be confused with `setDecorationSeed`, which ORs the scales with 1 and *adds* the
 /// two products (see `get_population_seed`).
--- a/crates/pumpkin-util/src/random/mod.rs
+++ b/crates/pumpkin-util/src/random/mod.rs
@@ -149,7 +149,19 @@ pub const fn seed_slime_chunk(x: i32, z: i32, seed: u64, salt: u64) -> u64 {
 
 /// Generates a carver seed for cave and ravine generation.
 ///
-/// Carver seeds are used for terrain carving features like caves and ravines.
+/// This is vanilla's `WorldgenRandom.setLargeFeatureSeed(seed, chunkX, chunkZ)`, which is
+/// shared by the carvers, the structure `GenerationContext` random and the `legacy_type_3`
+/// structure frequency reduction:
+///
+/// ```text
+/// this.setSeed(seed);
+/// long xScale = this.nextLong();
+/// long zScale = this.nextLong();
+/// this.setSeed(chunkX * xScale ^ chunkZ * zScale ^ seed);
+/// ```
+///
+/// Not to be confused with `setDecorationSeed`, which ORs the scales with 1 and *adds* the
+/// two products (see `get_population_seed`).
 ///
 /// # Arguments
 /// - `world_seed` – The base world seed (plus carver index).
@@ -162,11 +174,9 @@ pub const fn seed_slime_chunk(x: i32, z: i32, seed: u64, salt: u64) -> u64 {
 #[must_use]
 pub fn get_carver_seed(world_seed: u64, chunk_x: i32, chunk_z: i32) -> u64 {
     let mut random = LegacyRand::from_seed(world_seed);
-    let l = random.next_i64() | 1;
-    let m = random.next_i64() | 1;
-    ((chunk_x as i64)
-        .wrapping_mul(l)
-        .wrapping_add((chunk_z as i64).wrapping_mul(m)) as u64)
+    let x_scale = random.next_i64();
+    let z_scale = random.next_i64();
+    ((i64::from(chunk_x).wrapping_mul(x_scale)) ^ (i64::from(chunk_z).wrapping_mul(z_scale))) as u64
         ^ world_seed
 }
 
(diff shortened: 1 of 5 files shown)


IndrajeethY:
@xudoyberdi0410 that is actually a lot of ai generated code... are you sure you play tested all of these changes?

xudoyberdi0410:
The parity work isn't verified by hand-testing — it's verified by byte-exact comparison. I generate the same seed with a real vanilla 26.2 server, dump its region files, and diff Pumpkin's chunks against them block for block. Before this batch, 3 of the 256 chunks in the test area matched vanilla exactly; after it, 253 do. Every commit is a single measured change with the mismatch count moving in the right direction, which is why the diff is large but mechanical.

I also ran this branch as a real server today, and it's worth being precise about what that did and didn't show. With no player connected it generated nothing at all: `forceload` marks the chunks in `forced_chunks` and calls `update_active_chunks`, which only recomputes what ticks — it never requests a load. So my first runs sat at 0 chunks and 0 region files, which is a separate gap and not something these PRs touch. Once I connected an actual client, 285 chunks streamed, every one with all 24 sections and no empty chunks, no panics and no errors in the log, and a save + restart round-trip returned the same chunks.

Happy to share the comparison tooling and the vanilla dumps if that would help review.

CyanStarlight:
Alright. And how does this affect performance? More xoroshiro? Less xoroshiro?

xudoyberdi0410:
I measured it rather than guessed. `cargo bench -p pumpkin-world --bench chunk_gen`, release+LTO, merge-base `8d0d0d31` vs the branch head, single chunk, seed 42:

| stage | base | branch | delta |
|---|---|---|---|
| full_chunk_generation | 38.36 ms | 39.78 ms | +3.7% |
| carvers | 2.267 ms | 1.923 ms | −15.2% |
| features | 1.617 ms | 1.821 ms | +12.6% |
| structure_starts | 2.50 µs | 6.46 µs | +4 µs |
| structure_references | 48.9 µs | 59.6 µs | +11 µs |
| noise | 10.13 ms | 10.36 ms | +2.3% |
| surface | 5.558 ms | 5.639 ms | +1.4% |
| lighting | 11.48 ms | 11.58 ms | +0.8% |

A caveat on reading those: biomes and lighting are untouched by this work and still move +2.0% and +0.8%, so roughly ±2% here is build-layout noise between two different binaries. The honest reading is ~1.5–2% on a full chunk, with carvers measurably *faster* (the `replaceable` / `lava_level` fixes reject positions earlier) and features measurably slower. The structure stages move a lot in percentage terms and almost nothing in absolute terms — 4 µs and 11 µs against a 38 ms chunk.

On xoroshiro specifically, the mineshaft draws are close to a wash. The corridor chest now costs two draws instead of one (`nextBoolean` for the rail shape, `nextLong` for the loot seed, which is what vanilla does), the cobweb gate consumes none when the piece isn't interior where it previously drew unconditionally, and `isInInvalidLocation` rejects a piece before it is built, which saves the whole piece rather than a single draw. The structure stages above are the ceiling on all of it: ~15 µs out of a 38 ms chunk.

Note these numbers cover the whole parity branch, not this PR alone — piece placement happens inside the features stage together with the geode, spring, multiface and ore-vein changes. Isolating this PR on its own base would mean two more full LTO builds, so I'll leave it at the branch-level number unless that distinction actually matters for the review.

PR #3315, opened 2026-09-07
fix(worldgen): place fossils and ruined portals the way vanilla does 🤖🤖🤖

xudoyberdi0410:
## Description

**Builds on the heightmaps PR (#3314) and the structure-seeding PR (#3311); their commits are included here and drop out once those merge.** The ruined portal writes its mound through the whole generation window (which needs the live `_WG` heightmaps on neighbour chunks) and is placed from the per-structure random stream.

**1. Fossils burn the palette draw.** `StructureTemplate.placeInWorld` opens with `settings.getRandomPalette(palettes, position)`, which is `palettes.get(random.nextInt(paletteSize))` unconditionally. Every fossil template ships one palette, but `nextInt(1)` still consumes one `nextInt()` off the feature random before the first `BlockRotProcessor` draw. Pumpkin's `place_fossil_template` went straight into the block loop, so both the fossil pass and the ore-overlay pass read the rot stream one draw early.

**2. Ruined portal.** Four things kept it (and its netherrack mound) off vanilla's position; all four are needed before any of them shows in a dump:
* `RuinedPortalStructure.findGenerationPoint` was not ported: origin is the chunk's world position (not centre − pivot), the surface is sampled at the centre of the rotated box, the weighted `setups` list, `sample(airPocketProbability)`, the 5 % giant-portal branch and `findSuitableY` (whose corner scan needs real noise columns; `NoiseHeightSampler` now caches the whole `getBaseColumn` per column and answers `column_is_opaque`).
* `BoundingBox.getCenter()` is `min + span / 2` with an inclusive span, not the midpoint of the corners (x = 5, not 4, on the 10-wide box). Added `BlockBox::center`.
* `placeInWorld` takes the chest's `LootTableSeed` off the *feature* random (`random.nextLong()`), before `spreadNetherrack`. Pumpkin derived it from `hashBlockPos` and ran the whole mound one draw early.
* `RuinedPortalPiece.postProcess` writes the mound through the `WorldGenLevel`, up to 14 blocks into neighbouring chunks. Pumpkin handed pieces the centre `ProtoChunk` and clipped (and, through `x & 15`, silently *wrapped*) every write. Structure placement now gets the whole `GenerationCache` window, and pieces that write outside their chunk do it in a new `StructurePieceBase::place_across_chunks` hook.

## Measurement

| | mismatching blocks | chunks identical |
|---|---|---|
| fossil palette draw | 874 → 784 | 156 → 160/256 |
| ruined portal (all four) | 784 → **669** | 160 → **161/256** |

Every fossil in the box is bit-identical after (1) (`bone_block` / `deepslate` / `deepslate_diamond_ore` confusions 38 → 0). For (2): chunk (−1, 15) goes 107 → 1 and (−1, 14) 9 → 0; no other chunk moves. Intermediate steps: 784 → 784 (points 1–2 invisible while writes are clipped) → 800 (region writes, old centre) → 787 (`getCenter`) → 669 (loot-seed draw).

## Testing

`cargo test -p pumpkin-world -p pumpkin-util --lib` green, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean. Unit tests: `placing_a_template_burns_the_palette_draw_before_the_rot_draws`, `a_ruined_portal_is_anchored_at_the_chunk_corner`, `the_desert_setup_matches_the_datapack` and `a_bounding_box_centre_uses_the_inclusive_span`.

**Known impact:** ruined portals and fossils move to vanilla's positions.

### DIFF
--- a/crates/pumpkin-util/src/math/block_box.rs
+++ b/crates/pumpkin-util/src/math/block_box.rs
@@ -232,6 +232,25 @@ impl BlockBox {
         self.max.y - self.min.y + 1
     }
 
+    /// Vanilla `BoundingBox.getCenter()`:
+    ///
+    /// ```java
+    /// return new BlockPos(this.minX() + this.getXSpan() / 2,
+    ///                     this.minY() + this.getYSpan() / 2,
+    ///                     this.minZ() + this.getZSpan() / 2);
+    /// ```
+    ///
+    /// The span is inclusive (`max - min + 1`), so on an even span this sits one block
+    /// past the arithmetic midpoint of the corners.
+    #[must_use]
+    pub const fn center(&self) -> Vector3<i32> {
+        Vector3::new(
+            self.min.x + (self.max.x - self.min.x + 1) / 2,
+            self.min.y + (self.max.y - self.min.y + 1) / 2,
+            self.min.z + (self.max.z - self.min.z + 1) / 2,
+        )
+    }
+
     /// Expands this box to encompass another box.
     ///
     /// # Arguments
--- a/crates/pumpkin-util/src/random/mod.rs
+++ b/crates/pumpkin-util/src/random/mod.rs
@@ -149,7 +149,19 @@ pub const fn seed_slime_chunk(x: i32, z: i32, seed: u64, salt: u64) -> u64 {
 
 /// Generates a carver seed for cave and ravine generation.
 ///
-/// Carver seeds are used for terrain carving features like caves and ravines.
+/// This is vanilla's `WorldgenRandom.setLargeFeatureSeed(seed, chunkX, chunkZ)`, which is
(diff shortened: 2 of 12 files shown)


crates/pumpkin-util/src/math/block_box.rs
```diff
@@ -232,6 +232,25 @@ impl BlockBox {
         self.max.y - self.min.y + 1
     }
 
+    /// Vanilla `BoundingBox.getCenter()`:
+    ///
+    /// ```java
```
RoosterBooster007, line 237: You appear to have included actual mc java code in here, which we cannot accept. Also, please cut down on the AI comments before opening a new PR.

PR #3318, opened 2026-09-07
fix(chest): unpack loot tables for both halves of double chests

ydw1904:
## Problem

Structure loot chests (e.g. bastion remnants) that generate as double chests only had loot in one half. Worldgen stores a deferred `LootTable` tag on **each** chest block's block entity, but `get_chest_screen_handler_factory` only unpacked the loot table of the chest half that was clicked:

- the clicked half filled with loot
- the other half stayed empty and kept its pending loot table forever, since a double chest can never be opened one half at a time

## Fix

Compute the connected-half direction first, then unpack deferred loot tables for **both** halves on first open (still skipping spectators), matching vanilla's `CompoundContainer` behavior where both halves unpack lazily through the combined container.

## Testing

- `cargo check -p pumpkin` clean
- Manually verified on a bastion: both halves of double chests now contain loot on first open

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed double-chest opening behavior so deferred loot is unpacked correctly across both halves.
  * Connected chest loot is now unpacked only when the opening area is unobstructed.
  * Preserved spectator behavior by preventing connected-half loot from being unpacked for spectators.
  * Ensured the first chest’s deferred loot is available before obstruction checks are performed.
  * Improved consistency when opening chests with connected halves and obstructed areas.

### DIFF
--- a/crates/pumpkin/src/block/blocks/chests.rs
+++ b/crates/pumpkin/src/block/blocks/chests.rs
@@ -166,6 +166,10 @@ fn get_chest_comparator_output(args: &GetComparatorOutputArgs<'_>) -> Option<u8>
     }
 }
 
+/// Returns a screen handler factory for opening a chest or double chest inventory.
+///
+/// Unpacks deferred loot tables on first open (for non-spectator players) and combines
+/// connected inventories for double chests if neither chest half is obstructed.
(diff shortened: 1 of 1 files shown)


RoosterBooster007:
Thanks, @ydw1904!

PR #2786, opened 2026-08-05
perf: preallocate direct palette data

nobuildersnotools:
## Description

Preallocates the exactly known packed-data capacity when serializing direct block and biome palettes, avoiding vector growth without changing output

## Testing

Temporary block-palette serialization bench:
1.177 µs -> 1.001 µs

PR #3320, opened 2026-09-07
perf(world): avoid repeated scans when packing dense palettes

JulesB40:
## Problem and change

Saving a chunk section with more than 256 distinct block states uses dense storage. `to_palette_and_packed_data` previously scanned the entire palette to find the index of each of the 4096 blocks. At 4096 states this performs about 8.4 million comparisons per section.

Build a temporary `FxHashMap` for dense palettes larger than 256 entries and reuse it while packing. Keep the indexed path and small dense scans. Preserve palette order and the first-index behavior for duplicate palette entries. This trades O(palette size) temporary memory for expected O(block count + palette size) lookup work.

## Validation

- All 175 world-crate unit tests passed, including new exact-output tests at 16/256/257/1024/4096 states, partial packed words, reversed palette order and duplicate values.
- `cargo clippy -p pumpkin-world --all-targets --all-features --release --locked -- -D warnings`, formatting and `git diff --check` passed.
- Added the Criterion `palette_packing` benchmark. It invokes the production method with fixture construction outside timing and lookup allocation inside timing. Whole-server TPS and full chunk-save latency are not measured.

Criterion point estimates, per section:

| Distinct states | Before | After (repeat) | Speedup |
| --- | --- | --- | --- |
| 16, indexed control | 1.304 us | 1.331 us | Within noise threshold |
| 256, indexed control | 1.620 us | 1.595 us | No significant change |
| 257, dense | 136.13 us | 7.167 us | 19x |
| 1024, dense | 475.26 us | 9.307 us | 51x |
| 4096, dense | 1781.8 us | 21.973 us | 81x |

The initial 16-state comparison flagged a roughly 5% slowdown; the longer repeat classified it within the noise threshold. These synthetic cases establish the dense-packing improvement, not a universal performance gain.

Base: `786e2fbf782f92113543fcb46eaf95ba53835d41`. Intel i9-13900HX, Windows, Rust 1.98.1; both versions use the bench profile with `CARGO_PROFILE_RELEASE_LTO=false` and `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=16`, without lowering world-crate optimization. Baseline: 20 samples, 1-second warmup, 2-second measurement; final repeat: 40 samples, 2-second warmup, 4-second measurement.

Reproduce from repository root:

```text
cargo bench -p pumpkin-world --bench palette_packing -- --save-baseline before
# Apply the source fix, keeping the same benchmark/profile.
cargo bench -p pumpkin-world --bench palette_packing -- --baseline before
```

Searched existing PRs for overlap. Merged #2492 addresses bulk palette construction and #2786 direct-network output preallocation; this change targets dense palette index packing during saving. Prepared with Codex assistance.

## Summary by CodeRabbit

- **Performance**
  - Improved serialization efficiency for large block palettes, reducing lookup overhead as palette sizes grow.
  - Preserved efficient handling for smaller palettes.

- **Bug Fixes**
  - Ensured duplicate palette values consistently resolve to their first matching index.
  - Improved correctness when packing palette data, including partial word boundaries and large palettes.

### DIFF
--- a/crates/pumpkin-world/Cargo.toml
+++ b/crates/pumpkin-world/Cargo.toml
@@ -65,6 +65,10 @@ name = "chunk"
 harness = false
 
 [[bench]]
+name = "palette_packing"
+harness = false
+
+[[bench]]
 name = "chunk_io"
 harness = false
 
--- a/crates/pumpkin-world/benches/palette_packing.rs
+++ b/crates/pumpkin-world/benches/palette_packing.rs
@@ -1,0 +1,28 @@
+use std::hint::black_box;
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use pumpkin_world::chunk::palette::BlockPalette;
+
+fn palette_packing(c: &mut Criterion) {
+    let mut group = c.benchmark_group("palette_packing");
+    for size in [16usize, 256, 257, 1024, 4096] {
+        let palette: Vec<_> = (0..size as u16)
+            .map(pumpkin_data::BlockStateId::new_or_air)
+            .collect();
+        let bits = pumpkin_util::encompassing_bits(size).max(4);
+        let per_word = 64 / bits as usize;
+        let mut packed = vec![0i64; 4096usize.div_ceil(per_word)];
+        for index in 0..4096 {
+            packed[index / per_word] |=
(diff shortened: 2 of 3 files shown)


RoosterBooster007:
Looks good. Waiting on a review from @Snowiiii.

PR #3329, opened 2026-09-07
fix(entity): implement sulfur cube AI and synced size 🤖🤖🤖

IsaiahJaiden:
## Description

Sulfur cubes (the new 26.2 cube mob) were completely inert, and spawned smaller than they should. Two root causes:

1. `Entity::from_type` had no `SULFUR_CUBE` arm, so sulfur cubes fell back to the generic `LivingEntity` wrapper with no goals and no move control — they never moved, jumped, or did anything.
2. The slime family's `Size` tracked data was stored internally but never synced to clients, so cube mobs always rendered at the default size 1 on the client (living entities are spawned via the living-entity packet, which ignores spawn-packet data).

This PR implements a `SulfurCubeEntity` modeled on the existing slime implementation, following vanilla `AbstractCubeMob`/`SulfurCube` (26.2):

- Goals: float (water/lava), random direction (only while grounded or in water/lava and no target), keep-on-jumping — the passive subset; sulfur cubes are not aggressive and have no target/attack goals.
- Hop move control with a 10–30 gametick jump delay and yaw lerping, matching vanilla `CubeMobMoveControl`.
- `set_size` with vanilla semantics: clamped 1–127, health = 4 × size, movement speed = 0.2 + 0.1 × size, scaled bounding box/eye height, and (vanilla `SulfurCube.setSize` override) size 1 adults become babies.
- The `Size` tracked data is now actually synced via `set_synced_data`, so the client renders cubes at their real size.
- Baby flag (`Age` + `DATA_BABY_ID` tracked data) with vanilla polarities (size-1 on `set_size` → baby; adult size 2 on spawn, matching vanilla `finalizeSpawn`/`setSpawnSize`, so `/summon` gives a large cube).
- Death splitting: cubes larger than size 1 split into 2 half-size babies at the vanilla offsets.
- Small/large jump/squish/hurt sounds at the vanilla category (neutral) and volume (0.4 × size).
- NBT round-trip for `Size` (stored size-1, like vanilla), `Age`, and `wasOnGround`.

Bonus fix from the same root cause: `SlimeEntity::set_size` now syncs its `Size` tracked data too, so slimes and magma cubes render at their actual size instead of always size 1.

Out of scope (left as TODOs, matching vanilla behavior that needs more infrastructure): bucketing, item swallowing/equipment archetypes, primed explosion behavior, shearing, and the tempt/search-for-items goals.

## Testing

- `cargo fmt --check` clean
- Summoned sulfur cubes via `/summon`: they now hop around with vanilla-like jump cadence, render at large size, and split into two small babies on death; slimes/magma cubes render at their real size.

## Summary by CodeRabbit

- **New Features**
  - Added sulfur cube entities with configurable sizes, baby variants, movement, jumping, floating, and random-direction behaviors.
  - Sulfur cubes now squish, play size-specific sounds, save their state, and split into smaller cubes after death.
  - Added support for creating and spawning sulfur cubes in the game world.

- **Bug Fixes**
  - Slime size changes now remain synchronized with displayed entity data.

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -31,6 +31,7 @@ use crate::entity::attributes::ModifierOperation;
 use crate::entity::combat::{CombatRules, CombatTracker, FallLocation, knockback_after_resistance};
 use crate::entity::mob::equipment::DEFAULT_EQUIPMENT_DROP_CHANCE;
 use crate::entity::mob::slime::SlimeEntity;
+use crate::entity::mob::sulfur_cube::SulfurCubeEntity;
 use crate::entity::player::statistics::{CustomStatistic, StatisticCategory};
 use crate::server::Server;
 use crate::world::loot::LootContextParameters;
@@ -2311,6 +2312,8 @@ impl LivingEntity {
     fn hurt_sound(&self) -> Sound {
         if self.entity.entity_type == &EntityType::SLIME {
             SlimeEntity::hurt_sound_for_size(self.entity.data.load(Relaxed))
+        } else if self.entity.entity_type == &EntityType::SULFUR_CUBE {
+            SulfurCubeEntity::hurt_sound_for_size(self.entity.data.load(Relaxed))
         } else {
             Self::hurt_sound_for_entity(self.entity.entity_type)
         }
--- a/crates/pumpkin/src/entity/mob/mod.rs
+++ b/crates/pumpkin/src/entity/mob/mod.rs
@@ -62,6 +62,7 @@ pub mod silverfish;
 pub mod skeleton;
 pub mod slime;
 pub mod spider;
+pub mod sulfur_cube;
 pub mod vex;
 pub mod vindicator;
 pub mod warden;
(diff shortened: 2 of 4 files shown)


IsaiahJaiden:
Valid finding — fixed in 2dd9012e.

The three goals (`SulfurCubeFloatGoal`, `SulfurCubeRandomDirectionGoal`, `SulfurCubeKeepOnJumpingGoal`) stored `Arc<SulfurCubeEntity>` while being owned by the entity's own goal selector, forming a `MobEntity -> goal selector -> goals -> Arc` reference cycle. They now hold `Weak<SulfurCubeEntity>` and no-op when the reference is gone, matching what `SulfurCubeMoveControl` in the same file already does.

Worth noting this cycle pattern also exists in `slime.rs`/magma cube (`SlimeFloatGoal` etc. store `Arc<SlimeEntity>`) — I left that alone since it's upstream scope for this PR, but it'd make a good small follow-up.

RoosterBooster007 (approved):

RoosterBooster007:
Looks good. Btw, the hop control and float/random-direction goals are now a near-duplicate of slime's. Vanilla shares those via `AbstractCubeMob`, so a shared cube-mob base would be a nice future cleanup, but not required here.

IsaiahJaiden:
Thanks for the approval! Agreed on the cleanup — the hop control and float/random-direction goals are structurally identical to slime's because they came from vanilla's shared `AbstractCubeMob`. If a shared cube-mob base (move control + float/random-direction/keep-on-jumping goals parameterized over the cube type) lands later, sulfur cube, slime, and magma cube could all delegate to it — happy to help with that in a follow-up if it's wanted.

Issue #3264, opened 2026-09-05
300-500% CPU spike lasting ~1.5 min on first player join per server session (only ~5s on subsequent rejoins); persistent 55-60% idle floor afterward


C1ht:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.45

### Operating System

Debian 13 (Trixie), running inside Docker (ghcr.io/pterodactyl/yolks:alpine build image) on a Proxmox VM. Host: dual Intel Xeon E5-2680 v4 (28c/56t total).

### Bug Description & Expected Behavior

<img width="702" height="233" alt="Image" src="https://github.com/user-attachments/assets/3505aafc-3476-40ca-9ca8-d6f13ccd588b" />

With zero players connected, the server idles at roughly 5% CPU, which is expected and reasonable.

The high-CPU behavior appears to be specifically tied to the *first* player join since the server process was last started — not to server uptime, world age, or whether the world is fresh or previously saved.

- On the very first player join after a server (re)start: CPU immediately jumps to 300-500% (roughly 3-5 full cores) and stays elevated for approximately 1.5 minutes before beginning to decline.
- On subsequent joins by the same or a different player (leaving and rejoining, without restarting the server process) during that same server session: the same type of spike still occurs, but is dramatically shorter — only around 5 seconds.
- After the initial long spike settles, CPU usage stabilizes at a steady-state of around 55-60% for a single idle, stationary, connected player, and remains there indefinitely regardless of player activity.
- Restarting the server process resets this: the next first-join after any restart reproduces the full ~1.5 minute spike again, regardless of whether the world is new or previously existing/saved.

Expected: whatever one-time work is happening on first join shouldn't require a sustained ~1.5 minute period of 300-500% CPU load, and the subsequent steady-state of 55-60% for one idle player also seems too high given Rust's architectural advantages over JVM-based alternatives like Paper/Purpur.

<img width="725" height="276" alt="Image" src="https://github.com/user-attachments/assets/29894a54-ec29-4b7b-b97f-879bfc9484c8" />

### Steps to Reproduce

1. Start a fresh Pumpkin server process (world can be new or previously existing — doesn't affect the result). Confirm baseline CPU usage (~5%) with zero players connected.
2. Have one player join for the first time since this server process started.
[...]

PR #3333, opened 2026-09-07
perf: batch chunk inhabited time updates

JulesB40:
## Description

Inhabited-time updates perform a DashMap lookup and one atomic increment per active chunk every tick. Letting Rayon split this small workload freely creates avoidable scheduling overhead with a stationary player. This was found while investigating the sustained CPU reported in #3264.

Set a minimum parallel task length of 1024 for this iterator. Small active sets execute without subdivision; larger sets retain parallelism. Chunk selection, relaxed atomic increments, and tick frequency are unchanged.

## Testing

On Windows / i9-13900HX (32 logical processors), a local server test with one stationary Java client, view/simulation distance 10, prebuilt air chunks, and mob spawning/random ticks disabled before joining measured:

| 30-second stationary sample | Before, % of one core | After, % of one core |
| --- | --- | --- |
| First pair | 14.17 | 0.89 |
| Repeat pair | 11.09 | 1.25 |

The mean is 12.63% before and 1.07% after, about 91.5% lower CPU **in this controlled workload**. Runs were sequential, without concurrent compilation, with successful client connections and clean shutdowns. Both binaries were built from `786e2fbf782f92113543fcb46eaf95ba53835d41` using Rust 1.98.1, release opt-level 3, LTO disabled and 16 codegen units; the candidate differs only by this patch.

A separate 20 Hz model of the lookup/increment loop checked all counters after 200 ticks. Minimum task length 1024 reduced mean work time from 523 to 33 microseconds at 441 chunks, 611 to 210 at 4225, and 778 to 607 at 16641. This models the loop with atomic counters, not full ChunkData; short-run process CPU readings were variable, so these are elapsed-work measurements.

The normal generated-world control was mixed: its warming phase measured 48.65% before / 37.19% after, but its subsequent stationary phase measured 25.78% / 30.73%. This does **not** establish an overall normal-world CPU improvement. This PR addresses the isolated scheduling overhead and does not claim to resolve the first-join generation spike or all symptoms in #3264.

Validation: all 759 workspace nextest tests, release all-target/all-feature Clippy, formatting, whitespace checks, and eight local harness tests passed. Functional tests use pumpkin opt-level 0; performance measurements use the optimized binaries described above. No overlapping inhabited-time PR was found when checking open PRs.

## Summary by CodeRabbit

* **Refactor**
  * Improved the efficiency of world update processing by batching background work more effectively.
  * Gameplay behavior and visible functionality remain unchanged in this release.

### DIFF
--- a/crates/pumpkin/src/world/mod.rs
+++ b/crates/pumpkin/src/world/mod.rs
@@ -2032,14 +2032,18 @@ impl World {
             });
         }
 
-        // Update chunk inhabited time for active chunks in parallel with Rayon
+        // Batch these cheap lookups and atomic increments to avoid waking Rayon
+        // workers for tiny tasks every tick, while retaining parallelism for large sets.
         let loaded_chunks = self.level.loaded_chunks.clone();
         let active_chunks_vec: Vec<_> = active_chunks.iter().copied().collect();
-        active_chunks_vec.par_iter().for_each(|pos| {
-            if let Some(chunk) = loaded_chunks.get(pos) {
-                chunk.inhabited_time.fetch_add(1, Relaxed);
-            }
-        });
+        active_chunks_vec
+            .par_iter()
+            .with_min_len(1024)
+            .for_each(|pos| {
+                if let Some(chunk) = loaded_chunks.get(pos) {
+                    chunk.inhabited_time.fetch_add(1, Relaxed);
+                }
+            });
     }
 
     pub fn check_fluid_collision(&self, bounding_box: BoundingBox) -> bool {
--- a/crates/pumpkin/src/world/mod.rs
+++ b/crates/pumpkin/src/world/mod.rs
@@ -1896,6 +1896,7 @@ impl World {
     #[expect(clippy::too_many_lines)]
     pub fn tick_chunks(self: &Arc<Self>, server: &Arc<Server>) {
         const BATCH_SIZE: usize = 32;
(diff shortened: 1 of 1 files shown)


RoosterBooster007 (approved):

crates/pumpkin/src/world/mod.rs
```diff
@@ -2032,14 +2032,18 @@ impl World {
...
         let loaded_chunks = self.level.loaded_chunks.clone();
         let active_chunks_vec: Vec<_> = active_chunks.iter().copied().collect();
-        active_chunks_vec.par_iter().for_each(|pos| {
-            if let Some(chunk) = loaded_chunks.get(pos) {
-                chunk.inhabited_time.fetch_add(1, Relaxed);
-            }
-        });
+        active_chunks_vec
+            .par_iter()
+            .with_min_len(1024)
```
RoosterBooster007: Looks good. My only nit: the rest of the file names its batch sizes (`BATCH_SIZE`, `ENTITY_TICK_BATCH_SIZE`), so pulling 1024 out into a `const` would fit the house style.

JulesB40: Done in b5c821ddf: extracted INHABITED_TIME_BATCH_SIZE alongside BATCH_SIZE.

RoosterBooster007:
Thanks, @JulesB40!

PR #3334, opened 2026-09-07
fix(loot): support block_state_property conditions and pool-level functions in loot tables 🤖🤖🤖

### LINKED ISSUE #3292: Berries duplicate when placing them and breaking them without letting them grow
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

55762a5/release

### Operating System

So when you plant a berry and break it quickly it duplicates the amount of berries.

I believe it should rather just just give you 1 berry back instead of 2 berries, which does the duplication

https://github.com/user-attachments/assets/51a4bb67-a31a-4438-b490-a43c18f7c8be

### Steps to Reproduce

1. Get berries
2. Plant berry
3. Quickly destroy the planted berry
4. You get 2 berries instead of just 1

### LINKED ISSUE #3293: Composter gives you free bonemeal every time you break it
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

55762a5/release

### Operating System

### Steps to Reproduce

1. Get a composter
2. Place composter
3. Break composter

IsaiahJaiden:
## Description

The loot table codegen silently mapped every condition type it didn't know to `LootCondition::None` (i.e. unconditional). This broke every loot table gated on `minecraft:block_state_property`:

- Breaking a **composter** at any fill level dropped a bone meal — the drop is gated on `level=8` (#3293).
- Breaking a **sweet berry bush** always dropped 2 berries at every age instead of following the vanilla age gates (#3292).

The codegen also only read `set_count` / `apply_bonus` functions from the entries themselves, ignoring the pool-level and table-level function layers that vanilla data uses. Vanilla `sweet_berry_bush.json` puts its functions on the pools, so berry counts collapsed to a constant 1 and the fortune bonus was dropped entirely.

### Changes

**Codegen (`tools/pumpkin-codegen`):**
- Parse `minecraft:block_state_property` conditions into a new `LootCondition::BlockStateProperty` variant.
- Thread pool-level and table-level functions through entry extraction so `set_count` / bonus formulas from any layer apply (entry → pool → table, first match wins, matching vanilla resolution).
- Normalize path separators when collecting JSON files so the generated keys/identifiers are identical on Windows and Unix hosts (previously a Windows regen produced `minecraft:blocks\foo` keys).

**Runtime (`pumpkin` / `pumpkin-util`):**
- Evaluate `BlockStateProperty` against the loot context's block state: the block must match and every specified property must match (unspecified properties ignored), matching vanilla semantics.

**Regenerated `pumpkin-data` loot tables** now honor block state conditions everywhere — composter (`level=8`), sweet berry bush (`age=2`/`age=3` with counts 1-2/2-3 + fortune), snow (`layers=1`), and ~30 other previously-unconditional pools.

## Testing

- `cargo fmt --check` clean
- Verified regenerated tables: `BLOCKS_COMPOSTER` bone meal pool now carries `level=8`; `BLOCKS_SWEET_BERRY_BUSH` pools carry `age=3` (2-3 berries) / `age=2` (1-2 berries) + `UniformBonusCount` fortune; age 0-1 bushes drop nothing (no pool condition passes)

Fixes #3293
Fixes #3292

## Summary by CodeRabbit

- **New Features**
  - Loot tables can apply conditions based on a block’s type and state properties.
  - Block-state conditions support optional namespace prefixes and multiple property checks.
  - Conditions without additional properties succeed when the specified block type matches.
  - Loot-table functions, including count and bonus formulas, are preserved through nested and combined entries.
  - Repeated random conditions now resolve independently, preserving separate rolls.
  - Loot-table files are processed consistently across platforms.

### DIFF
--- a/crates/pumpkin-data/src/generated/loot_table.rs
+++ b/crates/pumpkin-data/src/generated/loot_table.rs
@@ -711,7 +711,10 @@ static BLOCKS_ACACIA_DOOR_POOL0_ENTRIES: &[LootEntry] = &[LootEntry {
     weight: 1i32,
     min_count: 1i32,
     max_count: 1i32,
-    condition: LootCondition::None,
+    condition: LootCondition::BlockStateProperty {
+        block: "minecraft:acacia_door",
+        properties: &[("half", "lower")],
+    },
     bonus_formula: None,
 }];
 static BLOCKS_ACACIA_DOOR_POOLS: &[LootPool] = &[LootPool {
@@ -1397,7 +1400,10 @@ static BLOCKS_BAMBOO_DOOR_POOL0_ENTRIES: &[LootEntry] = &[LootEntry {
     weight: 1i32,
     min_count: 1i32,
     max_count: 1i32,
-    condition: LootCondition::None,
+    condition: LootCondition::BlockStateProperty {
+        block: "minecraft:bamboo_door",
+        properties: &[("half", "lower")],
+    },
     bonus_formula: None,
 }];
 static BLOCKS_BAMBOO_DOOR_POOLS: &[LootPool] = &[LootPool {
@@ -1768,7 +1774,10 @@ static BLOCKS_BEETROOTS_POOL0_ENTRIES: &[LootEntry] = &[
         weight: 1i32,
         min_count: 1i32,
         max_count: 1i32,
-        condition: LootCondition::None,
+        condition: LootCondition::BlockStateProperty {
+            block: "minecraft:beetroots",
+            properties: &[("age", "3")],
+        },
         bonus_formula: None,
     },
     LootEntry {
@@ -1804,7 +1813,10 @@ static BLOCKS_BEETROOTS_POOLS: &[LootPool] = &[
         min_rolls: 1i32,
         max_rolls: 1i32,
         empty_weight: 0i32,
-        condition: LootCondition::None,
+        condition: LootCondition::BlockStateProperty {
+            block: "minecraft:beetroots",
+            properties: &[("age", "3")],
+        },
     },
 ];
 pub static BLOCKS_BEETROOTS: LootTable = LootTable {
@@ -1887,7 +1899,10 @@ static BLOCKS_BIRCH_DOOR_POOL0_ENTRIES: &[LootEntry] = &[LootEntry {
     weight: 1i32,
     min_count: 1i32,
     max_count: 1i32,
-    condition: LootCondition::None,
+    condition: LootCondition::BlockStateProperty {
+        block: "minecraft:birch_door",
+        properties: &[("half", "lower")],
+    },
     bonus_formula: None,
 }];
 static BLOCKS_BIRCH_DOOR_POOLS: &[LootPool] = &[LootPool {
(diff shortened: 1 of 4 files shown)


IsaiahJaiden:
Thanks @coderabbitai — verified both findings against the code:

**Finding 1 (nested loot table pool conditions) — valid, fixed in f23126d7.**
The three nested `minecraft:loot_table` recursion sites were overwriting the pool condition with `pool_cond = parsed`, which dropped the parent entry's condition and all pool conditions except the last one. All three sites now collect `entry_cond` plus every parsed pool condition into a `Vec<LootCondition>` and merge through a new `merge_conditions` helper (drops `None`, dedupes, falls back to `AllOf`), and `combine_conditions` now delegates to that same helper. Regenerating produced no diff on the current dataset (no nested-condition tables are present yet), so this was a latent fix for future data — exactly the kind of thing worth catching.

**Finding 2 (alternatives children dropping conditions) — false positive.**
A child entry's own conditions are never dropped in this path. `extract_entries_with_depth` re-combines the inherited condition with the child's own conditions on *every* call, before any `entry_type` dispatch:

```rust
let entry_cond = match (inherited_condition, combine_conditions(&entry.conditions)) {
    (LootCondition::None, cond) | (cond, LootCondition::None) => cond,
    (first, second) if first == second => first,
    (first, second) => LootCondition::AllOf(Box::leak(vec![first, second].into_boxed_slice())),
};
```

So an `alternatives` child with a `block_state_property` condition still ends up as `AllOf([entry_cond, block_state_property])` (or just the child's condition when the parent is `None`) — the `else` branch it falls into only decides which child iterator gets walked, not the resulting condition.

One genuine nuance that survives (pre-existing): inside the `alternatives` branch, when a silk-touch/shears variant has already been seen, the negative tool condition on later children *replaces* rather than combines with the entry condition. That's a much rarer case, but worth a follow-up if we want it exact.

IsaiahJaiden:
Thanks for the review — one valid catch and one I believe is already correct:

**1. Repeated stochastic conditions** — Valid and fixed in 20fbd818. `merge_conditions` deduplicated equal conditions and the inherited/own merge collapsed identical pairs, so two identical `random_chance` terms in an `all_of` would roll once instead of twice. The merger now only drops `None`s, and equal pairs are combined rather than collapsed. Regeneration wraps the previously-collapsed deterministic duplicates (silk touch / shears pairs) in `AllOf`, which is semantically identical for those and matches vanilla keeping both condition nodes.

**2. Retaining ancestor functions when entering a nested loot table** — I checked this against vanilla 26.2 and I don't believe outer functions should cascade. `NestedLootTable#createItemStack` delegates straight to the referenced table's `getRandomItemsRaw` (NestedLootTable.java:51-54), and the output is decorated only with the nested table's own composite function (LootTable.java:90). The outer entry/pool/table functions are never applied to items produced by a referenced table — nested tables are self-contained in vanilla (loop protection works via visited-element tracking there vs our depth cap here). So the current generator behavior (passing the *nested table's own* pool/table functions into the recursion, not the caller's) matches vanilla. Happy to be proven wrong with a data file where it matters, but I didn't want to add cascading that vanilla doesn't do.

IsaiahJaiden:
Both valid — fixed in da901a5d:

**1. Nested-table merge on every path** — Correct, and thanks for catching this: my earlier fix (f23126d7) only actually rewrote the name-resolved path; the direct-reference and inline paths still did `pool_cond = parsed`, dropping the inherited condition and all but the last pool condition. All three nested paths now build the condition list starting from `entry_cond` and run it through `merge_conditions`.

**2. Functions on tag entries** — Also correct: the tag branch hardcoded `min_count: 1, max_count: 1, bonus: None`, ignoring entry/pool/table-level functions. The set_count + apply_bonus/enchanted_count_increase resolution is now a shared `resolve_functions` helper (first layer providing a value wins), used by both the item and tag branches.

Regeneration produced no diff against the committed tables (the current 26.2 dataset has no tag entries carrying functions and no nested-table conditions), so these are latent-correctness fixes. CI parity green (check/fmt/clippy/test).

PR #3337, opened 2026-09-08
fix(command): include command name in unknown-command errors

### LINKED ISSUE #3265: Placeholder error
### Before Submitting

### Minecraft Edition

Both / Crossplay

### Server Version / Commit

0.1.0-dev+26.2-26.45 (Commit: 8d0d0d3/release)

### Operating System

CachyOS 7.2.2-1

current:
```
$ x
Unknown command: %s. Please check that the command exists and that you have permission to use it.
x<--[HERE]
```
expected:
```
$ x
Unknown command: x. Please check that the command exists and that you have permission to use it.
x<--[HERE]
```

### Steps to Reproduce

1. Start the server.
2. Type an invalid command.
3. Notice the %s

JulesB40:
Unknown commands currently render as `Unknown command: %s...` in the console
and Bedrock output because the Bedrock translation expects one argument but
`DISPATCHER_UNKNOWN_COMMAND` supplies none.

Give that error one argument and supply the first command token through a shared
dispatcher helper. Unknown input, commands without an executor, and disabled
commands now include the command name. Extra command arguments are excluded
from the substitution, and the original error context is preserved.

Fixes #3265. This differs from closed #2623: it supplies the missing translation
argument rather than changing translation-key selection for every console/RCON
message. No overlapping open fix was found.

The regression checks exact console and Bedrock text, the Java translation key
and argument, and original context across four unknown/unavailable input cases.

Validation: all 79 command library tests pass. The new regression fails on
original code with the exact literal `%s` output, then passes with the fix.
Functional tests use release mode with `pumpkin-data` and `pumpkin-world`
opt-level 0 to avoid unnecessary optimization of generated dependencies.
Command-crate all-target/all-feature Clippy, workspace formatting, and whitespace
checks pass. This is validated at the dispatcher/text-renderer layer; no new
full-server binary was needed for this fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Unknown-command errors now include the entered command token for clearer feedback.
  - Improved unknown-command messaging across execution, parsing, unregistered, unavailable, and disabled command scenarios.
  - Updated localized messages for Java and Bedrock representations.

### DIFF
--- a/crates/pumpkin-command/src/errors/error_types.rs
+++ b/crates/pumpkin-command/src/errors/error_types.rs
@@ -96,7 +96,7 @@ pub const LONG_TOO_HIGH: CommandErrorType<2> = CommandErrorType::new(
     translation::java::ARGUMENT_LONG_BIG,
 );
 
-pub const DISPATCHER_UNKNOWN_COMMAND: CommandErrorType<0> = CommandErrorType::new(
+pub const DISPATCHER_UNKNOWN_COMMAND: CommandErrorType<1> = CommandErrorType::new(
     translation::java::COMMAND_UNKNOWN_COMMAND,
     translation::bedrock::COMMANDS_GENERIC_UNKNOWN,
 );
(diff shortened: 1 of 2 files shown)


RoosterBooster007 (approved):

RoosterBooster007:
Thanks, @JulesB40!

PR #3338, opened 2026-09-08
fix(entity): play per-entity hurt and death sounds with correct sound categories 🤖🤖🤖

IsaiahJaiden:
## Description

Animals damaged by players or mobs currently always play the generic hurt sound ("oof"), and no death sound is ever played. This PR fixes both:

**Root causes**
1. `entity_type.hurt_sound` is `None` for most entities (cow, sheep, pig, chicken, ...), so the runtime falls back to `Sound::EntityGenericHurt` for everything.
2. Vanilla plays the death sound **server-side** in `LivingEntity#hurt` (dying branch calls `makeSound(getDeathSound())`), while entity status 3 (`broadcastEntityEvent`) only triggers the client-side death animation. Pumpkin never played a death sound anywhere.
3. Hurt sounds were broadcast with a hardcoded `SoundCategory::Players`, so even entities that do play sounds are placed in the wrong category (vanilla uses the entity's sound source: monsters -> Hostile, players -> Players, others -> Neutral).

**Changes**
- `assets/entities.json`: added `hurt_sound` + `death_sound` for the entities whose vanilla sounds map 1:1 onto existing `Sound` variants (48 entities with hurt sounds, 46 with death sounds). Entities with variant/size-based sounds (cat, parrot, axolotl, tropical_fish, frog, goat, slime, magma_cube, ender_dragon, ...) intentionally keep the generic fallback for now — that would need per-variant tracked data and can be a follow-up. `cave_spider` inherits `entity.spider.*` and `mooshroom` inherits `entity.cow.*`, matching vanilla.
- Codegen: emit the new `death_sound` field on the generated `EntityType`.
- Runtime (`LivingEntity`): per-entity `hurt_sound()`/`death_sound()` resolution with generic fallbacks; server-side death sound in `on_death` played before the death status broadcast (vanilla order); new `sound_category()` mapping (MONSTER -> Hostile, player -> Players, everything else -> Neutral) used for both hurt and death sounds.

**Verification note**: I wired a real-protocol 26.2 test client to the server and confirmed the previously-sent packets (damage event, hurt animation, sound effect) were already well-formed on the wire — the issue is purely about *which* sound is played and in which category, which this PR fixes.

## Testing

- `cargo fmt --check` clean
- `cargo clippy --workspace --all-targets --all-features` with `RUSTFLAGS=-Dwarnings` clean
- `cargo test -p pumpkin`: 203 passed, 0 failed (includes new tests: death sound resolution + fallback, sound category mapping)
[...]

PR #3342, opened 2026-09-08
fix(entity): break goal reference cycles in slime goals 🤖🤖🤖

IsaiahJaiden:
## Description

The slime goals (`SlimeFloatGoal`, `SlimeAttackGoal`, `SlimeRandomDirectionGoal`, `SlimeKeepOnJumpingGoal`) store `Arc<SlimeEntity>`. Because the entity owns its goal selector, which owns these goals, this forms a `MobEntity → goal selector → goals → Arc<MobEntity>` reference cycle, so slime (and magma cube, which wraps the same entity) never get dropped.

This converts the goals to `Weak<SlimeEntity>` — matching `SlimeMoveControl`, which already used `Weak` in the same file — and the goals no-op or return false gracefully when the upgrade fails. Slime was the only mob holding a strong goal reference; every other mob already follows the `Arc::downgrade` convention.

Found via the CodeRabbit review of the sulfur cube PR (#3329), which fixed the identical pattern there.

## Testing

- `cargo fmt --check` clean
- Behavior unchanged: goals upgrade to the entity each tick; jump/attack/wander logic identical to before

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved slime behavior when the entity is no longer available, preventing related actions from continuing incorrectly.
  - Slimes now stop attack behavior appropriately when they become tired, even if a target remains present.
  - Improved stability of slime movement and jumping behaviors during entity lifecycle changes.

### DIFF
--- a/crates/pumpkin/src/entity/mob/slime.rs
+++ b/crates/pumpkin/src/entity/mob/slime.rs
@@ -70,10 +70,16 @@ impl SlimeEntity {
                 .lock()
                 .unwrap_or_else(std::sync::PoisonError::into_inner);
 
-            goal_selector.add_goal(1, Box::new(SlimeFloatGoal::new(mob_arc.clone())));
-            goal_selector.add_goal(2, Box::new(SlimeAttackGoal::new(mob_arc.clone())));
-            goal_selector.add_goal(3, Box::new(SlimeRandomDirectionGoal::new(mob_arc.clone())));
-            goal_selector.add_goal(5, Box::new(SlimeKeepOnJumpingGoal::new(mob_arc.clone())));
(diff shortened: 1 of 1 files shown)


IsaiahJaiden:
Hi — noticed 8 of my open PRs (#3326, #3329, #3330, #3334, #3336, #3338, #3341, #3342) were closed today without merging or comment. If they were closed due to a policy on agent-authored PRs, a size/rebase concern, or because the approach should change (e.g. pending the pumpkin-command refactor), I'm happy to rework and resubmit — just let me know what direction you'd prefer. All branches are still available on my fork if any of the fixes are useful.

RoosterBooster007:
Many of your PRs were useful, but it would be better if they were more human-written and tested. And maybe open fewer at a time?

PR #3346, opened 2026-09-08
fix(bedrock): restore hunger synchronization and eating feedback

ZlordHUN:
## Description
Fixes Bedrock hunger synchronization and eating feedback, including an incorrect hunger bar and interrupted food consumption.

## What
- Keeps the hunger bar synchronized when joining and eating.
- Restores reliable eating completion and burp sounds.
- Corrects rejected eating at full hunger while preserving always-edible foods.

## How
- Resends health and hunger after the Bedrock client finishes initialization.
- Uses zero for the attribute-update input tick instead of the server tick counter.
- Prevents repeated item-use packets from restarting the eating timer.
- Uses the correct main hand and resynchronizes hunger when eating is rejected.
- Restores the completion burp and adds regression tests through the actual packet handlers.

## Testing
- `cargo test`
- `cargo clippy`
- Verified in game that Bedrock hunger and eating now work correctly.

## Summary by CodeRabbit

- **New Features**
  - Eating food now plays a burp sound after hunger is restored.
  - Player health and hunger values are synchronized when joining or becoming ready.
  - Food-use actions now correctly use the active hand and provide health updates when hunger is full.

- **Bug Fixes**
  - Repeated use inputs no longer restart an ongoing food-use action.
  - Foods that are not edible at full hunger are prevented from being consumed, while always-edible foods remain usable.
  - Rejected food-use corrections are rate-limited to reduce redundant updates.

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -3091,6 +3091,11 @@ impl EntityBase for LivingEntity {
                     player
                         .hunger_manager
                         .eat(player, food.nutrition as u8, food.saturation);
+                    self.entity.world.load().play_bedrock_level_sound(
+                        "burp",
+                        &self.entity.pos.load(),
+                        -1,
+                    );
                 }
 
                 self.apply_consumable_effects(caller, item);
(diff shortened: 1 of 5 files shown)


Q2297045667:
Please compress the code when merging it

ZlordHUN:
Noted for next time.

PR #3350, opened 2026-09-08
fix: advertise current protocol to generic status probes

### LINKED ISSUE #3340: Problems with the proxy server
### Before Submitting

### Minecraft Edition

Both / Crossplay

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.45 (Comm 62a5/release)

### Operating System

linux x64

Velocity PumpkinMC still has many issues and is incomplete. Specifically, my problem is with the Geyser Velocity plugin; when Bedrock players switch to the Pumpkin server, they are kicked out and cannot transfer.I have already enabled all the proxy configurations.

[latest.log](https://github.com/user-attachments/files/31939651/latest.log)

### Steps to Reproduce

1 2 3

JulesB40:
## Description

Fixes #3340.

Generic status probes currently identify Pumpkin as its oldest supported version (1.7.2/protocol 4). Velocity/ViaVersion then select the legacy login UUID decoder for a modern backend connection, disconnecting players; a leading-zero UUID produces the reported empty UUID error.

Advertise the current native protocol when a probe does not name a supported client version. Supported clients still receive their own protocol, and the advertised version-range name remains intact.

## Testing

- Both new status regressions fail before the fix (protocol 4 versus 776); all 203 Pumpkin library tests pass afterward.
- Real Velocity 4.1.0-SNAPSHOT build 20 + ViaVersion 5.11.0: the original server fails in the legacy UUID decoder; the candidate automatically detects and saves `ana: 776`, then completes login and sends teleport/world chunks. The explicitly pinned control also passes.
- Release all-target/all-feature Clippy, formatting and whitespace checks.

The functional validation binary also contains an independent last-block placement-effect fix; it uses unoptimized generated data/world and server code, with release LTO disabled. No performance claim is made. The user previously verified real Bedrock 26.45 login through Geyser/Floodgate with the configuration workaround; this code fix was verified with the protocol observer, not a new graphical Bedrock run.

## Summary by CodeRabbit

- **Bug Fixes**
  - Generic server status probes now receive the current Minecraft protocol version.
  - Supported clients continue to receive their requested protocol version.
  - Cached status responses remain consistent across repeated requests.

### DIFF
--- a/crates/pumpkin/src/server/connection_cache.rs
+++ b/crates/pumpkin/src/server/connection_cache.rs
@@ -204,7 +204,8 @@ impl CachedStatus {
         StatusResponse {
             version: Some(Version {
                 name: format!("{LOWEST_SUPPORTED_MC_VERSION}-{CURRENT_MC_VERSION}"),
-                protocol: LOWEST_SUPPORTED_MC_VERSION.protocol_version() as u32,
+                // Generic proxy probes must detect our native protocol, not the oldest client.
+                protocol: CURRENT_MC_VERSION.protocol_version() as u32,
             }),
             players: Some(Players {
                 max: max_players,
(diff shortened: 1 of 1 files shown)


JulesB40:
The only failing Actions check on this revision is the [Android AArch64 release job](https://github.com/Pumpkin-MC/Pumpkin/actions/runs/34248280941/job/102142369888). Its annotation reports: "The hosted runner lost communication with the server." The build step has no completion result and its log is unavailable, so there is no compiler/test diagnostic identifying a source failure.

Android tests and all other build, test, and lint checks passed. Could a maintainer rerun the failed Android release job? I attempted the targeted rerun, but GitHub returned HTTP 403 (repository admin rights required).

QuantumSegfault:
Multi-version support has been removed, so now `LOWEST_SUPPORTED_MC_VERSION` the same as `CURRENT_MC_VERSION`. Making this PR NFC.

You should double check the issue, but otherwise this fix is no longer necessary.

PR #3355, opened 2026-09-08
feat(plugin,QOL): plugin show information, reload, and tab-auto-completion

zinedinp:
## Description
During plugin devloping/testing i encoutered a tedious flaw, that's why I added some QOL plugin commands:

- Add PluginStatus enum to track states (Active, Loading, Unloaded, Disabled, Unsigned, PermissionDenied, Failed, NoLoader)
- Add PluginEntry struct to unify plugin view across all sources (running, inactive, failed, no-loader)
- Add /plugin list showing all known plugins with color-coded status (green=active, yellow=loading, red=inactive)
- Add /plugin info <name> with full plugin details (version, authors, description, dependencies, permissions)
- Add /plugin reload <name> to unload and reload a plugin (blocks reload if loader cannot unload at runtime)
- Add /plugin reloadall to unload all plugins, prune missing files, and rescan the plugin directory (also picks up new files)
- Add /plugin load <name> and not only load from path.
- Add tab-auto-completion for all plugin subcommands
- Add PermissionRegistry::unregister_namespace() to remove all permissions in a namespace on plugin unload/reload
- Migrate inactive plugins map from async RwLock to sync SyncRwLock to support synchronous tab-completion provider

## Testing

- cargo clipply --release
- gametesting

Servus

## Summary by CodeRabbit

- **New Features**
  - Plugin management now displays installed, active, disabled, failed, unsigned, and unavailable plugin statuses.
  - Added richer plugin listings and details, including metadata, permissions, status indicators, and hover information.
  - Plugin commands now support names, paths, quoted names, status-aware suggestions, and improved load, unload, reload, and information handling.
  - Plugins can be cleanly unloaded and reloaded without retaining their registered permissions.

- **Bug Fixes**
  - Improved handling of failed plugin initialization, unload failures, and duplicate plugins during rescans or hot reloads.


crates/pumpkin/src/plugin/mod.rs
```diff
@@ -180,12 +184,95 @@ pub enum PluginState {
...
+    #[must_use]
+    pub const fn is_active(&self) -> bool {
+        matches!(self, Self::Active)
+    }
+}
+
+impl std::fmt::Display for PluginStatus {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Active => f.write_str("Running"),
```
QuantumSegfault, line 215:
```suggestion
            Self::Active => f.write_str("Active"),
```

For consistency

crates/pumpkin/src/plugin/mod.rs
```diff
@@ -180,12 +184,95 @@ pub enum PluginState {
...
 }
 
+/// State of a plugin file the manager knows about, running or not.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PluginStatus {
+    Active,
+    Loading,
+    /// Unloaded on request
+    Unloaded,
+    /// Turned off in the configuration, or a `.deactivated` file
```
QuantumSegfault, line 194: Not sure there's much value in `.deactivated`...

Rather specific something to remember.

QuantumSegfault (approved):
LGTM. Just a couple nit-picks

But I'll defer to @vyPal here.

crates/pumpkin/src/plugin/mod.rs
zinedinp: I can remove the extension handler if you say so.
My tought was, if someone doesnt wants to configure it and just deactivate constantly during developement, where I often reseted the config, but I dont think its bad to have.
How should I proceed?

PR #3374, opened 2026-09-11
fix(worldgen): record fancy trunk logs to stop leaf decay

### LINKED ISSUE #2445: Block state of leaves doesn't work as expected

### Operating System

Windows 11

f4815f5

### What happened?

The value of block state "distance" should be dynamically generated, but it's actually the default value "7" in the game.
Look at the left part of the screenshot:

Image

### To Reproduce

1. Launch the server and the game
2. Join the server
3. Press F3+F6, Search for the option "looking_at_block"
4. Switch the option to "Overlay"
5. Close this Screen, and press F3
6. Look at the block state "distance"

### Expected behavior

The value should be dynamically chosen by the distance between the leaves and the logs.

matheustavarestrindade:
## Description

Big-oak trees decay their leaves after world generation, with no player nearby. Issue #2445 shows the same cause: the leaf `distance` state stays at 7, and a leaf at distance 7 decays on a random tick.

The bug is in `FancyTrunkPlacer` (`crates/pumpkin-world/src/generation/feature/features/tree/trunk/fancy.rs`). The placer writes the main trunk and the branches to the world, but it does not add those logs to the log list that it returns. A big oak without branches returns an empty list.

`TreeFeature::update_leaves` starts its distance search from this log list. An empty list gives no start point, so the search does not run and every leaf keeps distance 7.

The fix captures the logs from the main trunk and from `make_branches`. The list now holds every placed log, and the leaf search runs as intended.

Fixes #2445.

## Testing

- `cargo test -p pumpkin-world --lib`: 187 passed.
- `RUSTFLAGS=-Dwarnings cargo clippy -p pumpkin-world --all-targets --all-features`: no warnings.
- A local diagnostic generated 64 chunks with seed 42 and counted leaves with distance 7. The count went from 676 to 1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected fancy tree trunk generation so log positions from both the main trunk and branches are consistently tracked and returned.
  - Improved the reliability of log placement data for complex tree trunks, supporting more consistent generated trees.


RoosterBooster007 (approved):

RoosterBooster007:
Looks good. Thanks, @matheustavarestrindade!

PR #3398, opened 2026-09-11
fix: Pterodactyl egg port sync and several gameplay bugs

Laggy60:
### Pterodactyl egg: allocation port never updated after the first boot

The egg's `file` config parser matches a **line prefix** and replaces the **whole line**. The old find key `address = "0.0.0.0:25565"` therefore only matched on the first boot. After that, the line holds the allocated port, so later allocation changes were never applied. A looser prefix (`address = `) also matches the RCON, Bedrock and query `address` lines. Pterodactyl has no TOML parser, and the yolks entrypoint runs `exec env ${PARSED}` (no shell), so the startup command can't rewrite the file either.

- New `PUMPKIN_JAVA_ADDRESS` env var that overrides `networking.java.address` at startup. It is applied after the logger is initialised and never written back to `pumpkin.toml`; invalid values are logged and ignored.
- Egg startup: `PUMPKIN_JAVA_ADDRESS=0.0.0.0:{{SERVER_PORT}} ./pumpkin`.
- Egg install:
  - `VERSION=latest` resolves to `releases/latest` (it pointed at nightly).
  - Downloads to a temp file with `--retry 3`, so a failed download never replaces a working binary.
  - Seeds a minimal `pumpkin.toml` instead of running the server for 2s; drops unused `jq`.
- The existing `file` parser entry is kept, so older binaries without the override still get the right port on the first boot.

### Gameplay fixes

- **Huge mushrooms never grew from bone meal.** The space check in `grow_mushroom` tested the mushroom's own position, which is only replaced with air after the check. Vanilla removes the block first.
- **Brewing stand slots.** Hoppers use `is_valid_slot_for`, which accepted any non-fuel item in the ingredient slot (a hopper could jam it with cobblestone) and rejected glass bottles in the potion slots. It now follows the brewing screen's rules: real ingredients only, and potion items one per slot. Fuel stays out of the ingredient slot because hoppers don't have vanilla's per-side slots yet (`TODO check WorldlyContainer`).
- **Brewing stand chained brews.** `do_brew` started the next brew itself, which skipped `BrewingStartEvent`. Like vanilla, the timer now ends at 0 and `tick` starts the next brew.
- **Hopper stacking.** `add_one_item` merged stacks by item type only (`TODO check Components equal`), so renamed or enchanted items were merged into plain ones and lost their components. It now uses `are_items_and_components_equal`, the same check the extraction path got in #3147.
- **Armor stands.** A named armor stand dropped a plain item. And with `mobGriefing=false`, a player's projectile couldn't break one, because the check looked at the direct source (the arrow). Vanilla only blocks damage when the causing entity is a `Mob`.

## Testing

- `cargo test -p pumpkin -p pumpkin-config -p pumpkin-inventory`: 210 + 16 + 21 passed, 0 failed.
- `cargo test --workspace --no-fail-fast`: 781 passed, 0 failed, 20 ignored.
- Egg JSON validated; the install script round-trips exactly and passes `sh -n`.
- Checked the Pterodactyl wings `file` parser (prefix match, full-line replace) and the yolks entrypoint against upstream source.
- After rebasing onto current `master` (no conflicts), re-ran on the rebased code: clippy clean; `cargo test -p pumpkin -p pumpkin-config -p pumpkin-inventory`: 273 passed, 0 failed.

### DIFF
--- a/crates/pumpkin-config/src/lib.rs
+++ b/crates/pumpkin-config/src/lib.rs
@@ -141,6 +141,35 @@ impl LoadConfiguration for PumpkinConfig {
     }
 }
 
+/// Environment variable that overrides `networking.java.address` at startup.
+///
+/// Hosting panels such as Pterodactyl assign a port per server; this lets them pass
+/// the bind address on every start without having to rewrite `pumpkin.toml`.
+pub const JAVA_ADDRESS_ENV: &str = "PUMPKIN_JAVA_ADDRESS";
+
+impl PumpkinConfig {
+    /// Applies overrides from environment variables on top of the loaded configuration.
+    ///
+    /// Overrides only affect the running server and are never written back to `pumpkin.toml`.
+    pub fn apply_env_overrides(&mut self) {
+        let Ok(value) = std::env::var(JAVA_ADDRESS_ENV) else {
+            return;
+        };
+        let value = value.trim();
+        if value.is_empty() {
+            return;
+        }
+        let address: std::net::SocketAddr = match value.parse() {
+            Ok(address) => address,
+            Err(err) => {
+                warn!("Ignoring {JAVA_ADDRESS_ENV}={value:?}: {err}");
+                return;
+            }
+        };
+        self.advanced.networking.java.address = address;
+    }
+}
+
 /// Advanced configuration for optional and feature-specific server settings.
 ///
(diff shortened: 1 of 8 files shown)


RoosterBooster007:
Thanks for these fixes, but they're unrelated, so please split them into their own PRs for review.

Issue #3390, opened 2026-09-11
Fire aspect still doesnt drop cooked food


darwincereska:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.45

### Operating System

Gentoo Linux

### Bug Description & Expected Behavior

As brought up in this issue #2366 cooked food doesnt drop when using fire aspect and as of this version it still does not work

Mobs I tested on:
- Cow
- Hoglin

### Steps to Reproduce

1. Kill mob with fire aspect weapon

Issue #3377, opened 2026-09-11
Bottom dripstone does not deal damage


darwincereska:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.45

### Operating System

Gentoo Linux

### Bug Description & Expected Behavior

When landing on a stalagmite (one on the floor), it does not deal damage to the player.

https://github.com/user-attachments/assets/0d8173c2-e427-45eb-8afb-a8407898dd40

### Steps to Reproduce

1. Fall on stalagmite
2. No damage happens

Issue #3240, opened 2026-09-04
Incorrect music disc drops


Sqrilizz:
### Before Submitting

- I have searched existing issues and confirmed this is not a duplicate

### Minecraft Edition

Both / Crossplay

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.45 (Commit: 4c4cf06/release)

### Operating System

Linux

### Bug Description & Expected Behavior

Music discs can drop from mobs under incorrect conditions.

A zombie dropped a music disc after being killed normally by the player. The same behavior also happend with a creeper killed directly by the player.

Zombies should not drop music discs as normal loot. Creepers should only drop a music disc when killed by a skeleton or stray.

https://www.youtube.com/watch?v=FrDkyNKfKFE

timestamp ```00:01:50```

### Steps to Reproduce

1.Spawn or find a zombie.
2.Kill it normally as a player.
3.Check the dropped loot.
4.Repeat the test with a creeper and kill it directly as a player.

PR #3399, opened 2026-09-11
fix: loot table conditions, creative drops, stalagmites and falling blocks

### LINKED ISSUE #3293: Composter gives you free bonemeal every time you break it

Laggy60:
## Description

Fixes #3293, #3292, #3120, #3377, #3240, #3390, #3153, and the drop part of #3109.

### Loot tables ignored `block_state_property` and other conditions

`tools/pumpkin-codegen/src/loot_table.rs` turned every condition it didn't recognise into `LootCondition::None`, which means "always true". 155 of the 1356 vanilla loot tables use `minecraft:block_state_property` (259 times), so all of those conditions were dropped. Three more flattening problems made it worse:

- **`alternatives` with non-tool conditions.** Every child became eligible at the same time and the pool then picked one at random, e.g. mature wheat gave wheat *or* seeds 50/50.
- **Conditional `set_count` functions.** Their conditions were ignored. Slabs say "count 2 only if `type=double`", so **every broken single slab dropped 2 slabs**.
- **Pool-level `functions`.** These were never read, e.g. the sweet berry bush's `set_count` and Fortune bonus.

Visible results in game:

- Composters dropped bone meal every time they were broken (#3293).
- Sweet berry bushes dropped berries at any age, with the wrong counts (#3292).
- Snow layers dropped a random mix of snowballs, snow blocks and layers (the drop part of #3109).
- Slabs duplicated (1 → 2); both halves of doors, beds and tall plants were eligible to drop; candles and sea pickles ignored their count; crops dropped their produce while still growing.

Changes:

- `pumpkin-util`: new `LootCondition::BlockStateProperty { properties }`, `Inverted(&Self)` and `AnyOf(&[Self])`.
- `world/loot.rs`: evaluates them. `BlockStateProperty` checks the broken block's state, which block breaking and explosions already pass in `LootContextParameters::block_state`.
- Codegen:
  - Parses `block_state_property` (exact values; ranges stay unconditional as before), generic `inverted` and generic `any_of`.
  - `alternatives` now follows vanilla's "first matching child wins": a later child gets the negation of every earlier **deterministic, fully parsed** condition. Random conditions (table bonus, random chance) aren't negated, since that would re-roll them. Conditions the codegen doesn't understand yet (e.g. sheep colour or chicken variant `entity_properties`) aren't negated either, so those tables stay as before; so gravel/flint and leaves/sapling generate exactly the same conditions as before.
  - A `set_count` with deterministic conditions is split into one entry per possible count (slabs, candles, sea pickles, …).
  - Pool-level functions apply to the pool's entries.

### Entity loot: creeper music discs and cooked drops

`entity_properties` conditions were treated as "always true" too, so **every creeper dropped a music disc**, whoever killed it (#3240). `furnace_smelt` was ignored, so mobs killed while burning or with Fire Aspect dropped raw food (#3390).

- New `ThisOnFire`, `KillerType { types }` and `ToolEnchanted { enchantments }` conditions, evaluated from `is_on_fire`, `killer_entity` and the attacker's weapon (`tool`), which the death loot context already fills in.
- The codegen parses the three `entity_properties` shapes vanilla uses for these: `this` + `flags.is_on_fire`, `attacker` + `entity_type`, and `direct_attacker` + main-hand enchantments. Entity type and enchantment tags are expanded. Every other shape stays unconditional, as before.
- `furnace_smelt` emits a cooked and a raw variant, using the vanilla smelting recipes. Smelting is skipped if any of its conditions can't be parsed, so nothing gets cooked by accident.

`loot_table.rs` is regenerated with `cargo run -p pumpkin-codegen -- loot_table`.

### Loot ignored `limit_count` and `explosion_decay`

The codegen also ignored `limit_count` and `explosion_decay`, and never read table-level functions, which is where vanilla usually puts `explosion_decay`. Two effects:

- **Fortune exceeded vanilla's caps:** up to 7 glowstone dust instead of 4, and 10 melon slices instead of 9.
- **Explosions dropped full counts:** blocks destroyed by an explosion dropped everything, instead of each item surviving with a `1 / radius` chance.

The fix:

- `LootEntry` gets `limit` and `explosion_decay`, applied after bonuses.
- Pool- and table-level functions now apply to every entry.
- Conditional ones are left out, as before.

### Blocks that lose support didn't drop after a creative break (#3120)

A creative break passes `SKIP_DROPS`, and `replace_with_state_for_neighbor_update` forwarded the same flags when breaking a torch, lever or button that lost its support. Vanilla clears `UPDATE_SUPPRESS_DROPS` before neighbour shape updates (`Level.setBlock`, `flags & -34`), so those blocks drop. The flag is now removed there. Door, bed and tall-plant halves don't dupe, because the loot tables above now only drop the lower half or the head.

### Landing on a stalagmite did no extra damage (#3377)

Pointed dripstone had no `on_landed_upon`. Like vanilla's `PointedDripstoneBlock.fallOn`, landing on an upward tip now calls `causeFallDamage(fallDistance + 2.5, 2.0, stalagmite)`. `OnLandedUponArgs` gets the landing `position`, and `LivingEntity::handle_fall_damage_from` takes the damage type (`handle_fall_damage` keeps using `FALL`).

### Falling blocks replaced partial blocks (#3153)

A landing falling block always replaced the block at its position. Like vanilla's `FallingBlockEntity`, it now only lands where the current block is replaceable and the falling block can survive, and otherwise drops as an item (when `doEntityDrops` is on).

## Testing

- `cargo test -p pumpkin -p pumpkin-util`: 307 passed, 0 failed.

- New loot tests in `world/loot.rs`, each over 16 seeds:
  - composter only drops bone meal at level 8;
  - sweet berry bush drops nothing at age 0 and 2–3 berries at age 3;
  - single slabs drop 1, double slabs 2;
  - wheat only drops wheat at age 7;
  - creepers only drop music discs when a skeleton killed them;
  - cows drop cooked beef only when burning;
  - glowstone with Fortune III never drops more than 4 dust;
  - blocks destroyed by a huge explosion drop (almost) nothing.
- Regenerated `loot_table.rs` compared with `master`: leaving aside the two fields every entry gains in the last commit, the entries of 168 of 1312 tables change, namely 151 block tables, 16 entity tables that gain cooked-drop variants and the sweet berry harvest table. The creeper change is in its music-disc pool condition. Oak leaves, gravel, diamond ore and the abandoned mineshaft chest are byte-for-byte identical, as are tables whose conditions the codegen still doesn't understand (charged creeper heads, chicken eggs, sheep and mooshroom shearing).

Built on Windows with `x86_64-pc-windows-gnullvm` + llvm-mingw.

## Summary by CodeRabbit

* **New Features**
  * Expanded loot-table support for block states, entity conditions, enchanted tools, item limits, and explosion decay.
  * Added more accurate loot generation for smelting, count variations, alternatives, tags, and nested tables.
  * Falling blocks now drop as items when they cannot be placed safely.

* **Bug Fixes**
  * Pointed dripstone now damages living entities that land on upward-facing tips.
  * Blocks losing support now correctly drop items when appropriate.
  * Loot drops now better respect vanilla limits and explosion survival behavior.

### DIFF
--- a/crates/pumpkin-data/src/generated/loot_table.rs
+++ b/crates/pumpkin-data/src/generated/loot_table.rs
@@ -711,7 +711,9 @@ static BLOCKS_ACACIA_DOOR_POOL0_ENTRIES: &[LootEntry] = &[LootEntry {
     weight: 1i32,
     min_count: 1i32,
     max_count: 1i32,
-    condition: LootCondition::None,
+    condition: LootCondition::BlockStateProperty {
+        properties: &[("half", "lower")],
+    },
     bonus_formula: None,
 }];
 static BLOCKS_ACACIA_DOOR_POOLS: &[LootPool] = &[LootPool {
@@ -941,14 +943,28 @@ static BLOCKS_ACACIA_SIGN_POOLS: &[LootPool] = &[LootPool {
 pub static BLOCKS_ACACIA_SIGN: LootTable = LootTable {
     pools: BLOCKS_ACACIA_SIGN_POOLS,
 };
-static BLOCKS_ACACIA_SLAB_POOL0_ENTRIES: &[LootEntry] = &[LootEntry {
-    item: "minecraft:acacia_slab",
-    weight: 1i32,
-    min_count: 2i32,
-    max_count: 2i32,
-    condition: LootCondition::None,
-    bonus_formula: None,
-}];
+static BLOCKS_ACACIA_SLAB_POOL0_ENTRIES: &[LootEntry] = &[
+    LootEntry {
+        item: "minecraft:acacia_slab",
+        weight: 1i32,
+        min_count: 2i32,
+        max_count: 2i32,
+        condition: LootCondition::BlockStateProperty {
+            properties: &[("type", "double")],
+        },
+        bonus_formula: None,
+    },
+    LootEntry {
+        item: "minecraft:acacia_slab",
+        weight: 1i32,
+        min_count: 1i32,
+        max_count: 1i32,
+        condition: LootCondition::Inverted(&LootCondition::BlockStateProperty {
+            properties: &[("type", "double")],
+        }),
+        bonus_formula: None,
+    },
+];
 static BLOCKS_ACACIA_SLAB_POOLS: &[LootPool] = &[LootPool {
     entries: BLOCKS_ACACIA_SLAB_POOL0_ENTRIES,
     min_rolls: 1i32,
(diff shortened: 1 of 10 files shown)


RoosterBooster007:
Please split this up into smaller PRs for each fix.

PR #3400, opened 2026-09-11
fix(effect): apply instant effects once and never show them as active

### LINKED ISSUE #3381: Dragon's breath from ender dragon effect stays after the duration ends
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

0.1.0-dev+26.2-26.45

### Operating System

Gentoo Linux

The dragons breath effect (it shows up as instant damage right now) gets applied to the player and then does not go away even after the duration ends 

### Steps to Reproduce

1. Get with with dragon's breath

Laggy60:
## Description

Fixes #3381.

Instant effects (instant damage and instant health) had two problems:

- **The effect never went away on the client.** `LivingEntity::add_effect` applies an instant effect immediately and, correctly, doesn't store it in `active_effects`. It still broadcast the "add mob effect" packet, though. Since the effect was never stored, the server never sent a removal, so the client kept showing it forever, e.g. after standing in dragon's breath.
- **Potions applied instant effects twice.** `PotionContents::apply_effects_to` (splash and lingering potions, area effect clouds, drinking) applied the instant damage or heal itself and then called `add_effect` for a "short visual effect entry", which applied it again.

Vanilla never adds instant effects to the active effects: `PotionContents.applyToLivingEntity` calls `applyInstantenousEffect` for them instead of `addEffect`. Now:

- `apply_effects_to` no longer adds an entry for instant effects.
- `add_effect` returns right after applying an instant effect, so `/effect` and plugins don't leave a stuck effect either.

## Testing

- `cargo test -p pumpkin`: 235 passed, 0 failed.

Built on Windows with `x86_64-pc-windows-gnullvm` + llvm-mingw.

## Summary by CodeRabbit

* **Bug Fixes**
  * Instant health and damage effects are now applied immediately without being broadcast or retained as active effects.
  * Instant potion effects now scale correctly with their amplifier.
  * Ongoing effects, including attribute modifiers, absorption, invisibility, and glowing, continue to be handled correctly.

### DIFF
--- a/crates/pumpkin/src/entity/living.rs
+++ b/crates/pumpkin/src/entity/living.rs
@@ -848,6 +848,8 @@ impl LivingEntity {
         if effect.effect_type == &StatusEffect::INSTANT_HEALTH {
             let heal_amount = 4.0 * (1 << effect.amplifier) as f32;
             self.heal(heal_amount);
+            // Like vanilla, instant effects are never sent or stored as active effects.
+            return;
         } else if effect.effect_type == &StatusEffect::INSTANT_DAMAGE {
             let damage_amount = 6.0 * (1 << effect.amplifier) as f32;
             let dyn_self = self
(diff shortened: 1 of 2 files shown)


Laggy60:
Thanks - fixed, the direct path now uses the same `<< amplifier` scaling as `LivingEntity::add_effect`.

RoosterBooster007:
Looks good. Thanks, @Laggy60!

PR #3152, opened 2026-08-30
fix(tnt, client synchronization): entity tracker registration and explosion knockback

zinedinp:
**Summary**
Fixed three interconnected bugs:

- chunk-loaded entities not registering with tracker (freezing mobs)
- dimension changes leaving stale player pairings (invisible entities)
- TNT knockback not visible client-side (missing periodic resync).

_Changes_

- **world/mod.rs**: Route chunk-loaded entities through entity tracker; fix raycast DDA gate; remove unused `broadcast_entity_spawn()`.
- **entity_tracker.rs**: Add `drop_player_pairings()` to clear stale UUIDs on dimension leave; add `try_restore_vehicle()` in `add_pairing`.
- **entity/player.rs**: Call `drop_player_pairings()` when unloading chunks.
- **entity/tnt.rs**: Reorder `send_meta_data` before velocity check; use periodic `send_tracked_position()` for client resync.
- **entity/mod.rs**: Add `send_tracked_position()` method for periodic position/velocity updates.
- **world/explosion.rs, block/blocks/tnt.rs, pumpkin-util/**: Supporting explosion and raycast fixes.

_Results_

- TNT now damages/knocks back nearby entities
- Mobs despawn correctly instead of becoming ghosts
- Dimension travel keeps entities visible
- Client-side TNT physics stays synced with server

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TNT ignition, fuse handling, explosion damage, knockback, and power persistence.
  * Corrected entity tracking during movement, world changes, vehicle restoration, and unloading.
  * Reduced redundant movement updates and improved position, velocity, rotation, and metadata synchronization.
  * Corrected raycasts that begin in air and collision movement calculations.
  * Improved rail protection behavior to match vanilla gameplay.

* **Improvements**
  * Explosion updates now provide affected players with accurate knockback information.
  * TNT movement applies more consistent drag and impulse behavior.
  * Added synchronization for global world events.

PR #3401, opened 2026-09-11
fix(entity resync): vanilla sendChanges resync through the entity tracker

zinedinp:
## Premise
This is part 2 of the seperation of https://github.com/Pumpkin-MC/Pumpkin/pull/3152 

## Description
Entity position and velocity sync moves from each entity's tick into `TrackedEntity::send_changes`, which runs after all entities tick like vanilla `ServerEntity.sendChanges`. TNT physics in later changes, so there is a client prediction missmatch.

- Position, velocity and metadata are sent at vanilla cadence and with vanilla encoding.
- Per-entity sync code and duplicate syncs are removed.
- Bedrock gets per-tick movement updates with the correct per-type height offset.
- Falling blocks no longer briefly disappear on landing.
- TODO: Bedrock spawn metadata for non-mob entities.

## Summary by CodeRabbit

- **New Features**
  - Improved Bedrock player spawning, including appearance, equipment, abilities, and metadata.
  - Improved entity visibility and tracking during chunk loading, movement, respawning, and world transfers.
  - Added spawn protection for player-caused block breaking.
  - Minecarts now properly activate detector rails.

- **Bug Fixes**
  - Improved block updates when falling blocks land.
  - Improved entity synchronization and positioning across Java and Bedrock clients.
  - Improved chunk delivery reliability and reduced stale entity visibility after respawning or teleporting.
  - Improved player positioning and rotation updates across supported clients.


zinedinp:
I found out, after digging deeper, that the dragon fight is broken hard, after coderabbit asked too. The dragon can despawn and the boss bar never faids, and the dragon dies too fast rather a die and open the portal. I guess it is a regression from one the larger rewrites in the last time.

Follow up PR for the dragon will come.

zinedinp:
I encoutered rather interesting bugs when debugging bedrock after coderabbit suggested to skip non bedrock watcher. The Bedrock sends now skip entities without Bedrock watchers, which only works if everything a Bedrock client sees goes through the tracker's "seen_by".

That wasn't true for players:
The join paths spawned players directly with "AddPlayer", bypassing "seen_by".
The tracker's own player spawn used "AddActor", which Bedrock doesn't render. Java players became invisible after a teleport or relog, and skins went missing because of the duplicate spawns.

## Further Changes:

- Tracker as the only spawn path: Player overrides send_bedrock_spawn_packet ("PlayerList" + "AddPlayer" + held item). The join paths only send the tab list, and Bedrock joins pair through "pair_new_player_with_tracked_entities".
- Respawn refresh only goes to "seen_by".
- Name tags: "ALWAYS_SHOW_NAME_TAG" is set now
- all-air sub-chunks above are no longer sent, which fixes the dark sky light in new chunks when chunk_caching = false

Helpers in "player.rs": 
- the packets are built from player state only, and the tracker only sees "dyn EntityBase", so it reaches them through the Player override. They replace **5 copies in world/mod.rs**.

crates/pumpkin/src/entity/mod.rs
```diff
@@ -2670,9 +2368,17 @@ impl Entity {
...
+            (self.pos.load(), self.velocity.load())
+        } else {
+            (self.last_sent_pos.load(), self.last_sent_velocity.load())
+        }
+    }
+
     pub fn create_spawn_packet(&self) -> CSpawnEntity {
-        let entity_loc = self.pos.load();
-        let entity_vel = self.velocity.load();
+        let (entity_loc, entity_vel) = self.spawn_pos_vel();
```
RoosterBooster007: One nit: the rotation arguments a few lines down are still reading live values, whereas vanilla uses `getLastSentXRot/YRot` and `getLastSentYHeadRot` unpacked from the cache. Because `send_changes` diffs rotations against that very same `last_sent_*` cache, reading live here can cause a desync in which the client ends up with a rotation the server assumes it already received if the entity snaps back before the next sync tick.

Might be worth bundling rotation into that same `spawn_pos_vel` helper (keeping it live only for players, since the tracker skips `send_changes` for them).

RoosterBooster007 (approved):

crates/pumpkin/src/net/bedrock/level_chunk.rs
```diff
@@ -76,24 +76,27 @@ impl CLevelChunk<'_> {
...
-        cache_enabled.write(&mut writer)?;
-
-        let mut blobs = Vec::new();
-
         let block_sections = chunk
             .section
             .block_sections
             .read()
             .map_err(|_| Error::other("block_sections read lock poisoned"))?;
+        // all-air sub-chunks above the highest non-air one are not sent.
```
RoosterBooster007: Also, this one feels like it belongs in its own PR. Trimming trailing all-air sub-chunks changes what the Bedrock client is told about the chunk, and it isn't something the resync needs, so it'll get better testing isolated (it'd also flip an existing test's expectation, which is a good sign it deserves its own review). If you did hit a Bedrock bug while working on this that it fixes, lmk and you can keep it here.

zinedinp: This is, because the bedrock client showed false skylight shadows while changeing the code, and this seemed to fix it. I make it its own PR, because the error fixes itdelf on bedrock when moving.

crates/pumpkin/src/entity/mod.rs
zinedinp: i made a helper for enteties and live enteties, but live enteties get a live update. I added some TODOs where this helper can be implemente. this would save roundtrips and cleaner code. This also seems to fix the error where player dont have a head yaw on join.

zinedinp:
https://github.com/user-attachments/assets/d6313ddb-8a14-47eb-a595-ef48929e0edb

sand now falls smoothly and more things are now smooth and synced

ZlordHUN:
I'll review this one tomorrow

ZlordHUN (approved):

ZlordHUN:
LGTM

kefu84510-cmd (commented):
Hey

Issue #3250, opened 2026-09-05
Leaves drop saplings when broken too often


hydroko13:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0-dev+26.2-26.45 (Commit: 8d0d0d3/release) - 287 Contributors

### Operating System

Pumpkin running on raspberry pi 4 8gb ram

### Bug Description & Expected Behavior

When I break the leaves on a tree, it drops saplings for EVERY SINGLE leaf block i break

Expected Behavior:
Saplings should be rare drop when you break leaf block

### Steps to Reproduce

1. Find a tree
2. Break the leaves (not the log)
3. You will see that you get TONS of saplings (more than vanilla)

PR #3418, opened 2026-09-12
fix(loot): honour block state conditions in generated loot tables

### LINKED ISSUE #3293: Composter gives you free bonemeal every time you break it

Laggy60:
Split out of #3399 as requested. This is the base fix; the entity-loot and `limit_count`/`explosion_decay` parts follow in separate PRs once this one is in, because they touch the same generated file.

Fixes #3293, #3292, #3250 and the drop half of #3109.

`tools/pumpkin-codegen/src/loot_table.rs` turned every condition it didn't recognise into `LootCondition::None`, which means "always true". 155 of the 1356 vanilla tables use `minecraft:block_state_property` (259 times), so all of those were dropped. Three flattening problems made it worse:

- **`alternatives` with non-tool conditions.** Every child became eligible at once and the pool picked one at random, so mature wheat gave wheat *or* seeds, 50/50.
- **Conditional `set_count`.** The conditions were ignored. Slabs say "count 2 only when `type=double`", so every single slab dropped 2.
- **Pool-level `functions`** were never read, e.g. the sweet berry bush's `set_count`.

In game: composters dropped bone meal every time they were broken, sweet berry bushes dropped berries at any age, snow layers dropped a random mix of snowballs, blocks and layers, slabs duplicated, both halves of doors and beds could drop, and crops dropped their produce while still growing.

Changes:

- `pumpkin-util`: `LootCondition::BlockStateProperty { properties }`, `Inverted(&Self)` and `AnyOf(&[Self])`.
- `world/loot.rs` evaluates them; `BlockStateProperty` checks the broken block's state, which block breaking and explosions already pass in `LootContextParameters::block_state`.
- Codegen parses `block_state_property` (exact values; ranges stay unconditional as before), generic `inverted` and generic `any_of`.
- `alternatives` follows vanilla's "first match wins": a later child gets the negation of every earlier condition that is both deterministic and fully parsed. Random conditions (table bonus, random chance) aren't negated, since that would re-roll them, and conditions the codegen doesn't understand yet aren't either, so those tables stay exactly as they were.
- A `set_count` with deterministic conditions is split into one entry per possible count.
- Pool-level functions apply to the pool's entries.
- The table keys are now built with `/` on every OS. They were made from file paths, so generating on Windows produced keys like `minecraft:blocks\composter` and every lookup failed.

Testing: `cargo clippy --workspace --all-targets` and `cargo fmt --all --check` are clean, `cargo test -p pumpkin -p pumpkin-util` passes. New tests in `world/loot.rs` run each case over 16 seeds: composter only drops bone meal at level 8, sweet berry bush drops nothing at age 0 and 2-3 berries at age 3, single slabs drop 1 and double slabs 2, wheat only drops wheat at age 7.

Comparing the regenerated `loot_table.rs` with `master`: 152 of 1312 tables change, 151 block tables plus the sweet berry harvest table. Oak leaves, gravel, diamond ore and the abandoned mineshaft chest come out byte-for-byte identical, so the tables that were already right are untouched.

## Summary by CodeRabbit

* **New Features**
  * Loot tables now support block-state-based conditions, including properties such as crop growth, slab type, and composter level.
  * Added support for inverted and “any of” condition logic in loot tables.
  * Loot outcomes now better reflect conditional item counts, alternatives, and pool-level effects.
* **Bug Fixes**
  * Corrected loot drops for state-dependent blocks, including composters, sweet berry bushes, slabs, and wheat.
  * Improved handling of generated loot-table conditions and item functions.

### DIFF
--- a/crates/pumpkin-data/src/generated/loot_table.rs
+++ b/crates/pumpkin-data/src/generated/loot_table.rs
@@ -711,7 +711,9 @@ static BLOCKS_ACACIA_DOOR_POOL0_ENTRIES: &[LootEntry] = &[LootEntry {
     weight: 1i32,
     min_count: 1i32,
     max_count: 1i32,
-    condition: LootCondition::None,
+    condition: LootCondition::BlockStateProperty {
+        properties: &[("half", "lower")],
+    },
     bonus_formula: None,
 }];
 static BLOCKS_ACACIA_DOOR_POOLS: &[LootPool] = &[LootPool {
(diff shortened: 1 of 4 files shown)


Laggy60:
Thanks for the review. I checked all three codegen findings; they're real gaps, but they're all pre-existing rather than introduced here, and since this PR was already split out on request I'd rather keep them as follow-ups than grow it again.

**Conditional `set_count` fallback.** When a `set_count` carries a random or not-yet-parsed condition, the generator keeps the old behaviour and applies the first `set_count` unconditionally. That's what `master` does for every `set_count` today, so nothing regresses; the cases this PR does split (slabs, candles, sea pickles) all have deterministic conditions. Getting the remaining ones exact means evaluating functions at runtime instead of baking counts in, which is a bigger change.

**First-match semantics with random alternatives.** Deliberate. Negating a random condition (`table_bonus`, `random_chance`) would roll it a second time, so gravel/flint and leaves/sapling would change probability. Those tables therefore generate exactly the conditions they do on `master`. Doing it properly needs ordered alternatives at runtime, or rolling each random condition once and reusing the result.

**Functions on tag entries and nested tables.** Correct: `minecraft:tag` pushes its expanded items without the function pipeline, and `minecraft:loot_table` recurses with only the inner pool's functions. Also true on `master` today.

The `direct_attacker` / `ToolEnchanted` point is about the entity-loot part, which isn't in this PR - that's in a follow-up branch, and I'll carry the selector through there.

Happy to do any of these as their own PR after this one lands.

PR #3436, opened 2026-09-13
Fix item stacks in entity metadata for older clients 🤖🤖🤖

### LINKED ISSUE #3317: Dropped item entities render as the wrong item (double item-ID remap in entity metadata)

**Minecraft Edition:** Java Edition

**Server Version / Commit:** 0.1.0-dev+26.2-26.45 / 55762a50d (master, release binary)

**Operating System:** Linux (Fedora 44)

**Bug Description & Expected Behavior:**
When an item is dropped in the world (ItemEntity), it renders to the client as a completely different, unrelated item (e.g. a log and a door both render as unrelated block-like sprites). Picking the item back up restores the correct item in the inventory, so the underlying item data (ID, count, NBT) is not corrupted — this is purely a client-display bug in how dropped item entities sync their appearance. Expected: the dropped entity renders as the actual item, matching the inventory icon.

**Steps to Reproduce:**
1. Start the server and join with a Java client.
2. Drop any item from the inventory (Q) — e.g. an oak log or a door.
3. Observe the dropped entity sprite: it shows an unrelated item/block texture instead.
4. Pick it back up: the inventory shows the correct item again.

**Suspected cause (from code inspection, not yet fixed):**
`Metadata::write` in `crates/pumpkin-protocol/src/java/client/play/entity_metadata.rs` has an `ITEM_STACK` special case that re-decodes the already-serialized bytes and applies `remap_item_id_for_version` a second time — but `ItemStackSerializer::write_with_version` already remapped the ID. The sent ID is therefore `remap(remap(id))`, which is not idempotent for older protocol versions (only V_26_2 is identity), so the client resolves an unrelated item. Inventory slot packets use `write_with_version` directly (single remap) and render correctly, and server-side data is untouched (pickup restores it). The adjacent `BLOCK_STATE` special case in the same function has the same double-remap shape and may need the same treatment.

potatosips:
Item stack serialization already remaps item IDs for the target protocol version. Entity metadata was decoding that serialized stack and remapping the ID a second time, so dropped items could render as the wrong item on older clients.

This removes the duplicate metadata-level remap and adds a 1.21.4 regression test that verifies the encoded item ID is remapped exactly once.

Validated with the focused protocol test, cargo check -p pumpkin-protocol, formatting, and git diff --check.

Fixes #3317

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected item stack metadata serialization for version 1.21.4.
  - Prevented item identifiers from being remapped more than once during encoding.

- **Tests**
  - Added coverage to verify correct item stack metadata encoding for version 1.21.4.


MidnightDoes:
Hey, this was already fixed in: https://github.com/Pumpkin-MC/Pumpkin/pull/3055, which has already been pushed into the main (may want to sync with upstream)

PR #3438, opened 2026-09-13
fix(worldgen): gate freeze-top-layer snow on face support and light

### LINKED ISSUE #3404: Complete snow placement validation in FreezeTopLayerFeature
## Summary

Complete the snow placement validation in `FreezeTopLayerFeature`.

The current snow check in `crates/pumpkin-world/src/generation/feature/features/freeze_top_layer.rs` verifies that the target block is air, that the block below is not air, and that the block below does not have `MINECRAFT_CANNOT_SUPPORT_SNOW_LAYER`. This does not cover all required placement conditions.

## Required changes

Update the snow placement logic to account for:

- Whether the supporting block face is full (`isFaceFull`).
- The `SUPPORT_OVERRIDE_SNOW_LAYER` tag.
- A brightness gate that permits placement only when brightness is below 10.

## Rationale

The current non-air support check can place snow on blocks that do not meet vanilla snow-layer support rules.

## Affected area

- `crates/pumpkin-world/src/generation/feature/features/freeze_top_layer.rs`
- Supporting block-state, tag, face-shape, and brightness APIs as needed.

## Acceptance criteria

- Snow placement uses the required support-face and tag conditions.
- Snow placement applies the `brightness < 10` condition.
- The behavior matches vanilla rules for blocks with non-full faces and for blocks covered by `SUPPORT_OVERRIDE_SNOW_LAYER`.
- Existing water-freezing behavior remains unchanged.

## Backlinks

- Pull request: https://github.com/Pumpkin-MC/Pumpkin/pull/3183
- Review comment: https://github.com/Pumpkin-MC/Pumpkin/pull/3183#discussion_r3984290959
- Requested by: @RoosterBooster007

Kurenaiiiii:
## What was changed?

In `crates/pumpkin-world/src/generation/feature/features/freeze_top_layer.rs`, the snow-placement condition gained three vanilla requirements on top of the existing air/non-air/`CANNOT_SUPPORT_SNOW_LAYER` checks (which are unchanged, as is the `snowy` property update):

- a face-full check (`is_top_face_full`, duplicating `snow.rs`'s `can_place_at` collision-shape logic due to the `pumpkin-world`/`pumpkin` crate dependency direction preventing a direct import),
- the `MINECRAFT_SUPPORT_OVERRIDE_SNOW_LAYER` tag as an override to that face-full requirement,
- and a brightness `< 10` gate via `BlockLightProvider::get_light_proto` with section-index math.

The separate water-freezing-to-ice logic earlier in the same function is untouched.

Fixes #3404

The previous snow placement check only verified the target was air and the block below wasn't air and lacked one specific tag, missing three vanilla requirements: a full top face on the support block, the override tag exception for blocks like honey/soul sand/mud, and a light-level gate. This allowed snow to incorrectly place on non-full-face blocks like slabs and stairs.

`FreezeTopLayerFeature::generate` has exactly one call site (`configured_features.rs`, during one-time proto-chunk worldgen decoration) with no tick-time/live-snow-accumulation caller in this codebase. During actual worldgen, light data is not yet computed, so `get_light_proto` returns 0 and the brightness gate passes unconditionally — this exactly matches vanilla's own behavior (unlit block light reads as 0, which is `< 10`, so the check is naturally a no-op at this specific call time in both vanilla and here). The face-full and override-tag checks are NOT inert and take effect immediately during worldgen. The brightness check is included for correctness and will become active wherever/whenever this function might be invoked with real light data in the future.

- Exact commands run: `cargo fmt` (clean), `cargo clippy --all-targets` (zero warnings), `cargo test --workspace` (all passing, including new test `snow_placement_rules`).
- No existing test harness existed for this feature; built one on `ProtoChunk` with a snowy-plains biome map (matching the pattern used in the earlier cave carver tests). Four cases: full-face block in darkness places snow, a slab (non-full-face, no override) does not, soul sand (non-full-face, has override tag) does, and a full-face block in bright light does not. Verified with a negative control: temporarily reverting to the old logic reproduces the reported bug (test fails), confirming the test actually exercises the new conditions rather than passing vacuously.
- Note the face-full check is a duplicated ~7-line block (not a shared import) due to the crate dependency direction — flag this as a known minor duplication for maintainer awareness, not hidden.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snow placement to require precipitation, freezing temperatures, a valid supporting block, and sufficient top-surface coverage or approved support behavior.
  * Prevented snow from appearing where block light is too bright or no valid air position exists.

* **Tests**
  * Added coverage for snow placement on solid blocks, slabs, soul sand, and brightly lit areas.


MegalithOfficial (changes requested):
I compared this with the 26.3 server code and ran isFaceFull over every block state, the union check gives the same result everywhere. The logic is fine.

The comment saying this is duplicated from `LayeredSnowBlock::can_place_at` isnt true. `can_place_at` looks for one box covering the top, this one doesnt, they disagree on ~2600 states, mostly top half stairs. Yours is the right one. Side effect is worldgen puts snow on top stairs now and the block's own check removes it on the first neighbor update, so `snow.rs` needs the same logic at some point. Can be a follow up, but drop the comment.

`shouldSnow` also checks `getPrecipitationAt` before temperature, we dont. No vanilla biome hits it, datapack ones can. It is the same condition so might as well add it.

The light check does nothing right now. ProtoChunk light is zero until the lighting stage and features run before that. Keep it, vanilla reads 0 there too, but  the description shouldnt say it is active. That is also why the first CodeRabbit comment is wrong, lighting before features would be the non vanilla behavior.

Kurenaiiiii:
@MegalithOfficial Thanks for the review! Fixed as suggested dropped the inaccurate `can_place_at` comment and added the `has_precipitation` gate before the temperature check. Ready for a re-review when you have a moment.

PR #3445, opened 2026-09-14
ci: Optimize CI for Pull Requests w/ affected checks only

Phoenixxo:
## Description

Reworks the Rust CI workflow into a prepare -> test/lint -> build pipeline that distinguishes pull requests from full `master` runs.

For pull requests, a new planning script reads Cargo workspace metadata and the Git diff to identify changed packages, then includes their reverse dependents so downstream crates are still validated. Clippy, nextest, doctests, and debug builds receive that affected package list. Unknown or workspace-wide inputs such as `Cargo.toml`, `Cargo.lock`, toolchain configuration, or the CI workflow itself conservatively fall back to the full workspace.

Affected pull requests run tests on Linux x64, Windows x64, and macOS, followed by debug builds on the existing five native GitHub runners and Android. They do not produce artifacts or run release builds. Documentation and repository-management-only changes can skip Rust CI entirely.

Pushes to `master` and manually triggered workflows continue running the complete workspace, including ARM test coverage, Android testing, release linting, release builds, and artifacts. Concurrency cancellation prevents superseded runs for the same workflow and ref from continuing.

The Nix workflow also gains concurrency cancellation and limits push-triggered builds to relevant changes on `master`, while retaining its existing pull-request path filters.

>[!note]
> This PR will of course run the full suite still as it is not a crate specific change, for reasons as stated above.

## Testing

- [Affected-package example](https://github.com/Phoenixxo/Pumpkin/actions/runs/34787766266): changing only `pumpkin-api-macros` selected that package. Planning, formatting, dependency checks, Clippy, tests, all five native debug builds, and the Android build passed. Release jobs and artifacts were correctly skipped.
- [Nix example](https://github.com/Phoenixxo/Pumpkin/actions/runs/34787766193): flake formatting, evaluation, and the default package build passed.
- Verified full-run selection reports all 20 workspace packages for a `master` push.

## Summary by CodeRabbit

- **Chores**
  - CI checks now identify affected packages and run targeted formatting, linting, testing, and builds where possible.
  - Workspace-level changes trigger full validation.
  - Pull request builds include separate Android and package-specific checks, with improved Android caching.
  - WIT binding validation is included in CI.
  - Superseded workflow runs are canceled automatically.
  - Push-based workflows run only for the `master` branch.
  - Nightly release naming is clearer.


Phoenixxo:
Apologies for duplicate commits, I was fighting git :/... Since this is a squash merge though, it shouldn't matter. you can verify the diff is only the in scope changes.

Phoenixxo:
Also one question before merging: Should the nix flake job run on PRs or only master? It feels redundant to have it run on PR, but it can help in case something breaks (though ideally another job catches something breaking)

.github/scripts/ci_packages.py
```diff
@@ -0,0 +1,291 @@
...
+    )
+
+
+def changed_paths(base: str, head: str) -> list[Path]:
+    if not base or not head:
+        raise ValueError("pull request planning requires both base and head SHAs")
+    output = run_command(
+        ["git", "diff", "--name-only", "--diff-filter=ACMRD", base, head, "--"],
+        capture=True,
+    )
```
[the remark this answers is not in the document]
Phoenixxo: Good catch, thanks!

.github/workflows/rust.yml
```diff
@@ -14,66 +16,97 @@ env:
...
           submodules: recursive
-      - run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
+      - run: rustup update stable && rustup default stable
       - run: rustup component add clippy
       - uses: Swatinem/rust-cache@v2
-      - name: Clippy (release)
-        run: cargo clippy --release --all-targets --all-features
+      - run: cargo clippy --release --all-targets --all-features
+      
   check_wit:
```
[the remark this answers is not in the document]
Phoenixxo: @coderabbitai Should be fixed now with the recent commit [ec7a063](https://github.com/Pumpkin-MC/Pumpkin/pull/3445/commits/ec7a063a7371eaefa2798024de06ba152659d056)

Phoenixxo:
@Snowiiii Pinging for review since it never assigned you to it, sorry

.github/scripts/ci_packages.py
```diff
@@ -0,0 +1,308 @@
...
+    "Cargo.toml",
+    "rust-toolchain.toml",
+}
+
+NON_RUST_ROOTS = {
+    ".github/ISSUE_TEMPLATE",
+    "docs",
+}
+
+NON_RUST_FILES = {
```
QuantumSegfault: I'd say `.github/workflows/sync-wit.yml` is non-Rust as well now.

QuantumSegfault: Missing others possibly? More Markdown? The Docker file? The Pterydactl(?) egg. etc.

Also as a nit-pick, maybe there's a better name than "NON_RUST"...? I'd say "NON_CODE" but then there's still the assets...

.github/workflows/reviewers.yml
```diff
@@ -48,6 +48,7 @@ jobs:
 
               { pattern: 'crates/pumpkin-world/src/generation/feature/', reviewers: ['Snowiiii', 'RoosterBooster007', 'MegalithOfficial'] },
               { pattern: 'crates/pumpkin-world/src/generation/noise/', reviewers: ['Snowiiii', 'RoosterBooster007', 'MegalithOfficial'] },
+              { pattern: '.github/workflows/*', reviewers: ['Snowiiii'] },
```
QuantumSegfault:
```suggestion
              { pattern: '.github/workflows/', reviewers: ['Snowiiii'] },
```

.github/workflows/rust.yml
```diff
@@ -14,106 +16,158 @@ env:
   RUSTFLAGS: "-Dwarnings"
 
 jobs:
+  prepare:
+    name: Plan CI
+    runs-on: ubuntu-latest
```
QuantumSegfault, line 28: `ubuntu-slim` might be good for this.

.github/workflows/rust.yml
```diff
@@ -14,106 +16,158 @@ env:
...
+          --head "$CI_HEAD_SHA"
+          --output "$GITHUB_OUTPUT"
+
   format:
     name: Check formatting
+    needs: prepare
+    if: needs.prepare.outputs.has_code == 'true'
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
```
QuantumSegfault, line 21: Matrix maybe not the right tool, but maybe having a variable to control this in a single place is still good? In case we decide to pin Rust versions later?

.github/workflows/rust.yml
```diff
@@ -14,106 +16,158 @@ env:
...
+          python-version: "3.x"
+      - run: rustup update stable && rustup default stable
       - run: rustup component add clippy
       - uses: Swatinem/rust-cache@v2
-      - name: Clippy (debug)
-        run: cargo clippy --all-targets --all-features
+        with:
+          shared-key: debug-${{ runner.os }}-${{ runner.arch }}
+      - name: Clippy affected packages
+        run: >-
```
QuantumSegfault, line 100: Python for this step is overkill. This should be doable with plain Bash. 

Just make the planner output a flat, newline separated list of packages (or similar), rather than JSON. Have to be careful with spaces in file names though...

Or you might even be able to use GHA's own expression evaluation instead? To read the JSON, and rewrite it into a series of switches and package names (in single-quotes to account for potential whitespace).

.github/workflows/rust.yml
```diff
@@ -14,106 +16,158 @@ env:
...
       - uses: actions/checkout@v7
         with:
           submodules: recursive
-      - run: rustup update ${{ matrix.toolchain }} && rustup default ${{ matrix.toolchain }}
+      - run: rustup update stable && rustup default stable
       - uses: bnjbvr/cargo-machete@main
-  clippy_debug:
-    name: Run lints (debug)
+
+  clippy:
```
QuantumSegfault, line 80: Why not keep the `_debug` specifier?

.github/workflows/rust.yml
```diff
@@ -14,106 +16,158 @@ env:
...
-  build_and_test:
-    name: Build project and test
-    runs-on: ${{ matrix.os }}
-    needs: [format, machete, clippy_debug, clippy_release, check_wit]
+
+  test:
+    name: Test (${{ matrix.name }})
+    needs: [prepare, format, machete, clippy, check_wit]
+    if: >-
+      always() &&
```
QuantumSegfault, line 134: `always()`???

.github/workflows/rust.yml
```diff
@@ -14,106 +16,158 @@ env:
...
+      - run: rustup update stable && rustup default stable
       - uses: taiki-e/install-action@nextest
       - uses: Swatinem/rust-cache@v2
-      - name: Cargo nextest
-        run: cargo nextest run --verbose
-      - name: Cargo doctest
-        run: cargo test --doc --verbose
+        with:
+          shared-key: debug-${{ runner.os }}-${{ runner.arch }}
+      - name: Test affected packages
```
QuantumSegfault, line 158: Python overkill.

.github/workflows/rust.yml
```diff
@@ -144,37 +199,111 @@ jobs:
...
           submodules: recursive
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.x"
+      - run: rustup update stable && rustup default stable
+      - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: debug-${{ runner.os }}-${{ runner.arch }}
+      - name: Build affected packages
+        run: >-
```
QuantumSegfault, line 242: Python overkill.

.github/workflows/rust.yml
```diff
@@ -144,37 +199,111 @@ jobs:
...
+        with:
+          ndk-version: r27c
       - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: android-debug-${{ runner.os }}-${{ runner.arch }}
+      - run: cargo install cargo-ndk
+      - name: Build affected packages
+        env:
+          CARGO_NDK_PLATFORM: 24
+        run: >-
```
QuantumSegfault, line 277: Python overkill

.github/workflows/rust.yml
Phoenixxo: sure, either works, it only takes like 5 seconds to run regardless so it's a very negligible change

.github/workflows/rust.yml
Phoenixxo: more of an implied thing since PRs only run debug now, can add back for clarity if you'd like

QuantumSegfault: Ah right. I made that comment before I fully realized that (`run_full`). 

Yeah, nevermind.

.github/workflows/rust.yml
Phoenixxo: we can easily make it a CI variable

.github/workflows/rust.yml
Phoenixxo: It's there so that github always evaluates the check regardless if a dependency fails or not

QuantumSegfault: Why do we want that? If the dependency fails, the check is moot, no?

.github/workflows/rust.yml
Phoenixxo: Python might be overkill in some ways but it contains all of necessary processing data and handling there, makes CI workflow file look cleaner... Also allows for more power in what we can do in some ways

.github/scripts/ci_packages.py
Phoenixxo: true

Phoenixxo: yeah :/ its hard to go through everything though and write down everything, i could also do assets/* and what not but yeah

Phoenixxo: tried my best to add some more that I can think of for now, we can always tweak it later

.github/workflows/rust.yml
Phoenixxo: For now, python is our cleanest solution, this can be refactored another time but it would significantly delay the goal of this PR even further if we do it now

.github/workflows/rust.yml
Phoenixxo: specifically, it lets tests proceed when check_wit is intentionally skipped, while the rest of the condition still requires format, machete, and clippy to succeed.

QuantumSegfault: Oh...okay, I see now

Makes sense.

QuantumSegfault (approved):
LGTM.

-----

Python vs Bash was discussed on Discord. Py seems the best way, all considered.

Snowiiii:
LGTM, Thanks @Phoenixxo :D

Issue #3367, opened 2026-09-10
game_rules/weather/wandering_trader read from root only, Paper-Purpur worlds silently lose them


RoosterBooster007:
Paper and Purpur place several world data files inside the overworld container (`dimensions/minecraft/overworld/data/minecraft/`) instead of the world root (`data/minecraft/`), per the [Paper migration docs](https://docs.papermc.io/paper/migration). 

Now that #3366 allows these worlds to load, our other readers still only check the root path. Because they don't find the Paper files, Pumpkin falls back to defaults and writes them to the root on save, permanently shadowing the real data (most notably resetting gamerules).

### Tracker

- [x] `world_gen_settings.dat` (resolved in #3366)
- [ ] `game_rules.dat` - currently falls back to defaults and shadows real rules on save (`anvil.rs:386`, `:454`)
- [ ] `weather.dat` - `clear_weather_time` defaults; root copy created on save (`anvil.rs:393`, `:481`)
- [ ] `wandering_trader.dat` - spawn delay/chance reset to defaults on save (`anvil.rs:404`, `:489`)
- [ ] `scheduled_events.dat` - needs fallback lookup once a reader is added

*(Other saved data like `scoreboard.dat`, `world_clocks.dat`, `custom_boss_events.dat`, and `maps/` are root-only across all platforms and don't need changes.)*

### Proposed Approach

Add a helper in `world_info::data_files` that checks root first (so existing Pumpkin saves stay authoritative), falling back to the overworld directory:

```rust
/// Checks root first, falling back to Paper's overworld dimension container.
fn find_data_file(level_folder: &Path, name: &str) -> Option<PathBuf>

PR #3366, opened 2026-09-10
fix(world): support 26.x world generation settings layouts

Phoenixxo:
## Description

Support world generation settings layouts used by vanilla and Paper-derived 26.x servers, as per an issue brought up by `@Stigstille`in `#general`. The loader now checks both the global and canonical overworld locations for `world_gen_settings.dat` and unwraps nested `data` or `Data` compounds when locating the world seed.

Fixes an existing issue with 26.x worlds failing to start with `No world seed found` even though their world generation settings contain a valid seed. Also added a regression test reproduces the affected nested file layout.

Tested with the following:

- `cargo test -p pumpkin-world`: 183 passed
- `cargo clippy -p pumpkin-world --all-targets -- -D warnings`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved world information loading for newer 26.x world layouts.
  - World seeds are now detected from both standard and nested world-generation data locations.
  - Added support for seed values nested within `data` or `Data` sections.
  - Improved warnings to identify the file that could not be parsed.

PR #3447, opened 2026-09-14
fix(world): preserve Paper overworld settings on import

younghyun1:
## Description

Importing a Paper or Purpur world can silently reset gamerules, weather, and wandering-trader settings. Pumpkin checks only the root data directory, while these worlds store the files under the overworld dimension. Saving then writes default root copies that shadow the imported data.

- **Read fallback:** Look for the three settings files at the root first, then in `dimensions/minecraft/overworld/data/minecraft/`. Update the Anvil loader's presence checks as well as the individual readers.
- **Precedence:** Keep existing root files authoritative and leave the imported files unchanged. This prevents new bad imports; it does not guess whether existing root settings came from a previous bad import or intentional changes.
- **Scheduled events:** Avoid creating an empty root stub over an existing imported file. Loading and executing scheduled events are not implemented by this change.
- **Scope:** Changes are confined to `world_info/data_files.rs` and `world_info/anvil.rs`, including regression tests. World-generation settings and root-only data such as world clocks and scoreboards are unchanged.
- **Related work:** The merged world-generation settings fix linked below covers a different set of files. This change extends import support to gamerules, weather, and wandering-trader settings, and prevents scheduled-event shadowing.

## Testing

- **Failing reproduction:** Before the fix, the Paper-layout round-trip and scheduled-event shadowing tests fail. The conflicting-root precedence control passes before and after.
- **Unit tests:** `cargo test --locked -p pumpkin-world --lib`: 219 passed.
- **Static checks:** `cargo clippy --locked -p pumpkin-world --all-targets -- -D warnings` and `cargo fmt --check`: passed.
- **Integration build:** `cargo build --locked -p pumpkin`: development build passed.
- **Real migration:** Created a world with official Paper 26.2 build 123. Unpatched Pumpkin changed `keep_inventory=true` and `random_tick_speed=19` to `false` and `3`. The patched server retains `true` and `19` on initial import and after save/restart.
- **Coverage limits:** Weather and trader preservation are covered by NBT fixtures. The full workspace suite and other platforms were not tested.

## References

- [Pumpkin-MC/Pumpkin#3367: Paper/Purpur world settings are ignored and shadowed on save](https://github.com/Pumpkin-MC/Pumpkin/issues/3367)
- [Pumpkin-MC/Pumpkin#3366: merged world-generation settings import fix](https://github.com/Pumpkin-MC/Pumpkin/pull/3366)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with Paper-style world data layouts when loading game rules, weather, and wandering trader data.
  - Preserved the correct priority of root-level world data files.
  - Prevented unrelated Paper data files from being affected.
  - Avoided overwriting existing scheduled-event data during saves.
  - Improved reliability when world data exists in either supported location.


younghyun1:
@RoosterBooster007, could you review this follow-up to #3366?

PR #3453, opened 2026-09-14
feat(scheduler): add stackless global domain scheduling

Phoenixxo:
## Summary

Introduces `pumpkin-scheduler`, a standalone crate for scheduling async work through execution domains. This gives us a foundation for letting tasks wait for results without blocking the worker, while keeping control over ordering and how much work is allowed at once.

## Description

The scheduler accepts owned async tasks and returns an awaitable result. Tasks and wakeups are queued in FIFO order, with one task polled at a time. When a task returns Pending, other ready work can run, including nested calls that the original task is waiting on.

Nested submissions preserve their parent's call-chain context. Ordering applies to individual turns, so an entire async operation isn't atomic. Callers need to release locks before awaiting and recheck state when they resume.

Capacity includes ready, running, and pending tasks. Once full, new submissions receive an explicit error. The crate also handles duplicate wakeups, reports task panics, and provides queue counts and slow-poll diagnostics. CPU-heavy work still needs to be handled separately, since the scheduler can't interrupt a poll.

The executor is supplied by the caller. The crate doesn't create a Tokio runtime or depend on Wasmtime, WIT, or Pumpkin's game types.

### Where this fits into the overall plan

This is the first implementation step toward the stackless async scheduler planned for the next plugin runtime. This has been discussed at length with @QuantumSegfault and a few others in `#plugin-dev`. Starting with Global lets us establish submission, ordering, and suspension behavior before connecting it to plugins and game state.

The next steps are cancellation, deadlines, barriers, and graceful shutdown. After that, we can connect async plugin dispatch and gradually move plugin-capable work onto the scheduler. Plugin-free CPU work will continue using Rayon.

World, region, and entity domains can follow once their ownership rules are defined. This PR only accepts global submissions.

## Validation

Three integration tests cover progress with eight pending tasks on one worker, nested call chains, FIFO turns, wakeups, capacity limits, and failure handling.

Scheduler tests, Clippy, and formatting checks pass.

## Summary by CodeRabbit

* **New Features**
  * Added a global task scheduler with FIFO execution, bounded capacity, nested tasks, and configurable polling.
  * Added task submission, awaitable handles, execution domains, cancellation controls, task context, and scheduler status snapshots.
  * Added lifecycle handling for shutdown, task panics, invalid submissions, queue limits, and executor failures.

* **Tests**
  * Added coverage for scheduling, nested task chains, cancellation, wake handling, capacity limits, and error scenarios.


Phoenixxo:
@Snowiiii @QuantumSegfault pinging for review, not sure who else could review it and the workflow doesnt have this tracked to auto assign reviewers so

crates/pumpkin-scheduler/src/global.rs
```diff
@@ -0,0 +1,326 @@
...
+        let queued = {
+            let mut queue = shared.lock();
+            if let Some(entry) = queue.tasks.get_mut(&signal.id)
+                && !entry.queued
+            {
+                entry.queued = true;
+                queue.ready.push_back(signal.id);
+                true
+            } else {
+                false
```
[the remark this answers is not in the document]
Phoenixxo: @coderabbitai, this is an intentional implementation, the PR only mentioned handling them, not reporting them. This is part of the coalescence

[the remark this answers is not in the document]
Phoenixxo: If the task is already queued, another wake doesn't need to add it again. If the task no longer exists, there's nothing left to wake. Both cases require no further action here.

The PR description says we handle duplicate wakeups, not report them. That's referring to coalescing them into one queued turn, which the integration test covers. Separate counters could be useful later, but aren't required for correctness.

crates/pumpkin-scheduler/src/global.rs
PigTurtleDev: this is `usize::from(bool)`. Is that intentional?

PigTurtleDev: This could be extracted into a separate (private) function for brevity. Not a big deal though.

PigTurtleDev:
```suggestion
            // The signal will mark the task ready again once woken by the future
            let wake = waker_ref(&signal);
```
It took me a bit to understand that the work will re-enqueue itself due to the signal used as its waker. It's likely because I'm not that familiar with custom async executors, but a comment wouldn't hurt.

Cargo.toml
QuantumSegfault, line 1: Lots of reformatting causing diff noise. Please undo these so we can see the actual changes.

crates/pumpkin-scheduler/src/task.rs
```diff
@@ -0,0 +1,124 @@
...
+impl SchedulerTaskId {
+    #[must_use]
+    pub const fn get(self) -> u64 {
+        self.0
+    }
+}
+
+/// Owned causal metadata preserved across suspension and worker moves
+///
+/// A child inherits this chain only while if and only if it's parent is still valid
```
QuantumSegfault:
```suggestion
/// A child inherits this chain only while it's parent is still valid
```

While makes more sense here as an ongoing check; and "while and only while" would just be redundant.

QuantumSegfault (approved):
Looks good as far as I'm concerned. In isolation at least. Still will need to see how this all fits together on the grand scheme.

Seems to be a lot going on here...but it seems like mostly boilerplate to interface with Rust's concept of async and with an outer scheduler (i.e. Tokio).

Otherwise its as simple as running one task a time a time per scheduler poll, and making children tasks wait for parent tasks.(..right?)

QuantumSegfault (approved):
Alright. Just need to fix CI failures now.

QuantumSegfault:
@coderabbitai full review

PR #3459, opened 2026-09-15
fix(elytra): vanilla-parity fixes for gliding and flight state

darkzmaj:

Vanilla-parity fixes for elytra gliding and flight state:

- Stop gliding immediately when the elytra is unequipped mid-flight, not just on landing - covers both movement packet handlers, one of which had no fall-flying-stop logic at all.
- Starting a glide now actually requires a usable elytra (equipped, not broken, not underwater, no Levitation) instead of just being airborne - previously you could toggle flight with any chestplate contents.
- Creative/spectator flight and elytra gliding can no longer coexist, matching vanilla - enabling flight force-stops any active glide, and gliding can't start while already flying.
- No fall damage landing right after gliding, matching vanilla.
- `set_fall_flying` no longer panics on a redundant call - it's reachable directly from the plugin API with no guard of its own, so a no-op call (two plugins reacting to the same event) would've crashed the server thread.
- Skin-parts metadata write failures now log instead of being silently swallowed.

## Summary by CodeRabbit

- **New Features**
  - Added safer Elytra gliding checks for underwater status, levitation, equipment condition, and durability.
  - Added off-hand swing animations across Java and Bedrock.
  - Preserved living-entity equipment when saving and loading.
  - Improved compatibility of player skin-part metadata across Java client versions.

- **Bug Fixes**
  - Prevented flight and Elytra gliding from being active simultaneously.
  - Gliding now ends when players land or remove their Elytra.
  - Fall-damage handling now accounts for the previous gliding state.
  - Repeated flight-state updates no longer cause errors.
  - Creative players now receive eligible block-entity drops.
  - Metadata write failures are now logged.


darkzmaj:

PR #3460, opened 2026-09-15
fix(item-drops): apply air drag once in grounded item friction

darkzmaj:

Grounded item friction was applied as `0.98 * slipperiness * 0.98` - air drag baked into the starting value and then multiplied in again. Vanilla applies it once (`slipperiness * 0.98`), so items here stopped ~2%/tick sooner than the client's own prediction of the same item.

This PR is now **only** that friction fix, plus a unit test. Following review feedback from @RoosterBooster007, the earlier per-tick position sync and the rest-velocity/buoyancy handling were dropped: #2832 deliberately keeps vanilla's 20-tick item tracking on Java (Bedrock gets per-tick updates via `send_bedrock_pos`), and per-tick sync made item movement laggier in testing.

Verification: `cargo fmt --check` clean, new unit test passes, clippy reports nothing in `item.rs`. (My local clippy can't run the workspace's full `-Dwarnings` pass because it doesn't know a lint the workspace config enables, so I'm relying on CI for that.) I have not tested in-game.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved grounded friction behavior for more predictable horizontal item movement.


darkzmaj:

RoosterBooster007:
Hey @darkzmaj, thanks for splitting this off from #3093. Before removing the 20-tick throttle, I'd try the physics first: `move_and_apply_friction` multiplies horizontal velocity by `0.98 * slipperiness * 0.98`, where vanilla only does `airDrag * slipperiness` once, so our items stop ~2%/tick sooner than the client's own prediction, and the next position packet is what snaps them back. Also, the clamp on `velo.y` eats the +5.0e-4 fluid nudge (4.9e-4 after friction, under the 1e-3 epsilon), so items in water or lava would hang instead of bobbing up (CodeRabbit's comment). Skipping the y clamp while buoyancy is active should cover that. Vanilla also refreshes the item's velocity every interval (`trackDeltas` is true for items), which we never do, so if the drift survives the friction fix, that's the next place to look.

darkzmaj:
Good catch, thanks. Fixed both:

- `move_and_apply_friction` was doing `0.98 * slipperiness * 0.98` for grounded horizontal velocity - the trailing `*= ` was compounding onto the already-set 0.98 air drag baseline instead of replacing it. Now it's just `slipperiness * 0.98`, applied once, matching vanilla.
- The rest-snap on `velo.y` (anything under `1.0e-3` gets zeroed to stop settled items from spamming sync packets) was also eating the buoyancy nudge, since `5.0e-4` becomes `4.9e-4` after the vertical drag multiply - under the epsilon. Now the snap is skipped while the item is actually submerged enough for buoyancy to be active, so it can bob up instead of hanging.

Left the per-tick sync as-is for now per your suggestion - didn't touch it. If drift still shows up after this, I'll look at the `trackDeltas`-style velocity refresh next.

Pushed as 6b287366f.

RoosterBooster007:
Nice, both fixes look right. However, could you drop the sync half and keep this as the friction fix for now? #2832 deliberately kept vanilla's 20-tick item tracking on Java and only gave Bedrock per-tick updates (which is why `send_bedrock_pos` exists). Also, the comment above the sync block says clients don't predict item motion, which isn't quite right (Java clients appear to tick items locally and then re-anchor them with `SetEntityMotion` each ~20t interval). I tested your sync changes, and item movement seems laggier.

darkzmaj:
Thanks @RoosterBooster007, that makes sense, and thanks for testing it.

Done, pushed as 454d85287. I dropped the sync half entirely: `sync_motion_if_dirty` and the `ITEM_UPDATE_INTERVAL` throttle are back to exactly what's on master, so Java keeps vanilla's 20-tick tracking and Bedrock keeps `send_bedrock_pos`. You're right about the comment too - I had that backwards, so it's gone along with the code it described.

I also removed the rest-velocity snap and the buoyancy exception. Both were only there to stop the per-tick sync from spamming packets, and master has no y-clamp at all, so with the sync reverted they'd be dead weight (and the buoyancy problem you and CodeRabbit spotted goes away with them).

What's left is just the friction fix: `slipperiness * 0.98` applied once instead of `0.98 * slipperiness * 0.98`, plus a small unit test. I updated the title and description to match. I haven't run this in-game, so I can't say yet whether the remaining drift is gone. If it isn't, I'll look at the `trackDeltas`-style velocity refresh you mentioned as a separate follow-up.

PR #3461, opened 2026-09-15
fix(setworldspawn): actually persist the new spawn point to level.dat

darkzmaj:

`level_info.store()` only updates the in-memory copy - the change looks like it worked (feedback shows the new coords, in-session respawn/compass behavior uses it) but silently reverts to the old value on the next restart, since nothing actually reached level.dat. Now calls `save_world_info()` right after, matching the pattern every other `level_info`-mutating command already uses (see `/datapack`).

## Summary by CodeRabbit

* **Bug Fixes**
  * World spawn changes now persist after the server restarts.
  * Save failures are logged while the command continues to report success.


darkzmaj:

PR #3462, opened 2026-09-15
fix(firework-rocket): render the rocket and always allow launching

darkzmaj:

The rocket entity never sent its item-stack metadata (what the client renders the model/trail from), so a boost-fired rocket was completely invisible - no model, no trail - even though the boost itself worked correctly server-side. Now stores the item stack used to fire it and syncs that on spawn, same pattern as the other thrown-item entities (egg, splash potion, etc).

Also, a firework rocket is a functional item on its own (decorative on the ground, a boost while gliding), not something that only works while already flying - both `use_on_block` and `normal_use` were wrongly gated on `is_fall_flying()`, which broke plain non-gliding launches. The boost itself stays correctly gated separately, inside the spawned entity's own tick logic based on the owner's fall-flying state at the time.

## Summary by CodeRabbit

- **New Features**
  - Firework rockets preserve their specific item data, enabling custom rocket effects and visuals.
  - Rockets can be launched using a rocket held in either hand.
  - Rockets may be launched during normal use without requiring fall-flying.
  - Block-placed rockets now appear at the player’s position.

- **Bug Fixes**
  - Rocket launches now correctly validate available ammunition before spawning.
  - The selected rocket stack is properly consumed after use.


darkzmaj:

PR #3463, opened 2026-09-15
fix(protocol): fix stream desync and lost text for named items in bundles

darkzmaj:
Split out from #3011 per RoosterBooster007's request (1 of 2 - the other is the stale block-entity guards).

Bundle contents (`minecraft:bundle_contents`) encode nested items using the plain Slot format, which has no per-component byte-length prefix. `deserialize_item_stack_template` read and discarded a VarInt as if such a prefix existed, consuming the first bytes of the component's actual payload and desyncing every read after it. Any bundle holding an item with a non-empty component patch - e.g. a crafted firework rocket with real explosion data - corrupted the rest of the stream, which then surfaced as an unrelated "component_id_N (TODO)" disconnect once the garbage bytes were misread as a component ID.

Separately, `ItemNameImpl` serializes its text as raw NBT (matching the wire format), but its `deserialize` read a plain length-prefixed wire string instead - an asymmetry that broke any custom-named item that round-trips through the generic codec (e.g. nested in a bundle), even after the length-prefix bug above is fixed. Uses the `get_nbt_with_version` helper `CustomNameImpl` already uses for the same field.

Added a regression test that serializes a bundle containing a renamed, non-default firework rocket and asserts it decodes back correctly.

Built and verified against current master: clippy clean on pumpkin-protocol, regression test passing, dependent `pumpkin` crate still builds.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility when displaying item names from translated and text-based data.
  - Fixed item stack decoding for bundled components, preventing nested firework items from causing subsequent data to be read incorrectly.
  - Preserved compatibility with existing string-formatted item names and safely handles invalid name data.


darkzmaj:
@RoosterBooster007 @Snowiiii @aure31 @xRookieFight this is the first of the two split-out PRs from #3011, ready for review. The second (block-entity guards) is coming shortly.

PR #3465, opened 2026-09-15
Guard stale block-entity property decoding (split 2/2 of #3011)

darkzmaj:
Second half of the split from #3011, per RoosterBooster007's request there: the stale block-entity guards.

Several block entities decode their `BlockProperties` from whatever block currently occupies their position without first checking it's still the block they expect. If the block gets swapped out between ticks, or a container GUI interaction races a block removal, `from_state_id` happily decodes garbage bits into the wrong properties type instead of the tick/interaction backing off cleanly.

Added a `handles_block_id` check before each `from_state_id` call in:
- `barrel.rs` (`set_open`, `play_sound`)
- `brewing_stand.rs` (`tick`'s bottle-presence update)
- `chest_like_block_entity.rs` (shared `play_sound` helper, covers chest + trapped chest)
- `furnace_like_block_entity.rs` (shared `tick` helper, covers furnace/blast furnace/smoker)
- `jigsaw_block.rs` (`generate`)

This matches the guard `hopper.rs` already has. `daylight_detector.rs` and `hopper.rs` didn't need any changes since they no longer hit this pattern.

Verified with `cargo check`/`cargo clippy --all-targets --all-features` (debug + release) and `cargo fmt --check` on the `pumpkin` crate, all clean.

Part 1 of this split is #3463.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented unsupported barrel, chest, furnace, brewing stand, and jigsaw blocks from triggering invalid state changes or interactions.
  - Corrected sound behavior so unsupported blocks no longer play container-related sounds.
  - Prevented unsupported jigsaw blocks from generating structures.
  - Preserved existing behavior for valid supported blocks, including brewing stand updates and furnace state handling.


darkzmaj:
@RoosterBooster007 @Snowiiii @aure31 @xRookieFight this is part 2/2 of the #3011 split (the stale block-entity guards). Part 1 (bundle desync + ItemName fix) is #3463. I don't have push access to this repo so I can't formally request review — tagging you here instead. Happy to address any feedback!

xRookieFight:
@Snowiiii Not sure on this, topic LGTM but I don't know about the implementation

darkzmaj:
Thanks @xRookieFight, and no worries about not knowing the implementation. It's a small change, so here's what it does in case it helps you or @Snowiiii judge it.

**The problem:** a block entity can outlive its block for a moment, for example when the block is replaced by something else. The entities here (barrel, chest, brewing stand, furnace, jigsaw) decode the block's properties directly from its current state id. If that block is no longer the type they expect, they'd interpret an unrelated block's state as their own, then write a state back or play a sound based on garbage.

**The fix:** before touching properties, each one checks the block currently at its position is actually the kind it expects, and bails out if not. Behavior for a valid block is unchanged.

I just pushed an update (6e3c92785) that:
- resolves the merge conflict with master (an import line in `brewing_stand.rs`)
- fixes two real issues Copilot found in review:
  - **Brewing stand:** the "last potion count" cache was updated *before* the guard, so a stale entity that bailed out would think the update had already happened and never retry once the stand was restored. It's now only recorded after the state is actually written.
  - **Chests:** the guard used a check that also matches the copper chest variants, which these entities don't represent. It now compares against the exact block each entity belongs to (`CHEST` / `TRAPPED_CHEST`).

Copilot also flagged that the furnace check and the state write are two separate world accesses. That's technically true, but the world only exposes an unconditional `set_block_state`, so a real compare-and-set would need new engine API. I'd rather not widen a small guard PR to add that. Happy to look at it as a follow-up if @Snowiiii thinks it's worth it.

`cargo fmt` is clean and the block-entity tests pass locally. I haven't run this in-game.

PR #3469, opened 2026-09-16
fix: remove killed entities outside simulation

### LINKED ISSUE #3435: /kill on entities outside simulated chunks leaves permanent 0-HP corpses (vanilla removes immediately); corpses also get persisted on chunk unload
### Before Submitting

Related: #2503, #3401 (entity resync rework — the client-visible desync side), #1698 (items invisible after relog).

*(Body rewritten 2026-09-14 after deeper testing — the original "duplication / resurrection from persisted storage" theory was wrong. See the correction comments for the full trail. What remains is a much narrower but real vanilla-parity gap.)*

### Minecraft Edition

Java Edition

master `4de1744`

### Operating System

Windows 11 x86_64

### What happens

**Entities killed outside simulated chunks (e.g. via `/kill` with no player nearby) are never removed.** They stay in the live entity list as 0-HP corpses indefinitely, keep matching `@e` selectors, inflate `/tp`/`/kill` counts, and are written to the chunk's entity storage on unload, so they re-enter the world on the next load.

### Root cause

- `/kill` on a living entity routes through `damage(f32::MAX)` → `on_death` (loot, pose `Dying`) → and then relies on the **tick-driven** removal in `LivingEntity::tick` (`crates/pumpkin/src/entity/living.rs:3339-3368`): when `health <= 0`, `death_time` increments each tick and at 20 the entity is `remove()`d. This mirrors vanilla's `tickDeath` (`LivingEntity.java:572-578`) and works fine inside simulated chunks.
- Vanilla, however, removes the entity **synchronously** in `Entity.kill(ServerLevel)` (`Entity.java:410-413` → `remove(RemovalReason.KILLED)`), independent of ticking. Pumpkin's `EntityBase::kill` (`crates/pumpkin/src/entity/mod.rs:725-732`) diverges here for living entities.
- When the entity's chunk is not being simulated (no player nearby; `forceload add` alone does not make entities tick — verified live below), `death_time` never advances and the corpse stays. `@e` still matches it because `Entity::is_alive()` is `!is_removed()` (`entity/mod.rs:3322-3328`) and `removal_reason` is only set by `World::remove_entity` (`world/mod.rs:5245-5263`), which the frozen corpse never reaches.
- `World::save_entity` (`world/mod.rs:580`) only skips entities with a `removal_reason`, so a frozen corpse whose chunk does unload is saved with `Health: 0` and re-materializes on the next load — the "invisible entities" and inflated selector counts seen in play.

### Evidence (RCON only, 0 players online, `forceload add 0 0 128 128`)

1. Summon + `/kill` an iron golem → the corpse remains selectable forever; `kill @e[type=iron_golem,nbt={DeathTime:0s}]` matches it minutes later, i.e. the death branch in `tick` never ran even once.
2. A live zombie summoned into the same forced area kept exactly `Health: 20.0f` through 20 seconds of daytime (no burn ticks, no physics) — entities in forced-but-unsimulated chunks do not tick at all, so nothing (including the death timer) can advance.
3. After a server restart the same area contains zero golems and the entity region files (`entities/r.*.mca`, netty-style `LZ4Block` framing) contain no golem records from that session — the corpses were live-list state, not duplicated data.
4. In-game consequence observed earlier: after long headless test sessions, `/tp @e[type=minecraft:iron_golem] @s` reported 48 entities while one was visible, and repeated `/kill` kept reporting the same corpses.

### Suggested fix

Make `kill()` on living entities remove synchronously like vanilla (`Entity.kill` → `remove(RemovalReason.KILLED)`), or at minimum force-removal when the entity's chunk is not simulated. Optionally skip saving entities with `health <= 0` in `save_entity` so already-frozen corpses from older sessions can't be re-added to new ones.

### Expected (vanilla)

`/kill` removes the entity immediately (no death animation, no loot) regardless of simulation state; natural deaths only occur in simulated chunks where `tickDeath` completes.

HP-network:
Fixes #3435

`/kill` currently routes living entities through the normal damage and death-animation path. Entities outside the simulation distance do not tick, so non-player corpses can remain in the world's entity list and be saved during chunk unload.

Keep the existing damage, death-event, loot, and statistic flow, then remove non-player living entities immediately after `/kill` completes. Players retain the existing death and respawn lifecycle.

Testing:
- `cargo test -p pumpkin --lib --no-fail-fast` (249 passed)
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Non-player entities are now detached from vehicles and their passengers are removed before the entities are removed.
  * Player kills continue to apply the standard kill damage behavior.


HP-network:
Updated in e55c247.

Non-player `/kill` targets now follow the vanilla synchronous removal path instead of entering the damage/death animation. This avoids drops and death events for non-player kill targets and works when their chunks are outside simulation. Player targets keep the existing damage path.

Validation: `cargo fmt --check`, `git diff --check`, and `cargo test -p pumpkin --lib` (249 passed).

HP-network:
Updated in 35fe2c0.

Before direct removal, non-player kill targets now detach from their vehicle and clear each passenger's reciprocal vehicle reference. This prevents stale mount state after killing a vehicle or passenger, including when the entity is outside simulation distance.

Validation: `cargo fmt --check`, `git diff --check`, and `cargo test -p pumpkin --lib` (249 passed).

PR #3471, opened 2026-09-16
fix(entity): stop resending player collision velocity to self

### LINKED ISSUE #3464: Pressing against a mob after taking knockback launches the player upwards (server keeps sending the player its own stale SetEntityMotion y)
### Before Submitting

Related: #3127 (player could not push entities — fixed by #3180, which introduced the velocity-friction code involved here), #3401 (entity resync through the tracker).

### Minecraft Edition

Java Edition

master `02b16e15b` (playtest build = master + my #3434/#3458 branches, which only touch `entity/ai/goal/*` and `entity/passive/*` goal registration — none of the paths below)

### Operating System

Windows 11 x86_64 (client Minecraft 26.2)

### What happened?

Standing against a mob after taking a knockback launches the player into the air and keeps feeding upward velocity, instead of the normal vanilla push-out.

The server sends `SetEntityMotion` **for the player's own entity** on every push, carrying a `y` velocity the server never decays. The client dutifully applies it, so the player gets repeatedly shoved upward while pressing against a mob.

### Steps to reproduce

1. Be in **survival** (creative players aren't pushable — `crates/pumpkin/src/entity/player.rs:6718-6720` returns false for Creative/Spectator, so the whole chain doesn't run).
2. Take one knockback from any mob — e.g. let a zombie hit you once. This sets the server-side `velocity.y` to `0.4`.
3. Press against any mob (cow, iron golem, …) for a few seconds.

Result: the player is launched upward and keeps getting pushed up while in contact.

### Evidence (packet capture)

A TCP proxy between client and server decoded the clientbound `SetEntityMotion` packets (`id=101`, packed `LpVec3` format). The player's own entity id was read from the `play/login` packet (`id=49`, first field is a 4-byte big-endian entity id). After one knockback followed by pressing against a cow:

```
[21:21:54] SetEntityMotion id=13755 [*** SELF ***] v=(+0.2046, +0.4000, +0.0764)
[21:22:45] SetEntityMotion id=13755 [*** SELF ***] v=(-0.0422, +0.4000, +0.0396)
[21:22:45] SetEntityMotion id=13755 [*** SELF ***] v=(-0.0800, +0.4000, +0.0759)
[21:22:46] SetEntityMotion id=13755 [*** SELF ***] v=(-0.0452, +0.4000, +0.0623)
[21:22:48] SetEntityMotion id=13755 [*** SELF ***] v=(+0.0005, +0.4000, +0.0924)
[21:22:50] SetEntityMotion id=13755 [*** SELF ***] v=(+0.0129, +0.4000, +0.1665)
...
[21:23:36] SetEntityMotion id=13755 [*** SELF ***] v=(+0.0667, +0.4000, +0.0678)
```

`id=13755` is the player's own entity. **22 packets with `y = +0.4000`** were sent, spanning 21:21:54 → 21:23:36 (**102 seconds**), i.e. not a one-off knockback packet — a sustained per-push feed. The `x`/`z` components jitter as expected from being pushed, while `y` stays pinned at the knockback value.

### Root cause

Three things combine:

1. **Push marks the velocity dirty, and the player's tick then broadcasts it — to everyone including the entity itself.**

   `Entity::push` modifies `x`/`z` of the velocity and sets `velocity_dirty` (`crates/pumpkin/src/entity/mod.rs:587` / `:595`). The player's tick consumes that flag and calls `send_velocity()` (`crates/pumpkin/src/entity/living.rs:3137-3140` on master), which broadcasts through `broadcast_to_chunk_editioned`:

   ```rust
   // crates/pumpkin/src/entity/mod.rs:1268
   pub fn send_velocity(&self) {
       let velocity = self.velocity.load();
       let chunk_pos = self.chunk_pos.load();
       self.world.load().broadcast_to_chunk_editioned(
           chunk_pos,
           &CEntityVelocity::new(self.entity_id.into(), velocity),
           ...
   ```

   The recipient filter in `broadcast_to_chunk_editioned` (`crates/pumpkin/src/world/mod.rs:7096-7110`) only checks chunk distance — it **does not exclude the entity itself**, so a player receives a velocity packet about itself.

2. **Vanilla never sends a player its own periodic velocity.**

   In vanilla, the per-tick velocity sync only happens when `trackDelta` is set (`ServerEntity.java:177-193`), and `EntityType.trackDeltas()` returns `this != EntityTypes.PLAYER` (`EntityType.java:432-433`) — so players are never delta-tracked. Entity→player pairing also skips self (`ChunkMap.java:1413`: `if (player != this.entity)`). The only self-directed motion packet in vanilla is the one-shot `hurtMarked` send (`ServerEntity.java:225-228` → `sendToTrackingPlayersAndSelf`).

3. **The player's `y` velocity is never decayed server-side, so the stale knockback value stays forever.** *(corrected below: this part is vanilla behaviour too, so it is not the divergence that causes the bug)*

   Players take a different tick branch from mobs: `LivingEntity::tick` only calls `apply_travel_friction()` for players (`crates/pumpkin/src/entity/living.rs:3316-3320`), and that function **only touches `x`/`z`** — it even early-returns when `x == z == 0` (`crates/pumpkin/src/entity/living.rs:1445-1467`). Mobs instead go through `tick_movement()`, which calls `check_zero_velo()` (`crates/pumpkin/src/entity/living.rs:1324`) to zero out tiny components. So the `y = 0.4` left by `apply_knockback` (`crates/pumpkin/src/entity/mod.rs:1540-1567`, called from the damage path at `crates/pumpkin/src/entity/living.rs:3152`) is never cleared. Client movement packets don't clear it either — `progress_motion` (`crates/pumpkin/src/entity/player.rs:2926-2946`) only adds exhaustion.

   **[edit] Correction:** vanilla is the same here. `Player.canSimulateMovement()` returns `true` on the server (`Player.java:1290-1292`), so the `scale(0.98)` branch in `LivingEntity.aiStep` (`LivingEntity.java:3038-3042`) never applies to a server-side player, and `travel()` is skipped because the server player is not the movement authority. A stale `y = 0.4` in the server-side `deltaMovement` is normal vanilla behaviour and is harmless as long as it is never sent back. So the divergence that causes this bug is (1) alone; (3) only explains why the value that gets echoed back stays at `+0.4` instead of fading away.

So: one knockback leaves a permanent `y = 0.4`; every later touch of a mob re-dirties the velocity; each push then broadcasts that full vector to the player itself; the client keeps applying `+0.4` upward.

### Expected (vanilla)

A player standing against a mob is pushed out horizontally. The player only ever receives its own `SetEntityMotion` as a one-shot when actually knocked back (`hurtMarked`), and never gets a repeated feed of a stale `y`.

### Suggested fix

Mirror vanilla's `hurtMarked` semantics: broadcast `send_velocity` to tracking players other than the entity, and send to the entity itself only when it has actually been knocked back.

One trap to watch: simply excluding self from `send_velocity` would break knockback delivery for players, because the damage path calls `Entity::apply_knockback` (`entity/mod.rs:1540`), which only stores velocity + sets `velocity_dirty` and does **not** send anything itself (unlike `Player::set_velocity`, which does send directly at `player.rs:2421`). So the knockback needed a separate self-send path.

Also worth considering: give players the same small-value `y` cleanup that mobs get via `check_zero_velo`, and/or decay `y` in `apply_travel_friction`, so a stale component can't linger indefinitely.

HP-network:
## Description

Player entity pushing currently sends the full velocity vector back to the player whenever collision marks velocity dirty. After a knockback leaves a vertical component in the velocity, repeated mob collisions resend that stale component and can launch the player upward.

Route player velocity updates through tracked viewers and send a velocity packet to the player only for the knockback applied by a successful damage hit. Other entities keep the existing broadcast path.

Fixes #3464

## Testing

- `cargo test -p pumpkin --lib --locked --offline`
- `cargo check -p pumpkin --lib --locked --offline`
- `cargo clippy -p pumpkin --lib --locked --offline -- -D warnings`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved velocity updates after entities receive knockback.
  - Player movement updates are now sent only to relevant tracking players.
  - Non-player entity velocity updates continue to be broadcast as before.


HP-network:
Closing this in favor of #3476, which covers the same self-velocity issue with the complete tracking behavior.

PR #3476, opened 2026-09-16
fix(entity): stop sending a player its own stale motion on every push 🤖🤖🤖

### LINKED ISSUE #3464: Pressing against a mob after taking knockback launches the player upwards (server keeps sending the player its own stale SetEntityMotion y)

Developers13:
## Problem

Standing against a mob **after taking a knockback** launches the player into the air and keeps feeding it upward velocity, instead of the normal vanilla push-out. Reported as #3464.

The server sends the player **its own** `SetEntityMotion` on every push, carrying a `y` the server never decays, so the client keeps applying `+0.4` upward while in contact.

## Root cause

Three things line up:

1. `Entity::push` marks the velocity dirty (`entity/mod.rs` `push()`), and the player's tick consumes that flag through `send_velocity()`, which broadcasts with `broadcast_to_chunk_editioned`. That recipient filter (`world/mod.rs`) only checks chunk distance — it does **not** exclude the entity itself, so a player gets a velocity packet about itself every tick it is pressed against a mob.
2. Vanilla never does that. The per-tick motion sync only runs when `needsSync || trackDelta` and goes to `sendToTrackingPlayers` (`ServerEntity.java:177-195`); `EntityType.trackDeltas()` is `false` for players (`EntityType.java:432-433`), `ChunkMap` skips self when pairing (`ChunkMap.java:1413`), and the only motion packet an entity ever receives about itself is the one-shot `hurtMarked` send (`ServerEntity.java:225-228`). Pumpkin had no `hurtMarked` concept at all.
3. The knockback `y` (`Entity::apply_knockback`, max `0.4`) is never cleared for players, because players skip `tick_movement`/`check_zero_velo` and only get `apply_travel_friction`, which handles `x`/`z` only. So the value echoed back in (1) stays pinned at `+0.4`.

## Changes

Mirror vanilla's `hurtMarked` split:

- `entity/mod.rs`: new `hurt_marked: AtomicBool`. `send_velocity()` now goes to **tracking players only** (`send_to_tracking_players_editioned`) — matching vanilla's `needsSync` path. New `send_velocity_and_self()` mirrors `ServerEntity`'s `sendToTrackingPlayersAndSelf` `hurtMarked` branch. Packet construction factored into `velocity_packets()`.
- `entity/living.rs`: the damage path sets `hurt_marked` where vanilla calls `markHurt()` (same block as the knockback), and the entity's tick consumes it right after the existing `velocity_dirty` coalescing — i.e. at the same point `ServerEntity.tick` ends, so the packet carries the tick's final velocity (including the extra sprint/enchantment knockback that `combat::handle_knockback` adds afterwards).

**Trap we avoided:** simply dropping the entity from the broadcast would have broken knockback delivery, because the damage path calls `Entity::apply_knockback`, which only stores velocity and sets the dirty flag — it does not send anything (unlike `Player::set_velocity`, which sends to the client directly). That is exactly why the self send had to be modelled explicitly.

**Not changed on purpose:** we do **not** add a `y` decay / `check_zero_velo` for players, and we do not remove `apply_travel_friction`. Vanilla has the same "stale `y` stays on the server" property for players — `Player.canSimulateMovement()` is `true` server-side (`Player.java:1290-1292`), so `LivingEntity.aiStep`'s `scale(0.98)` branch (`LivingEntity.java:3038-3042`) does not apply, and `travel()` is skipped because the server player is not the local authority. The stale value is harmless as long as it is never echoed back to the player, which is the part that was missing.

## Testing

`cargo fmt` clean, `cargo clippy --all-targets` clean, `cargo test` all green.

Behavior verified A/B on a local server with a **headless protocol client** (no human needed — the push/pushable logic is entirely server-side), with a TCP proxy decoding clientbound `SetEntityMotion` (`LpVec3`) and tagging packets addressed to the player's own entity id (read from `play/login`).

Scenario: survival player on the ground → one `damage ... minecraft:mob_attack` knockback → cows kept inside the player's hitbox (re-teleported every second, offset ~0.6 blocks so `Entity::push` actually fires) for 40 s.

| | motion packets addressed to the player itself | of those, `y ≈ +0.4` | packets about other entities |
|---|---|---|---|
| before | **347** | **347** (all, `y = 0.39999`) | 7883 |
| after | **1** | 1 | 5608 |

- Before: every self packet had `y` pinned at the quantized `0.4` while `x`/`z` jittered with the pushes — the same shape as the capture in #3464.
- After: exactly one, at the moment of the knockback (`v = (-0.85, +0.40, -0.85)`, i.e. with the knockback direction). Nothing while pressed.
- The "other entities" column stays populated in both runs, so the second run is not a silent zero-packet failure.

Reverse check (no regression): knocking a cow back still delivers its motion to the players tracking it — the nearby player received 12 packets with `y ∈ [0.3, 0.5]` after the hit.

Fixes #3464

🤖🤖🤖

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed knockback movement updates so affected entities receive their own velocity update.
  * Improved velocity synchronization to ensure tracking players receive consistent movement updates.
  * Prevented duplicate self-velocity updates within a single game tick.


gitKBDL:
This no longer merges since #3401. I needed it on my server, so here is a rebase onto current master with your authorship kept: gitKBDL/Pumpkin-Core@a45d2226 (branch [`fix/player-self-velocity`](https://github.com/gitKBDL/Pumpkin-Core/tree/fix/player-self-velocity)), with lints and tests passing on it ([run](https://github.com/gitKBDL/Pumpkin-Core/actions/runs/35897512040)). Feel free to take it as is.

The only conflict: #3401 made `send_velocity` the immediate `hurtMarked` path behind `set_velocity`, so that stays, and the per-tick flush calls a new `send_velocity_to_watchers` instead.

Possibly the same bug: a player there was launched upward on joining. Their saved `Motion` had y≈2.45, and since `velocity_dirty` starts out true, the first flush sent that motion back to them. With this rebase that no longer happens.

PR #3483, opened 2026-09-16
fix(world): accept byte booleans in flat generator settings

### LINKED ISSUE #3482: Flat generator settings from vanilla worlds fail to parse because NBT bytes are read as numbers
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

nightly (Minecraft 26.2)

### Operating System

Linux

Pumpkin does not use the flat generator settings of vanilla worlds. It uses the built-in preset `classic_flat` instead. New chunks contain bedrock, dirt and grass. The world layers, the biome and the structure settings do not apply. The console shows no warning.

Expected result: Pumpkin generates the layers and the biome from the world settings.

Example: a vanilla superflat world uses these settings:

```json
{
  "biome": "minecraft:lukewarm_ocean",
  "features": 0,
  "lakes": 0,
  "layers": [
    {"block": "minecraft:air", "height": 64},
    {"block": "minecraft:bedrock", "height": 1},
    {"block": "minecraft:sandstone", "height": 46},
    {"block": "minecraft:sand", "height": 4},
    {"block": "minecraft:water", "height": 1}
  ]
}
```

Pumpkin generates bedrock, dirt and grass for this world.

Cause:

NBT has no boolean type. Vanilla writes the boolean values `features` and `lakes` as bytes with the value 0 or 1. Pumpkin reads the settings as JSON. It converts the bytes to numbers. Then it reads the numbers as booleans. This step fails. Pumpkin rejects the complete flat settings object and selects `classic_flat`.

The relevant code:

```rust
// crates/pumpkin-world/src/world_info/mod.rs
#[serde(default)]
pub features: bool,
#[serde(default)]
pub lakes: bool,
```

```rust
// crates/pumpkin-world/src/world_info/data_files.rs
NbtTag::Byte(b) => serde_json::Value::Number((*b).into()),
```

The protocol code already contains a byte-to-boolean conversion. The world info code does not use it.

Impact: every vanilla flat world loses its preset. Superflat maps and maps with a custom flat preset are affected. Stored chunks stay correct. Only new chunks use the wrong layers.

Related: #1499 fixed the case where `settings` was a string. This report covers the remaining parse failure.

### Steps to Reproduce

1. Create a superflat world in vanilla Minecraft with the settings above.
2. Start Pumpkin with that world.
3. Load a chunk that is not stored in the region files.
4. The new chunk contains bedrock, dirt and grass.

matheustavarestrindade:
Fixes #3482.

Problem:

Vanilla writes the flat settings values `features` and `lakes` as bytes with the value 0 or 1. Pumpkin converts the bytes to numbers. Then it reads the numbers as booleans. This step fails, so Pumpkin rejects the settings and uses `classic_flat`. All vanilla flat worlds lose their preset.

Change:

- The fields `features` and `lakes` accept a boolean or the numbers 0 and 1.
- A missing field still uses `false`.
- `as_flat_settings` logs a warning when the settings do not parse.

```rust
// crates/pumpkin-world/src/world_info/mod.rs
#[serde(default, deserialize_with = "deserialize_bool_from_byte")]
pub features: bool,
#[serde(default, deserialize_with = "deserialize_bool_from_byte")]
pub lakes: bool,
```

Tests:

- `flat_settings_accept_byte_booleans`
- `flat_settings_accept_booleans`

## Summary by CodeRabbit

- **Bug Fixes**
  - Added clearer warning messages when compound generator settings cannot be deserialized.
  - Preset feature and lake settings now support native booleans and numeric values, with `0` treated as false and nonzero integers treated as true.

- **Tests**
  - Expanded coverage for boolean and numeric setting formats, including NBT byte conversion and native boolean values.


MegalithOfficial (approved):
Looks good. Vanilla saves lakes and features as bytes, so this was breaking every vanilla flat world. I also ran the test without your change and it fails.

One small thing: vanilla counts any nonzero number as true instead of erroring. With this a 2 still drops the whole preset back to classic_flat. `Number(n) => Ok(n != 0)` would match vanilla.

Would also be nice to have a test that starts from an actual NBT compound and goes through `nbt_tag_to_json` and `as_flat_settings`. Right now the tests only use serde_json.

QuantumSegfault (approved):

PR #3488, opened 2026-09-16
fix(world): read chunks without yPos

matheustavarestrindade:
Upgraded worlds store the section `Y` with different numeric tags inside the same chunk: the sections that came from the original map use a byte, and the sections the datafixer adds use an int. The chunk reader reads `Y` with `get_byte`, which only matches byte tags, so every int section is treated as `Y = 0`. They all land on the same slot and overwrite each other, and the fallback for a missing `yPos` then decides from those wrong values.

This change reads `Y` from any numeric tag and uses what vanilla does for a chunk without `yPos`: the lowest section that stores biomes, and never above 0.

**Problem**

A real chunk from a 1.17.1 world upgraded to 26.2 by the vanilla server:

```text
yPos: absent
section Y tags: 3 byte, 21 int
section that holds the island: byte Y = 4
all 24 sections store biomes
```

With the byte-only read, `get_byte("Y")` finds only 3 sections. The fallback returns `4`, so the chunk starts at y=64 instead of y=-64, and the 21 int sections all collapse into one slot. Their terrain is lost and the real section 0 is overwritten with air. On a normal converted world that shreds the map.

The Nether chunk from the same world: 2 byte and 14 int tags, all sections store biomes, lowest biome section is 0.

A second case from the review: old chunks keep a light section at `Y = -1`. It stores no biomes, and the fallback must ignore it, otherwise the chunk starts at -1 and the client shows everything one section too high. Our scans did not find this case in 2,528 chunks per dimension, so it is covered by a synthetic test.

**Change**

- `section_y` reads `Byte`, `Short`, `Int` and `Long`.
- `lowest_biome_section_y` takes the lowest section that stores biomes, and never above 0. That gives `-4` in the Overworld and `0` in the Nether and End, the section minimums of those dimensions.
- The parser loop and the `max_y_section` loop use the same helper, so a chunk cannot be read with two different rules.
- The old comment claimed vanilla reads the lowest section. It does not; it uses the dimension minimum. The comment now says what the code does.

**Tests**

- `chunk_without_y_pos_uses_the_lowest_biome_section` (int `Y`, no `yPos`)
- `chunk_with_int_y_sections_keeps_every_section` (two int sections stay at their heights)
- `chunk_with_mixed_numeric_y_tags_keeps_sections` (byte and int tags in one chunk)
- `fallback_ignores_light_only_sections_below_zero`
- `fallback_never_goes_above_zero`
- `chunk_without_y_pos_or_biomes_still_fails` (the `Missing yPos` error stays for chunks with no usable section)

All six fail against the old behavior and pass with the change. `cargo test -p pumpkin-world` passes with 225 tests, `cargo fmt --check` is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chunk parsing when chunk-level vertical position data is missing.
  * Supports byte, short, integer, and long section position values.
  * Applies a safe fallback of `0` for unsupported or out-of-range values.
  * Derives missing vertical positions only from sections containing biome data, capped at `0`.
  * Ignores light-only sections when deriving vertical position.
  * Preserves an error when no valid vertical position data is available.
  * Applies section-based vertical positioning consistently when loading chunk data.


MegalithOfficial (changes requested):
I just checked this against 26.3. You're right that `forceUpgrade` doesn't add `yPos`, but vanilla doesn't read the lowest section either; it just uses the dimension's minimum section. So the comment in the code isn't accurate.

The bigger issue is that this doesn't work on actual upgraded chunks. I upgraded a 1.17.1 world with 26.3 to test it. The sections added by `ChunkHeightAndBiomeFix` store `Y` as an int, not a byte, while `get_byte` only reads byte values. Because of that, all of those sections end up being treated as `Y = 0`, which causes them to overwrite the real `Y = 0` section with air. `Y` should be read from any numeric type, like vanilla does.

The Nether and End are broken as well. Old chunks still retain their `Y = -1` light section after the upgrade, so this logic picks `-1` instead of `0`. As a result, everything appears one section too high on the client, and the wrong `yPos` gets saved.

Vanilla's `ChunkProtoTickListFix` only considers sections that contain biomes and never goes above `0`. Doing the same here gives the expected values: `-4` in the Overworld and `0` in the Nether and End.

The current test uses a byte `Y`, so it passes despite the issue. It would be good to add a test using int `Y` values, matching what the datafixer actually writes.

matheustavarestrindade:
Addressed. Y is read from any numeric tag now (byte, short, int, long), and the fallback uses the lowest section that stores biomes, never above 0, so an upgraded chunk gets -4 in the Overworld and 0 in the Nether and End. The light section at Y = -1 is ignored.

Checked against a real 1.17.1 world upgraded to 26.2 by the vanilla server: the island chunk has 3 byte and 21 int Y tags, all 24 sections store biomes, and the fallback lands on -4. Six tests cover int Y, mixed tags, the light section, the cap and the remaining Missing yPos error, and all of them fail against the old behavior. Description updated.

MegalithOfficial (approved):
I checked it again, all three things from my last review are fixed.

Only thing left is the `fallback_never_goes_above_zero` test. It passes even if you remove the `.min(0)`, since the stone at y=16 is found either way. Check that y=0 returns air instead, without the cap that would be None.

Small note: when no section has biomes vanilla just uses 0, here it still errors. This shouldnt happen on upgraded worlds anyway.

Good to merge from my side once the test is fixed.

PR #3492, opened 2026-09-16
fix(world): keep block entities when a chunk is resumed from disk

### LINKED ISSUE #3491: Chests that come from a world file are invisible and cannot be opened
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

nightly (Minecraft 26.2)

### Operating System

Linux

A chest that comes from an existing world is not usable. The block data is in the chunk, but Pumpkin does not create the block entity. The client draws nothing for the chest (it looks transparent) and the server has no container, so the chest cannot be opened.

Image

A chest that a player places in the same session works. It is visible and it opens.

Expected result: a chest from an existing world is visible and opens like a placed chest.

**Example.** Two chests, one block apart, in the same chunk. Both were checked with the same client in one session:

- Chest A at `-67, 67, 0`, from the world file: transparent, cannot be opened.
- Chest B at `-66, 67, 0`, placed live with `/setblock`: visible, opens normally.

**The stored data is present.** The chunk at `(-5, 0)` in `r.-1.0.mca` contains:

```text
block_entities: minecraft:chest at (-67, 67, 0)
  Items: minecraft:obsidian x10, minecraft:melon_slice x1, minecraft:pumpkin_seeds x1
  keepPacked: 0
```

The record also survives a normal visit: after the player loaded the chunk and the chunk was saved again, the entity and all three items are still in the file. Only the runtime block entity is missing.

**Cause, from reading the code.** The placement path creates the block entity directly, so a placed chest works. A chunk loaded from disk must instantiate its block entities through the migration path: `migrate_pending_block_entities` in `crates/pumpkin/src/world/mod.rs`, which calls `get_block_entity` and then `update_block_entity`. For chunks that come from a world file this does not happen, or the result does not reach the client. The console shows no warning or error.

**Impact.** Every chest, barrel, furnace, sign, and other block entity that comes from an existing world is dead until a player breaks and replaces it. The data is not lost, but it cannot be reached in the game.

Related, but different: #2626 reports that the chunk writer drops block entities on save. Here the writer keeps them; the stored data stays intact. #3104 covered opening chests that were placed in the session.

### Steps to Reproduce

1. Use a world that contains a chest with items. Do not place that chest in the current server session.
2. Start Pumpkin on that world.
3. Join the server and go to the chest.
4. The chest is transparent and cannot be opened.
5. Place a new chest next to it. The new chest is visible and opens normally.

matheustavarestrindade:
Chunks that go through the proto-chunk path lose their block entities. A chest from a world file is invisible to the client and cannot be opened on the server, and the next save writes an empty `block_entities` list, so the data is erased on disk. This change keeps the block entities when the chunk data is turned into a proto chunk.

**When this happens**

`ProtoChunk::from_chunk_data` runs for two kinds of chunks:

- chunks saved below `Full` status, which continue generation on load;
- `Full` chunks whose lighting is uniform, which Pumpkin downgrades to relight.

Converted and upgraded worlds land in the second case. Their lighting is uniform (a superflat or void world is all sky light), so Pumpkin relights them on the first load, and the block entities are lost there.

**Problem**

A world with a chest does this:

1. A player joins and goes to the chest.
2. The chest is transparent and cannot be opened.
3. A chest placed in the same session, in the same chunk, works fine.
4. After a save, the chest and its contents are gone from the file.

Server side, the world really has no container:

```text
> item replace block -67 67 0 container.0 with minecraft:emerald 1
Target position -67, 67, 0 is not a container

> item replace block -66 67 0 container.0 with minecraft:emerald 1
Replaced a slot at -66, 67, 0 with [Emerald]
```

The first chest came from the world file. The second one was placed in the session.

**Cause**

`ChunkData::from_bytes` reads the block entities into the chunk's pending list. `ProtoChunk::from_chunk_data` then copies the blocks, the biomes, the heightmaps, the light and the generation stage, but not the block entities. The pending list of the proto chunk stays empty, so the chunk that reaches the world has no block entities and the save writes none.

A debug build shows both ends:

```text
[DBG] migrate Vector2 { x: 0, y: -1 } read=Some(0)
[DBG] save chunk (0, -1) block_entities=0
```

**Change**

Copy the entities in `from_chunk_data`:

```rust
proto_chunk.pending_block_entities = chunk_data
    .pending_block_entities
    .lock()
    .unwrap_or_else(std::sync::PoisonError::into_inner)
    .values()
    .cloned()
    .collect();
```

**Tests**

`block_entities_survive_chunk_data_resume` builds a chunk with one chest, upgrades it to a level chunk, resumes it with `from_chunk_data` and checks the entity in the proto chunk and again in the level chunk. The test fails without the change.

**Verification**

Local server on a converted 26.2 world, which is relit on load and hits this path. Before: the chest is transparent, `item replace block` says "not a container", and the save writes zero entities. After: the chest renders, opens, keeps its two items, and the saved chunk still has the entity.

Fixes #3491. Also fixes the loss of block entities on save described in #2626.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved pending block entities, such as chests, when loading and reconstructing chunk data.
  * Ensured block entity identity and contents remain intact across chunk conversion and reloading.

* **Tests**
  * Added regression coverage for block entity persistence.


MegalithOfficial (approved):
I checked against 26.3, vanilla keeps block entities the same way when loading a proto chunk. Test fails without the fix. LGTM.

Description is a bit off though. Normal full chunks don't go through `from_chunk_data`, only unfinished chunks and chunks that get relit. So it's not "every block entity".

matheustavarestrindade:
Description updated. Thanks for the correction.

QuantumSegfault (approved):

PR #3493, opened 2026-09-16
feat(scheduler): add cancellation and completion tracking to scheduler

Phoenixxo:
## Summary

Adds cancellation and shared completion handling to the Global domain scheduler
This is the second step of the v0.2 scheduler work and builds on the Global domain foundation PR

### Long-term: 

The goal is to move plugin execution toward stackless async scheduling with explicit ownership of work and state
Instead of holding a Rayon thread through an entire plugin call chain, work can suspend while waiting on a child task, host operation, or I/O result
The scheduler keeps execution ordered within a domain while allowing unrelated work to make progress elsewhere
This PR adds the task lifecycle pieces needed before plugins can be routed through that model

---
## Changes
- Adds CancellationHandle for explicit task cancellation
- Cancels active descendants while leaving parents, siblings, and unrelated chains running
- Uses one completion path for success, errors, cancellation, panics, and driver shutdown
- Releases task admission after cleanup
- Adds completion, cancellation, and failure counters to scheduler snapshots
- Documents task ownership, suspension, cancellation, and cleanup behavior
- Adds integration coverage for cancellation races, descendant chains, cleanup panics, and driver teardown

###  To continue on: 
Later work will add domain routing, deadlines, lifecycle draining, and the bridge from plugin calls into scheduler-owned work
Once domains are introduced, plugin work can be routed to the owner of the relevant world, region, entity, or global state instead of treating every plugin call as one global blocking operation

>[!note]
> This is depending on #3453 to be merged first (as it includes it's changes). But once it is merged in this can be fully opened up, until then it will sit as a draft. I tried to make it a stacked PR but it doesn't work across repos :(


Phoenixxo:
@coderabbitai full review

PR #3494, opened 2026-09-16
fix(world): save entity chunks as a snapshot instead of appending

wiflow:
Stumbled over this while testing #3490. `save_entity` only ever appends to a chunk's entity list, and both `/save-all` and the unload path call it, so a mob that walked into another chunk in between gets written into both chunks. Same-chunk copies collapse on reload because of the UUID guard, but the stale copy in the old chunk comes back as a second mob with old position and health as soon as that chunk loads on its own. My test world had 36 mobs saved in two chunks after a few sessions, 22 of them in one nether region file.

Chunks whose entities went live are now rebuilt from what is actually in them on every save (vanilla writes a section's entities as a snapshot too), and chunks that were loaded but never spawned keep their records with fresh ones merged in by UUID. `/save-all` still doesn't flush the entity files to disk, that's a separate thing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved entity persistence during world saves, shutdowns, and chunk unloading.
  - Ensured entities retain their current chunk location and are restored correctly when the world is loaded.
  - Empty active chunks are now saved accurately, preventing stale entities from reappearing.
  - Preserved entities in inactive areas while updating records for entities that have moved or changed.


RoosterBooster007 (approved):

RoosterBooster007:
Looks good. Thanks, @wiflow!

PR #3500, opened 2026-09-17
fix(protocol): reject overflowing final byte in varint decoders 🤖🤖🤖

### LINKED ISSUE #3499: VarInt decoders accept overflowing final-byte payload bits
### Description

The synchronous and asynchronous `VarInt`/`VarUInt` decoders enforce the maximum byte count, but do not validate the unused high bits in the final byte. The same issue exists for `VarLong`/`VarULong` in their tenth byte.

As a result, out-of-range encodings are silently truncated and accepted as valid values. For example:

```text
00                   -> 0
80 80 80 80 10       -> 0  (should be rejected: bit 32 is set)

ff ff ff ff 0f       -> -1
ff ff ff ff 7f       -> -1 (should be rejected: bits above bit 31 are set)
```

I reproduced those results by applying the current decoder operations to the byte sequences above. In the fifth iteration, `(byte & 0x7f) << 28` is evaluated as an `i32`/`u32`, so bits above bit 31 are discarded. The decoder then returns the truncated value because the continuation bit is clear.

The same pattern is present in:

- `VarInt::decode` and `decode_async`
- `VarUInt::decode` and `decode_async`
- `VarLong::decode`
- `VarULong::decode` and `PacketRead::read`

This matters in particular for Bedrock packet framing, where `VarUInt::decode` reads both the packet length and header. Malformed encodings should not be accepted as aliases for canonical values.

### Expected behavior

Reject a five-byte `VarInt`/`VarUInt` when the fifth byte contains payload bits above the type width. Likewise, reject a ten-byte `VarLong`/`VarULong` when the tenth byte contains payload bits above bit 63.

For 32-bit values, the fifth byte may contain at most the low four payload bits. For 64-bit values, the tenth byte may contain at most one payload bit.

### Suggested fix

Before OR-ing the final byte, validate its payload mask. Apply the same check to both synchronous and asynchronous paths, and add regression tests for the examples above plus their 64-bit equivalents.

Relevant source: https://github.com/Pumpkin-MC/Pumpkin/blob/844e2871e2091c14b9991cb6534e97a6c8e09480/crates/pumpkin-protocol/src/codec/var_int.rs#L56-L95

tunglambk:
## Description

Fixes #3499.

The varint decoders enforce the maximum byte count but never look at the high payload bits of the final byte. Those bits get shifted past the end of the type and dropped, so an out-of-range encoding comes back as a valid value: `80 80 80 80 10` decodes to `0` even though bit 32 is set, and `ff ff ff ff 7f` decodes to `-1`.

This rejects a final byte whose payload is wider than the bits left in the type:

- `VarInt`/`VarUInt`: the fifth byte only has four payload bits, so a payload above `0x0f` is an error;
- `VarLong`/`VarULong`: the tenth byte only has one payload bit, so a payload above `0x01` is an error.

The check is in every decoder that had the hole: `decode` and `decode_async` for `VarInt`/`VarUInt`, `decode` for `VarLong`/`VarULong`, and the `PacketRead::read` impls for all four. The `PacketRead` paths matter too, since Bedrock packet parsing and item data read through `VarInt::read`/`VarUInt::read`. All of them use the error the surrounding code already returns for an oversized varint (`ReadingError::TooLarge` in the `decode` paths, `std::io::Error` with `ErrorKind::InvalidData` in `PacketRead::read`).

I also removed the four stale `// TODO: Validate that the first byte will not overflow a i32/i64` comments, since this is that validation.

## Testing

`cargo fmt --all -- --check`, `cargo clippy -p pumpkin-protocol --all-targets --all-features` with `RUSTFLAGS=-Dwarnings` (the CI gate) and `cargo test -p pumpkin-protocol` all pass locally; the crate's 125 unit tests are green.

The tests cover the largest legal value, one bit past it and an all-ones final byte for each type, on the sync path, the async path where it exists, and `PacketRead::read`. Before the change `VarInt::decode(&[0x80, 0x80, 0x80, 0x80, 0x10])` returned `0`, `VarInt::decode(&[0xff, 0xff, 0xff, 0xff, 0x7f])` returned `-1`, and the 64-bit decoders accepted `ff ff ff ff ff ff ff ff ff 7f` as `-1`/`u64::MAX`; all of those now return an error. Round-trip tests keep the legal range unchanged, including `i32::MIN`/`i64::MIN`, which need the full fifth/tenth byte.

🤖🤖🤖

## Summary by CodeRabbit

* **Bug Fixes**
  * Added overflow validation for variable-length 32-bit and 64-bit integer decoding.
  * Invalid encodings that exceed supported numeric ranges are now rejected consistently across synchronous, asynchronous, and packet-based decoding.

* **Tests**
  * Added coverage for maximum valid values, boundary cases, round trips, and overflowing encodings.


QuantumSegfault:
Closing in favor of #3536

PR #3524, opened 2026-09-17
Fire `EntityChangeBlockEvent` when falling block lands

Neol3108:
Inline with how Bukkit does it

## Summary by CodeRabbit

- **Bug Fixes**
  - Falling blocks now respect block-change cancellation when landing.
  - Cancelled placements remove the falling block without placing it in the world.


crates/pumpkin/src/entity/falling.rs
```diff
@@ -69,6 +69,21 @@ impl EntityBase for FallingEntity {
...
+            );
+            if let Some(server) = world.server.upgrade() {
+                server.plugin_manager.fire_blocking(&server, &mut event);
+            }
+            if event.cancelled {
+                self.entity.remove();
+                return;
+            }
+
             world.set_block_state(&landing_pos, state_id, BlockFlags::NOTIFY_ALL);
```
QuantumSegfault: How about moving this inside of `if !event.cancelled`, instead of the duplicate remove and early return?

Neol3108: Fair, updated.

PR #3541, opened 2026-09-18
fix: preserve gamemode feedback argument order

HP-network:
The gamemode feedback translation uses different argument orders on Java and Bedrock. Keep the platform-specific order explicit so player and mode values are rendered in the correct positions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bedrock translation rendering with platform-specific substitution parameters.
  * Corrected argument ordering for translated game mode feedback messages.
  * Added fallback behavior to preserve existing translations when Bedrock-specific parameters are unavailable.
* **New Features**
  * Added support for translations with separate Java and Bedrock keys and arguments, improving message accuracy across platforms.


HP-network:
Closing this duplicate follow-up. The same gamemode translation change was already submitted in #3540; I will not keep two PRs for the same fix.

PR #3555, opened 2026-09-18
fix(block): relight campfires hit by burning projectiles

### LINKED ISSUE #3550: Burning projectiles cannot relight extinguished campfires
## Summary

Extinguished campfires cannot be relit by flaming arrows or other burning projectiles, unlike vanilla Java.

## Current behavior

`CampfireBlock` in `crates/pumpkin/src/block/blocks/campfire.rs` has no `on_projectile_hit` implementation and ends with an explicit `// TODO: onProjectileHit`.

Projectile collision is already wired through the block registry, and other blocks such as TNT inspect `args.projectile.get_entity().is_on_fire()` in their `on_projectile_hit` handler. Because campfires use the default no-op handler, every burning projectile hit is ignored.

## Vanilla behavior

In vanilla Java, an unlit, non-waterlogged campfire is relit when hit by a burning projectile. This includes Flame arrows and other projectiles that are on fire. A waterlogged campfire must stay extinguished.

References:

- https://minecraft.wiki/w/Campfire
- https://maven.fabricmc.net/docs/yarn-1.21.1+build.3/net/minecraft/block/CampfireBlock.html

## Reproduction

1. Place a campfire or soul campfire.
2. Extinguish it so `lit=false` and `waterlogged=false`.
3. Shoot it with a Flame arrow.

Pumpkin leaves it extinguished. Vanilla sets `lit=true`.

Also verify that a waterlogged campfire remains unlit when struck.

## Suggested fix

Implement `CampfireBlock::on_projectile_hit` and set the existing campfire state's `lit` property to true only when:

- the projectile is on fire,
- the campfire is currently unlit, and
- it is not waterlogged.

Preserve facing and signal-fire state, notify clients/neighbors, and cover regular and soul campfires plus the waterlogged case with tests.

Verified against current `master` at `1ac447f2918b7dec7d19a5836fa3673d4fe3c9f5`; targeted GitHub searches for campfire projectile ignition, flaming arrows, and `onProjectileHit` found no existing issue.

HP-network:
Burning projectiles should relight an unlit campfire, but the default projectile handler currently ignores the hit.

Relight regular and soul campfires when the projectile is on fire, while leaving lit and waterlogged campfires unchanged.

Checks: cargo fmt --check; cargo check -p pumpkin

Fixes #3550


matheustavarestrindade:
https://github.com/user-attachments/assets/4f7363a0-8123-4cfa-b02f-e8513b5aaf8a

MidnightDoes (approved):
LGTM, only issue would be arrows but that's arrow issue not campfire issue.

PR #3556, opened 2026-09-18
fix(block): collapse big dripleaves hit by projectiles

### LINKED ISSUE #3552: Projectiles do not collapse big dripleaf leaves
## Summary

Hitting a big dripleaf with a projectile does nothing. Vanilla immediately tilts the leaf fully down for 100 ticks, even when the leaf is powered by redstone.

## Current behavior

`BigDripleafBlock` in `crates/pumpkin/src/block/blocks/plant/big_dripleaf.rs` implements delayed tilting from entity steps and reset behavior from redstone, but has no projectile handler. The implementation contains an explicit:

```rust
//TODO: onProjectileHit
```

The default `BlockBehaviour::on_projectile_hit` is a no-op, so arrows, snowballs, wind charges, and other projectiles never change the leaf's tilt.

Pumpkin already has the needed `set_tilt_and_schedule_tick` helper: `Tilt::Full` schedules the reset after 100 ticks.

## Vanilla behavior

In vanilla Java, any projectile hit immediately makes a big dripleaf fully tilted and temporarily non-solid. Redstone power prevents normal entity-triggered tilting, but intentionally does not prevent projectile-triggered collapse.

References:

- https://minecraft.wiki/w/Big_Dripleaf
- https://maven.fabricmc.net/docs/yarn-23w33a+build.7/net/minecraft/block/BigDripleafBlock.html

## Reproduction

1. Place a big dripleaf.
2. Shoot its leaf with an arrow or throw a snowball at it.
3. Repeat while powering the leaf with redstone.

Pumpkin leaves the leaf upright in both cases. Vanilla immediately sets `tilt=full` in both cases, then resets it after 100 ticks.

## Suggested fix

Implement `BigDripleafBlock::on_projectile_hit` and call the existing tilt helper with `Tilt::Full`, without applying the redstone-power guard used by entity stepping. Preserve facing and waterlogged state, play the vanilla tilt-down sound, and test powered and unpowered leaves with multiple projectile types.

Verified against current `master` at `1ac447f2918b7dec7d19a5836fa3673d4fe3c9f5`; targeted GitHub searches for big dripleaf projectiles, arrow tilting, and `onProjectileHit` found no existing issue.

HP-network:
Projectiles currently have no effect on big dripleaves. Vanilla makes the leaf fully tilt immediately, including when it is powered.

Use the existing full-tilt transition so the normal 100-tick reset and tilt-down sound are preserved.

Checks: cargo fmt --check; cargo check -p pumpkin

Fixes #3552

## Summary by CodeRabbit

- **Bug Fixes**
  - Big dripleaf now reacts correctly when struck by a projectile: it fully tilts, plays the appropriate sound, and resets after a short delay.
  - Previously scheduled tilt updates are cancelled before a new update is queued, preventing conflicting or outdated state changes.
  - Block tick scheduling is cleaned up more reliably, improving consistency when timed world updates are cancelled.


HP-network:
Updated the projectile path to cancel any existing big dripleaf block tick before scheduling the 100-tick reset. This prevents a prior 10-tick entity-step update from shortening the projectile tilt.

Validation: `cargo fmt --all`, `cargo test -p pumpkin-world remove_tick_clears_the_queued_entry --lib`, `cargo check -p pumpkin`, and `git diff --check`.

PR #3559, opened 2026-09-18
Allow lily pads on waterlogged blocks

### LINKED ISSUE #3557: Lily pads cannot be placed on waterlogged blocks
## Summary

Lily pads cannot be placed on waterlogged blocks, including waterlogged top slabs, even though vanilla Java treats the water fluid in those blocks as valid support.

## Current behavior

`LilyPadBlock::can_plant_on_top` in `crates/pumpkin/src/block/blocks/plant/lily_pad.rs` has an explicit TODO to use fluids rather than blocks. It reads only the support block:

```rust
let block = block_accessor.get_block(pos);
...
block.has_tag(&tag::Fluid::MINECRAFT_SUPPORTS_LILY_PAD)
    || block.has_tag(&tag::Block::MINECRAFT_SUPPORTS_LILY_PAD)
```

The block support tag contains ice and frosted ice. The fluid support tag contains water, but testing that fluid tag against the block object does not inspect the block's waterlogged fluid state. A waterlogged slab therefore fails both support checks.

## Vanilla behavior

Vanilla Java allows lily pads on water and on waterlogged blocks, including waterlogged top-half slabs. Mojang classified placement on waterlogged blocks as working as intended.

References:

- https://minecraft.wiki/w/Lily_Pad
- https://mojira.dev/MC-127588

## Reproduction

1. Place a top-half slab and waterlog it.
2. Try to place a lily pad on top.

Pumpkin rejects the placement. Vanilla allows it.

Ice and frosted ice should continue to support lily pads without needing a water fluid state.

## Suggested fix

Expose or use the support position's fluid state in the plant-placement path. Accept support when either:

- the fluid at the position is in `minecraft:supports_lily_pad`, or
- the block is in the block tag of the same name.

Keep the existing requirement that the lily pad's destination position is air, and add tests for water, waterlogged top slabs, ice, frosted ice, and dry slabs.

Verified against current `master` at `1ac447f2918b7dec7d19a5836fa3673d4fe3c9f5`; live issue-list review plus a targeted repository web search found no existing lily-pad/waterlogged issue.

HP-network:
Fixes #3557

Lily pads currently only check the support block tag, so waterlogged blocks are rejected even though they contain source water.

Use the support block state to recognize water and waterlogged blocks while preserving the existing ice support and air check for the placement position.

Validation:
- `cargo check -p pumpkin`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved lily pad placement on supported waterlogged blocks and fluids recognized as compatible with lily pads.
  - Lily pads can now be placed consistently when the space above the supporting block is clear.


MegalithOfficial (changes requested):
I tested this in game and waterlogged blocks work now. The source vs flowing part is right too, the fluid tag only has water and `from_state_id` maps the flowing states to flowing_water.

The issue is the above check. `can_plant_on_top` isnt only used for placement, `get_state_for_neighbor_update` goes through it too, and there the block above is the lily pad itself. So `is_air()` is always false and the pad gets removed on any neighbor update. Put two lily pads next to each other on water and both break. Same on ice, place a pad and put any block next to it. This is already broken on master, but this PR rewrites that line so it should be fixed here. Vanilla 26.3 checks `fluidAbove.is(Fluids.EMPTY)`, so no fluid above, not air above.

Kelp, kelp_plant, seagrass, tall_seagrass and bubble_column also hold source water in vanilla, but they have no waterlogged property so `is_waterlogged()` misses them. `World::get_fluid_from_state_id` already has that list. Can be a follow up.

We use conventional commits here, so please redo the commit and the PR title. `fix(block): allow lily pads on waterlogged blocks` would be it.

Small note: the PR only lists fmt and check. For block behavior changes I suggest trying it in game too.

HP-network:
Thanks for catching this.

Updated the neighbor-update check to require an empty fluid above the pad, so adjacent lily pads and blocks no longer invalidate each other. The check also covers water-bearing plants without a waterlogged property.

I also updated the commit and PR title to the repository convention. `cargo fmt --all -- --check` and `cargo check -p pumpkin` pass. I have not run a full in-game session yet.

HP-network:
Thanks for testing this and spelling out the neighbor-update case. I see the distinction: placement still needs an air destination, while neighbor updates need to check for empty fluid above the support. I'll separate those checks, add coverage for the cases you listed, and test the behavior in-game before asking for another review.

PR #3564, opened 2026-09-18
fix(entity): send paintings with a horizontal 3D facing

### LINKED ISSUE #3563: Paintings from a world file disconnect the client with a network protocol error
### What happens

A painting that comes from a world file disconnects the client with a network protocol error as soon as it enters view. The crash report points at `Painting.recreateFromPacket` and `HangingEntity.setDirection`:

```
java.lang.IllegalArgumentException: The validated expression is false
	at net.minecraft.world.entity.decoration.HangingEntity.setDirection(HangingEntity.java:64)
	at net.minecraft.world.entity.decoration.painting.Painting.recreateFromPacket(Painting.java:229)
	at net.minecraft.client.multiplayer.ClientPacketListener.handleAddEntity(ClientPacketListener.java:610)
```

### Why

A painting stores its facing in the world as a horizontal value, `Direction.get2DDataValue`: 0 south, 1 west, 2 north, 3 east.

Pumpkin reads that value into the entity data (`read_custom_nbt` in `crates/pumpkin/src/entity/decoration/painting.rs`) and `create_spawn_packet` sends it as the spawn packet's object data. The client reads that field as a 3D direction, where 0 is down. A painting facing down fails the horizontal check in `HangingEntity.setDirection`, and the client closes the connection.

```
world NBT:        facing = 0   # south
packet data:           0      # the client reads down
expected:              3      # south in 3D indices
```

### Example

A parkour map from 1.12.2, upgraded by vanilla, contains 2 paintings and 50 item frames. Every client crashes when a painting enters view. Item frames are fine because their facing is already stored in the 3D encoding (north 2, south 3, west 4, east 5).

The painting NBT is valid vanilla data:

```
id = minecraft:painting
facing = 0
block_pos = [40, 224, 5]
variant = minecraft:bust
```

### Fix

Convert between the world file encoding and the 3D index when reading and writing the painting NBT. PR: #3564

matheustavarestrindade:
### Problem

A painting facing exists in two encodings in the protocol and in world files:

- The **world file** stores a horizontal value: `0` south, `1` west, `2` north, `3` east (vanilla `Direction.get2DDataValue`).
- The **entity data and spawn packet** use the 3D index: north `2`, south `3`, west `4`, east `5`.

Pumpkin uses the world value as the entity value, and the entity value as the world value.

**On load.** A painting saved as `0` (south) is sent as `0`. The client reads that field as a 3D direction, where `0` is down. A painting must be horizontal, so the client disconnects on the spot:

```
java.lang.IllegalArgumentException: The validated expression is false
	at net.minecraft.world.entity.decoration.HangingEntity.setDirection(HangingEntity.java:64)
	at net.minecraft.world.entity.decoration.painting.Painting.recreateFromPacket(Painting.java:229)
	at net.minecraft.client.multiplayer.ClientPacketListener.handleAddEntity(ClientPacketListener.java:610)
```

This is why upgraded maps crash: their paintings carry horizontal values, and `0` and `1` are vertical in the 3D encoding. Paintings placed in game do not crash, because the placement path stores the 3D index.

**On save.** The reverse mix-up writes non-vanilla data. A painting facing south (3D `3`) is saved as `facing = 3`, which vanilla reads as east, and west and east (3D `4` and `5`) are saved as values the format does not define.

### Change

`crates/pumpkin/src/entity/decoration/painting.rs`

- `facing_from_horizontal` converts the world value to the 3D index and is used by `read_custom_nbt`, so the spawn packet is correct.
- `facing_to_horizontal` converts back and is used by `write_custom_nbt`, so saved worlds match vanilla.

### Test

- `horizontal_and_3d_facings_round_trip`: every horizontal value maps to a horizontal direction and back.
- `south_two_is_not_down_zero`: the reported case, south `0` must become index `3`.
- `out_of_range_values_wrap_into_horizontal_directions`: `4` and `255` stay horizontal.

Run: `cargo test --ignore-rust-version -p pumpkin --lib painting`

### Migration

Worlds saved by the current code store the 3D index in the file, so after this change those paintings read one direction off: a painting saved as `3` was south and now reads as east. It only affects paintings placed with Pumpkin, not vanilla or upgraded worlds, and there is no crash either way. A datafixer can be added if you want old saves adjusted.

Fixes #3563

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed painting orientation handling so paintings consistently face the correct direction when saved and loaded.
  - Added reliable fallback behavior for missing or invalid orientation data, defaulting paintings to face south.
  - Improved handling of orientation values outside the standard range by wrapping them to a valid direction.


matheustavarestrindade:
One migration note for reviewers: paintings placed and saved by the current code store the 3D index in the NBT `facing` field, which expects the horizontal encoding. After this change those worlds read one direction off (for example a painting saved as 3 was south, now reads as east). It only affects paintings placed by Pumpkin itself, not vanilla or upgraded worlds, and there is no crash either way. I can add a datafixer or a one-time conversion if you want the old saves adjusted too.

matheustavarestrindade:
Fixed in 8b77fdc06. A missing field now goes through `facing_from_nbt` with a south default, matching vanilla's `getByte`, and `missing_facing_defaults_to_south` covers it: an empty compound gives South (index 3), and `facing = 1` still gives West.

MegalithOfficial (approved):
Looks right to me. Vanilla stores the painting facing as 0 south, 1 west, 2 north, 3 east and uses south when the field is missing, so this lines up with it.

One small thing, the test name `south_two_is_not_down_zero` is wrong. South is 0 in the file and 3 in the packet, there's no 2 anywhere. Please rename it. Correct me if Im wrong.

No need for a fixer for the old Pumpkin saves imo. Those were already wrong for vanilla, so Im fine with them being one direction off.

ZlordHUN:
LGTM

PR #3570, opened 2026-09-18
fix(item): only open signed books through the server

matheustavarestrindade:
### Problem

A book and quill cannot be written in. The client shows the editor for a moment and then switches to the read-only view.

### Why

The client opens the editor by itself when the player uses a book and quill. The open-book packet makes it open the book again, and that second open lands in the read-only view.

### Change

Send the open-book packet only for a signed book, which the client does not open on its own. The page-turn sound is unchanged.

### Tested

- Book and quill: writing, signing and reopening a signed book all work.
- Signed book: opens the reader through the packet as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected book-opening behavior so signed books open explicitly while writable books use the client’s automatic opening behavior.
  * Ensured book interactions respect the hand used, including correct signed-book opening behavior.
  * Preserved the page-turn sound for both book types.
  * Improved offhand item use so items are read from and updated in the correct hand.


matheustavarestrindade:
The small path is in 9152c65af: the hand now flows through `ItemRegistry` into a new `normal_use_with_hand` default, which forwards to `normal_use_with_rotation` so no other item changes behaviour, and the book sends `hand.to_packet_id()`.

The full trait change (adding the hand to `normal_use` and `normal_use_with_rotation` for every implementor) is a wider refactor, so I opened a separate issue for it: https://github.com/Pumpkin-MC/Pumpkin/issues/3575

MegalithOfficial (approved):
LGTM.

Issue #3561, opened 2026-09-18
Attempt to attack invalid entity when hitting ender dragon


zhangxuan2011:
### Before Submitting

- [x] I have searched existing issues and confirmed this is not a duplicate.

### Minecraft Edition

Java Edition

### Server Version / Commit

1ac447f

### Operating System

Android

### Bug Description & Expected Behavior

1. Description
When attacking ender dragon in the end, it will report this error:

```
21:57:03 ERROR Player id 1 interacted with entity id 1391, which was not found.
21:57:03  WARN Closing connection for 0: Attempting to attack an invalid entity
```

And server kicks me with the message which is contained in title.

2. Expected behavior
I should NOT be kicked and the ender dragon got hurt correctly

### Steps to Reproduce

1. Go to the end
2. Hit the ender dragon (for multiple times)

zhangxuan2011:
NOTE: This program does not panick, It just kicked me instead and runs correctly...

@RoosterBooster007

SurvivorsDev:
This should already be covered by #3407, which registers the dragon parts so they are addressable by id without being ticked, saved or tracked — that is what makes `get_entity_by_id` fail for id 1391 today, since it only walks `entities` and `players`. It is stacked on #3401, which is now approved and mergeable, so the chain looks unblocked.

Separately, #3583 removes the disconnect itself: `handle_attack` currently kicks whenever the entity lookup returns nothing, which vanilla ignores — it guards the whole branch with `target != null` and disconnects only for item entities, experience orbs, the player themselves and non-attackable arrows. That stops the kick regardless of the dragon work, but on its own it would leave the hit doing nothing rather than damaging the dragon.

PR #3583, opened 2026-09-19
fix(net): stop disconnecting players who attack a despawned entity

SurvivorsDev:
## Description

**What was changed?** `handle_attack` no longer disconnects the player when the attacked entity cannot be found, and it now disconnects for the targets vanilla actually rejects.

**Why were these changes necessary?** The entity is routinely gone by the time the attack packet lands — a falling block that just settled, a mob that just died, an item that was just picked up. `crates/pumpkin/src/net/java/play/attack.rs` kicked unconditionally in that case:

```rust
let Some(target) = target else {
    self.try_kick(&TextComponent::translate_cross(
        translation::java::MULTIPLAYER_DISCONNECT_INVALID_ENTITY_ATTACKED, ...
    ));
    return;
};
```

Vanilla guards the whole branch with `target != null` and simply ignores the packet when the entity is missing. It disconnects only for targets that are never attackable:

```java
Entity target = level.getEntityOrPart(packet.entityId());
if (target != null && level.getWorldBorder().isWithinBounds(target.blockPosition())) {
   ...
   if (target instanceof ItemEntity || target instanceof ExperienceOrb || target == this.player
       || target instanceof AbstractArrow arrow && !arrow.isAttackable()) {
      this.disconnect(Component.translatable("multiplayer.disconnect.invalid_entity_attacked"));
   }
```

So the deviation ran both ways: Pumpkin disconnected where vanilla ignores, and allowed attacking item entities and experience orbs where vanilla disconnects.

**What is the impact of this change?** Losing the race between entity removal and an in-flight attack packet no longer drops the connection. Item entities and experience orbs are now rejected the way vanilla rejects them; the self-attack case was already handled here.

**Are there any known issues or limitations?** The non-attackable-arrow case is not covered, since there is no `isAttackable` equivalent on arrows yet. Vanilla's world-border bound check on the target is also not replicated.

**Redundancy.** I found no open or closed PR touching this handler.

## Testing

- `cargo nextest run -p pumpkin` — 314 passed
- Manual: reproduced the disconnect by left-clicking a falling block as it landed, on a local build where the landing block leaves nothing behind, and confirmed it no longer disconnects with this change. On unmodified `master` the same click lands on the block that was just placed, so it becomes a dig rather than an attack — the race is reachable from the code path rather than from that particular reproduction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Players are no longer disconnected when attacking an entity that cannot be found.
  * Attacks against items or experience orbs now disconnect the player.
  * Other attack targets continue through normal attack handling.


zhangxuan2011:
@SurvivorsDev

Is this a fix of #3561 ?

SurvivorsDev:
Partly — it stops the disconnect, but not the rest of it.

The kick comes from `handle_attack` disconnecting whenever the entity lookup returns nothing, which this PR removes; vanilla ignores that case. So you would no longer be kicked.

The dragon still would not take damage, though. Its parts are built in `EnderDragon::new` but never registered:

```rust
// crates/pumpkin/src/entity/boss/ender_dragon.rs:242
// TODO: world.add_entity_silent(part.clone() as Arc<dyn EntityBase>);
```

so `World::get_entity_by_id` — which only walks `entities` and `players` — cannot find id 1391, which is exactly the "interacted with entity id 1391, which was not found" line. Vanilla looks these up through `Level.getEntityOrPart`, which falls back to the part map, and then routes the hit to the parent dragon.

So #3561 needs that TODO done as well, and an attack on a part forwarded to the dragon. Happy to leave that one to you since you filed it, or take a look if you would rather not.

xRookieFight (approved):

SurvivorsDev:
Bumping this — it's a small, self-contained fix (single file, no other open PR touches this handler per the description above), CI is green, and it fixes a real disconnect race (attacking a block right as it settles, or an item/orb the moment before pickup). @Snowiiii would appreciate a look when you have a moment.

PR #3588, opened 2026-09-19
fix(block): break suspicious sand and gravel when they land

SurvivorsDev:
## Description

**What was changed?** `BrushableBlock` now owns its own falling behaviour, and `FallingEntity` gained a `cancel_drop` flag plus a `land()` branch that mirrors vanilla's `FallingBlockEntity`.

**Why were these changes necessary?** Suspicious sand and gravel settled as a block after falling instead of breaking — the second half of #3067. Both were listed in `FallingBlock::ids()` while `BrushableBlock` claimed the same ids, and `BlockManager::register` keeps one behaviour per block id (`block_indices[i] = idx`), so the later registration silently replaced the earlier one. `FallingBlock` won, which also left `BrushableBlock::placed` and `broken` unreachable: the `BrushableBlockBlockEntity` was never created, and a buried item was lost when the block was broken.

Vanilla does not extend `FallingBlock` here either — `BrushableBlock` implements `Fallable` itself so it can call `disableDrop()`, which is what makes the block break on landing and drop nothing.

**What is the impact of this change?** Suspicious sand and gravel break on landing, the registry conflict is gone so their block entity and buried loot work again, and falling blocks that cannot settle where they land now drop as an item the way vanilla does.

**Are there any known issues or limitations?** Vanilla additionally requires `canSurvive` and `!wouldContinueFalling` before settling; only the replaceable check is implemented here. `callOnBrokenAfterFall` is not run on the drop-as-item path.

**Redundancy.** This carries the landing rework from #3156 by @ydw1904, credited as co-author. That PR was closed and they confirmed they are not returning to it. Two things are added on top: the registry conflict it did not reach, and preserving the concrete powder solidification that has since landed on `master` — rebasing #3156 as-is would have dropped it. The block-id list inside the entity is replaced by the `cancel_drop` flag so the decision lives on the block, as vanilla does.

## Testing

- `cargo nextest run -p pumpkin` — 316 passed
- Manual, on a 26.3 client against this branch: suspicious sand and gravel break on landing with break particles and drop nothing; concrete powder still hardens when it lands in water; sand dropped onto a torch drops as an item; brushing still completes and yields its item

One thing worth flagging from testing, though it is not caused by this change: a **26.2** client connected to a 26.3 server never renders server-sent block break particles, for any block. Breaking stone from the server with `/setblock <pos> air destroy` shows no particle either, while the same command on a matched-version pair does. So the missing particle when suspicious blocks break on landing is that general cross-version gap, not something specific to these blocks -- on a matched 26.3 client the particle renders correctly.


SurvivorsDev:
Bumping this — heads up that it's showing a merge conflict against current master (just drift from master moving, not a design issue); happy to rebase it as soon as it's going to get looked at, just didn't want to rebase repeatedly while it sits idle. @Snowiiii

PR #3591, opened 2026-09-19
fix(world): send the destroy effect vanilla actually uses

### LINKED ISSUE #3589: Older clients never render server-sent block break particles
### Before Submitting

### Minecraft Edition

Java Edition

### Server Version / Commit

Pumpkin 0.1.0+26.3-26.51 (Commit: c31eb836)

### Operating System

CachyOS (Arch)

A client older than the server never renders server-sent block break particles. Tested with a **26.2** client against a **26.3** server, which is inside the advertised `1.7.2-26.2` range.

The distinction is where the particle comes from:

- Breaking a block **yourself** shows particles. The Java client predicts that locally, and `World::break_block` deliberately skips the causing player (`broadcast_to_chunk_except_editioned`), so the server event is not involved.
- Any particle that has to come **from the server** does not render. `/setblock <pos> air destroy` on stone, dirt or suspicious gravel produces no particle at all on the older client, while the same command on a matched-version pair renders it normally.

Because it is the server-sent path that fails, an older client also would not see block break particles from *other* players, since those go through the same broadcast. That is easy to miss when testing alone.

Things I checked that are **not** the cause, in case it saves someone time:

- The block state remap is correct. `CWorldEvent::write_packet_data` remaps the `ParticlesDestroyBlock` data, and for a 26.2 client `suspicious_sand` 122 becomes 119 and `suspicious_gravel` 128 becomes 125 — exactly their native ids on a 26.2 build.
- It is not limited to states whose id shifts. Stone is one of only 27 of 35723 states that map to themselves, and it fails too.
- Packets are serialised per version: `World::broadcast_java_grouped` calls `serialize_packet_for_version` for each version group rather than encoding once.
- Packet ids are version aware through `MultiVersionJavaPacket::to_id(version)`.

I did not track down the actual cause.

### Steps to Reproduce

1. Run a 26.3 server and join with a 26.2 client
2. Run `/setblock <pos> stone replace`, then `/setblock <pos> air destroy`
3. No break particles appear; breaking the same block by hand does show them
4. Repeat with a 26.3 client and the particles appear in both cases

SurvivorsDev:
## Description

**What was changed?** Every call site that emits the block destruction effect now sends `WorldEvent::ParticlesAndSoundDestroyBlock` (2001) instead of `ParticlesDestroyBlock` (2014).

**Why were these changes necessary?** 26.3 renamed the old effect and handed its former name to a new one, so the generated enum now reads:

```rust
ParticlesAndSoundDestroyBlock = 2001   // what 26.2 called PARTICLES_DESTROY_BLOCK
ParticlesDestroyBlock         = 2014   // new in 26.3
```

The call sites kept the old name and silently moved from 2001 to 2014 with the port. Vanilla emits 2014 nowhere; 26.3 still uses 2001 for block destruction:

```java
EatBlockGoal.java:73         this.level.levelEvent(2001, below, Block.getId(...));
Rabbit.java:642              level.levelEvent(2001, cropsPos, Block.getId(blockState));
TurtleEggBlock.java:84       level.levelEvent(2001, pos, Block.getId(state));
CarvedPumpkinBlock.java:117  level.levelEvent(2001, block.getPos(), Block.getId(block.getState()));
```

**What is the impact of this change?** Two separate symptoms go away. A client older than the server has no event 2014 and silently drops the packet, so it renders no break particles at all — including other players' breaks, which travel the same broadcast. And since 2014 is the particles-only variant, server-sent breaks are currently silent on *every* client; you only hear your own because the client predicts it.

**Are there any known issues or limitations?** The two `write_packet_data` guards that decide whether to remap the block state id now accept either event, so the data keeps being remapped for older clients. The Bedrock `LevelEvent` is unaffected — 2001 is correct in its own numbering.

**Redundancy.** No open or closed PR touches these call sites.

## Testing

- `cargo clippy --workspace --all-targets`
- `cargo nextest run -p pumpkin` — 314 passed
- Manual: on a 26.3 server with a **26.2** client, `/setblock <pos> air destroy` produced no particle before and does now; the same holds for blocks that break on landing. Verified against a 26.3 client as well.

Fixes #3589

## Summary by CodeRabbit

- **Bug Fixes**
  - Block destruction effects now consistently display the associated sound alongside particles for golem and wither construction, door breaking, grass eating, and failed block-breaking events.
  - Java clients now receive correctly mapped block-state data for block-destruction effects across supported event formats.
  - Bedrock clients continue receiving the existing block-destruction effect behavior.


Q2297045667:
Now that the multi-version module has been extracted as a standalone plugin, this issue and the related PR are no longer applicable.

SurvivorsDev:
Thanks for the note on #3589 — could you point me at where the multi-version module landed? `level_event.rs`/`worldevent.rs` and their `write_packet_data` remap guards are still in this repo on current `master`, so I want to make sure I understand what actually moved before deciding whether to close this:

- Is the destroy-block event remap itself (the code this PR touches) moving out to the standalone plugin, or is it a different piece of the multi-version layer?
- If it's the former, is there a timeline/PR for that extraction I should rebase this against, or should this just be closed as superseded?

Happy to close it myself once I know which of those it is.

SurvivorsDev:
Bumping this — small diff, but the impact is bigger than it looks: right now every block break is silent for every player except the one who broke it (client-side prediction masks it locally). Fixing the event id restores sound for everyone else. @Snowiiii

PR #3596, opened 2026-09-19
fix(command): accept score holders in execute if score

matheustavarestrindade:
### Problem

`execute if score` only accepted a plain holder name, so datapacks that test the score of the executing entity failed with a syntax error:

> execute if score @s detectGenTime matches 0.. run time query daytime
Expected whitespace to end one argument, but found trailing data

### Change

The `target` and `source` arguments now use the same score holder argument as the scoreboard commands, so names and selectors both work.

This branch is based on the scoreboard branch, so the first commit is PR A. Merge that first.

### Tested

With this and PR A, the One Block Skyblock datapack runs end to end on Pumpkin, and the command above parses with `@s`.

## Summary by CodeRabbit

- **New Features**
  - Added scoreboard operations for assignment, arithmetic, swapping, minimum, and maximum values.
  - Added support for fake names, entity selectors, and wildcard score holders.
  - Expanded scoreboard commands to set, get, add, remove, reset, and operate on scores.
  - Score conditions now support flexible score-holder arguments.
  - Added suggestions and client-side metadata for scoreboard arguments.

- **Bug Fixes**
  - Score calculations now wrap safely on integer overflow.
  - Division and modulo by zero leave the target score unchanged.


matheustavarestrindade:
Added the two related score-condition fixes in c2a31fb77:

- Validate objective existence with `ARGUMENTS_OBJECTIVE_NOTFOUND`, so a missing objective no longer incorrectly passes `unless`. Missing scores in existing objectives remain a false test, and read-only objectives remain valid inputs.
- Support standalone `execute if/unless score` conditions: success returns 1 with non-broadcast `Test passed` feedback, and failure uses `COMMANDS_EXECUTE_CONDITIONAL_FAIL`.
- Reuse the same condition logic for standalone and chained forms, using the existing fork builder for Java-compatible chained results and error handling.

Added Rust regressions for lookup behavior, read-only objectives, feedback, result values, error propagation, and command registration. Verified with 359 Rust tests, formatting, Clippy, the server build, and 195 live RCON checks on each of Pumpkin and Mojang Java 26.3.

The server-global scoreboard change will be a separate follow-up.

PR #3597, opened 2026-09-19
fix(command): accept namespaced function names and tags in /function

matheustavarestrindade:
### Problem

`/function` only accepted a single word, so any namespaced function failed to parse:

> function nores:tick
Expected whitespace to end one argument, but found trailing data

Function tags could not be used at all.

### Change

The name argument now reads a full function name or `#tag`, and reports itself as `minecraft:function` to the client, which is what vanilla does. The executor resolves names and tags exactly as before.

### Tested

`function nores:tick` and map style calls such as `run function nores:detect` now parse and run on Pumpkin. Inline tests cover namespaced names, tags, whitespace handling and the empty name error.

## Summary by CodeRabbit

* **New Features**
  * The function command now accepts function names in formats such as `name`, `namespace:name`, and `#tag`.
  * Function arguments are parsed consistently during command execution and registration.
  * Function names can include leading or trailing whitespace appropriately, while empty names receive a localized syntax error.
  * Added clearer feedback when a function name is missing or invalid.


MegalithOfficial (changes requested):
The fix works for namespaced names and tags, but reading everything up to the next space skips the identifier parsing vanilla does. `FunctionArgument` skips the # and then calls `Identifier.read`, and we already have that as `Identifier::from_reader`.

Without it, names with no namespace break. `function ping` or `function #mcgrp` should resolve to `minecraft:ping` / `#minecraft:mcgrp`, but functions and tags are stored under the full `namespace:path` key, so they come back as unknown. I tested this on a server, and the namespaced versions of the same calls work fine. Invalid names like `Rev:hello!` also get through parsing and only fail later as an unknown function. Vanilla rejects them right away with `argument.id.invalid`.

Switching to `Identifier::from_reader` would also get rid of the empty-name error. That one uses `arguments.function.unknown`, which expects an argument.

matheustavarestrindade:
Thanks for the review, @MegalithOfficial. Updated in 001fa5eff to strip the optional `#` and use `Identifier::from_reader`. Bare functions and tags now resolve through the default `minecraft` namespace; identifier validation is delegated to the shared parser, and the custom empty-name error is removed.

Verified with 335 Rust tests and 28 live RCON assertions on both Pumpkin and Mojang Java 26.3, including execution of bare/namespaced functions and tags. Formatting, Clippy, and the build also pass.

MegalithOfficial (approved):
LGTM, thank you @matheustavarestrindade !

PR #3598, opened 2026-09-19
fix: cancel test CI jobs correctly

Phoenixxo:
## Description
Fixes CI test jobs continuing to run despite the job being cancelled. This was due to the use of `always()`, which overrides and doesn't care about cancellation. Switching to `!cancelled()` instead gives us all of those benefits of `always()` , but doesn't continue to run after being cancelled. 

## Summary by CodeRabbit

* **Chores**
  * Updated automated test and Android build workflow conditions to prevent jobs from running after cancellation.
  * Cleaned up formatting in the release lint workflow.


MegalithOfficial (changes requested):
LGTM. but `build_android_pull_request` (line 250) still has always(), switch that one too?

MegalithOfficial (approved):
LGTM, thanks.

PR #3600, opened 2026-09-19
fix(command): accept block states in block arguments

SurvivorsDev:
## Description

`setblock`, `fill`, `clone` and `execute if block` advertise `block_state` and `block_predicate` to the client, so clients send `minecraft:wheat[age=3]`. The parser read only the block name: `[` ended the token and the rest came back as *Expected whitespace to end one argument, but found trailing data*. None of those commands could place or match anything but a block's default state.

A second bug sat next to it. `BlockPredicate` did parse properties, but `test` compared block ids alone, so `fill … replace minecraft:oak_log[axis=x]` matched every log regardless of axis.

Both are fixed here.

The `[key=value, …]` suffix is now parsed the way vanilla's `BlockStateParser.readProperties` does, rejecting unknown property names, values the property does not accept, a repeated name, a missing value and an unclosed bracket, with vanilla's messages. Pumpkin's generated block data carries no property table, so the accepted names and values are recovered by walking the block's own states.

Placing applies the written properties over the default state; matching compares only the properties that were written and leaves the rest as wildcards. That is the split vanilla draws between `BlockStateArgument` and `BlockPredicateArgument`.

`commands.data`-style arity was checked against the 26.3 client's `en_us.json`: `argument.block.property.novalue` takes two arguments, not the one it was declared with, so that is corrected too.

Part of #2592.

## Testing

Five unit tests, including one that walks every `oak_stairs` state to check that unwritten properties really do stay free. Reverting the fix fails three of them.

In game over RCON on 26.3:

- `setblock 0 80 0 minecraft:oak_stairs[facing=east,half=top]` places that exact state
- `fill 4 80 0 6 80 0 minecraft:oak_stairs[facing=north,half=top]` fills 3
- `execute if block` truth table over seven cases, all correct: `[facing=east]`, `[half=top]` and the bare name match; `[facing=west]`, `[half=bottom]` and `minecraft:stone` do not
- with `oak_log[axis=x]` and `oak_log[axis=z]` side by side, `fill … replace minecraft:oak_log[axis=x]` fills **1**, `fill … replace #minecraft:logs[axis=y]` fills **1**, and `clone … filtered minecraft:oak_log[axis=x]` clones **1**
- all five error messages match vanilla, including the cursor context

## Summary by CodeRabbit

* **New Features**
  * Block arguments now support specifying block properties, including validation for unknown, duplicate, missing, invalid, and incomplete values.
  * Omitted properties act as wildcards when matching blocks.
* **Bug Fixes**
  * Block conditions in clone and execute commands now evaluate block properties and states correctly.
  * Fill and setblock commands now apply the selected block state and properties instead of defaulting to the base state.
  * Block argument examples and parsing behavior have been expanded and covered by additional tests.


SurvivorsDev:
Bumping this — right now `setblock`/`fill`/`clone`/`execute if block` can't accept the `[key=value]` block-state syntax at all (parser error), and `fill ... replace` filters ignore properties entirely. Both are fixed here with vanilla's own parsing/error messages. @Snowiiii

PR #3601, opened 2026-09-19
fix(command): merge a repeated literal instead of replacing it

SurvivorsDev:
## Description

`data get entity @p` and `data get block ~ ~ ~` were parse errors rather than commands. Only `storage` worked.

`ArgumentBuilder::then` inserted each child into a map keyed by name, so registering the same literal twice discarded the first node together with its whole subtree. Brigadier's `CommandNode.addChild` merges instead: an incoming executor replaces the current one and the incoming children are folded in recursively.

`data` describes one verb over three targets by registering `get`, `merge`, `remove` and `modify` once per target kind, so only the last kind registered — storage — survived. The same applied to `modify … from` and `… string`. Instrumenting registration counts **68** nodes lost this way, all of them in `/data`; no other command registers a duplicate literal, so merging can only recover what was already broken.

The second commit fixes something the first one uncovers. `commands.data.block.query` takes four arguments and `commands.data.block.modified` takes three, one per coordinate, but both were handed the coordinates joined into a single string. With the block branch unreachable nobody had seen the result:

```
0, 90, 0, {Items: [], id: "minecraft:chest", …}, {Items: [], …} has the following block data: {Items: [], …}
```

## Testing

Two regression tests registering the same literal twice: one checks that both branches remain reachable, the other that a later registration installs its executor without discarding the branch an earlier one hung below it. Reverting `then` to the old insert fails both.

In game over RCON on 26.3:

- `data get block 0 90 0` → `0, 90, 0 has the following block data: {Items: [], id: "minecraft:chest", x: 0, y: 90, z: 0}`
- `data get entity @e[type=armor_stand,limit=1] Pos` → `[0.5d, 87d, 0.5d]`, and `… Health` → `20f`
- `data modify entity … Invisible set value 1b` persists, `0b` → `1b`
- `data merge`, `data remove` and `data modify … set value` / `set from entity` / `set string entity` all work on storage
- `data modify block 0 90 0 …` reports success but does not persist; that is a separate bug in the write path, fixed in a follow-up rather than here

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved all command branches when multiple registrations share the same literal path.
  - Allowed later registrations to replace shared command behavior without removing existing branches.
  - Kept redirects valid when command nodes are merged, including across separate registrations.
  - Improved block-coordinate feedback by presenting X, Y, and Z values as separate arguments.


SurvivorsDev:
Superseded by #3602 — its branch (`data-block-write`) includes every commit from this one (`literal-node-merge`) plus the additional data-write fix on top, so nothing here is lost. Closing in favor of #3602 to keep the open-PR count down.

PR #3602, opened 2026-09-19
fix(command): make data writes reach the block entity

SurvivorsDev:
## Description

> Stacked on #3601 — its two commits are included here, so review that one first. This PR's own change is the last commit.

`data merge block` and `data modify block` reported success and changed nothing.

`set_data` handed the tag to `add_block_entity_nbt`, which stages NBT in the chunk's pending map for a block entity that does not exist yet. The migration that drains it skips any position whose block entity is already loaded:

```rust
let already_loaded = self.block_entities.get(&chunk_pos).is_some_and(|m| m.contains_key(&pos));
if !already_loaded && let Some(entity) = self.get_block_entity(&pos) {
```

That mechanism is for chunk loading and worldgen, where a structure writes chest contents before the entity exists. A write to a live block entity was dropped on the floor.

This rebuilds the block entity from the tag and puts it in place. Vanilla loads the tag into the entity that is already there, so the `id` and coordinates it carries can neither retype nor move it; there is no in-place load here, so both are pinned back to what is actually at the position before rebuilding.

One difference from vanilla remains, and I would rather flag it than hide it: the entity is replaced rather than loaded into, so a container being viewed at the moment the command runs keeps the old instance. Matching vanilla exactly would mean adding a load-in-place method to `BlockEntity` and implementing it across every type, which seemed worth separating from the fix itself.

## Testing

In game over RCON on 26.3, against a chest at `0 90 0`:

- `data merge block 0 90 0 {Items:[{Slot:0b,id:"minecraft:diamond",count:7}]}` persists, and the items come back from `data get`
- `data modify block 0 90 0 Items[0].count set value 3` persists, 7 → 3
- opening the chest in game shows the diamonds with the right count, so the `CBlockEntityData` broadcast reaches the client
- `data merge block 0 90 0 {id:"minecraft:furnace"}` leaves it a chest
- `data merge block 0 90 0 {x:99,y:99,z:99}` leaves it at `0 90 0`

410 tests pass.

## Summary by CodeRabbit

- **New Features**
  - Command registrations with the same name now preserve all previously registered child branches.
  - Updating an existing command keeps its nested subcommands while applying the latest command behavior.
  - Redirects to merged command nodes continue to resolve correctly.
  - Block data updates now rebuild valid block entities while preserving their type and position.

- **Bug Fixes**
  - Invalid or missing block entities now return a clear block-invalid error.
  - Block update success messages display coordinates consistently as separate X, Y, and Z values.


SurvivorsDev:
Now that #3601 is closed (its commits are included here, so nothing was lost — see the note on that PR), this one stands on its own: `data merge/modify block` currently reports success while silently changing nothing. @Snowiiii

PR #3604, opened 2026-09-19
fix(ci): allow builds after skipped optional check

Phoenixxo:
## Summary

Fix CI jobs being skipped when optional upstream checks are skipped.

* use `!cancelled()`  for so downstream jobs still evaluate their conditions
* explicitly require test success before builds run
* preserve cancellation behavior for superseded workflow runs

This should be the last fix now, small hiccups are hard to test with CI, I apologize

## Summary by CodeRabbit

* **Chores**
  * Updated pull request build checks to run only when applicable code changes are present and regular tests succeed.
  * Prevented build checks from running after workflow cancellation.


MegalithOfficial (approved):
LGTM

QuantumSegfault (approved):

